PR #8, opened 2021-08-11
Add e2e test framework

simonswine:
This adds the e2e test framework, as this one of things how Tempo/Loki depend on the newly added backoff method.

Any thoughts about how we could get rid of the `github.com/prometheus/prometheus` dependency? I am afraid the protobufs in `prompb` are only available through it...

This is the only use of it: 
https://github.com/grafana/dskit/pull/8/files#diff-a004dfdb3988c7eb31e0e060898b896853e0ed03fe5499e6536228b1b52aa19fR107


test/integration/e2e/images/images.go
```diff
@@ -0,0 +1,17 @@
+package images
+
+// If you change the image tag, remember to update it in the preloading done
+// by GitHub actions (see .github/workflows/*).
+
+// These are variables so that they can be modified.
+
+var (
```
pracucci, line 8: Here we should define only the images used by the framework.

test/integration/e2e/scenario_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/thanos-io/thanos/pkg/objstore/s3"
+	"gopkg.in/yaml.v2"
+
+	"github.com/cortexproject/cortex/integration/e2e"
+	e2edb "github.com/cortexproject/cortex/integration/e2e/db"
```
pracucci, line 19: Doesn't look correct.

test/integration/e2e/service_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"net"
+	"net/http"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/cortexproject/cortex/pkg/util"
```
pracucci, line 16: Shouldn't import cortex.

pracucci (commented):
Thanks @simonswine ! We have a couple of tests with `requires_docker` that we need to run in CI. Could you get them work, please?

simonswine:
I have started moving the labels package out of prometheus into prometheus/common:

https://github.com/prometheus/common/compare/main...simonswine:20210820_move-pkg-labels

In the meantime we could use: https://github.com/simonswine/prometheus-labels

We are currently still depending on cortex via Thanos:

https://github.com/grafana/dskit/pull/8/files#diff-d8682c232bddbc1ce0b4b704b0dd36f60da0f9d93f13cd728d0c3423c4f6afdbR15

pracucci:
> In the meantime we could use: https://github.com/simonswine/prometheus-labels

Could you explain me why we need this? I'm a bit concerned about using forks here. Why not vendoring prometheus until labels will be moved to prometheus/common?

PR #10, opened 2021-08-13
Fix linting issues after enabling CI

simonswine:
This is failing now in main, as it went in before CI was around.

(https://github.com/mgechev/revive relplaced golint in golangci-lint)


aknuds1 (approved):

56quarters (approved):
:+1:

PR #11, opened 2021-08-16
Add runtimeconfig to dskit

treid314:


runtimeconfig/manager.go
```diff
@@ -0,0 +1,210 @@
...
+
+	configMtx sync.RWMutex
+	config    interface{}
+
+	configLoadSuccess prometheus.Gauge
+	configHash        *prometheus.GaugeVec
+}
+
+// New creates an instance of Manager and starts reload config loop based on config
+func New(cfg Config, registerer prometheus.Registerer, logger log.Logger) (*Manager, error) {
```
treid314, line 61: Should we add a prefix here for the metric names?

runtimeconfig/manager.go
```diff
@@ -0,0 +1,210 @@
...
+type Config struct {
+	ReloadPeriod time.Duration `yaml:"period"`
+	// LoadPath contains the path to the runtime config file, requires an
+	// non-empty value
+	LoadPath string `yaml:"file"`
+	Loader   Loader `yaml:"-"`
+}
+
+// RegisterFlags registers flags.
+func (mc *Config) RegisterFlags(f *flag.FlagSet) {
```
treid314, line 37: We could add a prefix here, but it would a be a breaking change for cortex/tempo/loki

pstibrany: I would keep it as is for now. We can always add `RegisterFlagsWithPrefix` in the future, if needed.

runtimeconfig/manager.go
pstibrany: Generic is fine. People can pass custom registrerer that adds prefix, if they want to. (`prometheus.WrapRegistererWithPrefix`)

pstibrany (approved):
LGTM, thank you.

pracucci (approved):
Good job!

PR #12, opened 2021-08-17
Import module_service service wrapper with history

56quarters:
Imports the Cortex module_service wrapper that waits for module dependencies when starting and stopping, with history.


aknuds1 (approved):
LGTM

PR #14, opened 2021-08-18
Add kv package

aknuds1:
Add `kv` package.

TODOS:

- [x] Figure out if `kv.memberlist.ringBroadcast.Finished` needs to have a pointer receiver, since it mutates the receiver


pracucci:
What do you think to import `flagext` package in a preliminary PR?

aknuds1:
> What do you think to import `flagext` package in a preliminary PR?

@pracucci I'd be OK with that.

aknuds1:
> What do you think to import `flagext` package in a preliminary PR?

@pracucci done in #16.

kv/memberlist/broadcast.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (r ringBroadcast) Message() []byte {
+	return r.msg
+}
+
+func (r ringBroadcast) Finished() {
+	if r.finished != nil {
+		r.finished(r)
+	}
+	// TODO: Figure out whether the receiver needs to be made a pointer, so this mutation is valid
```
pstibrany, line 63: You're correct. If we want to set `r.msg` correctly, receiver needs to be a pointer. I think idea here was to clear `msg` pointer to make sure we don't leak memory, but this hasn't been a problem so far, so I think we can remove TODO and commented line. WDYT?

kv/memberlist/testlogger.go
```diff
@@ -0,0 +1,8 @@
+package memberlist
+
+type testLogger struct {
```
pstibrany, line 3: Shall we move this to one of _test files?

kv/testlogger.go
```diff
@@ -0,0 +1,8 @@
+package kv
+
+type testLogger struct {
```
pstibrany, line 3: This seems unused. Can we remove it?

go.mod
```diff
@@ -3,11 +3,35 @@ module github.com/grafana/dskit
...
+	github.com/hashicorp/go-cleanhttp v0.5.1
+	github.com/hashicorp/go-sockaddr v1.0.2
+	github.com/hashicorp/memberlist v0.2.3
+	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
+	github.com/miekg/dns v1.1.42 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
 	github.com/prometheus/common v0.26.0
 	github.com/stretchr/testify v1.7.0
+	github.com/thanos-io/thanos v0.22.0
```
pstibrany, line 26: Surprised to see Thanos as dependency here... with little extra work, I think we could remove it (esp. DNS provider). But maybe that's better left for another PR.

pstibrany (approved):
Good job! Thank you.

go.mod
aknuds1: @pstibrany same here. I was going to ask if we can do what you're suggesting, i.e. extract what we need from it.

kv/memberlist/testlogger.go
aknuds1: @pstibrany the drawback is that the definition is shared between the different test files. It's a bit weird that a shared definition is in one of them. What's the typical way of sharing code between test files? Stick a build tag in a shared .go file so it only builds for tests?

pstibrany: I don't see that as a big problem. Alternatively, we can rename `testlogger.go` to `logger_test.go` so that it's clear that it is only used from tests.

kv/memberlist/broadcast.go
aknuds1: @pstibrany if it's OK to remove the mutation I'm totally OK with that solution.

kv/memberlist/testlogger.go
aknuds1: @pstibrany OK, let me see what I can do.

go.mod
pstibrany: I think biggest part would be DNS provider. Instead of direcly using one from Thanos, we can define an interface in dskit, provide basic implementation, and allow dskit clients to replace it with one from Thanos. But I would suggest to do that in a separate PR, together with copying/replacing other thanos dependencies (apart from DNS provider, I only see tiny usage from tests). WDYT?

kv/testlogger.go
aknuds1: @pstibrany it's actually used by tests :) Same as the other one.

go.mod
aknuds1: A separate PR is OK by me @pstibrany 👍

kv/memberlist/broadcast.go
aknuds1: Removing the TODO and mutation.

pstibrany: > if it's OK to remove the mutation I'm totally OK with that solution.

It doesn't work in its current form anyway, so looks totally OK to remove it to me :)

kv/memberlist/testlogger.go
aknuds1: Done.

pstibrany:
Thanks for addressing my feedback!

aknuds1:
> Thanks for addressing my feedback!

You're welcome @pstibrany - happy for your review :)

kv/consul/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"context"
+
+	consul "github.com/hashicorp/consul/api"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/weaveworks/common/instrument"
+)
+
+var consulRequestDuration = instrument.NewHistogramCollector(prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
pracucci, line 11: A global metric slipped in. We shouldn't have it.

kv/multi.go
```diff
@@ -0,0 +1,361 @@
...
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+)
+
+var (
+	primaryStoreGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
pracucci, line 17: Global metrics here.

PR #15, opened 2021-08-18
Move moduleService and ErrStopProcess to modules.

pstibrany:
`moduleService` and `ErrStopProcess` were moved out of `services` package to `modules`.

Rationale: Idea behind `services` package is to provide clean reimplementation of Guava services, without too much cruft in it. `moduleService` is used by modules, and is not related to `services` package. Same goes for `ErrStopProcess`.

### DIFF
--- a/modules/module_service.go
+++ b/modules/module_service.go
@@ -1,4 +1,4 @@
-package services
+package modules
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


pracucci (approved):
👍  Thanks!

PR #16, opened 2021-08-19
Add flagext package

aknuds1:
Add `flagext` package from Cortex. Will be needed by the `kv` package.


flagext/deprecated.go
```diff
@@ -1,38 +0,0 @@
-package flagext
```
pstibrany: Why not keep this as well, perhaps with modified metric name? (It's used to mark some flags as deprecated).

pstibrany (approved):
Thank you!

flagext/deprecated.go
aknuds1: @pstibrany there was an issue with a dependency on a global logger, and it was the easy fix to remove it. I can put it back though, rewritten with parameterized logger.

pstibrany: I see. It seems generic enough to be usable for other projects too, but I'm not sure if Tempo/Loki already use it or not. That may guide the decision here.

aknuds1: @pstibrany I re-added this, with parameterized logger.

PR #17, opened 2021-08-19
Add README and CONTRIBUTING

simonswine:
This is my first stab at both a README and a CONTRIBUTING markdown. I have used the cortex contributing guide as source of inspiration :slightly_smiling_face:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,121 @@
+# Contributing
+
+Welcome! We're excited that you're interested in contributing. Below are some
+basic guidelines.
+
+## Workflow
+
+Dskit follows a standard GitHub pull request workflow. If you're unfamiliar
+with this workflow, read the very helpful [Understanding the GitHub
+flow][github-flow] guide from GitHub.
+
+[github-flow]: https://guides.github.com/introduction/flow/
+
+You are welcome to create draft PRs at any stage of readiness - this can be
+helpful to ask for assistance or to develop an idea. But before a piece of work
+is finished it should:
+
+* Be organised into one or more commits, each of which has a commit message
+  that describes all changes made in that commit ('why' more than 'what' - we
+  can read the diffs to see the code that changed).
+
+* Each commit should build towards the whole - don't leave in back-tracks and
+  mistakes that you later corrected.
+
+* Have unit tests for new functionality or tests that would have caught the bug
+  being fixed.
+
+* Include a CHANGELOG message if users of Dskit need to hear about what you
+  did.
+
+## Testing und Linting the Code
+
+To lint the code:
+
+```bash
+make lint
+```
+
+To run the unit tests suite:
+
+```bash
+make test
+```
+
+### Dependency management
+
+We use [go modules] to manage dependencies on external packages.  This requires
+a working Go environment with version 1.16 or greater, git and [bzr] installed.
+
+[Go modules]: https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more
+[bzr]: http://wiki.bazaar.canonical.com/Download
+
+To add or update a new dependency, use the `go get` command:
+
+```bash
+# Pick the latest tagged release.
+go get example.com/some/module/pkg
+
+# Pick a specific version.
+go get example.com/some/module/pkg@vX.Y.Z
+```
+
+Tidy up the `go.mod` and `go.sum` files:
+
+```bash
+go mod tidy
+git add go.mod go.sum
+git commit
+```
+
+You have to commit the changes to `go.mod` and `go.sum` before submitting the
+pull request.
+
+Dskit uses the `goimports` tool (`go get golang.org/x/tools/cmd/goimports` to
+install) to format the Go files, and sort imports. We use goimports with
+`-local github.com/grafana/dskit` parameter, to put Dskit internal imports into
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -0,0 +1,22 @@
+# Grafana Dskit
+
+This library contains utilities that are useful for building distributed
+services.
+
+## Current state
+
+This library is still in development. In a first stage we plan to move over
+utilities from the [Cortex] project.
```
aknuds1, line 9: Nit, but I think "during the first stage" sounds more precise:
```suggestion
This library is still in development. During the first stage we plan to move over
utilities from the [Cortex] project.
```

README.md
```diff
@@ -0,0 +1,22 @@
...
+## Contributing
+
+If you're interested in contributing to this project:
+
+- Start by reading the [Contributing guide](/CONTRIBUTING.md).
+
+## License
+
+[Apache 2.0 License](https://github.com/grafana/dskit/blob/main/LICENSE)
+
```
aknuds1, line 22: Nit, but can we remove this extra newline?

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+
+To run the unit tests suite:
+
+```bash
+make test
+```
+
+### Dependency management
+
+We use [go modules] to manage dependencies on external packages.  This requires
```
aknuds1, line 47: Nit, just spotting an extra space.
```suggestion
We use [go modules] to manage dependencies on external packages. This requires
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+a working Go environment with version 1.16 or greater, git and [bzr] installed.
+
+[Go modules]: https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more
+[bzr]: http://wiki.bazaar.canonical.com/Download
```
pstibrany, line 51: :) this looks old

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+go get example.com/some/module/pkg
+
+# Pick a specific version.
+go get example.com/some/module/pkg@vX.Y.Z
+```
+
+Tidy up the `go.mod` and `go.sum` files:
+
+```bash
+go mod tidy
```
pstibrany, line 66: should we include `go mod vendor` here too? Otherwise lint check will fail.

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+
+You have to commit the changes to `go.mod` and `go.sum` before submitting the
+pull request.
+
+Dskit uses the `goimports` tool (`go get golang.org/x/tools/cmd/goimports` to
+install) to format the Go files, and sort imports. We use goimports with
+`-local github.com/grafana/dskit` parameter, to put Dskit internal imports into
+a separate group. We try to keep imports sorted into three groups:
+imports from standard library, imports of 3rd party packages and internal
+Cortex imports. Goimports will fix the order, but will keep existing newlines
```
aknuds1, line 79: I think there's a stray Cortex reference there.
```suggestion
a separate group. We try to keep imports sorted into three groups:
imports from standard library, imports of 3rd party packages and internal
imports. Goimports will fix the order, but will keep existing newlines
```

pstibrany (approved):

aknuds1 (changes requested):
Thank you for adding these, they generally look good. Apart from nits, I just think you should fix what looks like a stray reference to Cortex in the contribution guide.

CONTRIBUTING.md
simonswine: Agreed and removed, haven't heard about bazaar in a while :joy:

CONTRIBUTING.md
simonswine: We currently don't have a vendor folder checked in and hence we also don't do a lint check on it.

https://github.com/grafana/dskit/blob/4710207529678a6e5f8954d1c280b1d099a3a70e/Makefile#L27-L32

CONTRIBUTING.md
```diff
@@ -0,0 +1,120 @@
...
+
+To run the unit tests suite:
+
+```bash
+make test
+```
+
+### Dependency management
+
+We use [go modules] to manage dependencies on external packages. This requires
```
aknuds1, line 47: Total nit, but looks as if we mostly write it as _Go_:
```suggestion
We use [Go modules] to manage dependencies on external packages. This requires
```

CONTRIBUTING.md
aknuds1: Christian is right I think.

CONTRIBUTING.md
```diff
@@ -0,0 +1,120 @@
...
+
+    ```
+    # Ubuntu/Debian
+    apt-get install git-filter-repo
+
+    # MacOSX
+    brew install git-filter-repo
+    ```
+
+### Import a particular files
```
aknuds1, line 95:
```suggestion
### Import particular files
```

aknuds1 (approved):
LGTM!

CONTRIBUTING.md
pstibrany: I see. I thought we follow the same practice as in Cortex, but forgot to check. Thanks!

pstibrany (approved):

PR #20, opened 2021-08-24
kv/consul: Close mock

aknuds1:
Make sure to close `kv/consul` mock.

### DIFF
--- a/kv/consul/client_test.go
+++ b/kv/consul/client_test.go
@@ -29,10 +29,13 @@ func writeValuesToKV(t *testing.T, client *Client, key string, start, end int, s
 }
 
 func TestWatchKeyWithRateLimit(t *testing.T) {
-	c := NewInMemoryClientWithConfig(codec.String{}, Config{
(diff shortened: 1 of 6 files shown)


kv/consul/client_test.go
```diff
@@ -82,7 +88,10 @@ func TestWatchKeyNoRateLimit(t *testing.T) {
 }
 
 func TestReset(t *testing.T) {
-	c := NewInMemoryClient(codec.String{}, testLogger{})
+	c, closer := NewInMemoryClient(codec.String{}, testLogger{})
+	t.Cleanup(func() {
+		_ = closer.Close()
```
pracucci, line 93:
```suggestion
		require.NoError(t, closer.Close())
```

_Same comment applies in other tests._

pracucci (approved):
LGTM, but please always assert errors in tests (even in cases where the current implementation can't return an error, because it may change in the future)

kv/consul/client_test.go
aknuds1: @pracucci OK will take care of it.

PR #21, opened 2021-08-24
Remove dependency on Thanos

aknuds1:
Remove dependency on Thanos, using a DNS provider interface instead of the concrete type from Thanos.


kv/memberlist/dnsprovider.go
```diff
@@ -0,0 +1,15 @@
+package memberlist
+
+import (
+	"context"
+)
+
+type DNSProvider interface {
```
56quarters, line 7: nit: I realize this exists so that we don't need to depend on Thanos, but could we include some examples of input addresses and results to help people that want to create an implementation of this?

56quarters (approved):
LGTM, one non-blocking nit

kv/kvtls/test/ca.go
```diff
@@ -43,15 +45,17 @@ func newCA(name string) *ca {
...
 	}
-
 }
 
 func writeExclusivePEMFile(path, marker string, mode os.FileMode, data []byte) (err error) {
 	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, mode)
 	if err != nil {
 		return err
 	}
-	defer runutil.CloseWithErrCapture(&err, f, "write pem file")
```
pracucci, line 54: We use this `CloseWithErrCapture()` util quite frequently. We may just copy it into dskit and use it here. It's a tiny util, sounds fine copying it.

kv/memberlist/memberlist_client_test.go
```diff
@@ -1295,3 +1295,16 @@ type testLogger struct {
 func (l testLogger) Log(keyvals ...interface{}) error {
 	return nil
 }
+
+type providerImpl struct {
```
pracucci, line 1299: [nit] Which provider? A DNS provider! So I would name it `dnsProviderMock`. "Mock" because it's not a real implementation but just a mock.

```suggestion
type dnsProviderMock struct {
```

pracucci (approved):

kv/kvtls/test/ca.go
aknuds1: @pracucci OK looking into it.

kv/memberlist/memberlist_client_test.go
aknuds1: @pracucci 👍

kv/kvtls/test/ca.go
aknuds1: Done.

kv/memberlist/memberlist_client_test.go
aknuds1: Done.

kv/memberlist/dnsprovider.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+)
+
+// DNSProvider supports storing or resolving a list of addresses.
+type DNSProvider interface {
+	// Resolve stores a list of provided addresses or their DNS records if requested.
+	// Addresses prefixed with `dns+` or `dnssrv+` will be resolved through respective DNS lookup (A/AAAA or SRV).
+	// For non-SRV records, it will return an error if a port is not supplied.
+	//
+	// Example: By providing the address of "dns+grafana.com:80"}, it should resolve into something like []string{"34.120.177.193:80", "[2600:1901:0:b3ea::]:80"}.
```
aknuds1, line 13: @56quarters what do you think of this example? Couldn't come up with anything that produces results for `dnssrv+`.

kv/memberlist/dnsprovider.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+)
+
+// DNSProvider supports storing or resolving a list of addresses.
+type DNSProvider interface {
+	// Resolve stores a list of provided addresses or their DNS records if requested.
+	// Addresses prefixed with `dns+` or `dnssrv+` will be resolved through respective DNS lookup (A/AAAA or SRV).
+	// For non-SRV records, it will return an error if a port is not supplied.
+	//
+	// Example: By providing the address of "dns+grafana.com:80", it should resolve into something like []string{"34.120.177.193:80", "[2600:1901:0:b3ea::]:80"}.
```
aknuds1, line 13: @56quarters what do you think of this example? Couldn't come up with anything that produces results for `dnssrv+`.

kv/memberlist/dnsprovider.go
aknuds1: See my comment below.

kv/memberlist/dnsprovider.go
pracucci: The prefixes documented here are implementation specific and shouldn't belong to the interface. The comment for this function should just be:

```
Resolve stores a list of provided addresses or their DNS records if requested.
```

pracucci (approved):

kv/memberlist/dnsprovider.go
```diff
@@ -0,0 +1,18 @@
+package memberlist
+
+import (
+	"context"
+)
+
+// DNSProvider supports storing or resolving a list of addresses.
+type DNSProvider interface {
+	// Resolve stores a list of provided addresses or their DNS records if requested.
+	// Addresses prefixed with `dns+` or `dnssrv+` will be resolved through respective DNS lookup (A/AAAA or SRV).
```
pstibrany, line 10: I think use "dns+" prefixes is an implementation detail of Thanos resolver. I wouldn't consider it part of interface contract.

I would say instead that addresses are supplied by user, and implementations may interpret these addresses in their own specific ways.

kv/memberlist/dnsprovider.go
aknuds1: @pracucci thanks for clarifying that.

aknuds1: I'll update it.

kv/memberlist/dnsprovider.go
aknuds1: Thanks @pstibrany .

kv/memberlist/dnsprovider.go
56quarters: I think that's great to start.

56quarters: Oh, just saw @pstibrany's comment. That's fine too

aknuds1: @56quarters sorry, we removed this documentation in the meantime since @pstibrany and @pracucci pointed out it's implementation specific :)

PR #29, opened 2021-09-02
Add util concurrency

treid314:
**What this PR does**: As part of moving the ring package we have to bring over a handful of decencies. Concurrency contains syncBuffer and Job Runners.

**Checklist**
- [-] Tests updated


concurrency/runner.go
```diff
@@ -0,0 +1,108 @@
+package concurrency
+
+import (
+	"context"
+	"sync"
+
+	tsdbErrors "github.com/prometheus/prometheus/tsdb/errors"
```
treid314, line 7: Sadly we use Prometheus/Prometheus here. Is there a better error collector type object to use here? Here's the [Prometheus implementation of multierror](https://github.com/prometheus/prometheus/blob/main/tsdb/errors/errors.go#L27)

pstibrany: Let's use `multierror.MultiError` from ... `dskit` :)

treid314: Oh, Doh!

go.mod
```diff
@@ -4,43 +4,31 @@ go 1.16
...
 	github.com/hashicorp/memberlist v0.2.3
-	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
-	github.com/miekg/dns v1.1.42 // indirect
-	github.com/mitchellh/mapstructure v1.4.1 // indirect
 	github.com/opentracing-contrib/go-grpc v0.0.0-20210225150812-73cb765af46e // indirect
-	github.com/opentracing-contrib/go-stdlib v1.0.0 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
-	github.com/prometheus/common v0.26.0
+	github.com/prometheus/common v0.29.0
```
pstibrany, line 22: Why do we need this update?

concurrency/runner.go
```diff
@@ -0,0 +1,108 @@
+package concurrency
+
+import (
+	"context"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
+
+	dskitMath "github.com/grafana/dskit/math"
```
pstibrany, line 9: no need for this alias

concurrency/runner.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// wait for ongoing workers to finish.
+	wg.Wait()
+
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	errsMx.Lock()
+	defer errsMx.Unlock()
```
pstibrany, line 61: No need for locking here. `wg.Wait()` call (together with using `defer wg.Done()`)  already establishes happens-before relation between goroutines adding errors, and main goroutine calling `errs.Err()`. (`errsMx` is still needed inside goroutines though)

concurrency/buffer.go
```diff
@@ -0,0 +1,25 @@
+package concurrency
+
+import (
+	"bytes"
+	"sync"
+)
+
+type SyncBuffer struct {
```
pstibrany, line 8: Can we add comment to this type?

pstibrany (commented):
Thanks for working on this.

concurrency/buffer.go
treid314: Yeah, I'm shocked that didn't have linter issues.

go.mod
treid314: I think it got pulled by goland or via a mod tidy.

concurrency/buffer.go
```diff
@@ -5,18 +5,21 @@ import (
 	"sync"
 )
 
+// SyncBuffer is a struct that contains a buffer and a mutex lock to allow for synchronous access to the contained buffer.
```
pstibrany, line 8: :) That describes implementation more than intent. What about...

```suggestion
// SyncBuffer is an io.Writer implementation with atomic writes. It keeps data only in the memory.
```

concurrency/buffer.go
```diff
@@ -5,18 +5,21 @@ import (
...
 type SyncBuffer struct {
 	mu  sync.Mutex
 	buf bytes.Buffer
 }
 
+// Write will write bytes to SyncBuffer within a mutex lock.
```
pstibrany, line 14: This comment duplicates code. I'd suggest to only keep comment on `SyncBuffer` itself, and remove the comments on methods.

pstibrany (approved):
Thank you.

PR #31, opened 2021-09-02
Add util/test package

treid314:
**What this PR does**: As part of moving the ring package we have to bring over a handful of decencies. This is one of them, the test package has poll util that evaluates condition until we either timeout, or it succeeds.

**Checklist**
- [-] Tests updated


test/poll.go
```diff
@@ -0,0 +1,26 @@
+package test
+
+import (
+	"reflect"
+	"testing"
+	"time"
+)
+
+// Poll repeatedly evaluates condition until we either timeout, or it succeeds.
```
pstibrany, line 9: This comment is not correct. `Poll` doesn't "evaluate condition", but calls a function and waits until it returns expected value, or times out.

pstibrany (approved):

PR #33, opened 2021-09-09
Feature: Add dslog and tenant packages

aknuds1:
**What this PR does**:
Add dslog and tenant packages, based on respectively pkg/util/log and pkg/tenant from Cortex.

I decided on the name "dslog" in order to disambiguate versus `go-kit/kit/log`.

**Checklist**
- [x] Tests updated


CHANGELOG.md
```diff
@@ -2,4 +2,6 @@
 
 * [CHANGE] Removed global metrics for KV package. Making a KV object will now require a prometheus registerer that will
   be used to register all relevant KV class metrics. #22
-* [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
\ No newline at end of file
+* [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
+* [ENHANCEMENT] Add `dslog` package #
```
treid314, line 6: I kind of like the look of this, do you think we should do this on all of these changes going forward? We could also add a new tag like [Addition] or something if we'd like.

aknuds1: @treid314 I'm actually not 100% sure, I just thought it was the right thing to do since @pracucci asked us to keep the changelog up to date.

treid314: I've mostly been using the changelog to track any constructor or method input changes, there just haven't really been any for a bit. I think it's great to track adding modules as well to show us whats in here.

treid314 (approved):
LGTM!

pracucci:
Who else uses the tenant stuff, other than Cortex?

aknuds1:
@pracucci at least Loki does, haven't checked Tempo.

aknuds1:
Update: Tempo doesn't use `tenant`.

aknuds1:
Pushed an update reducing the tenant package API to the bare minimum.

aknuds1:
Closing after discussing with @pstibrany; he thinks these packages might as well be copied into Loki and I agree.

PR #34, opened 2021-09-09
add util/http

treid314:
**What this PR does**: This PR adds the util/http files from cortex, it's required as part of moving the ring. The http module contains some helper methods for sending data via http in a number of different formats (HTML, JSON, Yaml, Text, and Protobuf)

**Checklist**
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,4 @@
 * [CHANGE] Removed global metrics for KV package. Making a KV object will now require a prometheus registerer that will
   be used to register all relevant KV class metrics. #22
 * [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
+* [ENHANCEMENT] Add 'pkg/util/http' http utilities from cortex #34
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,5 +3,5 @@
 * [CHANGE] Removed global metrics for KV package. Making a KV object will now require a prometheus registerer that will
   be used to register all relevant KV class metrics. #22
 * [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
-* [ENHANCEMENT] Add 'pkg/util/http' http utilities from cortex #34
+* [ENHANCEMENT] Add http package from cortex #34
 * [ENHANCEMENT] Add middleware package. #38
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,4 +2,5 @@
 
 * [CHANGE] Removed global metrics for KV package. Making a KV object will now require a prometheus registerer that will
   be used to register all relevant KV class metrics. #22
-* [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
\ No newline at end of file
+* [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
+* [ENHANCEMENT] Add 'pkg/util/http' http utilities from cortex #34
```
aknuds1, line 6: Wondering if it might be better to refer to the package name it gets in dskit, as opposed to the one in Cortex (pkg/util/http) to avoid confusion on part of the reader?

http/http_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"net/http/httptest"
+	"strconv"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
+
+	dskitHttp "github.com/grafana/dskit/http"
```
aknuds1, line 19: The norm with import aliases is to write them in all lowercase I believe.
```suggestion
	dskithttp "github.com/grafana/dskit/http"
```

pstibrany:
I'm not convinced we should make dskit a kitchen sink with all sorts of tiny util packages. Perhaps we can copy single function used by ring into the ring package instead, while moving ring to dskit?

aknuds1 (commented):
I see that `ParseProtoReader` is used by Loki, maybe we can port only that as public API?

http/http_test.go
treid314: Ah goland offers a refactor to camel case, so I usually take that short route.

pstibrany:
> I see that `ParseProtoReader` is used by Loki, maybe we can port only that as public API? Cfr. @pstibrany's comment.
> 
> Also @bboreham requested `ParseProtoReader` be part of dskit.

What about we move `ParseProtoReader` into `proto` package, and leave the rest out of dskit?

aknuds1:
> > I see that `ParseProtoReader` is used by Loki, maybe we can port only that as public API? Cfr. @pstibrany's comment.
> > Also @bboreham requested `ParseProtoReader` be part of dskit.
> 
> What about we move `ParseProtoReader` into `proto` package, and leave the rest out of dskit?

I think @pstibrany's suggestion sounds good.

treid314:
I ended up putting the protoreader into a proto package as suggested. I also pulled the required http functions into the ring package.

PR #35, opened 2021-09-09
add util/strings

treid314:
**What this PR does**: This PR ports util/strings from cortex. It's two utility functions to look for a string in a collection of strings and a function to make a map from a collection of strings. This is used by the ring package.

**Checklist**
- [-] Tests updated


pstibrany:
> This is used by the ring package.

Should we make a private copy of these functions in the ring package. Having them in dskit seems like overkill.

56quarters:
> > This is used by the ring package.
> 
> Should we make a private copy of these functions in the ring package? Having them publicly-exposed in a new package in dskit seems like overkill.

I agree. They seem pretty trivial to include publicly.

aknuds1 (commented):
@pstibrany might be right in that this functionality is too trivial for its own package (and to be part of public API).

treid314:
Yeah these are both pretty trivial functions, I don't disagree this is overkill. But, moving the files like this lets us keep the history and gives us a single location for these functions in the event of a bug fix or enhancement for cortex, this is my thought process with these PR's of util/*. I'm not opposed to just porting the functions, but I tend to favor keeping the history and everything in one location.

56quarters:
You could push another commit to this branch that moves them into the ring package and makes them private. That would still preserve the rest of the history. I'm not sure how many bugfixes these are going to be getting from Cortex since they're only a few lines of code.

treid314:
> I'm not sure how many bugfixes these are going to be getting from Cortex since they're only a few lines of code.

Yeah it's a little bit more of a philosophical argument than a practical one here since this is basically check all the strings in a loop. I'm not opposed to make it private but these functions are used in multiple spots in cortex, I guess we could keep both, but it's just something to be aware of.

pstibrany:
> Yeah these are both pretty trivial functions, I don't disagree this is overkill. But, moving the files like this lets us keep the history and gives us a single location for these functions in the event of a bug fix or enhancement for cortex, this is my thought process with these PR's of util/*. I'm not opposed to just porting the functions, but I tend to favor keeping the history and everything in one location.

Let's keep this out of public API, and focus on moving code useful for building "distributed systems" into dskit. If Ring code uses them, feel free to make a copy in the Ring PR, but let's keep the functions private.

pracucci:
> Let's keep this out of public API, and focus on moving code useful for building "distributed systems" into dskit. If Ring code uses them, feel free to make a copy in the Ring PR, but let's keep the functions private.

I agree with Peter.

treid314:
Closing this PR in favor of copying the required functions to the ring package

PR #36, opened 2021-09-09
Adding util/shard and util/yolobuf

treid314:
**What this PR does**: This add util/shard and util/yolobuf from cortex. Both of these are required for moving the ring package. Shard is utilities to handle shuffle sharding and yolo buf is an unsafe buffer with a safely great name.

**Checklist**
- [-] Tests updated

### DIFF
--- a/shard/shard.go
+++ b/shard/shard.go
@@ -1,13 +1,15 @@
-package util
+package shard
 
 import (
 	"crypto/md5"
 	"encoding/binary"
 	"math"
+
+	"github.com/grafana/dskit/yolo"
 )
 
+// Sharding Strategies.
 const (
-	// Sharding strategies.
 	ShardingStrategyDefault = "default"
 	ShardingStrategyShuffle = "shuffle-sharding"
 )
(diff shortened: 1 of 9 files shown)


shard/shard.go
```diff
@@ -0,0 +1,47 @@
...
+	ShardingStrategyShuffle = "shuffle-sharding"
+)
+
+var (
+	seedSeparator = []byte{0}
+)
+
+// ShuffleShardSeed returns seed for random number generator, computed from provided identifier.
+func ShuffleShardSeed(identifier, zone string) int64 {
+	// Use the identifier to compute an hash we'll use to seed the random.
```
aknuds1, line 23: Could we fix this typo?
```suggestion
	// Use the identifier to compute a hash we'll use to seed the random.
```

aknuds1 (approved):
LGTM, but I guess @pstibrany or @pracucci should have a look before merging.

pstibrany:
Let's please both `ShuffleShardSeed` and `yoloString` into the ring code. It's fine to export `ShuffleShardSeed`, but let's keep `yoloString` private. We don't want to encourage its use from dskit.

pracucci:
> Let's please move both `ShuffleShardSeed` and `yoloString` into the ring code. It's fine to export `ShuffleShardSeed`, but let's keep `yoloString` private. We don't want to encourage its use from dskit.

I agree with Peter.

shard/shard.go
```diff
@@ -0,0 +1,47 @@
...
+	"encoding/binary"
+	"math"
+
+	"github.com/grafana/dskit/yolo"
+)
+
+// Sharding Strategies.
+const (
+	ShardingStrategyDefault = "default"
+	ShardingStrategyShuffle = "shuffle-sharding"
```
pracucci, line 14: I think these should not be part of dskit.

treid314:
Closed in favor of moving these functions into the ring code

PR #39, opened 2021-09-10
Feature: Add gRPC related packages

aknuds1:
**What this PR does**:
Add pkg/util/grpc* from Cortex as grpcclient, grpcencoding and grpcutil. Also rename `kv/kvtls` to `crypto/tls` since it's used from more places.

**Checklist**
- [ ] Tests updated


grpcclient/grpcclient.go
```diff
@@ -0,0 +1,104 @@
...
+		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{NewRateLimiter(cfg)}, unaryClientInterceptors...)
+	}
+
+	return append(
+		opts,
+		grpc.WithDefaultCallOptions(cfg.CallOptions()...),
+		grpc.WithUnaryInterceptor(grpc_middleware.ChainUnaryClient(unaryClientInterceptors...)),
+		grpc.WithStreamInterceptor(grpc_middleware.ChainStreamClient(streamClientInterceptors...)),
+		grpc.WithKeepaliveParams(keepalive.ClientParameters{
+			Time:                time.Second * 20,
```
pstibrany, line 99: These very low ping times proved problematic, because they require correct server-side settings too. I think we should at least document that. See https://github.com/cortexproject/cortex/issues/3606#issuecomment-745179803 for details.

grpcutil/naming/naming.go
```diff
@@ -0,0 +1,43 @@
+package naming
```
pstibrany, line 1: We don't use this code in grpc context. Should we move it into top-level naming package instead?

grpcutil/naming/dns_resolver.go
```diff
@@ -0,0 +1,277 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"time"
+
+	"google.golang.org/grpc/grpclog"
```
pstibrany, line 13: If this becomes part of dskit, we should switch to go-kit logging.

grpcutil/carrier.go
```diff
@@ -0,0 +1,40 @@
+package grpcutil
```
pstibrany, line 1: This file would be more appropriately put into "httpgrpc" package. It has nothing with "gRPC" itself, but works with HTTP requests encoded in protobuf messages (ie. "httpgrpc" package from weaveworks/commons). (I have doubts if this should be in dskit)

aknuds1:
Thanks @pstibrany - will look into your feedback.

grpcutil/naming/dns_resolver.go
aknuds1: @pstibrany done.

grpcutil/naming/naming.go
aknuds1: @pstibrany done.

grpcutil/carrier.go
aknuds1: @pstibrany WDYT, should this file be removed from the PR?

pstibrany: > WDYT, should this file be removed from the PR?

Yes, I don't think it belongs to dskit.

aknuds1: Removed it.

grpcclient/grpcclient.go
aknuds1: @pstibrany I made these into parameters `PingTime` and `PingTimeout`. I didn't give them any default values, thinking that maybe it's best to use the `gRPC` defaults and let applications choose their own. Does that make sense to you, or should we use respectively 20 and 10 seconds for defaults?

aknuds1: Thanks @pstibrany, let's see what Marco says.

pracucci: My take on this topic:
1. This is quite a deep thing and I wouldn't let it configurable because will probably just pollute the config more than we do right now
2. Doing logic changes while porting features from Cortex to dskit makes review way more difficult. Even if we'll decide to do changes, these should be done in a separate PR in my opinion.
3. Even if we add it to the config, I would use the same naming used by grpc (eg. keepalive instead of ping, ...)

grpcutil/healthcheck/health_check.go
```diff
@@ -0,0 +1,53 @@
+package healthcheck
```
pracucci, line 1: [nit] Do we need this subpackage? I'm wondering if moving it to `grpcutil` is enough.

kv/memberlist/memberlist_client.go
```diff
@@ -135,6 +135,10 @@ type KVConfig struct {
 	DeadNodeReclaimTime time.Duration `yaml:"dead_node_reclaim_time"`
 	EnableCompression   bool          `yaml:"compression_enabled"`
 
+	// ip:port to advertise other cluster members. Used for NAT traversal
```
pracucci, line 138: Could you rebase please? These changes should go away from this PR.

naming/dns_resolver.go
```diff
@@ -0,0 +1,284 @@
+package naming
```
pracucci, line 1: I would move this to `grpcutil` package too.

pracucci (commented):
Good job! I compared the diff with Cortex and changes LGTM. I left a couple of minor comments. A part from this, I would suggest to rollback changes to `PingTime` and `PingTimeout` config. Whatever decision we'll take, I believe such change should be done in a separate PR and not buried in this "import PR".

kv/memberlist/memberlist_client.go
aknuds1: @pracucci will do.

grpcutil/healthcheck/health_check.go
aknuds1: @pracucci let me look into it.

PR #45, opened 2021-09-23
Move the ring package from cortex to dskit

treid314:
**What this PR does**: Adds the ring package from cortex to dskit. Also included is the proto package with ParseProtoReader function.

**Checklist**
- [X] Tests updated


treid314:
There's still a couple of things that need tidying up regarding package visibility and some other odds and ends. So I'm leaving this as a draft for the time being.

ring/lifecycler.go
```diff
@@ -0,0 +1,885 @@
...
+	f.DurationVar(&cfg.JoinAfter, prefix+"join-after", 0*time.Second, "Period to wait for a claim from another member; will join automatically after this.")
+	f.DurationVar(&cfg.ObservePeriod, prefix+"observe-period", 0*time.Second, "Observe tokens after generating to resolve collisions. Useful when using gossiping ring.")
+	f.DurationVar(&cfg.MinReadyDuration, prefix+"min-ready-duration", 15*time.Second, "Minimum duration to wait after the internal readiness checks have passed but before succeeding the readiness endpoint. This is used to slowdown deployment controllers (eg. Kubernetes) after an instance is ready and before they proceed with a rolling update, to give the rest of the cluster instances enough time to receive ring updates.")
+	f.DurationVar(&cfg.FinalSleep, prefix+"final-sleep", 30*time.Second, "Duration to sleep for before exiting, to ensure metrics are scraped.")
+	f.StringVar(&cfg.TokensFilePath, prefix+"tokens-file-path", "", "File path where tokens are stored. If empty, tokens are not stored at shutdown and restored at startup.")
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		level.Error(log.NewNopLogger()).Log("msg", "failed to get hostname", "err", err)
```
treid314, line 97: This was originally using the (util_logger) global logger, but now we don't wish to do that. I could add a logger to the config, but I'm not sure if that's the best path here. Right now I just create a new no-op logger which preserves the line, but doesn't log. Is there a system logger or a generic logger I could use in go? Or could I return an error and have that be uncaught since we send an os exit after this log?

56quarters: I think you're right, we definitely don't want `os.Exit` in library code. Returning an error is the right choice here.

56quarters: On the other hand, a panic feels fine in the case we can't look up the hostname of the machine this code is one. That seems like something that Should Never Fail.

ring/shard/yolo.go
```diff
@@ -0,0 +1,8 @@
+package shard
+
+import "unsafe"
+
+// Buf will return an unsafe pointer to a string, as the name yolo.yoloBuf implies use at your own risk.
+func yoloBuf(s string) []byte {
```
pracucci, line 6: You can move this to `shard.go` and remove this file.

ring/testutils/testutils.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+
+	"github.com/grafana/dskit/kv"
+	"github.com/grafana/dskit/ring"
+)
+
+// NumTokens determines the number of tokens owned by the specified
+// address
+func NumTokens(c kv.Client, name, ringKey string) int {
```
pracucci, line 15: Who is using it? I think none is using it in dskit. If confirmed, then I would not have it in dskit. Specific projects needing it can just keep their own copy.

proto/http.go
```diff
@@ -0,0 +1,158 @@
...
+// Values for CompressionType
+const (
+	NoCompression CompressionType = iota
+	RawSnappy
+)
+
+const messageSizeLargerErrFmt = "received message larger than max (%d vs %d)"
+
+// ParseProtoReader parses a compressed proto from an io.Reader.
+func ParseProtoReader(ctx context.Context, reader io.Reader, expectedSize, maxSize int, req proto.Message, compression CompressionType) error {
```
pracucci, line 28: Are functions in this file related to moving ring to dskit? I don't think so. If not, I would remove them from this PR and discuss them separately, so we can get the ring merge merged sooner.

pracucci (commented):
You did a good job! I diffed all files with their Cortex counterparts and left few comments here and there. I didn't comment on metrics with `cortex_` prefix because I would suggest to get this PR merged as soon as possible (with the least changes) and then work on removing the `cortex_` prefix in a follow up PR.

However, we need a CI step to make sure compiled protos are updated (see `check-protos` target in Cortex Makefile for reference).

ring/util.go
```diff
@@ -0,0 +1,222 @@
...
+					return ipAddr
+				}
+			}
+		}
+	}
+	return ipAddr
+}
+
+// StringsContain returns true if the search value is within the list of input values.
+func StringsContain(values []string, search string) bool {
```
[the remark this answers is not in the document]
treid314: Since we use StringsContain in client/pool.go removeStaleClients we have to expose it here. I can see if it helps to move it to a folder within the package.

ring/http.go
```diff
@@ -0,0 +1,216 @@
...
+		return ringDesc, true, nil
+	}
+	return r.KVClient.CAS(ctx, r.key, unregister)
+}
+
+func (r *Ring) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.Method == http.MethodPost {
+		ingesterID := req.FormValue("forget")
+		if err := r.forget(req.Context(), ingesterID); err != nil {
+			level.Error(r.logger).Log("msg", "error forgetting instance", "err", err)
```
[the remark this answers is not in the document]
treid314: This is waiting on dslog/wrappers in #42  https://github.com/grafana/dskit/pull/42/files#diff-81c5fa15cb0fdbd5f36ef1f9711920c590bb2bacca88040102dfda58e80d2a48R32

ring/lifecycler.go
treid314: I took @56quarters suggestion and used a panic here.

pracucci (approved):
LGTM, let's merge it! We can add the `check-protos` linter in a follow up PR (but please do it 🙏 )

PR #46, opened 2021-09-29
remove cortex_ prefix

treid314:
Signed-off-by: Tyler Reid <tyler.reid@grafana.com>

**What this PR does**: This PR removes the cortex_ prefix from metrics in the ring package. 

Fixes #<issue number>

**Checklist**
- [-] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,5 +5,6 @@
   * `-memberlist.advertise_port`
 * [CHANGE] Removed global metrics for KV package. Making a KV object will now require a prometheus registerer that will be used to register all relevant KV class metrics. #22
 * [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
+* [CHANGE] Remove `cortex_` prefix for metrics registered in the ring #46
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,6 @@
   * `-memberlist.advertise_port`
 * [CHANGE] Removed global metrics for KV package. Making a KV object will now require a prometheus registerer that will be used to register all relevant KV class metrics. #22
 * [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
-* [CHANGE] Remove `cortex_` prefix for metrics registered in the ring #46
+* [CHANGE] Remove `cortex_` prefix for metrics registered in the ring. #46
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -25,19 +25,19 @@ import (
 
 var (
 	consulHeartbeats = promauto.NewCounterVec(prometheus.CounterOpts{
-		Name: "cortex_member_consul_heartbeats_total",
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,5 +5,6 @@
   * `-memberlist.advertise_port`
 * [CHANGE] Removed global metrics for KV package. Making a KV object will now require a prometheus registerer that will be used to register all relevant KV class metrics. #22
 * [CHANGE] Added CHANGELOG.md and Pull Request template to reference the changelog
+* [CHANGE] Remove `cortex_` prefix for metrics registered in the ring #46
```
pstibrany, line 8:
```suggestion
* [CHANGE] Remove `cortex_` prefix for metrics registered in the ring. #46
```

pstibrany (approved):
Looks good, thank you.

ring/lifecycler.go
```diff
@@ -25,19 +25,19 @@ import (
 
 var (
 	consulHeartbeats = promauto.NewCounterVec(prometheus.CounterOpts{
-		Name: "cortex_member_consul_heartbeats_total",
+		Name: "member_consul_heartbeats_total",
```
pracucci, line 28: These metrics are registered globally (BTW, these are the type of metrics we don't want in dskit). Removing the prefix from here, it means when we'll update dskit in Cortex, these metrics will not have the `cortex_` prefix anymore and there will be no way to add it back.

ring/ring.go
```diff
@@ -237,31 +237,31 @@ func NewWithStoreClientAndStrategy(cfg Config, name, key string, store kv.Client
 		ringDesc:             &Desc{},
 		shuffledSubringCache: map[subringCacheKey]*Ring{},
 		memberOwnershipDesc: prometheus.NewDesc(
-			"cortex_ring_member_ownership_percent",
+			"ring_member_ownership_percent",
```
pracucci, line 240: These metrics are exposed by `ring.Collect()`. How are we going to add the prefix back in Cortex? What's the plan?

56quarters: Seems like this could be handled with something like `prometheus.WrapRegistererWithPrefix("cortex_", reg)` before calling `reg.MustRegister(ring)`. Am I missing something or are you specifically asking if there's an open issue / design for doing that?

56quarters: Chatted offline with @treid314, I now understand why this is not as easy as it seems.

treid314: There is no promauto.With(registerer).NewDisc, should we move to a Summary here? We already have the const labels for these metrics on register time.

pracucci: I think an option is to remove the `Describe()` and `Collect()` functions from the `Ring` and move all these metrics to `promauto.With(reg).NewGaugeFunc()`, setting the name label as a const label (whose value is is the ring `name`). Compared to `NewGauge()`, the `NewGaugeFunc()` takes a function which is called each time Prometheus client needs to collect the metric value. The current logic inside `Collect()` can be moved to separate functions like `getMemberOwnershipMetric() float64`, `getNumMembersMetric() float64`, and so on.

pracucci: _Tyler if the suggestion is not clear, please reach out and I can provide some more guidance._

treid314: Talked with @pracucci offline, there's not really a good way to use GaugeFuncs here since we have a constant and variable label for these metrics. I'm going to look into using a ticker to update GaugeVecs at a specified interval for the ring metrics.

PR #47, opened 2021-09-30
Add proto related tasks to the makefile, update the proto files, and add a drone step for check-protos

treid314:
Signed-off-by: Tyler Reid <tyler.reid@grafana.com>

**What this PR does**: This PR adds a check proto step for drone to run on PRs, we also re-generated the proto files. 

**Checklist**
- [-] Tests updated


.drone/drone.yml
```diff
@@ -23,7 +23,16 @@
          ],
          "image": "golang:1.16.7",
          "name": "make-test"
-      }
+      },
+     {
+       "commands": [
+          "apt-get update && apt-get -y install protobuf-compiler",
```
56quarters, line 29: It's not clear what version of the compiler and protoc-gen-gogoslick is being installed here. This seems like it would create issues when the version installed here doesn't match what a developer compiled with locally. Should dskit have its own build image ala Cortex?

treid314: Yeah I was trying to figure out if it was worth a build image for this as well. I'd prefer not to make a build image if we can avoid it. I think you can specify a version for at least protoc-gen-gogoslick.

treid314: We could also download the protobuf-c binary to ./tools and use that rather than apt-get here

56quarters: > We could also download the protoc binary to ./tools and use that rather than apt-get here

That seems reasonable to me since it would ensure that people were running the same version locally as CI but I'd be curious to hear what other people think

treid314: I took a quick run at trying that method, it seems to work with drone and locally. We can also set versions that way.

treid314:
Currently as of today (10/4) we're matching the versions of protogen and protoc we use for cortex to not update the proto files in this repo. As a future task we may want to update the protogen/protoc versions for cortex.

.drone/drone.yml
treid314: I tried to build this to match what we use in cortex today to prevent a potential versioning issue with our proto-files. Is there a way to make sure these versions are what get vendored for local builds? We seem to rely on our build image for cortex for maintaining the versions.

Makefile
```diff
@@ -1,5 +1,19 @@
 # put tools at the root of the folder
 PATH := $(CURDIR)/.tools/bin:$(PATH)
+# We don't want find to scan inside a bunch of directories, to accelerate the
+DONT_FIND := -name vendor -prune -o -name .git -prune -o -name .cache -prune -o -name .tools -prune -o
+# Generating proto code is automated.
+PROTO_DEFS := $(shell find . $(DONT_FIND) -type f -name '*.proto' -print)
+PROTO_GOS := $(patsubst %.proto,%.pb.go,$(PROTO_DEFS))
+# Download the proper protoc version for Darwin (osx) and Linux. If you need windows for some reason it's at
```
pracucci, line 8: Either we mention where it is or remove it ;)

```suggestion
# Download the proper protoc version for Darwin (osx) and Linux.
```

Makefile
```diff
@@ -1,5 +1,19 @@
 # put tools at the root of the folder
 PATH := $(CURDIR)/.tools/bin:$(PATH)
+# We don't want find to scan inside a bunch of directories, to accelerate the
```
pracucci, line 3: "To accelerate the"... I think it's missing the last part of the sentence.

Makefile
```diff
@@ -34,6 +48,17 @@ mod-check:
 	GO111MODULE=on go mod tidy
 	@git diff --exit-code -- go.sum go.mod
 
+clean-protos:
```
pracucci, line 51: These new targets should all be `.PHONY` (except for `%.pb.go`).

.drone/drone.yml
```diff
@@ -1,30 +1,43 @@
...
+      "name": "make-test"
+    },
+    {
+      "commands": [
+          "apt-get update && apt-get -y install unzip",
+          "make .tools/bin/protoc",
+          "go mod vendor",
+          "go get github.com/gogo/protobuf/protoc-gen-gogoslick@v1.3.0",
+          "go get github.com/golang/protobuf/protoc-gen-go@v1.3.1",
+          "go get github.com/gogo/protobuf/gogoproto@v1.3.0",
```
pracucci, line 34: Not a big fan of doing it here, cause then `make protos` wouldn't work out of the box on the CLI. For example, look what happens if I run the following:

```
$ make .tools/bin/protoc
$ make protos

.tools/protoc/bin/protoc -I /Users/marco/workspace:./vendor/github.com/gogo/protobuf:./vendor:./kv/memberlist --gogoslick_out=plugins=grpc,Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,:./kv/memberlist ./kv/memberlist/kv.proto
protoc-gen-gogoslick: program not found or is not executable
--gogoslick_out: protoc-gen-gogoslick: Plugin failed with status code 1.
make: *** [kv/memberlist/kv.pb.go] Error 1
```

What if we move these as a pre-requiste target for `protos`?

pracucci (approved):
LGTM!

PR #50, opened 2021-10-05
Remove ring global metrics

treid314:
**What this PR does**: This is a refactor of the global metrics in Lifecycler and Ring in the Ring package. Since this is a shared base library we want to remove all global metrics and allow uses to decide how they want to register metrics. This also allows us to use these metrics in a similar fashion for cortex.

**Checklist**
- [-] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/grafana/dskit/kv"
 	shardUtil "github.com/grafana/dskit/ring/shard"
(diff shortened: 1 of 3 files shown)


ring/ring.go
```diff
@@ -563,25 +565,15 @@ func (r *Ring) countTokens() (map[string]uint32, map[string]uint32) {
...
-			id,
-		)
-		ch <- prometheus.MustNewConstMetric(
-			r.numTokensDesc,
-			prometheus.GaugeValue,
-			float64(numTokens[id]),
-			id,
-		)
+		r.memberOwnershipGaugeVec.WithLabelValues(id).Set(float64(totalOwned) / float64(math.MaxUint32))
+		r.numMembersGaugeVec.WithLabelValues(id).Set(float64(numTokens[id]))
```
pracucci, line 576: This is not correct. It was `numTokensDesc` before.

PR #56, opened 2021-10-07
grpcclient: Make keepalive time and timeout into config parameters

aknuds1:
**What this PR does**:
In `grpcclient` package, make ping time and timeout into configuration parameters so each app can control them instead of having to use hardcoded values. Also add a test.

**Checklist**
- [x] Tests updated

### DIFF
--- a/crypto/tls/tls.go
+++ b/crypto/tls/tls.go
@@ -163,8 +163,8 @@ var WithInsecure = func() grpc.DialOption {
 }
 
 // GetGRPCDialOptions creates GRPC DialOptions for TLS.
-func (cfg *ClientConfig) GetGRPCDialOptions(enabled bool) ([]grpc.DialOption, error) {
-	if !enabled {
+func (cfg *ClientConfig) GetGRPCDialOptions(tlsEnabled bool) ([]grpc.DialOption, error) {
+	if !tlsEnabled {
 		return []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}, nil
 	}
 


grpcclient/grpcclient.go
```diff
@@ -91,13 +98,29 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
...
+		withStreamInterceptor(middleware.ChainStreamClient(streamClientInterceptors...)),
+		withKeepaliveParams(keepalive.ClientParameters{
+			Time:                time.Duration(cfg.KeepaliveTime) * time.Second,
+			Timeout:             time.Duration(cfg.KeepaliveTimeout) * time.Second,
 			PermitWithoutStream: true,
 		}),
 	), nil
 }
+
+var withDefaultCallOptions = func(cos ...grpc.CallOption) grpc.DialOption {
```
pstibrany, line 112: Instead of using global variables for these, can they be passed to `dialOption` as parameters? (Default publicly-exported `DialOption` would just use grpc-functions). This avoids needing setup and cleanup, and can be hidden into “fakeDialOption” method or similar.

grpcclient/grpcclient_test.go
```diff
@@ -0,0 +1,121 @@
...
+				Timeout:             20 * time.Second,
+				PermitWithoutStream: true,
+			},
+		},
+	}
+
+	opts, err := cfg.DialOption(nil, nil)
+	require.NoError(t, err)
+
+	assert.Empty(t, cmp.Diff(expOpts, opts))
```
pstibrany, line 120: assert.Equals will already print expected and real values if the check fails. Perhaps not as nice, but doesn’t require another dependency.

aknuds1: @pstibrany I can try `assert.Equals` again, but it wasn't working for me initially due to being too naïve (this is the benefit of go-cmp, that it's a more advanced comparer).

grpcclient/grpcclient.go
```diff
@@ -23,6 +23,15 @@ type Config struct {
 	GRPCCompression string  `yaml:"grpc_compression"`
 	RateLimit       float64 `yaml:"rate_limit"`
 	RateLimitBurst  int     `yaml:"rate_limit_burst"`
+	// KeepaliveTime is the number of seconds after which the client will ping the server in case of inactivity.
+	//
+	// See `google.golang.org/grpc/keepalive.ClientParameters.Time` for reference.
+	KeepaliveTime int64 `yaml:"ping_time"`
```
pracucci, line 29:
```suggestion
	KeepaliveTime int64 `yaml:"keepalive_time"`
```

grpcclient/grpcclient.go
```diff
@@ -23,6 +23,15 @@ type Config struct {
...
+	// KeepaliveTimeOut is the number of seconds the client waits after pinging the server, and if no activity is seen
+	// after that, the connection is closed.
+	//
+	// See `google.golang.org/grpc/keepalive.ClientParameters.Timeout` for reference.
+	KeepaliveTimeout int64 `yaml:"ping_timeout"`
```
pracucci, line 34:
```suggestion
	KeepaliveTimeout int64 `yaml:"keepalive_timeout"`
```

grpcclient/grpcclient.go
pracucci, line 29: Their CLI flags are not registered.

grpcclient/grpcclient_test.go
pstibrany: I won't block the PR on that, but will point out that expected outcome is no difference. Bringing in extra library that is only useful in case of test failure seems unnecessary.

pstibrany: Oh, I misunderstood the purpose then. I thought it's rich diff you're after, not just equality check.

aknuds1: @pstibrany yes exactly :) Rich diff is nice, but I'm only bringing it in here because it actually is needed.

pracucci:
I'm still a bit dubious this extra config is going to give a real value. Are we solving a real issue? If not, I've the feeling we're just making our config bigger.

aknuds1:
> I'm still a bit dubious this extra config is going to give a real value. Are we solving a real issue? If not, I've the feeling we're just making our config bigger.

@pracucci IIRC @pstibrany thought this config would make sense. Peter, do you still think we should parameterize these values?

pstibrany:
I think default client timeout values used by dskit are too aggressive, and incompatible with default values on gRPC server. We have seen many issues with them in multiple projects (Cortex, Loki). I think dskit should not use such a low defaults. Exposing config (not necessarily to user, but to projects using dskit) would give us option to change defaults in dskit to be more reasonable, while giving projects that have adjusted their gRPC server values an option to use lower values still.

aknuds1:
Peter's feedback sounds reasonable to me, although I lack the experience to tell whether the defaults are too aggressive or not.

pstibrany:
Change in Cortex: https://github.com/cortexproject/cortex/pull/4168, example user-reported issue: https://github.com/cortexproject/cortex/issues/3606
Change in Loki: https://github.com/grafana/loki/pull/4435, example user-reported issue: https://github.com/grafana/loki/issues/4375

All projects using dskit will need to adjust server-side settings (or modify dskit defaults, but right now they can't).

grpcclient/grpcclient.go
pracucci: I think this feedback hasn't been addressed, am I wrong?

pracucci (commented):
> I think default client timeout values used by dskit are too aggressive, and incompatible with default values on gRPC server. We have seen many issues with them in multiple projects (Cortex, Loki). I think dskit should not use such a low defaults. Exposing config (not necessarily to user, but to projects using dskit) would give us option to change defaults in dskit to be more reasonable, while giving projects that have adjusted their gRPC server values an option to use lower values still.

I see your point. I agree exposing such config to projects (but not users, at least not here). However, given the breaking change in the config (and the risk downstream projects miss it), I think it's safe if we break the `DialOption()` signature (and take it explicitly) so that it's a clear signal to downstream projects.

I also don't think such config should be in the `Config` struct.

pstibrany:
> I see your point. I agree exposing such config to projects (but not users, at least not here). However, given the breaking change in the config (and the risk downstream projects miss it), I think it's safe if we break the `DialOption()` signature (and take it explicitly) so that it's a clear signal to downstream projects.
> 
> I also don't think such config should be in the `Config` struct.

I agree that breaking API is safer to make sure clients notice the change and there is no need to expose it via `Config` struct.

grpcclient/grpcclient.go
aknuds1: @pracucci you're right! I was deferring addressing this until consensus on whether the PR should be closed or not.

aknuds1:
@pstibrany @pracucci As per your feedback, I moved the KeepAliveTime* parameters into `Config.DialOption` arguments. Also documented in the changelog that their defaults are the same as in the wrapped `google.golang.org/grpc/keepalive` library. Hope that looks good?

grpcclient/grpcclient.go
aknuds1: @pracucci resolving this as these config params are converted into method arguments now.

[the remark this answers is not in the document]
aknuds1: @pstibrany thanks for noticing this. I documented in the changelog and for `Config.DialOption` that the defaults are the same as for the underlying `google.golang.org/grpc/keepalive` package. Good enough?

grpcclient/grpcclient.go
aknuds1: @pracucci @pstibrany the drawback I see with stubbing these dependencies via arguments is that it adds complexity. The tests will no longer call `Config.DialOption` itself, and introduces the risk of having a bug in `Config.DialOption` where it passes the wrong dependencies. The test coverage will also be lower with this technique.

If you still prefer the injection via method arguments technique, I can change to that. Just want to clarify the problems I see with it.

grpcclient/grpcclient.go
```diff
@@ -72,14 +72,19 @@ func (cfg *Config) CallOptions() []grpc.CallOption {
...
-	tlsOpts, err := cfg.TLS.GetGRPCDialOptions(cfg.TLSEnabled)
+// DialOption returns the config as a slice of grpc.DialOptions.
+//
+// keepaliveTime is the number of seconds after which the client will ping the server in case of inactivity.
+// See `google.golang.org/grpc/keepalive.ClientParameters.Time` for reference.
+//
+// keepaliveTimeout is the number of seconds the client waits after pinging the server, and if no activity is
+// seen after that, the connection is closed. See `google.golang.org/grpc/keepalive.ClientParameters.Timeout`
+// for reference.
+func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor, keepaliveTime, keepaliveTimeout int64) ([]grpc.DialOption, error) {
```
pracucci, line 83: I think keepaliveTime and keepaliveTimeout should be `time.Duration`.

crypto/tls/tls.go
```diff
@@ -72,10 +72,17 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
 	return config, nil
 }
 
-// GetGRPCDialOptions creates GRPC DialOptions for TLS
+// WithInsecure wraps grpc.WithInsecure.
+//
+// Stubbable for tests.
+var WithInsecure = func() grpc.DialOption {
```
pracucci, line 78: Do we wanna to expose it? Looks weird to me.

crypto/tls/tls.go
```diff
@@ -72,10 +72,17 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
...
+	return grpc.WithInsecure()
+}
+
+// GetGRPCDialOptions creates GRPC DialOptions for TLS.
 func (cfg *ClientConfig) GetGRPCDialOptions(enabled bool) ([]grpc.DialOption, error) {
```
pracucci, line 83: [nit] Can we rename enabled to tlsEnabled to make it more clear?

crypto/tls/tls.go
aknuds1: It's technically necessary for grpcclient tests to be able to stub calling out to the `grpc` package. I'll look again into whether there's another way to do this (dependency injection?).

crypto/tls/tls.go
aknuds1: @pracucci done.

liguozhong:
great 👍

pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

PR #57, opened 2021-10-07
Drone: Sync Jsonnet and YAML

aknuds1:
**What this PR does**:
.drone/drone.yml is generated from .drone/drone.jsonnet, but they have got out of sync due to manual additions to .drone/drone.yml. This PR brings them back in sync.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -8,10 +8,10 @@ local pipeline = {
 };
 
 local step = {
-  make(target):: {
+  make(target, commands=[]):: {
     name: 'make-%s' % target,
     image: image,
-    commands: ['make %s' % target],
+    commands: commands + ['make %s' % target],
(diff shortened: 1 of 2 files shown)


pracucci (approved):
Good job! I approved the PR causing the out of sync and reason is that I wasn't aware of it. What do you think if we have a CI step to make sure the `drone.yml` matches the output of `drone.jsonnet`?

aknuds1:
> Good job! I approved the PR causing the out of sync and reason is that I wasn't aware of it. What do you think if we have a CI step to make sure the `drone.yml` matches the output of `drone.jsonnet`?

@pracucci I'm planning to add it :) @simonswine pointed me to a solution of his, but haven't had the time to check it yet. I think I already implemented this check for Grafana.

PR #64, opened 2021-10-14
We should update and start the ticker metrics before we return from an error.

treid314:
…

**What this PR does**: The previous code where we would return nil if the ring name was not yet in the kv store was causing a race condition where if the ring name was not in the kv store we would never start the ring metrics ticker and never update the ring metrics.

Fixes #<issue number>

**Checklist**
- [-] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -269,21 +269,8 @@ func NewWithStoreClientAndStrategy(cfg Config, name, key string, store kv.Client
 }
 
 func (r *Ring) starting(ctx context.Context) error {
-	// Get the initial ring state so that, as soon as the service will be running, the in-memory
-	// ring would be already populated and there's no race condition between when the service is
-	// running and the WatchKey() callback is called for the first time.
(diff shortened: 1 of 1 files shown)


ring/ring.go
```diff
@@ -269,21 +269,8 @@ func NewWithStoreClientAndStrategy(cfg Config, name, key string, store kv.Client
...
-		return nil
-	}
-
-	r.updateRingState(value.(*Desc))
+	// Update the ring metrics at start.
 	r.updateRingMetrics()
-
 	// Use this channel to close the go routine to prevent leaks.
 	r.metricsUpdateCloser = make(chan struct{})
 	go func() {
```
pracucci, line 276: If I remember correctly the service framework, if starting returns an error, the stopping won't be called. Could you double check it? If confirmed, then we should start the ticker at last, otherwise we'll leak the goroutine.

pstibrany: We can also start it in `loop`, and make `metricsUpdateCloser` scoped to `loop` only.

PR #66, opened 2021-10-19
Ring: Include unhealthy replicas in error message

aknuds1:
**What this PR does**:
Make the ring replication strategies include information on the unhealthy replicas in the error message, when an insufficient number of healthy replicas are found.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/replication_strategy.go
+++ b/ring/replication_strategy.go
@@ -2,11 +2,8 @@ package ring
 
 import (
 	"fmt"
+	"strings"
 	"time"
-
-	"github.com/go-kit/log"
-	"github.com/go-kit/log/level"
-	"github.com/pkg/errors"
 )
 
 type ReplicationStrategy interface {
@@ -16,12 +13,10 @@ type ReplicationStrategy interface {
 	Filter(instances []InstanceDesc, op Operation, replicationFactor int, heartbeatTimeout time.Duration, zoneAwarenessEnabled bool) (healthy []InstanceDesc, maxFailures int, err error)
 }
 
-type defaultReplicationStrategy struct {
-	logger log.Logger
-}
+type defaultReplicationStrategy struct{}
 
-func NewDefaultReplicationStrategy(logger log.Logger) ReplicationStrategy {
-	return &defaultReplicationStrategy{logger: logger}
+func NewDefaultReplicationStrategy() ReplicationStrategy {
+	return &defaultReplicationStrategy{}
 }
 
 // Filter decides, given the set of instances eligible for a key,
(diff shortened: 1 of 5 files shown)


ring/replication_strategy.go
```diff
@@ -65,25 +77,30 @@ func (s *defaultReplicationStrategy) Filter(instances []InstanceDesc, op Operati
...
-func NewIgnoreUnhealthyInstancesReplicationStrategy() ReplicationStrategy {
-	return &ignoreUnhealthyInstancesReplicationStrategy{}
+func NewIgnoreUnhealthyInstancesReplicationStrategy(logger log.Logger) ReplicationStrategy {
+	return &ignoreUnhealthyInstancesReplicationStrategy{logger: logger}
 }
 
 func (r *ignoreUnhealthyInstancesReplicationStrategy) Filter(instances []InstanceDesc, op Operation, _ int, heartbeatTimeout time.Duration, _ bool) (healthy []InstanceDesc, maxFailures int, err error) {
 	now := time.Now()
 	// Filter out unhealthy instances.
+	var skipped []string
```
pracucci, line 91: [nit] I would call it `unhealthy`.

ring/replication_strategy.go
```diff
@@ -65,25 +77,30 @@ func (s *defaultReplicationStrategy) Filter(instances []InstanceDesc, op Operati
...
 			i++
 		} else {
+			skipped = append(skipped, instances[i].Addr)
 			instances = append(instances[:i], instances[i+1:]...)
 		}
 	}
 
 	// We need at least 1 healthy instance no matter what is the replication factor set to.
 	if len(instances) == 0 {
+		level.Error(r.logger).Log("msg", "failed to find any healthy ring replicas", "unhealthy", skipped)
```
pracucci, line 103: The `Filter` function is used on the write path too and this function could be called at a very high rate. The risk of flooding logs because of this is high, so I'm not sure about it.

I would rather add the unhealthy replicas to the returned error, instead of logging it.

_Comment applies to `defaultReplicationStrategy` too (actually that one is used on write path, but I would keep the same approach here too)._

aknuds1: @pracucci did we see all these errors getting logged when we were investigating the ruler/ring failures? Guess we should make sure they do get logged for relevant scenarios. We can sync on this tomorrow.

pracucci: Yes, they're logged by ruler.

pracucci (approved):

ring/replication_strategy.go
```diff
@@ -74,17 +79,23 @@ func NewIgnoreUnhealthyInstancesReplicationStrategy() ReplicationStrategy {
 func (r *ignoreUnhealthyInstancesReplicationStrategy) Filter(instances []InstanceDesc, op Operation, _ int, heartbeatTimeout time.Duration, _ bool) (healthy []InstanceDesc, maxFailures int, err error) {
 	now := time.Now()
 	// Filter out unhealthy instances.
+	var unhealthy []string
```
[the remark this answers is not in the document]
aknuds1: @56quarters discussed with @pracucci - he agrees that the current algorithm with strings.Join should be fine (it's dubious that fine tuning for perf will give us much), using `strings.Builder` would require some rewriting with uncertain gain versus effort.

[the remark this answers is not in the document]
56quarters: My motivation was to reduce the amount of code we're responsible for more than performance, but that's fine :+1:

PR #67, opened 2021-10-20
Move ring metrics go routine and ticker to loop and remove close channel in favor of loop context

treid314:
**What this PR does**: This PR is a quick refactor of when we create the ring update ticker based on [the comment here](https://github.com/grafana/dskit/pull/64#discussion_r729863452). This better fits how we use tickers within services in the cortex codebase.

**Checklist**
- [-] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -263,7 +263,7 @@ func NewWithStoreClientAndStrategy(cfg Config, name, key string, store kv.Client
 		logger: logger,
 	}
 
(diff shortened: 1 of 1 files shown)


ring/ring.go
```diff
@@ -318,10 +314,6 @@ func (r *Ring) loop(ctx context.Context) error {
 }
 
 func (r *Ring) stopping(_ error) error {
```
pracucci, line 316: You can completely remove this function and pass `nil` instead of `r.stopping` to `services.NewBasicService()`.

treid314: Oh yeah, true!

PR #68, opened 2021-10-26
Improve gRPC error status label in metrics

danielblando:
**What this PR does**:
This is an attempt to continue a PR in cortex (https://github.com/cortexproject/cortex/pull/4519)
After a discussion(https://github.com/cortexproject/cortex/issues/4441), we noticed that client metrics sent from cortex could be improved.

The main purpose is to separate 5xx and 4xx error types in the metrics.

**Checklist**
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * [CHANGE] Remove `cortex_` prefix for metrics registered in the ring. #46
 * [CHANGE] Rename `kv/kvtls` to `crypto/tls`. #39
 * [CHANGE] spanlogger: Take interface implementation for extracting tenant ID. #59
-* [CHANGE] The `status_code` label on metrics from client has changed from '200' and '500' to '2xx', '5xx', '4xx', 'cancel' or 'error'. #68
+* [CHANGE] The `status_code` label on gRPC client metrics has changed from '200' and '500' to '2xx', '5xx', '4xx', 'cancel' or 'error'. #68
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
 * [ENHANCEMENT] Add limiter package. #41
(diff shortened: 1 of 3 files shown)


bboreham:
Currently grpc metrics have `success`, `cancel`, specific 4xx and 5xx codes and `error`.
`instrument.ErrorCode` is used by dozens of other histograms, and has just `200` and `500`.

Changing `instrument.ErrorCode` in line with this PR would break anyone who had alerts configured on `"200"` or `"500"`.
Also it is used in a lot of non-RPC contexts, so really you just get a different spelling of "success" and "error". Doesn't seem like a win.

I'm ok with changing the behaviour just for gRPC as in this PR, and would also be OK with doing it in `weaveworks/common`, since the set of values is already quite diverse and people must have written their rules to expect that. Maybe someone has a specific rule detecting 401 vs 404, say; they would have to update.

Grafana Labs already squashes 400, 401, etc., to 4xx in dashboards: https://github.com/grafana/jsonnet-libs/blob/460ba73ba5c736f2e60c023f4e0de556eaa4574c/grafana-builder/grafana.libsonnet#L386

danielblando:
Great, thanks for the input @bboreham 

Looking at the sugestion about moving to ```weaveworks/common```, it seems it was a point already raised from you on https://github.com/cortexproject/cortex/pull/4471#pullrequestreview-756172358 
I would suggest to go forward with this PR and follow up with moving the code to another package.

@pracucci are you ok with moving forwarding and approving this?
Could we migrate the code in another PR?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [CHANGE] Remove `cortex_` prefix for metrics registered in the ring. #46
 * [CHANGE] Rename `kv/kvtls` to `crypto/tls`. #39
 * [CHANGE] spanlogger: Take interface implementation for extracting tenant ID. #59
+* [CHANGE] The `status_code` label on metrics from client has changed from '200' and '500' to '2xx', '5xx', '4xx', 'cancel' or 'error'. #68
```
pracucci, line 11: [nit]
```suggestion
* [CHANGE] The `status_code` label on gRPC client metrics has changed from '200' and '500' to '2xx', '5xx', '4xx', 'cancel' or 'error'. #68
```

pracucci (approved):
Ok, let's do it.

danielblando:
Hey @pracucci,

Do we need another approve or could we merge the change?

bboreham (approved):

PR #69, opened 2021-11-01
Add info log when entering final sleep

trevorwhitney:
**What this PR does**:
Adds an info log indicating when the lifecycler is about to enter it's current sleep period.

Since the lifecycler defaults this values to `30s`, to someone using the default who may not be aware of this behavior, it can be confusing why on shutdown the process seems to be hanging. This provides a helpful log message indicating the final sleep is beginning, and for how long it will last.


56quarters (approved):
LGTM

PR #70, opened 2021-11-02
Remove the closer package since it's trivial

### LINKED ISSUE #51: 'closer' package is too small; remove it
It is four lines of code; should be copied into wherever it is needed.
"A little copying is better than a little dependency." - Go proverb.

56quarters:
The only uses found were the Consul mock kv client. Even in
Cortex/Loki/Tempo the only uses were the vendored version of
the same Consul mock kv client.

Fixes #51

**Checklist**
- [X] Tests updated


pstibrany (approved):
Thank you!

pracucci (approved):

PR #74, opened 2021-11-19
Chore: Replace all usages of ioutil

aknuds1:
**What this PR does**:
Since ioutil is [deprecated](https://golang.org/doc/go1.16#ioutil) since Go 1.16, replace all usages of it with preferred equivalents. Nothing important, but might as well clean up :)

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/lifecycler_test.go
+++ b/ring/lifecycler_test.go
@@ -3,7 +3,6 @@ package ring
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"sort"
 	"testing"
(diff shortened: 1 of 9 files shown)


ring/lifecycler_test.go
```diff
@@ -596,7 +595,7 @@ func TestTokensOnDisk(t *testing.T) {
 	require.NoError(t, services.StartAndAwaitRunning(context.Background(), r))
 	defer services.StopAndAwaitTerminated(context.Background(), r) //nolint:errcheck
 
-	tokenDir, err := ioutil.TempDir(os.TempDir(), "tokens_on_disk")
+	tokenDir, err := os.MkdirTemp("", "tokens_on_disk")
```
pstibrany, line 598: can we use `t.TempDir()`? No need to do manual deletion via `defer` then.

ring/tokens_test.go
```diff
@@ -60,7 +59,7 @@ func TestTokens_Equals(t *testing.T) {
 }
 
 func TestLoadTokensFromFile_ShouldGuaranteeSortedTokens(t *testing.T) {
-	tmpDir, err := ioutil.TempDir("", "test-tokens")
+	tmpDir, err := os.MkdirTemp("", "test-tokens")
```
pstibrany, line 62: Can we also replace this with `f.TempDir()`?

pstibrany (approved):
LGTM, thanks.

ring/lifecycler_test.go
aknuds1: @pstibrany good point, didn't think of it myself :)

ring/tokens_test.go
aknuds1: @pstibrany surely.

PR #78, opened 2021-11-25
ring: Use static types for ring descriptions in HTTP responses

aknuds1:
**What this PR does**:
This is just a tiny PR that aims to clean up ring/http.go a bit, and use static types instead of `interface{}` for data in HTTP responses. I think it's an improvement readability wise (and should reduce the risk of introducing bugs), but feedback welcome!

**Checklist**
- [ ] Tests updated


pracucci (approved):
👍

pstibrany (approved):

PR #79, opened 2021-11-26
ring.Lifecycler: Handle when previous ring state is leaving and the number of tokens has changed

### LINKED ISSUE #73: Ring lifecycler stuck when increasing tokens and instance is already in the ring
I've found a bug in the `ring.Lifecycler` which happens in the following scenario:
- Ingester is already registered in the ring with 64 tokens and `LEAVING` state
- Ingester is configured to store/load tokens from disk
- Ingester starts up with a tokens config change (from 64 to 128)

What we get is that ingester will never increase tokens to 128 and will never switch to ACTIVE in the ring.

_The issue should be easy to reproduce with a unit test._

aknuds1:
**What this PR does**:
When starting a `ring.Lifecycler`, handle when previous ingester state is found in ring, its state is leaving and the number of tokens has changed. Tests are included.

# Manual testing

I've tested manually, leaving ring member behind (in LEAVING state) on exit via `unregister_on_shutdown=false`, AFAICT it works as it should in all relevant test cases:

* Token count is unchanged
* Token count changes from 512 to 256
* Token count changes from 256 to 512

The ring member state always gets switched to ACTIVE.

Fixes #73

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -616,38 +616,43 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 			return ringDesc, true, nil
 		}
 
-		// If the ingester failed to clean its ring entry up it can leave its state in LEAVING
-		// OR unregister_on_shutdown=false
-		// Move it into ACTIVE to ensure the ingester joins the ring.
+		tokens := Tokens(instanceDesc.Tokens)
+
+		// If the ingester fails to clean its ring entry up or unregister_on_shutdown=false, it can leave behind its
+		// ring state as LEAVING. Make sure to switch to the ACTIVE state.
 		if instanceDesc.State == LEAVING {
-			tokens := Tokens(instanceDesc.Tokens)
-			setIsActive := true
 			if len(tokens) != i.cfg.NumTokens {
-				level.Debug(i.logger).Log("msg", "existing entry has different number of tokens", "existingTokens", len(tokens), "newTokens", i.cfg.NumTokens)
+				level.Debug(i.logger).Log("msg", "existing entry has different number of tokens",
+					"existing_tokens", len(tokens), "new_tokens", i.cfg.NumTokens)
 				if len(tokensFromFile) > 0 {
-					level.Debug(i.logger).Log("msg", "adding tokens from file", "tokens", len(tokensFromFile))
-					setIsActive = len(tokensFromFile) >= i.cfg.NumTokens
+					tokens = tokensFromFile
+					if len(tokens) < i.cfg.NumTokens {
+						needTokens := i.cfg.NumTokens - len(tokens)
+						newTokens := GenerateTokens(needTokens, ringDesc.GetTokens())
+						tokens = append(tokens, newTokens...)
+						sort.Sort(tokens)
+					}
+					level.Debug(i.logger).Log("msg", "adding tokens from file", "tokens", len(tokens))
 					tokens = tokensFromFile
 				} else if i.cfg.NumTokens > len(tokens) {
 					needTokens := i.cfg.NumTokens - len(tokens)
-					level.Debug(i.logger).Log("msg", "no tokens in file, generating new ones in addition to those of existing instance", "newTokens", needTokens)
+					level.Debug(i.logger).Log("msg",
+						"no tokens in file, generating new ones in addition to those of existing instance",
+						"new_tokens", needTokens)
 					newTokens := GenerateTokens(needTokens, ringDesc.GetTokens())
 					tokens = append(tokens, newTokens...)
 					sort.Sort(tokens)
 				} else {
-					level.Debug(i.logger).Log("msg", "no tokens in file, adopting a subset of existing instance's tokens", "numTokens", i.cfg.NumTokens)
+					level.Debug(i.logger).Log("msg", "no tokens in file, adopting a subset of existing instance's tokens",
+						"num_tokens", i.cfg.NumTokens)
 					tokens = tokens[0:i.cfg.NumTokens]
 				}
 			} else {
 				level.Debug(i.logger).Log("msg", "adopting tokens of existing instance")
 			}
 
-			if setIsActive {
-				level.Debug(i.logger).Log("msg", "switching state to active")
-				i.setState(ACTIVE)
-			} else {
-				level.Debug(i.logger).Log("msg", "not switching state to active", "state", i.GetState())
-			}
+			level.Debug(i.logger).Log("msg", "switching state to active")
+			i.setState(ACTIVE)
 
 			i.setTokens(tokens)
 			instanceDesc.State = i.GetState()
@@ -655,17 +660,12 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 		} else {
 			// We exist in the ring and not in leaving state, so assume the ring is right and copy tokens & state out of there.
 			i.setState(instanceDesc.State)
-			i.setTokens(instanceDesc.Tokens)
+			i.setTokens(tokens)
 		}
 
 		// We're taking over this entry, update instanceDesc with our values
 		instanceDesc.Addr = i.Addr
 		instanceDesc.Zone = i.Zone
-
-		// We exist in the ring, so assume the ring is right and copy out tokens & state out of there.
-		i.setState(instanceDesc.State)
-		tokens, _ := ringDesc.TokensFor(i.ID)
-		i.setTokens(tokens)
 
 		level.Info(i.logger).Log("msg", "existing entry found in ring", "state", i.GetState(), "tokens", len(tokens), "ring", i.RingName)
 
@@ -769,6 +769,7 @@ func (i *Lifecycler) autoJoin(ctx context.Context, targetState InstanceState) er
 		// At this point, we should not have any tokens, and we should be in PENDING state.
 		myTokens, takenTokens := ringDesc.TokensFor(i.ID)
 		if len(myTokens) > 0 {
+			// TODO: Should we log this case as an error, as we don't treat it as one?
 			level.Error(i.logger).Log("msg", "tokens already exist for this instance - wasn't expecting any!", "num_tokens", len(myTokens), "ring", i.RingName)
 		}
 
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -616,7 +616,7 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 			return ringDesc, true, nil
 		}
 
-		level.Info(i.logger).Log("msg", "existing entry found in ring", "state", i.GetState(), "tokens", len(instanceDesc.Tokens), "ring", i.RingName)
+		level.Info(i.logger).Log("msg", "existing entry found in ring", "state", instanceDesc.State, "tokens", len(instanceDesc.Tokens), "ring", i.RingName)
 
 		// If the ingester failed to clean its ring entry up it can leave its state in LEAVING
 		// OR unregister_on_shutdown=false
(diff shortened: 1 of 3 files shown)


ring/lifecycler.go
```diff
@@ -586,22 +586,48 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
-			instanceDesc.State = ACTIVE
-		}
+		if instanceDesc.State == LEAVING {
+			var tokens Tokens
+			isActive := true
+			if len(instanceDesc.Tokens) != i.cfg.NumTokens {
+				level.Debug(i.logger).Log("msg", "existing entry has different number of tokens", "existingTokens", len(instanceDesc.Tokens), "newTokens", i.cfg.NumTokens)
+				if len(tokensFromFile) > 0 {
+					level.Debug(i.logger).Log("msg", "adding tokens from file", "tokens", len(tokensFromFile))
+					isActive = len(tokensFromFile) >= i.cfg.NumTokens
```
krajorama, line 601: According to Owen the number of tokens may have gone down, wouldn't we leave the instance (~ingester) in LEAVING state?

If I understand correctly in case the instance wasn't in the ring, then this would lead to PENDING state (as that's where we start out from) and then transition to ACTIVE after a period of time.

krajorama:
I think in general the initRing function is way too long and complicated. We should either refactor / split it up or write some data driven testsuite to get full coverage of scenarios and code paths.
image

ring/lifecycler.go
aknuds1: @krajorama You could find confirmation (via a test) that the proposed does indeed not lead to leaving state when the number of tokens has gone down?

> So I've looked at a test with decreasing number of tokens and the state is ACTIVE, however the number of tokens is not actually reduced, remains 256, not sure if this is ok or not - at least does not get stuck

You mean you wrote a test for the scenario?

aknuds1:
> I think in general the initRing function is way too long and complicated. We should either refactor / split it up or write some data driven testsuite to get full coverage of scenarios and code paths.

@krajorama I tend to agree, but I'd at least like to receive confirmation that this PR is doing the right thing before I think about refactoring. Even then, I would probably suggest doing the refactoring in a dedicated PR (maybe to base this one on).

aknuds1:
@krajorama I added a test for the decreasing number of tokens scenario, and also code to handle it. I would need input from e.g. Marco, but I think it was a bug.

ring/lifecycler.go
krajorama: yes, and I see you've done the same. And I see you wait until the number goes down as well.

ring/lifecycler.go
```diff
@@ -585,22 +585,53 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 			return ringDesc, true, nil
 		}
 
-		// If the ingester failed to clean its ring entry up in can leave its state in LEAVING
+		level.Info(i.logger).Log("msg", "existing entry found in ring", "state", instanceDesc.State, "tokens", len(instanceDesc.Tokens), "ring", i.RingName)
+
+		// If the ingester failed to clean its ring entry up it can leave its state in LEAVING
 		// OR unregister_on_shutdown=false
 		// Move it into ACTIVE to ensure the ingester joins the ring.
```
krajorama, line 592: Comment does not match what we do anymore. There's an exception to the rule because of line 600. I wonder why we wouldn't just generate the necessary number of tokens and set state to active - which is what we do if there's no token file.

ring/lifecycler.go
```diff
@@ -585,22 +585,53 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
-		if instanceDesc.State == LEAVING && len(instanceDesc.Tokens) == i.cfg.NumTokens {
-			instanceDesc.State = ACTIVE
-		}
+		if instanceDesc.State == LEAVING {
+			var tokens Tokens
+			isActive := true
```
pstibrany, line 595:
```suggestion
			setActiveState := true
```

ring/lifecycler.go
```diff
@@ -585,22 +585,53 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
+			if len(instanceDesc.Tokens) != i.cfg.NumTokens {
+				level.Debug(i.logger).Log("msg", "existing entry has different number of tokens", "existingTokens", len(instanceDesc.Tokens), "newTokens", i.cfg.NumTokens)
+				if len(tokensFromFile) > 0 {
+					level.Debug(i.logger).Log("msg", "adding tokens from file", "tokens", len(tokensFromFile))
+					isActive = len(tokensFromFile) >= i.cfg.NumTokens
```
pstibrany, line 600: If there are tokens in the file, but less than `i.cfg.NumTokens` we don't set state to `ACTIVE`. This leaves entry in LEAVING state.

pstibrany:
> > I think in general the initRing function is way too long and complicated. We should either refactor / split it up or write some data driven testsuite to get full coverage of scenarios and code paths.
> 
> @krajorama I tend to agree, but I'd at least like to receive confirmation that this PR is doing the right thing before I think about refactoring. Even then, I would probably suggest doing the refactoring in a dedicated PR (maybe to base this one on).

I think we would prefer to deprecate and eventually remove `Lifecycler`, while migrating code to `BasicLifecycler` instead.

aknuds1:
> I think we would prefer to deprecate and eventually remove `Lifecycler`, while migrating code to `BasicLifecycler` instead.

@pstibrany is the fix in this PR still relevant, with that in mind?

DylanGuedes:
As we had a few internal problems this week, I'm interested in resurrecting this PR.

DylanGuedes:
@krajorama @pstibrany sorry for pinging you :smile: I recently resurrected this PR, rebased it, and fixed a few tests that started to break after the rebase. Do you mind having a look at it? We recently had an internal incident caused by this change in the number of tokens so I'm interested in fixing it.

ring/lifecycler.go
```diff
@@ -593,11 +593,46 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
-		if instanceDesc.State == LEAVING && len(instanceDesc.Tokens) == i.cfg.NumTokens {
-			instanceDesc.State = ACTIVE
+		if instanceDesc.State == LEAVING {
+			var tokens Tokens = instanceDesc.Tokens // way of forcing tokens to be of type Tokens instead of []uint32.
+			setIsActive := true
+			if len(instanceDesc.Tokens) != i.cfg.NumTokens {
+				level.Debug(i.logger).Log("msg", "existing entry has different number of tokens", "existingTokens", len(instanceDesc.Tokens), "newTokens", i.cfg.NumTokens)
+				if len(tokensFromFile) > 0 {
+					level.Debug(i.logger).Log("msg", "adding tokens from file", "tokens", len(tokensFromFile))
+					setIsActive = len(tokensFromFile) >= i.cfg.NumTokens
```
agardiman, line 606: Why does the instance become active only when the number of tokens in the file are at least i.cfg.NumTokens?
What if `len(tokensFromFile) < i.cfg.NumTokens`, shouldn't the needed tokens be generated and added to the existing ones like we do later, and then set it to `ACTIVE`? The same when `len(tokensFromFile) > i.cfg.NumTokens`, shouldn't the extra tokens be removed?

ring/lifecycler.go
```diff
@@ -593,11 +593,46 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
+					tokens = tokensFromFile
+				} else if i.cfg.NumTokens > len(instanceDesc.Tokens) {
+					needTokens := i.cfg.NumTokens - len(instanceDesc.Tokens)
+					level.Debug(i.logger).Log("msg", "no tokens in file, generating new ones in addition to those of existing instance", "newTokens", needTokens)
+					newTokens := GenerateTokens(needTokens, ringDesc.GetTokens())
+					tokens = append(instanceDesc.Tokens, newTokens...)
+					sort.Sort(tokens)
+				} else {
+					level.Debug(i.logger).Log("msg", "no tokens in file, adopting a subset of existing instance's tokens", "numTokens", i.cfg.NumTokens)
+					tokens = instanceDesc.Tokens[0:i.cfg.NumTokens]
```
agardiman, line 616: Are the tokens sorted? If so, I think that this selects the smallest N, basically increasing the change that only smaller hashes will be selected for this instance, and bigger ones never match. To still have an even chance of distribution of hashes associated to all ingesters, I think random tokens should be left out of the slice, so the tokens should first be reshuffled, then take the first `i.cfg.NumTokens` and before assigning them to the instance sort them again.

agardiman:
@DylanGuedes @pstibrany do you have any update on this by any chance?

CarolineCH:
Hi all, I was wondering if there was been on any movement on this PR, we've also experiencing this issue? Thanks.

DylanGuedes:
> @DylanGuedes @pstibrany do you have any update on this by any chance?

Tbh I kinda abandoned it because we never faced this issue again. 
@CarolineCH were you using the latest version of dskit there?

sarthaktyagi-505:
Hi @DylanGuedes , I reproduced this issue today again, so i can confirm its still a bug. If you increase the number of tokens on ingesters and trigger a rollout, the ingesters in that zone are stuck in LEAVING state, blocking the entire ingestion of the pipeline, I have posted some comments on your PR, please have a look it will be good to have one major bug resolved from the dskit.

aknuds1:
We've agreed to prioritize this issue again, and I will take point.

ring/lifecycler.go
aknuds1: @krajorama I implemented your suggestion, without remembering your suggestion :) I guess it should be a good idea if we both landed on it.

ring/lifecycler.go
aknuds1: Thanks @agardiman, please see revised/simplified implementation :)

ring/lifecycler.go
aknuds1: Good point @agardiman. I've implemented what you suggest, i.e. shuffle before dropping tokens. PTAL.

aknuds1:
I've pushed revised and simplified implementation, also incorporating a suggestion from @agardiman, AFAICT this should correctly handle switching from the LEAVING to the ACTIVE state (adding or dropping tokens as necessary to reflect the configured token count).

krajorama (approved):
Thank you, LGTM. Changes and tests make sense to me, plus it's backwards compatible.

pracucci:
Closes #289

pracucci (approved):
Good job Arve! The changes LGTM. I left few minor comments I would be glad if you could look at before merging. Thanks!

ring/lifecycler.go
```diff
@@ -616,24 +617,53 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
+			} else if delta < 0 {
+				// We have too many tokens
+				level.Debug(i.logger).Log("msg", "existing instance has too many tokens, removing difference",
+					"num_tokens", len(tokens), "need", i.cfg.NumTokens)
+				// Originally suggested by Andrea Gardiman, make sure we don't pick the N smallest tokens,
+				// since that would increase the chance of the instance receiving only smaller hashes
+				// https://github.com/grafana/dskit/pull/79#discussion_r1056205242
+				rand.Shuffle(len(tokens), func(i, j int) {
+					tokens[i], tokens[j] = tokens[j], tokens[i]
+				})
```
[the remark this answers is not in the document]
aknuds1: Absolutely. Nice surprise, that `Tokens` already has a `Swap` method!

PR #80, opened 2021-11-26
flagext: add RegisterFlagsWithLogger

bboreham:
For cases such as `DeprecatedFlag()` that need a logger, extend `RegisterFlags` to allow this while still working with existing code.

**Checklist**
- NA Tests updated
- [x] `CHANGELOG.md` updated.

### DIFF
--- a/flagext/register.go
+++ b/flagext/register.go
@@ -11,8 +11,8 @@ type Registerer interface {
 	RegisterFlags(*flag.FlagSet)
 }
 
-// Registerer2 is a thing that can RegisterFlags with a Logger
-type Registerer2 interface {
+// RegistererWithLogger is a thing that can RegisterFlags with a Logger
+type RegistererWithLogger interface {
 	RegisterFlags(*flag.FlagSet, log.Logger)
 }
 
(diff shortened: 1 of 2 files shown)


flagext/register.go
```diff
@@ -1,24 +1,55 @@
...
+	"github.com/go-kit/log"
+)
 
 // Registerer is a thing that can RegisterFlags
 type Registerer interface {
 	RegisterFlags(*flag.FlagSet)
 }
 
+// Registerer2 is a thing that can RegisterFlags with a Logger
+type Registerer2 interface {
```
treid314, line 15: Can this be called something like LoggingRegisterer, or RegistererWithLog?

flagext/register.go
```diff
@@ -1,24 +1,55 @@
...
-		r.RegisterFlags(fs)
+	logger := log.NewNopLogger()
+	for _, v := range rs {
+		switch r := v.(type) {
+		case Registerer:
+			r.RegisterFlags(fs)
+		case Registerer2:
+			r.RegisterFlags(fs, logger)
+		default:
+			panic("RegisterFlagsWithLogger must be passed a Registerer")
```
treid314, line 51: Is it worth calling out we have to pass a Registerer or Registerer2(or whatever we want to call it)

treid314 (approved):
Logically fine, had some comments on naming

bboreham:
Thanks for your feedback; I have updated and rebased to cope with the CHANGELOG conflict.

PR #83, opened 2021-11-30
Improve BenchmarkMemberlistReceiveWithRingDesc

pracucci:
**What this PR does**:
I'm profiling `NotifyMsg()` and I've realized `BenchmarkMemberlistReceiveWithRingDesc()` is not benchmarking the `NotifyMsg()` in the case the received message updates the ring. That's the case we want to benchmark for every run, because it's much slower than the no-op case.

N/A

**Checklist**
- [ ] Tests updated


treid314 (approved):
Nice benchmark! LGTM

PR #84, opened 2021-11-30
Add sync.pool to contain our token map in conflictingTokensExist

treid314:
**What this PR does**: This PR _should_ use a memory pool instead of allocating a map for every time we call conflictingTokensExists to allow us to gc that memory easily and prevent long lived maps.

Fixes #<issue number>

**Checklist**
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 * [ENHANCEMENT] Replace go-kit/kit/log with go-kit/log. #52
 * [ENHANCEMENT] Add spanlogger package. #42
 * [ENHANCEMENT] Add runutil.CloseWithLogOnErr function. #58
-* [ENHANCEMENT] Optimise memberlist receive path when used as a backing store for rings with a large number of members. #76 #77
+* [ENHANCEMENT] Optimise memberlist receive path when used as a backing store for rings with a large number of members. #76 #77 #84
 * [ENHANCEMENT] Memberlist: prepare the data to send on the write before starting counting the elapsed time for `-memberlist.packet-write-timeout`, in order to reduce chances we hit the timeout when sending a packet to other node. #89
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
--- a/ring/model.go
+++ b/ring/model.go
@@ -303,7 +303,7 @@ func normalizeIngestersMap(inputRing *Desc) {
 	}
 }
 
-var tokenMapPool = sync.Pool{New: func() interface{}{return make(map[uint32]bool)}}
+var tokenMapPool = sync.Pool{New: func() interface{} { return make(map[uint32]bool) }}
 
 func conflictingTokensExist(normalizedIngesters map[string]InstanceDesc) bool {
 	count := 0
@@ -312,6 +312,9 @@ func conflictingTokensExist(normalizedIngesters map[string]InstanceDesc) bool {
 	}
 	tokensMap := tokenMapPool.Get().(map[uint32]bool)
 	defer func() {
+		for k := range tokensMap {
+			delete(tokensMap, k)
+		}
 		tokenMapPool.Put(tokensMap)
 	}()
 	for _, ing := range normalizedIngesters {
--- a/ring/model.go
+++ b/ring/model.go
@@ -306,10 +306,6 @@ func normalizeIngestersMap(inputRing *Desc) {
 var tokenMapPool = sync.Pool{New: func() interface{} { return make(map[uint32]bool) }}
 
 func conflictingTokensExist(normalizedIngesters map[string]InstanceDesc) bool {
-	count := 0
-	for _, ing := range normalizedIngesters {
-		count += len(ing.Tokens)
-	}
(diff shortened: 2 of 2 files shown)


treid314:
Looking at a pprof with a base of main and my branch looks like this showing a huge improvement in CPU time here.
image
As well as a pretty big improvement in memory between main and my branch here.
image
These numbers are so large I'm unsure if my change here is correct.

ring/model.go
```diff
@@ -302,13 +303,17 @@ func normalizeIngestersMap(inputRing *Desc) {
...
 func conflictingTokensExist(normalizedIngesters map[string]InstanceDesc) bool {
 	count := 0
 	for _, ing := range normalizedIngesters {
 		count += len(ing.Tokens)
 	}
-
-	tokensMap := make(map[uint32]bool, count)
+	tokensMap := tokenMapPool.Get().(map[uint32]bool)
+	defer func() {
+		tokenMapPool.Put(tokensMap)
```
stevesg, line 315: You either need to clear the map after taking it out or before putting it back. You're getting previous state :(

stevesg (changes requested):
I don't think it's working because you've introduced calls to `resolveConflicts` that weren't there before. If it's working as intended, then there would not be any. Functionally it's correct because checking for conflicts is an optimization in itself. If you fix it you may improve performance even more, or make it worse.

Also, just a reminder about what we have already discussed: I think the real optimization we need to do here is eliminate building the map altogether. Though if this gets us some percentage of the way there, then by all means proceed. Just bear in mind that I may remove it again.

ring/model.go
treid314: Yeah, too bad there's not a good way to .clear() the map :/

stevesg: (Problem here being over time, eventually all tokens will be in the map, and the check will say there are always conflicting tokens, so negating the original optimization)

treid314:
So when we clear the map via deleting all of the keys we do see a pretty decent CPU perf improvement compared to main
image
And a pretty large memory improvement over main because we're immediately clearing this map
image

ring/model.go
pstibrany: do we still need the `count`? it was only used to size the map I think.

treid314: I went with the delete here to clear this map, but also we could just use a map[uint32]{} here instead of using a boolean value and check contains rather than the value, but it's just a space optimization.

pstibrany: Have we tried to create empty map with zero size instead?

treid314: We don't use count here, good point. I tried a make(...) and that was slower than clearing the map oddly, but maybe I was looking at it wrong.

pstibrany: we can also try using `map[uint32]struct{}` which should be smaller, but requires bit of extra code to use.

pstibrany: looks like I should reload the page before adding more comments :)

treid314:
Using a map[uint32]struct{} saves us a byte per entry, which in our testing was about 7 meg which is free memory!
image

ring/model.go
treid314: I tested again with the zero size make() and it was way worse on cpu and memory than deleting the keys, which is an interesting go question.

pstibrany: Thanks. Not sure if it is worth trying to find conflicting tokens without any map at all which should be easy since all tokens slices are sorted already.

pracucci:
I've run `BenchmarkMemberlistReceiveWithRingDesc` and compared preformances with `main` version:

```
name                              old time/op    new time/op    delta
MemberlistReceiveWithRingDesc-12    3.24ms ± 2%    3.04ms ± 1%   -6.05%  (p=0.012 n=3+3)

name                              old alloc/op   new alloc/op   delta
MemberlistReceiveWithRingDesc-12    1.01MB ± 0%    0.05MB ±15%  -94.87%  (p=0.000 n=3+3)

name                              old allocs/op  new allocs/op  delta
MemberlistReceiveWithRingDesc-12       669 ± 0%       681 ± 1%     ~     (p=0.132 n=3+3)
```

CHANGELOG.md
```diff
@@ -19,4 +19,5 @@
 * [ENHANCEMENT] Add spanlogger package. #42
 * [ENHANCEMENT] Add runutil.CloseWithLogOnErr function. #58
 * [ENHANCEMENT] Optimise memberlist receive path when used as a backing store for rings with a large number of members. #76 #77
+* [ENHANCEMENT] Optimize `conflictingTokensExist` called by `NotifyMsg` to use a memory pool instead of allocating a map for each call #84
```
stevesg, line 22: I would merge this with the previous item given it's in the same release, user perspective is that "memberlist receive got better".

```suggestion
* [ENHANCEMENT] Optimise memberlist receive path when used as a backing store for rings with a large number of members. #76 #77 #84
```

ring/model.go
stevesg: [nit] After think about this, I think clearing the map after `Get()` is preferable, because then you're not trusting the last user to do the clearing, but this is a nit.

stevesg (approved):
Happy for us to get this in as it's a low risk change and reduces GC pressure, but as the CPU saving is not great, we should continue working on other optimizations. So don't be offended if I change it :)

ring/model.go
stevesg: We can leave this as-is

PR #87, opened 2021-12-01
Handle -memberlist.packet-write-timeout as an idle timeout

pracucci:
**What this PR does**:
I'm investigating the following errors that we see sometime on the **receiving** side:
```
msg="TCPTransport: packet digest mismatch" expected=6af0ad4fbb98e34497fa5933b735793b received=40ff8e998d061a0872696e6744657363
```

I've noticed that when we have these errors, on the sender side we also have this log:
```
msg="TCPTransport: WriteTo failed" addr=REDACTED:7946 err="digest: write tcp REDACTED:34320->REDACTED:7946: i/o timeout"
```

It makes sense. It on the writing side we hit the configured `-memberlist.packet-write-timeout`, then the TCP connection is closed before we send the digest too.

When investigating this, I've noticed `-memberlist.packet-write-timeout` is the total time the whole packet send can take and not an idle timeout on each write operation. In this PR I propose to change it to be an idle timeout instead.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,5 +19,6 @@
 * [ENHANCEMENT] Add spanlogger package. #42
 * [ENHANCEMENT] Add runutil.CloseWithLogOnErr function. #58
 * [ENHANCEMENT] Optimise memberlist receive path when used as a backing store for rings with a large number of members. #76 #77
+* [EHHANCEMENT] Memberlist: `-memberlist.packet-write-timeout` config option is now handled as an idle timeout for individual write operations instead of the max time the whole packet transferring can take. #87
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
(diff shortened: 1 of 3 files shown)


pstibrany (approved):
> In this PR I propose to change it to be an idle timeout instead.

Why? I find single timeout for entire packet-writing operation easier to reason about. Since we're doing 4 writes in this method, we can just also increase the timeout 4x. (Naming `-memberlist.packet-write-timeout` no longer makes sense after this change, as "packet" refers to memberlist packet.)

I don't think this is necessary, but it's harmless.

pracucci:
> > In this PR I propose to change it to be an idle timeout instead.
> 
> Why?

The packet can be either very small or very large (we configure `UDPBufferSize` to 10MB). After sending the packet, we also compute the md5 which could take "some" time if packet is big. In a normal condition this shouldn't be an issue (5s timeout is very high) but I've seen this issue happening in compactors, so my guess it's that while compaction is running it's slow to do it and so we could hit the timeout.

Changing it to an idle timeout should help in this situation.

pstibrany:
> Changing it to an idle timeout should help in this situation.

We continued this discussion offline. I still think deadline for entire operation makes more sense than extending it for every Write operation, but I'm not going to block this.

In my opinion we should fix the order of operations (compute digest first), and do all writes close to each other. (Ideally as single tcp.Write operation).

pracucci:
Superseeded by https://github.com/grafana/dskit/pull/89.

PR #90, opened 2021-12-01
Relax memberlist probe interval and timeout

pracucci:
**What this PR does**:
The memberlist client probes a random node every `ProbeInterval` (defaults to 1s). If it fails to get probe ack response within `ProbeTimeout` (defaults to 500ms) it both initiates an indirect probe (via 3 other nodes) and a direct probe via TCP (opening a stream this time, not packet). The timeout for this 2nd attempt via indirect+TCP stream is `ProbeInterval-ProbeTimeout` (so other 500ms with default config).

I've noticed an high volume of logs like this:
```
level=warn msg="Was able to connect to ingester-REDACTED but other probes failed, network may be misconfigured"
```

This log happens is both direct + indirect pings timeout but the TCP stream one succeed. I think the main reason of having TCP stream only succeeding is that, due to how our `TCPTransport` works, TCP stream requires only opening 1 TCP connection while other methods requires 2 TCP connections (1 to send the PING from node A->B and then B needs to open another TCP connection to send the response back from node B->A). The 2 TCP connections introduce further overhead.

In this PR I propose to relax `ProbeInterval` and `ProbeTimeout` to hopefully get the 1st direct ping to succeed, without having to fallback at all.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [CHANGE] Rename `kv/kvtls` to `crypto/tls`. #39
 * [CHANGE] spanlogger: Take interface implementation for extracting tenant ID. #59
 * [CHANGE] The `status_code` label on gRPC client metrics has changed from '200' and '500' to '2xx', '5xx', '4xx', 'cancel' or 'error'. #68
+* [CHANGE] Memberlist: changed probe interval from `1s` to `5s` and probe timeout from `500ms` to `2s`. #90
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
 * [ENHANCEMENT] Add limiter package. #41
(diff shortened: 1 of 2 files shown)


pstibrany (approved):

pstibrany (approved):

pracucci:
I've tested this change in a dev cluster and I think it's a good improvement. The screenshot below shows the occurrences of the issue on average on each pod. I did **not** deploy the change to compactors, so we can have a comparison. The deployment to other pods took few hours and completed at about 21:00 UTC.

Legend:
- Green: compactor (without the fix)
- Yellow: others (with the fix)

Screenshot 2021-12-02 at 07 38 48

pstibrany:
Nice improvement.

PR #93, opened 2021-12-08
Memberlist/Ring: do not normalise tokens on subsequent Merge().

stevesg:
**What this PR does**:

Improves performance of `ring.Desc.Merge()`.

Memberlist uses the Merge() function on ring.Desc such that it is only ever
called subsequently on the result of itself. Therefore, because the Merge()
function always returns Desc in the normalised form, it does not have to
normalise again before doing another Merge().

Unit tests are fixed up and extended to reflect the precondition that the LHS
must be normalised, and the RHS does not have to be normalised.

I suspect it is also possible to remove normalisation on the incoming change
(the Desc being merged into the current Desc), but it would be safest to test
these changes incrementally. (Open to opinions as to omitting both now)

Performance difference:
```
MemberlistReceiveWithRingDesc-6     414µs ± 5%      31µs ± 7%  -92.49%  (p=0.000 n=10+10)

MemberlistReceiveWithRingDesc-6    22.7kB ± 0%     8.3kB ± 0%  -63.54%  (p=0.000 n=9+10)

MemberlistReceiveWithRingDesc-6       655 ± 0%        55 ± 0%  -91.60%  (p=0.000 n=10+10)
```

Fixes n/a

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/model.go
+++ b/ring/model.go
@@ -175,6 +175,12 @@ func (i *InstanceDesc) IsReady(now time.Time, heartbeatTimeout time.Duration) er
 //
 // This method is part of memberlist.Mergeable interface, and is only used by gossiping ring.
 //
+// The Desc on which this function is called must be normalised, that is, the token lists must
+// sorted and not contain duplicates. The function guarantees that the Desc will be left in this
+// normalised state, so multiple subsequent Merge calls on the same Desc is valid.
+//
+// The Mergeable passed as the parameter does not need to be normalised.
+//
 // Note: This method modifies d and mergeable to reduce allocations and copies.
 func (d *Desc) Merge(mergeable memberlist.Mergeable, localCAS bool) (memberlist.Mergeable, error) {
 	return d.mergeWithTime(mergeable, localCAS, time.Now())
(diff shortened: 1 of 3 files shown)


ring/model.go
```diff
@@ -175,6 +175,12 @@ func (i *InstanceDesc) IsReady(now time.Time, heartbeatTimeout time.Duration) er
 //
 // This method is part of memberlist.Mergeable interface, and is only used by gossiping ring.
 //
+// The Desc on which this function is called must be normalised, that is, the token lists must
```
pstibrany, line 178: [nit] Shall we call it "receiver"?

pstibrany (approved):
Makes sense, and looks good.

I think we should be careful about removing normalization of incoming message (RHS). While our code produces messages that are normalized already (eg. sorted tokens), there may be clients using dskit where this doesn't apply. If we want to guarantee that value stored in KV store is normalized, I think we cannot drop normalization of incoming messages.

stevesg:
> I think we should be careful about removing normalization of incoming message (RHS). While our code produces messages that are normalized already (eg. sorted tokens), there may be clients using dskit where this doesn't apply. If we want to guarantee that value stored in KV store is normalized, I think we cannot drop normalization of incoming messages.

Good point. I suspect that it will be safest, and necessary, to always normalise whatever state we receive from CAS() operations, though we could possibly avoid it for broadcasted changes which were generated by Merge()? Could explore it separately if it comes up in profiling.

pstibrany:
In a private discussion @pracucci suggested having an option to enable/disable normalization of incoming value. We know we can disable it in our code, but it can default to enabled just in case.

pracucci (approved):
LGTM!

PR #94, opened 2021-12-10
Kubernetes KV store implementation

dimitarvdimitrov:
#### Summary
As part of a hackathon project we implemented a KV store provider that uses a kubernetes configmap as a backing store - package `/kv/kubernetes`

#### Implementation
The ring description is stored as a protobuf-compressed and base64-encoded value in the configmap.

`CAS` is implemented via [JSON patches](http://jsonpatch.com/) to the kubernetes API server. The comparison is done on the values' hashes which are also stored in the configmap.

`Watch` is implemented via the [kubernetes watch API](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes).

We extracted the `Client.Watch*` implementation in memberlist into `kv/internal/watch` and reused it in `kv/kubernetes`.

#### Scaling
In a dev cluster the configuration scaled up to 300 instances in the ring heartbeating every 30s (10 CAS per second).

#### Future work
This PR in it's current state doesn't provide much benefit over other KV providers.

The benefits will come as we try to extract the information out of the ring and instead rely on k8s information to infer instances' health. That will remove the need to heartbeat so often (or at all) and will make the kubernetes alternative the only that doesn't require to CAS at every heart beat. Hopefully this will scale better than the current write-intensive ring implementations.

**Checklist**
- [x] Tests updated


kv/memberlist/memberlist_client.go
```diff
@@ -642,52 +640,15 @@ func (m *KV) get(key string, codec codec.Codec) (out interface{}, version uint,
...
-		case <-m.shutdown:
-			// stop watching on shutdown
-			return
-
-		case <-ctx.Done():
-			return
-		}
-	}
+func (m *KV) WatchKey(ctx context.Context, key string, f func(interface{}) bool) {
+	m.watcher.WatchKey(ctx, key, f)
```
dimitarvdimitrov, line 651: The change here as well as the one in `WatchPrefix` have the side effect that the new value passed to different watchers will be same. Previously each watcher will get a different deep copy of the value. 

This doesn't have implications within dskit because only `Ring` uses either of those methods and it uses only one at a time 

https://github.com/grafana/dskit/blob/4bbe2b279599a0af2507518c5e2b3faf7adb894a/ring/ring.go#L304

The deeply copied values aren't a document behaviour either, so clients shouldn't be relying on it.

pstibrany:
> As part of a hackathon project we implemented a KV store provider that uses a kubernetes configmap as a backing store - package `/kv/kubernetes`

I think this is great as a hackaton project, but I don't think it should become part of dskit. I don't think the benefits (nothing listed in the PR, I guess it would be simpler setup when using inside kubernetes) outweigh the added complexity and maintenance burden (adding kubernetes libraries as dskit dependencies (!!!), yet another ring implemenation, extra code to keep bug-free).

dimitarvdimitrov:
I'm not quite sure how to tackle the extra dependencies. This is roughly
400k new lines of code in the vendor folder after `go mod vendor`.

I think the benefits will come as we try to extract the information out of
the ring and instead rely on k8s information to infer instances' health.
That will remove the need to heartbeat so often (or at all) and will make
the kubernetes alternative the only alternative that doesn't require to CAS
at every heart beat.

I'll add the above to the PR description. The caveat is that this is a side
project and the "when" is a bit unclear.

Do you think it's better to close this PR in the meantime?

pstibrany:
> Do you think it's better to close this PR in the meantime?

Keeping it open may bring more attention to it by people interested in this.

dimitarvdimitrov:
I am closing this because I won't be working on it for the time being

PR #95, opened 2021-12-13
ring/util.go: make WaitRingStability more responsive

### LINKED ISSUE #96: ring/util.go WaitRingStability slow to respond
In Cortex the test TestStoreGateway_InitialSyncWithWaitRingStability/shuffle_sharding_strategy,_20_gateways,_RF_=_3,_SS_=_3_(bucket_index_enabled_=_false)
 takes 30 seconds. And most of it is spent in WaitRingStability, which is weird as everything runs locally in the same process.

Goal: improve the responsiveness of the function if possible

krajorama:
**What this PR does**:

In some use cases, such as the Cortex store gateway, it is enough for the
user to know that the token distribution didn't change and it is not
important to know if some member changed between allowed states.

This patch implements the new WaitRingTokensStability function to
be able to wait for tokens stability and ignore allowed state changes.

The original motivation was that store gateway tests took a minute longer
then strictly necessary.

Fixes #96 

**Checklist**
- [+] Tests updated
- [+] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,5 +23,6 @@
 * [ENHANCEMENT] Memberlist: prepare the data to send on the write before starting counting the elapsed time for `-memberlist.packet-write-timeout`, in order to reduce chances we hit the timeout when sending a packet to other node. #89
 * [ENHANCEMENT] flagext: for cases such as `DeprecatedFlag()` that need a logger, add RegisterFlagsWithLogger. #80
 * [ENHANCEMENT] Added option to BasicLifecycler to keep instance in the ring when stopping. #97
+* [ENHANCEMENT] Add WaitRingTokensStability function to ring, to be able to wait on ring stability excluding allowed state transitions. #95
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
--- a/ring/replication_set_test.go
+++ b/ring/replication_set_test.go
@@ -241,3 +241,71 @@ func TestReplicationSet_Do(t *testing.T) {
 		})
 	}
 }
+
+var (
+	replicationSetChangesInitialState = ReplicationSet{
+		Instances: []InstanceDesc{
+			{Addr: "127.0.0.1"},
+			{Addr: "127.0.0.2"},
+			{Addr: "127.0.0.3"},
+		},
+	}
+	replicationSetChangesTestCases = map[string]struct {
+		nextState                                  ReplicationSet
+		expectHasReplicationSetChanged             bool
+		expectHasReplicationSetChangedWithoutState bool
+	}{
+		"timestamp changed": {
+			ReplicationSet{
+				Instances: []InstanceDesc{
+					{Addr: "127.0.0.1", Timestamp: time.Hour.Microseconds()},
+					{Addr: "127.0.0.2"},
+					{Addr: "127.0.0.3"},
+				},
+			},
+			false,
+			false,
+		},
+		"state changed": {
+			ReplicationSet{
+				Instances: []InstanceDesc{
+					{Addr: "127.0.0.1", State: PENDING},
+					{Addr: "127.0.0.2"},
+					{Addr: "127.0.0.3"},
+				},
+			},
+			true,
+			false,
+		},
+		"more instances": {
+			ReplicationSet{
+				Instances: []InstanceDesc{
+					{Addr: "127.0.0.1"},
+					{Addr: "127.0.0.2"},
+					{Addr: "127.0.0.3"},
+					{Addr: "127.0.0.4"},
+				},
+			},
+			true,
+			true,
+		},
+	}
+)
+
+func TestHasReplicationSetChanged_IgnoresTimeStamp(t *testing.T) {
+	// Only testing difference to underlying Equal function
+	for testName, testData := range replicationSetChangesTestCases {
+		t.Run(testName, func(t *testing.T) {
+			assert.Equal(t, testData.expectHasReplicationSetChanged, HasReplicationSetChanged(replicationSetChangesInitialState, testData.nextState), "HasReplicationSetChanged wrong result")
+		})
+	}
+}
+
+func TestHasReplicationSetChangedWithoutState_IgnoresTimeStampAndState(t *testing.T) {
+	// Only testing difference to underlying Equal function
+	for testName, testData := range replicationSetChangesTestCases {
+		t.Run(testName, func(t *testing.T) {
+			assert.Equal(t, testData.expectHasReplicationSetChangedWithoutState, HasReplicationSetChangedWithoutState(replicationSetChangesInitialState, testData.nextState), "HasReplicationSetChangedWithoutState wrong result")
+		})
+	}
+}
--- a/ring/util.go
+++ b/ring/util.go
@@ -96,6 +96,19 @@ func WaitInstanceState(ctx context.Context, r ReadRing, instanceID string, state
 // WaitRingStability monitors the ring topology for the provided operation and waits until it
 // keeps stable for at least minStability.
 func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability, maxWaiting time.Duration) error {
+	return waitStability(ctx, r, op, minStability, maxWaiting, HasReplicationSetChanged)
+}
+
+// WaitRingTokensStability waits for the Ring to be unchanged at
+// least for minStability time period, excluding transitioning between
+// allowed states (e.g. JOINING->ACTIVE if allowed by op).
+// This can be used to avoid wasting resources on moving data around
+// due to multiple changes in the Ring.
+func WaitRingTokensStability(ctx context.Context, r *Ring, op Operation, minStability, maxWaiting time.Duration) error {
+	return waitStability(ctx, r, op, minStability, maxWaiting, HasReplicationSetChangedWithoutState)
+}
+
+func waitStability(ctx context.Context, r *Ring, op Operation, minStability, maxWaiting time.Duration, isChanged func(ReplicationSet, ReplicationSet) bool) error {
 	// Configure the max waiting time as a context deadline.
 	ctx, cancel := context.WithTimeout(ctx, maxWaiting)
 	defer cancel()
@@ -117,7 +130,7 @@ func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability,
 			// replication set which we use to compare with the previous state.
 			currRingState, _ := r.GetAllHealthy(op) // nolint:errcheck
 
-			if HasReplicationSetChanged(ringLastState, currRingState) {
+			if isChanged(ringLastState, currRingState) {
 				ringLastState = currRingState
 				ringLastStateTs = time.Now()
 			} else if time.Since(ringLastStateTs) >= minStability {
--- a/ring/util_test.go
+++ b/ring/util_test.go
@@ -79,7 +79,7 @@ func TestGenerateTokens(t *testing.T) {
 	}
 }
 
-func TestGenerateTokensIgnoresOldTokens(t *testing.T) {
+func TestGenerateTokens_IgnoresOldTokens(t *testing.T) {
 	first := GenerateTokens(1000000, nil)
 	second := GenerateTokens(1000000, first)
 
@@ -96,14 +96,7 @@ func TestGenerateTokensIgnoresOldTokens(t *testing.T) {
 	}
 }
 
-func TestWaitRingStabilityShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t *testing.T) {
-	t.Parallel()
-
-	const (
-		minStability = 2 * time.Second
-		maxWaiting   = 10 * time.Second
-	)
-
+func createStartingRing() *Ring {
 	// Init the ring.
 	ringDesc := &Desc{Ingesters: map[string]InstanceDesc{
 		"instance-1": {Addr: "127.0.0.1", State: ACTIVE, Timestamp: time.Now().Unix()},
@@ -123,14 +116,63 @@ func TestWaitRingStabilityShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t
 		strategy:            NewDefaultReplicationStrategy(),
 	}
 
+	return ring
+}
+
+func TestWaitRingStability_ShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t *testing.T) {
+	t.Parallel()
+
+	const (
+		minStability = 2 * time.Second
+		maxWaiting   = 10 * time.Second
+	)
+
+	ring := createStartingRing()
+
 	startTime := time.Now()
 	require.NoError(t, WaitRingStability(context.Background(), ring, Reporting, minStability, maxWaiting))
 	elapsedTime := time.Since(startTime)
 
-	assert.InDelta(t, minStability, elapsedTime, float64(2*time.Second))
+	assert.GreaterOrEqual(t, elapsedTime, minStability)
+	assert.Less(t, elapsedTime, 2*minStability)
 }
 
-func TestWaitRingStabilityShouldReturnOnceMinStabilityHasBeenReached(t *testing.T) {
+func TestWaitRingTokensStability_ShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t *testing.T) {
+	t.Parallel()
+
+	const (
+		minStability = 2 * time.Second
+		maxWaiting   = 10 * time.Second
+	)
+
+	ring := createStartingRing()
+
+	startTime := time.Now()
+	require.NoError(t, WaitRingTokensStability(context.Background(), ring, Reporting, minStability, maxWaiting))
+	elapsedTime := time.Since(startTime)
+
+	assert.GreaterOrEqual(t, elapsedTime, minStability)
+	assert.Less(t, elapsedTime, 2*minStability)
+}
+
+func addInstanceAfterSomeTime(ring *Ring, addInstanceAfter time.Duration) {
+	go func() {
+		time.Sleep(addInstanceAfter)
+
+		ring.mtx.Lock()
+		defer ring.mtx.Unlock()
+		ringDesc := ring.ringDesc
+		instanceID := fmt.Sprintf("127.0.0.%d", len(ringDesc.Ingesters)+1)
+		ringDesc.Ingesters[instanceID] = InstanceDesc{Addr: instanceID, State: ACTIVE, Timestamp: time.Now().Unix()}
+		ring.ringDesc = ringDesc
+		ring.ringTokens = ringDesc.GetTokens()
+		ring.ringTokensByZone = ringDesc.getTokensByZone()
+		ring.ringInstanceByToken = ringDesc.getTokensInfo()
+		ring.ringZones = getZones(ringDesc.getTokensByZone())
+	}()
+}
+
+func TestWaitRingStability_ShouldReturnOnceMinStabilityOfInstancesHasBeenReached(t *testing.T) {
 	t.Parallel()
 
 	const (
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -24,3 +24,4 @@
 * [ENHANCEMENT] flagext: for cases such as `DeprecatedFlag()` that need a logger, add RegisterFlagsWithLogger. #80
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
+* [ENHANCEMENT] ring/util.go: make WaitRingStability more responsive #96
```
aknuds1, line 27: This PR's number is 95:
```suggestion
* [ENHANCEMENT] ring/util.go: make WaitRingStability more responsive #95
```

ring/util.go
```diff
@@ -102,27 +102,28 @@ func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability,
...
 	ringLastState, _ := r.GetAllHealthy(op) // nolint:errcheck
-	ringLastStateTs := time.Now()
 
-	const pollingFrequency = time.Second
-	pollingTicker := time.NewTicker(pollingFrequency)
-	defer pollingTicker.Stop()
+	// This timer will be our state
+	stabilityTimer := time.NewTimer(minStability)
+
+	stateWatch, _ := r.statesWatcher.Register()
```
aknuds1, line 109: Why ignore the error?

ring/util.go
```diff
@@ -102,27 +102,28 @@ func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability,
...
+	defer r.statesWatcher.UnRegister()
 
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case <-pollingTicker.C:
-			// We ignore the error because in case of error it will return an empty
-			// replication set which we use to compare with the previous state.
```
aknuds1, line 117: Why remove this comment, isn't it still relevant?

ring/watcher.go
```diff
@@ -0,0 +1,54 @@
+package ring
+
+import (
+	"fmt"
+	"sync"
+)
+
+type Watcher interface {
```
aknuds1, line 8: Why do we need this public interface in the package? Can it just be removed since it's not in use?

ring/watcher.go
```diff
@@ -0,0 +1,54 @@
...
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	if s.notifier != nil {
+		close(s.notifier)
+	}
+	s.notifier = nil
+}
+
+// Handles notifying listeners about state change in the Ring
+// Assumes that the callee will check that there is meaningfull change in the states
```
aknuds1, line 40: I think the correct term would be "the caller", i.e. the caller of the method:
```suggestion
// Assumes that the caller will check that there is meaningful change to the state.
```

ring/watcher.go
```diff
@@ -0,0 +1,54 @@
...
+
+type Watcher interface {
+	Register() (<-chan interface{}, error)
+	UnRegister()
+	notify()
+}
+
+type watcher struct {
+	mtx      sync.Mutex
+	notifier chan interface{}
```
aknuds1, line 16: Isn't it more readable if the type of the channel is `struct{}`, i.e. signaling that no information is passed over the channel?
```suggestion
	notifier chan struct{}
```

aknuds1 (commented):
I left some questions, please have a look :)

CHANGELOG.md
krajorama: mixup, thought it was the issue to be referenced

ring/util.go
krajorama: fair

ring/util.go
krajorama: heavy handed, removed the whole case and rebuilt it. fixed

ring/watcher.go
krajorama: okok, I almost used it for something else as well, but it did not improve speed

ring/watcher.go
krajorama: depends on how you look at it I suppose. I'll change it to reflect the user experience more

aknuds1: The callee would be this method, wouldn't it since it's getting called? It's the caller of the method that should ensure there's meaningful state change right?

ring/watcher.go
krajorama: I'm just a noob at golang, fixed

krajorama:
@aknuds1 Thank you for your comments! What do you think about this solution instead of reducing the polling interval overall?

aknuds1:
> @aknuds1 Thank you for your comments! What do you think about this solution instead of reducing the polling interval overall?

Happy to help :) I don't know enough about the current solution yet to tell if this is a good idea or not, so it's better if someone like Peter or Marco f.ex. comments. I have some other obligations now, but I hope to return to take another pass at reviewing this PR.

ring/util.go
```diff
@@ -102,27 +102,37 @@ func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability,
...
+				stabilityTimer.Reset(minStability)
+			} else {
 				return nil
 			}
+		case <-stateWatch:
+			ringLastState, _ = r.GetAllHealthy(op) // nolint:errcheck
+			if !stabilityTimer.Stop() {
+				<-stabilityTimer.C
+			}
+			stabilityTimer.Reset(minStability)
```
pracucci, line 135: Here we reset the timer but the fact that we received a notification from the watch doesn't mean the ring has changed so that `HasReplicationSetChanged()` return true (for example, that function doesn't compare heartbeat `Timestamp` which change continuously but the watcher will notify whenever an heartbeat timestamp has changed too).

pracucci (commented):
Thanks for working on this! You raised a very good issue, however I think the offered solution is not the right one. Let me explain it.

This solution assumes that ring changes propagation are almost instantaneous between different members but that's not guaranteed. We have relatively short propagation times when running ring on Consul (but still in the order of 1s or 2s) and higher propagation times when running with memberlist (even few tens of seconds on very large clusters). If propagation times are not almost instantaneous, the issue you describe happen anyway.

After looking at this along with Peter and after looking at how we use this in Cortex, we think what we really want is "wait ring TOKENS stability". Our suggestion is:
1. Add `WaitRingTokensStability()` which is a copy of `WaitRingStability()`
2. In `WaitRingTokensStability()` instead of calling `HasReplicationSetChanged()`, call a different function which is a copy of `HasReplicationSetChanged()` but doesn't compare member states (as well as doesn't compare timestamps too as `HasReplicationSetChanged()`)

_In the suggestion above I mentioned "copy" to make it clear what we have in mind, but we can obviously refactor existing methods to avoid code duplication. Just keep in mind we shouldn't change behaviour and signature of `HasReplicationSetChanged()` and `WaitRingStability()` in order to not introduce breaking changes._

## Why this PR assumes almost instantaneous propagation times?

This about a simple setup with 2 members (A and B) and the following timeline:
- A and B wait for stability
- A reaches stability
- A moves from JOINING -> ACTIVE
- B receives all updates at once and, despite the ring stability was reached, it resets the counter because A reached the stability but already changed their state from JOINING to ACTIVE (and what we're trying to fix is that B shouldn't reset the timer in this case)

With memberlist it's even worse because there's no changes propagation ordering guaranteed (while with Consul CAS() changes are propagated in order, but multiple changes can be propagated together).

aknuds1:
Thanks @pracucci for chiming in! :)

krajorama:
I've updated the solution with the suggested way of working. Added new tests and tested cortex as well - the expected speedup in the tests is still there.

CHANGELOG.md
```diff
@@ -24,3 +24,4 @@
 * [ENHANCEMENT] flagext: for cases such as `DeprecatedFlag()` that need a logger, add RegisterFlagsWithLogger. #80
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
+* [ENHANCEMENT] Add WaitRingTokensStability function to ring, to be able to wait on ring stability excluding allowed state transitions. #95
```
pracucci, line 27: [nit] We're used to group changelog entries by type. Do you mind moving this up, grouping to the other `[ENHANCEMENT]` entries, please?

ring/util.go
```diff
@@ -96,6 +96,19 @@ func WaitInstanceState(ctx context.Context, r ReadRing, instanceID string, state
 // WaitRingStability monitors the ring topology for the provided operation and waits until it
 // keeps stable for at least minStability.
 func WaitRingStability(ctx context.Context, r *Ring, op Operation, minStability, maxWaiting time.Duration) error {
+	return waitStability(ctx, r, op, minStability, maxWaiting, HasReplicationSetChanged)
+}
+
+// WaitRingTokensStability waits for the Ring to be unchanged at
+// least for minStability time period, excluding transitioninig between
```
pracucci, line 103: [nit]
```suggestion
// least for minStability time period, excluding transitioning between
```

ring/replication_set_test.go
```diff
@@ -241,3 +241,60 @@ func TestReplicationSet_Do(t *testing.T) {
 		})
 	}
 }
+
+func TestHasReplicationSetChangedWithoutStateIgnoresTimeStampAndState(t *testing.T) {
+	// Only testing difference to underlying Equal function
+	rs1 := ReplicationSet{
```
pracucci, line 247: [nit] May calling this `initialState` (or something like this) may clarify this is the initial state we're comparing to?

ring/replication_set_test.go
```diff
@@ -241,3 +241,60 @@ func TestReplicationSet_Do(t *testing.T) {
...
+		Instances: []InstanceDesc{
+			{Addr: "127.0.0.1"},
+			{Addr: "127.0.0.2"},
+			{Addr: "127.0.0.3"},
+		},
+	}
+	tests := map[string]struct {
+		rs                                         ReplicationSet
```
pracucci, line 255: [nit] Maybe calling it `nextState` could help to clarify it. Also naming fields in the test case. They're nit, up to you whether picking it up or not.

ring/util_test.go
```diff
@@ -123,48 +116,50 @@ func TestWaitRingStabilityShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t
...
 	}
 
+	return ring
+}
+
+func TestWaitRingStabilityShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t *testing.T) {
+	t.Parallel()
+
+	const (
+		minStability = 2 * time.Second
```
pracucci, line 126: 2s is not enough here, because when we assert on the actual wait time we pass a delta of 2s, see:
```
assert.InDelta(t, minStability, elapsedTime, float64(2*time.Second))
```

If we remove the call to `WaitRingStability()` at all, the test will pass anyway. I know it's not something you changed, but I would be glad if you could fix it. One way to fix it is checking the elapsed time like we do in `TestWaitRingStabilityShouldReturnOnceMinStabilityOfInstancesHasBeenReached`.

_The same comment applies to tests below._

ring/util_test.go
```diff
@@ -240,6 +261,98 @@ func TestWaitRingStabilityShouldReturnErrorIfMaxWaitingIsReached(t *testing.T) {
 	assert.InDelta(t, maxWaiting, elapsedTime, float64(2*time.Second))
```
pracucci, line 261: [nit] I think we can assert on elapsedTime >= maxWaiting instead of using delta, because we expect at least maxWaiting time has been waited for.

_Same comment applies to tests below._

ring/util_test.go
```diff
@@ -240,6 +261,98 @@ func TestWaitRingStabilityShouldReturnErrorIfMaxWaitingIsReached(t *testing.T) {
...
+
+	// Keep changing the ring.
+	done := changeStatePeriodically(ring)
+	defer close(done)
+
+	startTime := time.Now()
+	require.Equal(t, nil, WaitRingTokensStability(context.Background(), ring, Reporting, minStability, maxWaiting))
+	elapsedTime := time.Since(startTime)
+
+	assert.InDelta(t, minStability, elapsedTime, float64(2*time.Second))
```
pracucci, line 353: Same comment here. If we skip calling `WaitRingTokensStability()` at all the test will pass anyway.

pracucci (approved):
Thanks for working on this! Good job. The refactoring is very clean. The logic changes LGTM, but I would just improve a bit test assertions before merging. Thanks!

CHANGELOG.md
krajorama: ack

ring/util.go
krajorama: thank you, I didn't realize that VS code has no spell check by default like Rider. Now installed.

ring/replication_set_test.go
krajorama: fair

ring/replication_set_test.go
krajorama: fair

ring/util_test.go
krajorama: Yes, I noticed this as well,  I'll add the checking for minimal time. I suspect the numbers here are chosen to avoid flakyness.  
I also think that these should not be running in real time, so I'll propose a change as part of my pet project to improve testability at a later time.

ring/replication_set_test.go
```diff
@@ -241,3 +241,60 @@ func TestReplicationSet_Do(t *testing.T) {
 		})
 	}
 }
+
+func TestHasReplicationSetChangedWithoutStateIgnoresTimeStampAndState(t *testing.T) {
```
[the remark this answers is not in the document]
krajorama: fixed

ring/replication_set_test.go
```diff
@@ -241,3 +241,60 @@ func TestReplicationSet_Do(t *testing.T) {
...
+					{Addr: "127.0.0.4"},
+				},
+			},
+			true,
+			true,
+		},
+	}
+	for testName, testData := range tests {
+		t.Run(testName, func(t *testing.T) {
+			assert.Equal(t, testData.expectHasReplicationSetChanged, HasReplicationSetChanged(rs1, testData.rs), "HasReplicationSetChanged wrong result")
```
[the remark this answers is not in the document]
krajorama: I'm trying to not repeat a lot of code and point out the difference between the two functions. That said, the naming was crap. Let me suggest a new way

ring/util_test.go
```diff
@@ -123,48 +116,50 @@ func TestWaitRingStabilityShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t
 		strategy:            NewDefaultReplicationStrategy(),
 	}
 
+	return ring
+}
+
+func TestWaitRingStabilityShouldReturnAsSoonAsMinStabilityIsReachedOnNoChanges(t *testing.T) {
```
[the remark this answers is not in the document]
krajorama: fixed

ring/util_test.go
krajorama: yes

pracucci (approved):
LGTM, thanks! 👏

ring/util_test.go
```diff
@@ -96,21 +96,14 @@ func TestGenerateTokensIgnoresOldTokens(t *testing.T) {
...
-
+func createStartingRing() *Ring {
 	// Init the ring.
 	ringDesc := &Desc{Ingesters: map[string]InstanceDesc{
-		"instance-1": {Addr: "127.0.0.1", State: ACTIVE, Timestamp: time.Now().Unix()},
-		"instance-2": {Addr: "127.0.0.2", State: PENDING, Timestamp: time.Now().Unix()},
-		"instance-3": {Addr: "127.0.0.3", State: JOINING, Timestamp: time.Now().Unix()},
-		"instance-4": {Addr: "127.0.0.4", State: LEAVING, Timestamp: time.Now().Unix()},
-		"instance-5": {Addr: "127.0.0.5", State: ACTIVE, Timestamp: time.Now().Unix()},
+		"instance-1": {Addr: "instance-1", State: ACTIVE, Timestamp: time.Now().Unix()},
```
pstibrany, line 102: In practice, this field contains an IP address, so previous example was bit better, although in this test, it doesn't matter.

ring/util_test.go
```diff
@@ -182,44 +194,40 @@ func TestWaitRingStabilityShouldReturnOnceMinStabilityHasBeenReached(t *testing.
...
-		cfg:                 Config{HeartbeatTimeout: time.Minute},
-		ringDesc:            ringDesc,
-		ringTokens:          ringDesc.GetTokens(),
-		ringTokensByZone:    ringDesc.getTokensByZone(),
-		ringInstanceByToken: ringDesc.getTokensInfo(),
-		ringZones:           getZones(ringDesc.getTokensByZone()),
-		strategy:            NewDefaultReplicationStrategy(),
-	}
+	// Add 1 new instance after some time.
+	go addInstanceAfterSomeTime(ring, addInstanceAfter)
```
pstibrany, line 209: This function already starts extra goroutine, so this `go` keyword is currently extra.

ring/util_test.go
```diff
@@ -232,15 +240,123 @@ func TestWaitRingStabilityShouldReturnErrorIfMaxWaitingIsReached(t *testing.T) {
...
+	const (
+		minStability = 2 * time.Second
+		maxWaiting   = 7 * time.Second
+	)
+
+	ring := createStartingRing()
+
+	// Keep changing the ring.
+	done := changeStatePeriodically(ring)
+	defer close(done)
```
pstibrany, line 349: Nit: This test leaves goroutine running even after test is finishes (for up to 1s). Other tests  have the same or similar problems (eg. leaving goroutines behind, if they fail early). Goleak took would flag such tests, even though goroutines finish soon after. Probably not worth fixing.

ring/util_test.go
```diff
@@ -232,15 +240,123 @@ func TestWaitRingStabilityShouldReturnErrorIfMaxWaitingIsReached(t *testing.T) {
...
+
+	startTime := time.Now()
+	require.Equal(t, nil, WaitRingTokensStability(context.Background(), ring, Reporting, minStability, maxWaiting))
+	elapsedTime := time.Since(startTime)
+
+	assert.GreaterOrEqual(t, elapsedTime.Milliseconds(), minStability.Milliseconds())
+	assert.Less(t, elapsedTime.Milliseconds(), 2*minStability.Milliseconds())
```
pstibrany, line 356:
```suggestion
	assert.Less(t, elapsedTime, 2*minStability)
```

(Everywhere, not just here.)

pstibrany (approved):
Changes look good. I left some non-blocking comments.

ring/util_test.go
```diff
@@ -232,15 +240,123 @@ func TestWaitRingStabilityShouldReturnErrorIfMaxWaitingIsReached(t *testing.T) {
...
+	)
+
+	ring := createStartingRing()
+
+	// Keep changing the ring.
+	done := changeStatePeriodically(ring)
+	defer close(done)
+
+	startTime := time.Now()
+	require.Equal(t, nil, WaitRingTokensStability(context.Background(), ring, Reporting, minStability, maxWaiting))
```
aknuds1, line 352: Nit:
```suggestion
	require.NoError(t, WaitRingTokensStability(context.Background(), ring, Reporting, minStability, maxWaiting))
```

aknuds1 (approved):
LGTM, just a nit.

ring/util_test.go
krajorama: fixed

ring/util_test.go
krajorama: lol , true, don't know why it was like that in the first place, removed

ring/util_test.go
krajorama: huh? I'm confused, the done channel is there specifically to avoid that. How would it leak?

pstibrany: Oh, somehow I read the code as if it was waiting 1s. My bad. The race between test end and goroutine exit is still there, but it's definitely not 1s.

ring/util_test.go
```diff
@@ -257,10 +257,10 @@ func TestWaitRingStability_ShouldReturnErrorIfInstancesAddedAndMaxWaitingIsReach
 	defer close(done)
 
 	startTime := time.Now()
-	require.Equal(t, context.DeadlineExceeded, WaitRingStability(context.Background(), ring, Reporting, minStability, maxWaiting))
+	require.Error(t, context.DeadlineExceeded, WaitRingStability(context.Background(), ring, Reporting, minStability, maxWaiting))
```
pstibrany, line 260: This currently tests that `context.DeadlineExceeded` is an error, not that `WaitRingStability` returned one.

pstibrany: Btw, Same problem is in other places of this PR (eg. lines 280, 331)

krajorama: dang, fixed

ring/util_test.go
krajorama: discussed offline, won't fix now

PR #101, opened 2021-12-17
change the ruler ring key

trevorwhitney:
**What this PR does**:

Changes the ruler's ring key from `ring` to `ruler` so it does not conflict with the ingester's ring key.

If the ruler and ingester's rings have the same prefix, then they will both register to the same ring. This can result in the querier trying to query the ruler, which fails. As I don't think there's any need to explicitly prevent rings from sharing a common prefix (for instance, if the KV store is being used for other types of data, they may want to prefix everything grafana related under `grafana`, or `loki`, for example), we should make sure each ring key is unique.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [CHANGE] spanlogger: Take interface implementation for extracting tenant ID. #59
 * [CHANGE] The `status_code` label on gRPC client metrics has changed from '200' and '500' to '2xx', '5xx', '4xx', 'cancel' or 'error'. #68
 * [CHANGE] Memberlist: changed probe interval from `1s` to `5s` and probe timeout from `500ms` to `2s`. #90
+* [CHANGE] Ring: change the ruler's ring key from `ring` -> `ruler` to avoid conflicts with ingester's ring key
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
 * [ENHANCEMENT] Add limiter package. #41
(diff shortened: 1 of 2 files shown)


pstibrany:
I don’t see why the ring keys are defined in dskit in the first place? Are they actually used anywhere in dskit?

trevorwhitney:
@pstibrany I couldn't find any usage is dskit, no. I would be on board with moving them out of dskit as well.

pstibrany:
> @pstibrany I couldn't find any usage is dskit, no. I would be on board with moving them out of dskit as well.

Thank you for checking. I would suggest to remove these keys from dskit, and reintroduce them in projects that actually use them.

trevorwhitney:
@pstibrany do you have an opinion on the correct order of operations here? I'm thinking we probably want to update the projects first so they no longer rely on these constants, then circle back here and remove them?

pstibrany:
> @pstibrany do you have an opinion on the correct order of operations here? I'm thinking we probably want to update the projects first so they no longer rely on these constants, then circle back here and remove them?

If we remove them from dskit first, then we get compilation errors when updating projects – that’s a good thing… we will know exactly what needs updating.

pracucci:
I agree with Peter that these keys shouldn't be defined here, but unfortunately we can't change the ruler key without introducing a breaking change, and I'm a bit dubious we should merge this PR. Let's move this to the specific project and then let's discuss how to deal with a smooth migration path.

owen-d:
Rather than change the key, I think we should _remove_ it from this project. It will be a breaking change, but also prevent consumers from _compiling_, which seems much better. They can be redefined in the consumer projects as necessary.

pracucci:
Closing because the key was removed in #111.

PR #102, opened 2021-12-21
grpcutil: Remove gRPC load balancer references

aknuds1:
**What this PR does**:
`grpcutil.Resolver.Resolve` assumes that [SRV records](https://en.wikipedia.org/wiki/SRV_record) will use "grpclb" for the service part (which typically corresponds to a Kubernetes service named port), but that isn't necessarily the case. I propose we parameterize it, in order to support systems not following this convention.

The PR, on request from @pstibrany, also generally removes references to gRPC load balancers as the code is no longer for this use case.

TODO: Consider adding test.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/grpcutil/dns_resolver.go
+++ b/grpcutil/dns_resolver.go
@@ -101,8 +101,11 @@ func parseTarget(target string) (host, port string, err error) {
 	return "", "", fmt.Errorf("invalid target address %v", target)
 }
 
-// Resolve creates a watcher that watches the name resolution of the target.
-func (r *Resolver) Resolve(target string) (Watcher, error) {
+// Resolve creates a watcher that watches the SRV/hostname record resolution of the target.
+//
+// If service is not empty, the watcher will first attempt to resolve an SRV record.
+// If that fails, or service is empty, hostname record resolution is attempted instead.
+func (r *Resolver) Resolve(target, service string) (Watcher, error) {
 	host, port, err := parseTarget(target)
 	if err != nil {
 		return nil, err
@@ -119,22 +122,24 @@ func (r *Resolver) Resolve(target string) (Watcher, error) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 	return &dnsWatcher{
-		r:      r,
-		logger: r.logger,
-		host:   host,
-		port:   port,
-		ctx:    ctx,
-		cancel: cancel,
-		t:      time.NewTimer(0),
+		r:       r,
+		logger:  r.logger,
+		host:    host,
+		port:    port,
+		service: service,
+		ctx:     ctx,
+		cancel:  cancel,
+		t:       time.NewTimer(0),
 	}, nil
 }
 
 // dnsWatcher watches for the name resolution update for a specific target
 type dnsWatcher struct {
 	r      *Resolver
-	logger log.Logger
-	host   string
-	port   string
+	logger  log.Logger
+	host    string
+	port    string
+	service string
 	// The latest resolved address set
 	curAddrs map[string]*Update
 	ctx      context.Context
@@ -203,8 +208,13 @@ func (w *dnsWatcher) compileUpdate(newAddrs map[string]*Update) []*Update {
 }
 
 func (w *dnsWatcher) lookupSRV() map[string]*Update {
+	if w.service == "" {
+		level.Debug(w.logger).Log("msg", "not looking up DNS SRV record since w.service is empty")
+		return nil
+	}
+
 	newAddrs := make(map[string]*Update)
-	_, srvs, err := lookupSRV(w.ctx, "grpclb", "tcp", w.host)
+	_, srvs, err := lookupSRV(w.ctx, w.service, "tcp", w.host)
 	if err != nil {
 		level.Info(w.logger).Log("msg", "failed DNS SRV record lookup", "err", err)
 		return nil
@@ -251,7 +261,7 @@ func (w *dnsWatcher) lookupHost() map[string]*Update {
 func (w *dnsWatcher) lookup() []*Update {
 	newAddrs := w.lookupSRV()
 	if newAddrs == nil {
-		// If failed to get any balancer address (either no corresponding SRV for the
+		// If we failed to get any balancer address (either no corresponding SRV for the
 		// target, or caused by failure during resolution/parsing of the balancer target),
 		// return any A record info available.
 		newAddrs = w.lookupHost()
--- a/grpcutil/dns_resolver.go
+++ b/grpcutil/dns_resolver.go
@@ -190,7 +190,6 @@ func (w *dnsWatcher) compileUpdate(newAddrs map[string]*Update) []*Update {
 
 func (w *dnsWatcher) lookupSRV() map[string]*Update {
 	if w.service == "" {
-		level.Debug(w.logger).Log("msg", "not looking up DNS SRV record since w.service is empty")
 		return nil
 	}
 
--- a/grpcutil/dns_resolver.go
+++ b/grpcutil/dns_resolver.go
@@ -169,26 +169,6 @@ func (i *ipWatcher) Close() {
 	close(i.updateChan)
 }
 
-// AddressType indicates the address type returned by name resolution.
-type AddressType uint8
-
-const (
-	// Backend indicates the server is a backend server.
-	Backend AddressType = iota
-	// GRPCLB indicates the server is a grpclb load balancer.
-	GRPCLB
-)
-
-// AddrMetadataGRPCLB contains the information the name resolver for grpclb should provide. The
-// name resolver used by the grpclb balancer is required to provide this type of metadata in
-// its address updates.
-type AddrMetadataGRPCLB struct {
-	// AddrType is the type of server (grpc load balancer or backend).
-	AddrType AddressType
-	// ServerName is the name of the grpc load balancer. Used for authentication.
-	ServerName string
-}
-
 // compileUpdate compares the old resolved addresses and newly resolved addresses,
 // and generates an update list
 func (w *dnsWatcher) compileUpdate(newAddrs map[string]*Update) []*Update {
@@ -232,8 +212,7 @@ func (w *dnsWatcher) lookupSRV() map[string]*Update {
 				continue
 			}
 			addr := a + ":" + strconv.Itoa(int(s.Port))
-			newAddrs[addr] = &Update{Addr: addr,
-				Metadata: AddrMetadataGRPCLB{AddrType: GRPCLB, ServerName: s.Target}}
+			newAddrs[addr] = &Update{Addr: addr}
 		}
 	}
 	return newAddrs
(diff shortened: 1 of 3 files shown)


replay (commented):
This makes sense to me, although I don't think I'm familiar with all the call-sites of `dnsResolver.Resolve()`

grpcutil/dns_resolver.go
```diff
@@ -204,7 +206,7 @@ func (w *dnsWatcher) compileUpdate(newAddrs map[string]*Update) []*Update {
 
 func (w *dnsWatcher) lookupSRV() map[string]*Update {
 	newAddrs := make(map[string]*Update)
-	_, srvs, err := lookupSRV(w.ctx, "grpclb", "tcp", w.host)
+	_, srvs, err := lookupSRV(w.ctx, w.service, "tcp", w.host)
```
pstibrany, line 209: Can we return early if `w.service == ""`?

grpcutil/naming.go
```diff
@@ -26,11 +26,11 @@ type Update struct {
 
 // Resolver creates a Watcher for a target to track its resolution changes.
 type Resolver interface {
-	// Resolve creates a Watcher for target.
-	Resolve(target string) (Watcher, error)
+	// Resolve creates a Watcher for target/service.
```
pstibrany, line 29: Why do we even *need* this interface in the first place? It is not *used* in dskit, so it should not exist here. We should export the implementation instead. (same for `Watcher`). That's probably a separate PR.

grpcutil/dns_resolver.go
```diff
@@ -101,8 +101,8 @@ func parseTarget(target string) (host, port string, err error) {
 	return "", "", fmt.Errorf("invalid target address %v", target)
 }
 
-// Resolve creates a watcher that watches the name resolution of the target.
-func (r *dnsResolver) Resolve(target string) (Watcher, error) {
+// Resolve creates a watcher that watches the SRV record resolution of the target.
```
pstibrany, line 104: It watches both. SRV record first, and if it fails, then it does hostname resolution. If we will skip SRV lookup for service == "", we should document it here.

aknuds1:
Thanks for the feedback @pstibrany - will take care of it :)

grpcutil/dns_resolver.go
aknuds1: @pstibrany Makes sense. Implemented it.

grpcutil/naming.go
aknuds1: @pstibrany Let's do it in a separate PR then?

grpcutil/dns_resolver.go
aknuds1: @pstibrany Thanks, improved the comment.

grpcutil/naming.go
pstibrany: Yes. Let's see what others think too.

grpcutil/dns_resolver.go
```diff
@@ -101,8 +101,11 @@ func parseTarget(target string) (host, port string, err error) {
 	return "", "", fmt.Errorf("invalid target address %v", target)
 }
 
-// Resolve creates a watcher that watches the name resolution of the target.
-func (r *dnsResolver) Resolve(target string) (Watcher, error) {
+// Resolve creates a watcher that watches the SRV/hostname record resolution of the target.
+//
+// First SRV record resolution is attempted for the specified service; if this fails, hostname record resolution is
+// attempted instead.
```
pstibrany, line 107:
```suggestion
// If service is not empty, watcher will first attempt to resolve SRV record.
// If that fails or service was empty, hostname record resolution is attempted instead.
```

grpcutil/naming.go
aknuds1: @pstibrany https://github.com/grafana/dskit/pull/105

I realized that the reason for the `Watcher` interface is that there are two different implementations: `dnsWatcher` and `ipWatcher`. I.e., getting rid if the interface is trickier. Do you still think we should do it?

grpcutil/dns_resolver.go
```diff
@@ -203,8 +208,13 @@ func (w *dnsWatcher) compileUpdate(newAddrs map[string]*Update) []*Update {
 }
 
 func (w *dnsWatcher) lookupSRV() map[string]*Update {
+	if w.service == "" {
+		level.Debug(w.logger).Log("msg", "not looking up DNS SRV record since w.service is empty")
```
pstibrany, line 212: If empty string was passed, that was deliberate decision and we don't need to log about it.

grpcutil/dns_resolver.go
```diff
@@ -251,7 +261,7 @@ func (w *dnsWatcher) lookupHost() map[string]*Update {
 func (w *dnsWatcher) lookup() []*Update {
 	newAddrs := w.lookupSRV()
 	if newAddrs == nil {
-		// If failed to get any balancer address (either no corresponding SRV for the
+		// If we failed to get any balancer address (either no corresponding SRV for the
```
pstibrany, line 264: If we're changing this line, we should remove mention of balancer (this code was doing loadbalancer lookup originally).

pstibrany: Note that there are other refences to GRPCLB in this code, that should be removed too.

grpcutil/dns_resolver.go
aknuds1: @pstibrany you probably want to say "If that fails, hostname record resolution is attempted instead." because your edited comment says that the watcher will only try SRV if service is non-empty.

grpcutil/dns_resolver.go
aknuds1: @pstibrany My reasoning for adding a _debug_ log line here is the library client may have made the mistake of passing an empty service parameter. Is it a problem to be a bit chatty on the debug level?

grpcutil/dns_resolver.go
aknuds1: @pstibrany Isn't looking up the SRV record a load balancing strategy? I.e. load balancing by picking one of the SRV record's targets?

aknuds1: @pstibrany If we want to remove other GRPCLB references, can we do that in another PR? Otherwise I feel as if we're mixing separate concerns.

grpcutil/dns_resolver.go
pstibrany: I find it unnecessary, but it's no big deal.

grpcutil/dns_resolver.go
pstibrany: It's a matter of usage... In Cortex we don't use it for load-balancing, but to discover *all* targets to connect to.

grpcutil/dns_resolver.go
aknuds1: Yeah, whether there's any point to it or not is up for discussion.

grpcutil/dns_resolver.go
pstibrany: > @pstibrany If we want to remove other GRPCLB references, can we do that in another PR? Otherwise I feel as if we're mixing separate concerns.

This PR started by removing "grpclb" string, so we may as well continue in that direction.

aknuds1: > It's a matter of usage... In Cortex we don't use it for load-balancing, but to discover all targets to connect to.

@pstibrany thanks for clarifying that.

I don't agree the PR is about removing the grpclb string though. It's about dropping the assumption that the service part of the SRV record is "grplb". If we remove GRPCLB references we'll reduce the public interface (of `Update`, which gets returned by `Watcher.Next`).

aknuds1: I don't know if the `AddrMetadataGRPCLB` type is useful in any way, but I'd rather deal with that separately.

pstibrany: My take is that we started with removal of hardcoded assumption that we *always* want to do "grpclb" service resolution, and instead may want to do other services too. However there is still lot of grpc-related naming in this package. I think we should drop the following:
- `AddressType`, `Default` and `GRPCLB` constants
- `AddrMetadataGRPCLB` (if we remove `AddressType`, there is only `ServerName` left, with comment "ServerName is the name of the grpc load balancer. Used for authentication." which makes no sense here.
- `Metadata` field from `Update`

I would stop there for this PR. WDYT?

aknuds1: @pstibrany I guess that makes sense. What do youj mean by the `Default` constant though? I dont' see any.

pstibrany: > @pstibrany I guess that makes sense. What do youj mean by the `Default` constant though? I don't see any.

Sorry, that's a typo. I meant `Backend`.

aknuds1: @pstibrany I removed the gRPC LB related code. Look good?

grpcutil/dns_resolver.go
```diff
@@ -101,8 +101,11 @@ func parseTarget(target string) (host, port string, err error) {
 	return "", "", fmt.Errorf("invalid target address %v", target)
 }
 
-// Resolve creates a watcher that watches the name resolution of the target.
-func (r *dnsResolver) Resolve(target string) (Watcher, error) {
+// Resolve creates a watcher that watches the SRV/hostname record resolution of the target.
+//
+// If service is not empty, the watcher will first attempt to resolve an SRV record.
```
pstibrany, line 106: This description is missing one more case:
```
// If target can be parsed as IP address, watcher will return it, and will not send any more updates afterwards.
```

pstibrany (approved):
LGTM, thank you for addressing my feedback!

pstibrany:
We should update PR title and description to mention that grpclb-specific code was removed.

One more (non-blocking) nit: can we drop "balancer" from log message and comments too?

aknuds1:
@pstibrany OK! I'm thinking of also adding a test.

aknuds1:
> We should update PR title and description to mention that grpclb-specific code was removed.
> 
> One more (non-blocking) nit: can we drop "balancer" from log message and comments too?

@pstibrany Done and done.

grpcutil/dns_resolver.go
aknuds1: Done.

pstibrany (approved):
Thank you again, nice work.

aknuds1:
@pstibrany Before I forget, should we test this change with Cortex in a test cell to see how it works out?

grpcutil/dns_resolver.go
pracucci: I also find this unnecessary. If we really want to keep it, then I would not reference `w.service` but have a more descriptive information, like "not looking up DNS SRV record because disabled".

pracucci (approved):

grpcutil/dns_resolver.go
aknuds1: @pracucci @pstibrany removing the log message since it's no big deal to me, at least until it turns out to be useful, and consensus seems to be it's not necessary.

PR #104, opened 2021-12-22
Remove package math

aknuds1:
**What this PR does**:
As we in the Cortex team have reached the consensus that the math package shouldn't have been included in dskit, I'm proposing to remove it.

**Checklist**
- [x] Tests updated


math/rate.go
```diff
@@ -1,59 +0,0 @@
...
-
-import (
-	"sync"
-	"time"
-
-	"go.uber.org/atomic"
-)
-
-// EwmaRate tracks an exponentially weighted moving average of a per-second rate.
-type EwmaRate struct {
```
pstibrany, line 11: Hmm, this actually looks useful across different projects.

pstibrany (approved):
Thank you!

math/rate.go
pstibrany: (I think we can remove it from now, and put it back with improved package structure in the future, if we want it).

PR #105, opened 2021-12-23
grpcutil: Convert Resolver into concrete type

aknuds1:
**What this PR does**:
Convert grpcutil.Resolver into concrete type instead of `interface`. Requested by @pstibrany.

**Checklist**
- [na] Tests updated

### DIFF
--- a/grpcutil/dns_resolver.go
+++ b/grpcutil/dns_resolver.go
@@ -29,8 +29,8 @@ var (
 
 // NewDNSResolverWithFreq creates a DNS Resolver that can resolve DNS names, and
 // create watchers that poll the DNS server using the frequency set by freq.
-func NewDNSResolverWithFreq(freq time.Duration, logger log.Logger) (Resolver, error) {
-	return &dnsResolver{
+func NewDNSResolverWithFreq(freq time.Duration, logger log.Logger) (*Resolver, error) {
+	return &Resolver{
 		logger: logger,
 		freq:   freq,
 	}, nil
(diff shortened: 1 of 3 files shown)


grpcutil/dns_resolver.go
```diff
@@ -131,7 +131,7 @@ func (r *dnsResolver) Resolve(target string) (Watcher, error) {
 
 // dnsWatcher watches for the name resolution update for a specific target
 type dnsWatcher struct {
```
pstibrany, line 133: I would remove `Watcher` too and expose `dnsWatcher` as Watcher. WDYT? Nothing in dskit needs `Watcher` interface.

pstibrany: https://github.com/grafana/dskit/pull/102#discussion_r774620029

From @aknuds1:
> I realized that the reason for the Watcher interface is that there are two different implementations: dnsWatcher and ipWatcher. I.e., getting rid of the interface is trickier. Do you still think we should do it?

aknuds1: @pstibrany The problem is we have a second implementation of `Watcher` that can also be returned from `Resolve`: `ipWatcher`.

pstibrany: You're right, `Watcher` hides the implementation. I should have checked it better. Let's keep it.

aknuds1: Implementation wise, we can solve it without an interface (fusing the two structs), but do we want to?

pstibrany: I don't think we want to merge them.

pstibrany (approved):
This makes sense to me. I don't think we should define interface for single-implementation. Projects needing such interface can easily do that. But let's see what others think.

aknuds1:
> This makes sense to me. I don't think we should define interface for single-implementation. Projects needing such interface can easily do that. But let's see what others think.

@pstibrany I agree that we shouldn't expose interfaces in this library unless for good reason.

pracucci (commented):
I don't have a strong opinion. This code was just copied from https://github.com/grpc/grpc-go/tree/v1.29.x/naming when it was removed from gRPC so, at the time, we kept it "as is".

pracucci (approved):
I see we're doing changes anyway in #102 so I'm good start diverging this from the original code.

PR #111, opened 2022-01-04
Remove ring component keys

owen-d:
Removes project-specific references from the `dskit` library.
Reference: https://github.com/grafana/dskit/pull/101


pstibrany (approved):
Thank you!

PR #114, opened 2022-01-11
Clarify the use of the passed-in context.

ywwg:
**What this PR does**:
Fixes a grammar typo, and also better communicates how the context will be used.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/services/service.go
+++ b/services/service.go
@@ -55,7 +55,7 @@ func (s State) String() string {
 //
 type Service interface {
 	// StartAsync starts Service asynchronously. Service must be in New State, otherwise error is returned.
-	// Context is used as a parent context for service own context.
+	// Context is used as a parent context for all of the service's own child operation contexts.
 	StartAsync(ctx context.Context) error
 
 	// AwaitRunning waits until service gets into Running state.


services/service.go
```diff
@@ -55,7 +55,7 @@ func (s State) String() string {
 //
 type Service interface {
 	// StartAsync starts Service asynchronously. Service must be in New State, otherwise error is returned.
-	// Context is used as a parent context for service own context.
+	// Context is used as a parent context for all of the service's own child operation contexts.
```
pstibrany, line 58: "All" - there is normally just one. Why do you think previous description wasn't clear?

aknuds1: From reading implementations of `StartAsync` (I can find two: `services.Manager` and `services.BasicService`), I have to agree with @pstibrany. AFAICS "child operation contexts" would only describe the `services.Manager` implementation, which just acts as a proxy to its managed Services. The current documentation describes the context usage well IMO.

pstibrany: > From reading implementations of `StartAsync` (I can find two: `services.Manager` and `services.BasicService`)

Note that `services.Manager` does *not* implement `Service` interface.

aknuds1: @pstibrany ugh, you're right. It only implements part of it (at least the `StartAsync` method). This is one reason I wish Go had explicit implementations (like Rust traits), would make reading code a lot easier :/

ywwg: firstly, "for service own context" is nongrammatical -- it should at least be "for the service's own context".

I lost a week because I was passing in a timeout context, thinking that it would be used for the startup phase only. I am trying to make it more clear that the context is persisted inside the service object and used repeatedly.

aknuds1: Agree that the language could be improved, but the PR changes the meaning of the comment :) I think the following would be the right phrasing (I refer to `ctx` as the argument):

```
// ctx is used as parent context for the service's own context.
```

pstibrany: Discussed privately with @ywwg. Paraphrased:

> There is only one method that starts the service, and it takes a context, that is then passed to the service.
>
> Context is "not used repeatedly" but during a lifetime of the service. If context becomes "finished", service will stop.
> This is a property of BasicService implementation.
>
> Let's fix the grammar part. Tying lifecycle of the service with cancellation of the context may be too strong requirement for generic Service contract.

PR #121, opened 2022-01-28
Allow to override default of `final-sleep` and change default to `0`

dimitarvdimitrov:
**What this PR does**:
Allows to override default of `final-sleep` and change default to `0`

Fixes #<issue number>

**Checklist**
- [ n/a ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@
 * [CHANGE] grpcutil.Update: Remove gRPC LB related metadata. #102
 * [CHANGE] concurrency.ForEach: deprecated and reimplemented by new `concurrency.ForEachJob`. #113
 * [CHANGE] ring/client: It's now possible to set different value than `consul` as default KV store. #120
-* [CHANGE] Lifecycler: It's now possible to change default value of lifecycler's `final-sleep` to other than `30s`. #121
+* [CHANGE] Lifecycler: Default value of lifecycler's `final-sleep` is now `0s` (i.e. no sleep). #121
+* [CHANGE] Lifecycler: It's now possible to change default value of lifecycler's `final-sleep`. #121
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
 * [ENHANCEMENT] Add limiter package. #41
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 * [CHANGE] grpcutil.Update: Remove gRPC LB related metadata. #102
 * [CHANGE] concurrency.ForEach: deprecated and reimplemented by new `concurrency.ForEachJob`. #113
 * [CHANGE] ring/client: It's now possible to set different value than `consul` as default KV store. #120
-* [CHANGE] Lifecycler: Change default value of `-[ingester.]final-sleep` (YAML: `[ingester.]final_sleep`) to `0s`. 
+* [CHANGE] Lifecycler: Change default value of `-[ingester.]final-sleep` (YAML: `[ingester.]final_sleep`) to `0s`.  #121
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
 * [ENHANCEMENT] Add limiter package. #41
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 * [CHANGE] grpcutil.Update: Remove gRPC LB related metadata. #102
 * [CHANGE] concurrency.ForEach: deprecated and reimplemented by new `concurrency.ForEachJob`. #113
 * [CHANGE] ring/client: It's now possible to set different value than `consul` as default KV store. #120
-* [CHANGE] Lifecycler: Change default value of `-[ingester.]final-sleep` (YAML: `[ingester.]final_sleep`) to `0s`.  #121
+* [CHANGE] Lifecycler: It's now possible to change default value of lifecycler's `final-sleep` to other than `30s`. #121
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
 * [ENHANCEMENT] Add limiter package. #41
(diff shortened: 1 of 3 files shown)


ring/lifecycler.go
```diff
@@ -68,7 +68,7 @@ func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.Flag
 	f.DurationVar(&cfg.JoinAfter, prefix+"join-after", 0*time.Second, "Period to wait for a claim from another member; will join automatically after this.")
 	f.DurationVar(&cfg.ObservePeriod, prefix+"observe-period", 0*time.Second, "Observe tokens after generating to resolve collisions. Useful when using gossiping ring.")
 	f.DurationVar(&cfg.MinReadyDuration, prefix+"min-ready-duration", 15*time.Second, "Minimum duration to wait after the internal readiness checks have passed but before succeeding the readiness endpoint. This is used to slowdown deployment controllers (eg. Kubernetes) after an instance is ready and before they proceed with a rolling update, to give the rest of the cluster instances enough time to receive ring updates.")
-	f.DurationVar(&cfg.FinalSleep, prefix+"final-sleep", 30*time.Second, "Duration to sleep for before exiting, to ensure metrics are scraped.")
+	f.DurationVar(&cfg.FinalSleep, prefix+"final-sleep", 0*time.Second, "Duration to sleep for before exiting, to ensure metrics are scraped.")
```
pstibrany, line 71: Please run `make doc`, and update CHANGELOG with affected options.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 * [CHANGE] grpcutil.Update: Remove gRPC LB related metadata. #102
 * [CHANGE] concurrency.ForEach: deprecated and reimplemented by new `concurrency.ForEachJob`. #113
 * [CHANGE] ring/client: It's now possible to set different value than `consul` as default KV store. #120
+* [CHANGE] Lifecycler: Change default value of `-[ingester.]final-sleep` (YAML: `[ingester.]final_sleep`) to `0s`.  #121
```
pstibrany, line 22: Let's not use ambiguous command line options here, we can just say "lifecycler's final-sleep" directly.

dimitarvdimitrov: cleaned it up a bit in https://github.com/grafana/dskit/pull/121/commits/66acab096ac4ccdb1523a9e9d6175e2e29516130

ring/lifecycler.go
```diff
@@ -63,12 +68,17 @@ func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.Flag
 		prefix = "ingester."
 	}
 
+	// Allow clients to override default value by setting it before calling this method.
+	if cfg.FinalSleep == 0 {
```
pstibrany, line 72: Hmm, 0 value for this check may actually not work that well... user (or in our case our package) may want to set it to 0 to *disable* final sleep.

pstibrany: I think we might need to bite the bullet, and set the default to 0 in dskit, but let it respect non-zero value when registering flag.

dimitarvdimitrov: technically the client can set it to a very short duration (1 ns). That will be the minimal sleep timer supported by the OS/CPU ([usually 1-2 _ms_](https://github.com/golang/go/issues/29485)). It's essentially disabled then.

dimitarvdimitrov: actully a better proposition. `time.Duration` is `int64`, so we can also accept `-1` as the disabled value. go's `time.Sleep` docs say that negative and zero values cause `time.Sleep` [to return immediately](https://pkg.go.dev/time#Sleep).

pstibrany: > technically the client can set it to a very short duration (1 ns). That will be the minimal sleep timer supported by the OS/CPU ([usually 1-2 _ms_](https://github.com/golang/go/issues/29485)). It's essentially disabled then.

I don't think asking user to set it to 1ms to disable is a good user experience.

> time.Duration is int64, so we can also accept -1 as the disabled value. go's time.Sleep docs say that negative and zero values cause time.Sleep to return immediately.

I considered this option, but the problem is that if user passes "0", he doesn't get "no wait", but 30 seconds instead. It's this weird behaviour that I'd like to avoid.

dimitarvdimitrov: so setting a negative value for the struct field will disable sleep (will still want to do an `if FinalSleep < 0 { FinalSleep = 0 }`) and `0` will be treated as unset and `=30s`

pstibrany: I propose that we actually change the default in `dskit` but also do:

```
f.DurationVar(&cfg.FinalSleep, prefix+"final-sleep", cfg.FinalSleep, "Duration to sleep for before exiting, to ensure metrics are scraped.")
```

This will also respect other values, set by code using `dskit`, so others can go back to 30s if they want. Negative and 0 values will cause no sleep.

dimitarvdimitrov: I agree that the `0 = 30s` is bad UX. Pushed in 
https://github.com/grafana/dskit/pull/121/commits/57a7fb141493ad06c0dac17f1ab0d31d823eecbb
the changes to remove that `if`

pstibrany (approved):
Thank you.

dimitarvdimitrov:
@pstibrany do you think we can merge this?

joe-elliott:
It makes sense to me to make this configurable, but is there a reason we'd change the default?

dimitarvdimitrov:
> It makes sense to me to make this configurable, but is there a reason we'd change the default?

I think the original reasoning was that this flag existed for the purpose of leaving time for one last final prometheus metrics scrape of the ingester. This scrape is important when using chunks storage, due to chunks transfers on shutdown. Chunks are deprecated and with blocks storage this is becoming less crucial. So `0s` is supposed to be a more sensible default.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 * [CHANGE] grpcutil.Update: Remove gRPC LB related metadata. #102
 * [CHANGE] concurrency.ForEach: deprecated and reimplemented by new `concurrency.ForEachJob`. #113
 * [CHANGE] ring/client: It's now possible to set different value than `consul` as default KV store. #120
+* [CHANGE] Lifecycler: Default value of lifecycler's `final-sleep` is now `0s` (i.e. no sleep). #121
+* [CHANGE] Lifecycler: It's now possible to change default value of lifecycler's `final-sleep`. #121
```
bboreham, line 23: Is this really a breaking change?

dimitarvdimitrov: good question. I don't think both are breaking changes. I'll submit a PR to change this

dimitarvdimitrov: opened https://github.com/grafana/dskit/pull/157 to address that

PR #126, opened 2022-02-01
autodetect interfaces on private networks

aldernero:
**What this PR does**:
This PR adds a new package `netutil` which includes a function `PrivateNetworkInterfaces` that scans all system network interfaces and return those that are on a private network. This will make some configuration steps easier, especially for systems that use consistent network device naming.

This is similar to [PR 100](https://github.com/grafana/dskit/pull/100) but takes advantage of the new `IsPrivate` function in Go 1.17 and includes unit tests.

Fixes #<issue number>

**Checklist**
- [X] Tests updated

### DIFF
--- a/netutil/netutil.go
+++ b/netutil/netutil.go
@@ -36,7 +36,6 @@ func privateNetworkInterfaces(all []net.Interface, logger log.Logger) []string {
 				level.Warn(logger).Log("msg", "error getting ip address", "inf", i.Name, "addr", s, "err", err)
 			}
 			if ip.IsPrivate() {
-				level.Info(logger).Log(i.Name, "is private with address", s)
 				privInts = append(privInts, i.Name)
 				break
 			}
--- a/netutil/netutil.go
+++ b/netutil/netutil.go
@@ -2,7 +2,6 @@ package netutil
 
 import (
 	"net"
-	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -12,11 +11,11 @@ var (
 	getInterfaceAddrs = (*net.Interface).Addrs
 )
 
-// Parses network interfaces and returns those having an address conformant to RFC1918
+// PrivateNetworkInterfaces lists network interfaces and returns those having an address conformant to RFC1918
 func PrivateNetworkInterfaces(logger log.Logger) []string {
 	ints, err := net.Interfaces()
 	if err != nil {
-		level.Warn(logger).Log("msg", "error getting interfaces", "err", err)
+		level.Warn(logger).Log("msg", "error getting network interfaces", "err", err)
 	}
 	return privateNetworkInterfaces(ints, logger)
 }
@@ -27,13 +26,13 @@ func privateNetworkInterfaces(all []net.Interface, logger log.Logger) []string {
 	for _, i := range all {
 		addrs, err := getInterfaceAddrs(&i)
 		if err != nil {
-			level.Warn(logger).Log("msg", "error getting addresses", "inf", i.Name, "err", err)
+			level.Warn(logger).Log("msg", "error getting addresses from network interface", "interface", i.Name, "err", err)
 		}
 		for _, a := range addrs {
 			s := a.String()
 			ip, _, err := net.ParseCIDR(s)
 			if err != nil {
-				level.Warn(logger).Log("msg", "error getting ip address", "inf", i.Name, "addr", s, "err", err)
+				level.Warn(logger).Log("msg", "error parsing network interface IP address", "inf", i.Name, "addr", s, "err", err)
 			}
 			if ip.IsPrivate() {
 				privInts = append(privInts, i.Name)
@@ -44,6 +43,6 @@ func privateNetworkInterfaces(all []net.Interface, logger log.Logger) []string {
 	if len(privInts) == 0 {
 		return []string{"eth0", "en0"}
 	}
-	level.Info(logger).Log("msg", "found interfaces on private networks:", "["+strings.Join(privInts, ", ")+"]")
+	level.Debug(logger).Log("msg", "found network interfaces with private IP addresses assigned", "interfaces", privInts)
 	return privInts
 }
--- a/netutil/netutil.go
+++ b/netutil/netutil.go
@@ -1,0 +1,50 @@
+package netutil
+
+import (
+	"net"
+	"strings"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+var (
+	getInterfaceAddrs = (*net.Interface).Addrs
+)
+
+// Parses network interfaces and returns those having an address conformant to RFC1918
+func PrivateNetworkInterfaces(logger log.Logger) []string {
+	ints, err := net.Interfaces()
+	if err != nil {
+		level.Warn(logger).Log("msg", "error getting interfaces", "err", err)
+	}
+	return privateNetworkInterfaces(ints, logger)
+}
+
+// private testable function that checks each given interface
+func privateNetworkInterfaces(all []net.Interface, logger log.Logger) []string {
+	var privInts []string
+	for _, i := range all {
+		addrs, err := getInterfaceAddrs(&i)
+		if err != nil {
+			level.Warn(logger).Log("msg", "error getting addresses", "inf", i.Name, "err", err)
+		}
+		for _, a := range addrs {
+			s := a.String()
+			ip, _, err := net.ParseCIDR(s)
+			if err != nil {
+				level.Warn(logger).Log("msg", "error getting ip address", "inf", i.Name, "addr", s, "err", err)
+			}
+			if ip.IsPrivate() {
+				level.Info(logger).Log(i.Name, "is private with address", s)
+				privInts = append(privInts, i.Name)
+				break
+			}
+		}
+	}
+	if len(privInts) == 0 {
+		return []string{"eth0", "en0"}
+	}
+	level.Info(logger).Log("msg", "found interfaces on private networks:", "["+strings.Join(privInts, ", ")+"]")
+	return privInts
+}
--- a/netutil/netutil.go
+++ b/netutil/netutil.go
@@ -17,11 +17,19 @@ func PrivateNetworkInterfaces(logger log.Logger) []string {
 	if err != nil {
 		level.Warn(logger).Log("msg", "error getting network interfaces", "err", err)
 	}
-	return privateNetworkInterfaces(ints, logger)
+	return privateNetworkInterfaces(ints, []string{}, logger)
+}
+
+func PrivateNetworkInterfacesWithFallback(fallback []string, logger log.Logger) []string {
+	ints, err := net.Interfaces()
+	if err != nil {
+		level.Warn(logger).Log("msg", "error getting network interfaces", "err", err)
+	}
+	return privateNetworkInterfaces(ints, fallback, logger)
 }
 
 // private testable function that checks each given interface
-func privateNetworkInterfaces(all []net.Interface, logger log.Logger) []string {
+func privateNetworkInterfaces(all []net.Interface, fallback []string, logger log.Logger) []string {
 	var privInts []string
 	for _, i := range all {
 		addrs, err := getInterfaceAddrs(&i)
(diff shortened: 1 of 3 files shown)


netutil/netutil_test.go
```diff
@@ -0,0 +1,126 @@
...
+	return testInts
+}
+
+func TestEmptyInterface(t *testing.T) {
+	ints := []net.Interface{}
+	getInterfaceAddrs = func(i *net.Interface) ([]net.Addr, error) {
+		return []net.Addr{}, nil
+	}
+	privInts := privateNetworkInterfaces(ints, logger)
+	if !reflect.DeepEqual(privInts, []string{"eth0", "en0"}) {
```
56quarters, line 59: Usually we use `github.com/stretchr/testify/require` or `github.com/stretchr/testify/assert` for this, e.g.

```go
assert.Equal(t, privInts, []string{"eth0", "en0"})
```

netutil/netutil.go
```diff
@@ -0,0 +1,50 @@
...
+			level.Warn(logger).Log("msg", "error getting addresses", "inf", i.Name, "err", err)
+		}
+		for _, a := range addrs {
+			s := a.String()
+			ip, _, err := net.ParseCIDR(s)
+			if err != nil {
+				level.Warn(logger).Log("msg", "error getting ip address", "inf", i.Name, "addr", s, "err", err)
+			}
+			if ip.IsPrivate() {
+				level.Info(logger).Log(i.Name, "is private with address", s)
```
56quarters, line 39: Logging in a library for something low-level like this feels a little too verbose to me. Maybe `debug` for most of these and returning an error in the top-level function? That way the caller can decide how to communicate the issue to the user.

bboreham: When you would you want this info logged, as a library user?  It feels like something you would want when you were coding the library.

bboreham (commented):
I'm not sure this code achieves what the name suggests.  Addresses can be NATted so even though the code sees an RFC1918 address it is actually public.

aldernero:
https://github.com/grafana/dskit/pull/126#discussion_r796821182
I agree, this was an oversight and should be removed

aldernero:
> I'm not sure this code achieves what the name suggests. Addresses can be NATted so even though the code sees an RFC1918 address it is actually public.

@bboreham Yes, that's true and I can see how the name would be misleading in that case. The two problems the function is trying to solve is to 1) return a list of real interface names in use on the system (especially since some distros don't us `en0` or `eth0`) and 2) filter that list based on IP's that tend to be associated with configuration options, e.g. ring configuration.

Do you think trying to address problem 2) in this way is useful and we just need to change the name so it doesn't imply the network is definitely private, or is the approach more fundamentally flawed? Is there some other filtering in 2) that would make configuration easier for users?

netutil/netutil.go
aldernero: Thanks for pointing this out, I meant to remove it.

zalegrala:
Hello, I don't have the context about how this will be used, but the pr looked interesting.  I'm one of those IPv6 dreamers, and just wanted to mention that perhaps some networks don't have an IPv4 address on them, and also don't conform to RFC1918, but would be considered private.

bboreham:
> Do you think trying to address problem 2) in this way is useful

Sure it sounds _useful_; I was worried about a potential security breach.  Many security headlines can be traced back to making something easier to use.

But I don't really know what you are trying to do - is there a more detailed spec somewhere?  This feature only sounds useful in the case where there is more than one non-local interface, which is not the common case.  I can't think of any foolproof way to tell which interface should be used for intra-system comms so I would let the admin do it.

netutil/netutil_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"os"
+	"testing"
+
+	"github.com/go-kit/log"
+	"github.com/stretchr/testify/assert"
+)
+
+// Setup required logger and example interface names and addresses
+var (
+	logger        log.Logger = log.NewLogfmtLogger(os.Stdout)
```
pracucci, line 14: I would remove the global var and just use `log.NewNopLogger()` whenever you need to pass the logger.

netutil/netutil_test.go
```diff
@@ -0,0 +1,116 @@
...
+		"privNetA": "10.6.19.34/8",
+		"privNetB": "172.16.0.7/12",
+		"privNetC": "192.168.3.29/24",
+		"pubNet":   "34.120.177.193/24",
+	}
+)
+
+// A type that implements the net.Addr interface
+// Only String() is called by netutil logic
+type MockAddr struct {
```
pracucci, line 25: [nit] No need to export it.

```suggestion
type mockAddr struct {
```

netutil/netutil_test.go
```diff
@@ -0,0 +1,116 @@
...
+			Flags:        0,
+		}
+		testInts = append(testInts, k)
+	}
+	return testInts
+}
+
+func TestEmptyInterface(t *testing.T) {
+	ints := []net.Interface{}
+	getInterfaceAddrs = func(i *net.Interface) ([]net.Addr, error) {
```
pracucci, line 55: It's a good practice cleaning up any change to global vars once the test has done. You can move the mocking logic to a function like this:

```

func mockGetInterfaceAddrs(t *testing.T, addrs []net.Addr) {
	prev := getInterfaceAddrs
	t.Cleanup(func() {
		getInterfaceAddrs = prev
	})

	getInterfaceAddrs = func(i *net.Interface) ([]net.Addr, error) {
		return addrs, nil
	}
}
```

netutil/netutil.go
```diff
@@ -0,0 +1,49 @@
...
+
+var (
+	getInterfaceAddrs = (*net.Interface).Addrs
+)
+
+// Parses network interfaces and returns those having an address conformant to RFC1918
+func PrivateNetworkInterfaces(logger log.Logger) []string {
+	ints, err := net.Interfaces()
+	if err != nil {
+		level.Warn(logger).Log("msg", "error getting interfaces", "err", err)
```
pracucci, line 19:
```suggestion
		level.Warn(logger).Log("msg", "error getting network interfaces", "err", err)
```

netutil/netutil_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"testing"
+
+	"github.com/go-kit/log"
+	"github.com/stretchr/testify/assert"
+)
+
+// Setup required logger and example interface names and addresses
+var (
+	logger        log.Logger = log.NewLogfmtLogger(os.Stdout)
+	testIntsAddrs            = map[string]string{
```
pracucci, line 15: Can we remove this global var at all? I think all your unit tests can be written in a single function using the go table testing pattern.

pracucci: I also would like to see the following tests:
- An interface with multiple addressed
- Unable to parse the address (shouldn't happen, but we handle it in the code so let's test it)

pracucci (commented):
I've read though the comments. I agree it's not guaranteed to list private network interfaces, but given the context where we wanna use it (populate config defaults) it's a good improvement over the current hardcoded config.

We need to move forward this PR. I left few comments.

netutil/netutil.go
```diff
@@ -0,0 +1,49 @@
...
+	"net"
+	"strings"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+var (
+	getInterfaceAddrs = (*net.Interface).Addrs
+)
```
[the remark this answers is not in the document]
replay: Ic, I have missed the access to the `.Name` property.
(one of my long-time pet-peeves with Go is that you can't define a property as part of an interface)

In that case I agree that this solution is easier, because otherwise we'd have to wrap `net.Interface` in something which allows us to return the name via a function call, just to make the mocking work. that gets unnecessarily complicated...

netutil/netutil.go
```diff
@@ -0,0 +1,49 @@
...
+	return privateNetworkInterfaces(ints, logger)
+}
+
+// private testable function that checks each given interface
+func privateNetworkInterfaces(all []net.Interface, logger log.Logger) []string {
+	var privInts []string
+	for _, i := range all {
+		addrs, err := getInterfaceAddrs(&i)
+		if err != nil {
+			level.Warn(logger).Log("msg", "error getting addresses", "inf", i.Name, "err", err)
```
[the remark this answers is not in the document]
aldernero: fixed in commit a0e79a212113d965ef90a75017145a0a4fa17074

netutil/netutil_test.go
aldernero: fixed in commit a0e79a212113d965ef90a75017145a0a4fa17074

netutil/netutil.go
aldernero: fixed in commit a0e79a212113d965ef90a75017145a0a4fa17074

netutil/netutil_test.go
aldernero: I got rid of the global variables and refactored the testing code using the go table testing pattern. I also added a test case for an address that cannot be parsed.

netutil/netutil_test.go
aldernero: I removed all global variables.

netutil/netutil_test.go
aldernero: fixed in commit a0e79a212113d965ef90a75017145a0a4fa17074

netutil/netutil.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+var (
+	getInterfaceAddrs = (*net.Interface).Addrs
+)
+
+// Parses network interfaces and returns those having an address conformant to RFC1918
+func PrivateNetworkInterfaces(logger log.Logger) []string {
```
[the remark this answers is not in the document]
aldernero: done in commit 55fff4203f126beebd5c3ad48def850e2940f6cb

pracucci:
I've upgraded dskit to golang 1.17 and rebased this PR.

ring/lifecycler.go
```diff
@@ -81,7 +82,7 @@ func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.Flag
 		panic(fmt.Errorf("failed to get hostname %s", err))
 	}
 
-	cfg.InfNames = []string{"eth0", "en0"}
+	cfg.InfNames = netutil.PrivateNetworkInterfaces(log.NewLogfmtLogger(os.Stdout))
```
pracucci, line 85: Take the logger in input to `RegisterFlagsWithPrefix()` (as 3rd parameter) and then in `RegisterFlags()` as 2nd parameter. It's a pattern we already use elsewhere and there's an interface defined for that (see `RegistererWithLogger` interface in `flagext/register.go`).

netutil/netutil.go
```diff
@@ -0,0 +1,48 @@
...
+	for _, i := range all {
+		addrs, err := getInterfaceAddrs(&i)
+		if err != nil {
+			level.Warn(logger).Log("msg", "error getting addresses from network interface", "interface", i.Name, "err", err)
+		}
+		for _, a := range addrs {
+			s := a.String()
+			ip, _, err := net.ParseCIDR(s)
+			if err != nil {
+				level.Warn(logger).Log("msg", "error parsing network interface IP address", "inf", i.Name, "addr", s, "err", err)
```
pracucci, line 35:
```suggestion
				level.Warn(logger).Log("msg", "error parsing network interface IP address", "interface", i.Name, "addr", s, "err", err)
```

pracucci, line 35: You should also `continue` here

netutil/netutil.go
```diff
@@ -0,0 +1,48 @@
...
+			}
+			if ip.IsPrivate() {
+				privInts = append(privInts, i.Name)
+				break
+			}
+		}
+	}
+	if len(privInts) == 0 {
+		return []string{"eth0", "en0"}
+	}
```
pracucci, line 45: Let's not hardcode it. Take the fallback in input.

netutil/netutil_test.go
```diff
@@ -0,0 +1,126 @@
...
+			HardwareAddr: []byte{},
+			Flags:        0,
+		}
+		testInts = append(testInts, k)
+	}
+	return testInts
+}
+
+func TestPrivateInterface(t *testing.T) {
+	testIntsAddrs := map[string][]string{
```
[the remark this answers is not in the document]
aldernero: There is another test below that one, `TestPrivateInterfaceError` that tests an invalid address.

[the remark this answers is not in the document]
pracucci: Doh!

pracucci (approved):
LGTM, thanks!

PR #128, opened 2022-02-01
lifecycler: Remove parameters from YAML

simonswine:
**What this PR does**:

Removes `lifecycler` parameters, which according to the comment should only be used in tests.

**Checklist**
- [N/A] Tests updated
- [N/A] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -43,9 +43,9 @@ type LifecyclerConfig struct {
 	ReadinessCheckRingHealth bool          `yaml:"readiness_check_ring_health" category:"advanced"`
 
 	// For testing, you can override the address and ID of this ingester
-	Addr string `yaml:"address" doc:"hidden"`
-	Port int    `doc:"hidden"`
-	ID   string `doc:"hidden"`
+	Addr string `yaml:"address" doc:"hidden" category:"advanced"`
+	Port int    `doc:"hidden" category:"advanced"`
+	ID   string `doc:"hidden" category:"advanced"`
 
 	// Injected internally
 	ListenPort int `yaml:"-"`


pracucci (commented):
I'm not sure they're just used in tests. They were exposed also as CLI flags and we actually allow to set them in our similar configs in Cortex. I don't think we should remove them from YAML.

Logiraptor:
I'm not sure under what circumstance you would want to set these, and I can't find any evidence of them being set by customers. But I can confirm that similar options are configurable in cortex for various rings. 

I would vote to mark them as advanced parameters, which would be more in line with the other ring configurations.

pracucci:
> I'm not sure under what circumstance you would want to set these

They're atypical use cases, but:
1. `Addr`: when the auto-detected address (querying the network interfaces) is not the one you want to register in the ring
2. `Port`: when the port you want to register in the ring is different than the gRPC listening port. No good use case is coming to my mind right now, I'm wondering if a case could be if you have a proxy like istio in between.
3. `ID`: it's the ID of the instance registered in the ring. By default it's the hostname, but you may want to customize it. For example, if you run 2 instances on the same bare metal. We also used it during some migrations (eg. between K8S clusters, to differentiate between ingesters in cluster A and the ones in cluster B).

simonswine:
I think I tend to agree with @pracucci's assessment. 

So should we classify those flags as `advanced` and also add a fitting `doc` annotation?

pracucci:
I would classify them as advanced. I wouldn't remove `doc:"hidden"` **yet** because the default value is dynamic (at least for ID and Addr) and we're not ready with the cortex doc auto-generator to address it (it's something we wanna address, discussed with @Logiraptor too).

pracucci (approved):

Logiraptor (approved):

Issue #131, opened 2022-02-04
Wrong codec when using multiple codec with in-memory kv store.


cyriltovena:
When creating multiple client with different codec, only the first registered codec will be used when using the in-memory KV.

This has caused me multiple issues while trying to use another in-memory store in loki.

see https://github.com/grafana/loki/blob/main/vendor/github.com/grafana/dskit/kv/client.go#L140-L142

PR #132, opened 2022-02-07
Allow in-memory kv-client to support multiple codec

cyriltovena:
**What this PR does**:

This PR allows to create multiple in-memory client with different codec. Before it would always the first codec registered because of the singleton.

Fixes https://github.com/grafana/dskit/issues/131

**Checklist**
- [x] Tests updated

### DIFF
--- a/kv/client.go
+++ b/kv/client.go
@@ -33,8 +33,10 @@ func (r *role) Labels() prometheus.Labels {
 // The NewInMemoryKVClient returned by NewClient() is a singleton, so
 // that distributors and ingesters started in the same process can
 // find themselves.
-var inmemoryStoreInit sync.Once
-var inmemoryStore Client
+var (
+	inmemoryStoreInit sync.Once
+	inmemoryStore     *consul.Client
+)
 
 // StoreConfig is a configuration used for building single store client, either
 // Consul, Etcd, Memberlist or MultiClient. It was extracted from Config to keep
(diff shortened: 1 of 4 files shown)


kv/client_test.go
```diff
@@ -170,3 +171,38 @@ func TestDefaultStoreValue(t *testing.T) {
...
+
+func TestMultipleInMemoryClient(t *testing.T) {
+	logger := log.NewJSONLogger(os.Stdout)
+	foo, err := NewClient(Config{
+		Store: "inmemory",
+	}, stringCodec{value: "foo"}, prometheus.NewRegistry(), logger)
+	require.NoError(t, err)
+	bar, err := NewClient(Config{
+		Store: "inmemory",
+	}, stringCodec{value: "bar"}, prometheus.NewRegistry(), logger)
```
cyriltovena, line 196: I can't reuse the same registry see https://github.com/grafana/dskit/issues/133

kv/client_test.go
```diff
@@ -170,3 +171,38 @@ func TestDefaultStoreValue(t *testing.T) {
...
+	return c.value, nil
+}
+
+func (c stringCodec) Encode(interface{}) ([]byte, error) {
+	return []byte(c.value), nil
+}
+func (c stringCodec) CodecID() string { return c.value }
+
+func TestMultipleInMemoryClient(t *testing.T) {
+	logger := log.NewJSONLogger(os.Stdout)
```
bboreham, line 189: Do we want this in the test forever?

kv/client.go
```diff
@@ -146,7 +148,8 @@ func createClient(backend string, prefix string, cfg StoreConfig, codec codec.Co
 		inmemoryStoreInit.Do(func() {
 			inmemoryStore, _ = consul.NewInMemoryClient(codec, logger, reg)
 		})
-		client = inmemoryStore
+		// however we swap the codec so that we can encode different type of values.
+		client = inmemoryStore.Clone().WithCodec(codec)
```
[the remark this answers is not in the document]
cyriltovena: I can get rid of Clone and just use `WithCodec` + a comment.

bboreham (approved):
seems legit

PR #139, opened 2022-02-11
Add doc:default tag to ring lifecycler network interfaces field

pracucci:
**What this PR does**:
https://github.com/grafana/dskit/pull/126 has introduced dynamic config default for network interfaces. This breaks the doc generator, cause the default value changes depending on which machine you're running on. I'm adding the support for `doc:"default=foo"` in the doc generator, to override it.

This PR leverages on that.

N/A

**Checklist**
- [ ] Tests updated


pstibrany (approved):
👍

PR #143, opened 2022-03-22
Add tenant package

simonswine:
We would really benefit for having a shared implementation of the tenant resolver across our products. It would make it easier to share code on the enterprise versions of Tempo and Loki with having a common implementation of the interface.

It was previously argued that dskit is the wrong place for this (see https://github.com/grafana/dskit/pull/33). I would like to restart this conversation (cc @jeschkies @krajorama)

### DIFF
--- a/tenant/resolver.go
+++ b/tenant/resolver.go
@@ -1,0 +1,132 @@
+package tenant
+
+import (
+	"context"
+	"net/http"
+	"strings"
+
+	"github.com/weaveworks/common/user"
+)
+
+var defaultResolver Resolver = NewSingleResolver()
+
+// WithDefaultResolver updates the resolver used for the package methods.
+func WithDefaultResolver(r Resolver) {
+	defaultResolver = r
+}
+
+// TenantID returns exactly a single tenant ID from the context. It should be
+// used when a certain endpoint should only support exactly a single
+// tenant ID. It returns an error user.ErrNoOrgID if there is no tenant ID
+// supplied or user.ErrTooManyOrgIDs if there are multiple tenant IDs present.
+//
+// ignore stutter warning
+//nolint:golint
+func TenantID(ctx context.Context) (string, error) {
+	return defaultResolver.TenantID(ctx)
+}
+
+// TenantIDs returns all tenant IDs from the context. It should return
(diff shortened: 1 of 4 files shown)


jeschkies:
> Closing after discussing with @pstibrany; he thinks these packages might as well be copied into Loki and I agree.

[This](https://github.com/grafana/dskit/pull/33#issuecomment-920216806) was the closing comment by @aknuds1. Turns out we need to have one package. I encountered an issue in GEL which imports Loki and [backend-enterprise-libs](https://github.com/grafana/backend-enterprise-libs). The imports result in two tenant packages one from Loki and one from Mimir via backend-enterprise-libs. As a result calls like `tenant.WithDefaultResolver` or `tenant.TanantIDs` have different behavior.

jeschkies (approved):

krajorama (approved):
Having the correct direction of dependencies and number of dependencies trumps any misgivings I have about putting this stuff into dskit. And if dskit becomes too bloated , we can still split it later. So I'm for this.

simonswine:
Despite having the approvals here I would like to give others a chance to raise their concerns. I think this has been found to be not too well aligned with the stated goal of: _This library contains utilities that are useful for building distributed services._

/cc @pstibrany @bboreham

A potential alternative could be to add this to `weaveworks/commons` instead together with renaming `user` to `tenant`: https://github.com/weaveworks/common/tree/master/user

pstibrany:
I would be more in favor of this PR if tenants were actually used by other parts of dskit.

If the primary purpose is "We would really benefit for having a shared implementation of the tenant resolver across our products", then we should perhaps introduce "backend-databases-commons", so that dskit can focus on what it was intended for. I think weaveworks-commons may be a good place for "tenant" package too, if PR was accepted there.

I will not block the PR. Multitenancy is often required in distributed systems, so having it in dskit isn't too far off.

pracucci (approved):
I both see good arguments that this is not really something to build distributed systems, but it's also a foundation of any multi-tenant system (which is a common requirement for distributed systems we build at Grafana Labs), so **this PR LGTM**.

> A potential alternative could be to add this to weaveworks/commons

I would discourage using any repo which is not under our direct control, cause may be more difficult (or slower) to iterate on.

PR #147, opened 2022-03-29
Extract status page templates to files

colega:
**What this PR does**:

Instead of having a const with a huge string, now we have `.gohtml` files that are properly understood by the IDEs (at least by Goland) and we embed them into global string vars using the `//go:embed` directive.

Also standarized how both status pages templates for ring and memberlist are initialized.

None, cleaning up for future refactor.

**Checklist**
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]` (not needed since this only changes unexported vars)

### DIFF
--- a/kv/memberlist/kv_init_service.go
+++ b/kv/memberlist/kv_init_service.go
@@ -2,6 +2,7 @@ package memberlist
 
 import (
 	"context"
+	_ "embed"
 	"encoding/json"
 	"fmt"
 	"html/template"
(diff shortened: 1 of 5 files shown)


pstibrany:
❤️ `go:embed` still feels like cheating, it's so nice to have that in the language.

Issue #148, opened 2022-03-29
Allow custom status page templates


colega:
# Issue

Upstream projects may want to render pages in different ways, applying custom branding where necessary. 

# Solution

Allow using custom templates when instantiating the ring or memberlist.

PR #149, opened 2022-03-29
Clarify and clean up template contract for status

colega:
**What this PR does**:

Ring was passing a mix of time.Time and formatted string for the time: unified to pass always time.Time, and decide what to do when rendering. This will allow us some extra customisation if replacing the template with a non-default one.

Also replaced the "Unhealthy" state by "UNHEALTHY" in the template, because the rest of states are uppercase. Not replacing the constant value because it's being used for metrics and we'd have to fix the recording rules in all the downstream projects.

Changed how heartbeat time is rendered: rendering the duration since the heartbeat and the time in parenthesis: heartbeats are so common that showing a date doesn't make sense, and since they're usually few seconds away, the "duration ago" is what makes more sense IMO.

Relates to https://github.com/grafana/dskit/issues/148

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * [CHANGE] ring/client: It's now possible to set different value than `consul` as default KV store. #120
 * [CHANGE] Lifecycler: Default value of lifecycler's `final-sleep` is now `0s` (i.e. no sleep). #121
 * [CHANGE] Lifecycler: It's now possible to change default value of lifecycler's `final-sleep`. #121
+* [CHANGE] Minor cosmetic changes in ring and memberlist HTTP status templates. #149
 * [ENHANCEMENT] Add middleware package. #38
 * [ENHANCEMENT] Add the ring package #45
 * [ENHANCEMENT] Add limiter package. #41
--- a/kv/memberlist/kv_init_service.go
+++ b/kv/memberlist/kv_init_service.go
@@ -171,21 +171,13 @@ func (kvs *KVInitService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if strings.Contains(accept, "application/json") {
 		w.Header().Set("Content-Type", "application/json")
 
-		data, err := json.Marshal(v)
-		if err != nil {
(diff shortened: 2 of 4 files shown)


kv/memberlist/kv_init_service.go
```diff
@@ -171,21 +171,13 @@ func (kvs *KVInitService) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-		// We ignore errors here, because we cannot do anything about them.
-		// Write will trigger sending Status code, so we cannot send a different status code afterwards.
-		// Also this isn't internal error, but error communicating with client.
-		_, _ = w.Write(data)
 		return
 	}
 
-	err := defaultPageTemplate.Execute(w, v)
-	if err != nil {
+	if err := defaultPageTemplate.Execute(w, v); err != nil {
```
pstibrany, line 180: We should set `Content-Type` in this branch too.

pstibrany (approved):
Nice, thank you!

kv/memberlist/kv_init_service.go
colega: Added it here: https://github.com/grafana/dskit/pull/149/commits/d2926ce863a81aad855a73401b9069c03f8d219f

pstibrany: Thank you.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 * [CHANGE] ring/client: It's now possible to set different value than `consul` as default KV store. #120
 * [CHANGE] Lifecycler: Default value of lifecycler's `final-sleep` is now `0s` (i.e. no sleep). #121
 * [CHANGE] Lifecycler: It's now possible to change default value of lifecycler's `final-sleep`. #121
+* [CHANGE] Minor cosmetic changes in ring and memberlist HTTP status templates. #149
```
bboreham, line 25: Is this really a breaking change?  Looks like the output is for humans.

colega: It is definitely not breaking, but it's a change. Would you prefer this to be an `[ENHANCEMENT]`?

PR #158, opened 2022-04-13
wrong comment

3Xpl0it3r:
**What this PR does**:
fix a wrong description in comment

+ // If RunningFn or StoppingFn return error
- // If RunningFn or Stopping return error

for Stopping is a state, is not a function


pracucci (approved):
Thanks!

PR #163, opened 2022-05-04
Customizable memberlist HTTP status handler

colega:
**What this PR does**:

Extracts the memberlist's HTTP status handler into a separate entity with an injectable template, so users can provide their own.

Code is backwards compatible so `KVInitService` still implements `http.Handler` with the default template.

None, followup on https://github.com/grafana/dskit/pull/144
This will allow customizing the template in Mimir with consistent CSS styling.

**Checklist**
- [ ] Tests updated


kv/memberlist/http_status_handler.go
```diff
@@ -0,0 +1,217 @@
...
+	SentMessages     []Message
+	ReceivedMessages []Message
+}
+
+// NewHTTPStatusHandler creates a new HTTPStatusHandler that will render the provided template using the data from StatusPageData.
+func NewHTTPStatusHandler(kvs *KVInitService, tpl *template.Template) HTTPStatusHandler {
+	return HTTPStatusHandler{kvs, tpl}
+}
+
+func (h HTTPStatusHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
```
pracucci, line 38: Note to reviewers: this function is the same exact code of the previous `KVInitService.ServeHTTP()`

kv/memberlist/http_status_handler.go
```diff
@@ -0,0 +1,217 @@
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "text/html")
+	if err := h.tpl.Execute(w, v); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
+}
+
+func getFormat(req *http.Request) string {
```
pracucci, line 127: Note to reviewers: these utility functions have just been copied as-is from `kv_init_service.go`

kv/memberlist/kv_init_service.go
```diff
@@ -95,191 +86,5 @@ func (kvs *KVInitService) stopping(_ error) error {
...
-
-	w.Header().Add("content-type", "application/octet-stream")
-	// Set content-length so that client knows whether it has received full response or not.
-	w.Header().Add("content-length", strconv.Itoa(len(encoded)))
-	w.Header().Add("content-disposition", fmt.Sprintf("attachment; filename=%d-%s", val.version, key))
-	w.WriteHeader(200)
-
-	// Ignore errors, we cannot do anything about them.
-	_, _ = w.Write(encoded)
+	NewHTTPStatusHandler(kvs, defaultPageTemplate).ServeHTTP(w, req)
```
pracucci, line 89: [nit] Instead of creating an handler for each request, what if we create it once and just reuse it?

pracucci (approved):
LGTM! I left a nit, but up to you if you want to address it (I don't feel strong).

kv/memberlist/http_status_handler.go
colega: Yes, it's actually straightforward if you check individual commits, I did the code extraction in the first commit.

kv/memberlist/kv_init_service.go
colega: Same pattern is followed in other places (ring status handler) and it doesn't really cause an overhead (just an interface allocation, and even if it did cause an overhead... this is an http handler, so the impact is neglible). 

OTOH, having an instantiated handler in the kvs would mean that there's a field in kvs that refers to kvs, which is something I dislike.

Since there's no hard reason to have the field, I'd prefer keeping it this way.

pracucci: Sure!

PR #164, opened 2022-05-05
Explain why memberlist message history is empty

colega:
This section is almost always empty, unless the buffer is not zero. Mention that on the status page to make troubleshooting easier.

I didn't want to mention the specific configuration option or link since that may depend on the service that is using the library.

This is how it would look now:

image

And when history is enabled:

image


pstibrany (approved):
Good idea.

PR #165, opened 2022-05-05
Lifecycler: if the ring backend storage is reset, the instance adds itself back to the ring with an updated registration timestamp set to current time

pracucci:
**What this PR does**:
This PR introduces a change to `Lifecycler` and `BasicLifecycler` to fix the issue described in this comment:
https://github.com/grafana/mimir/issues/1766#issuecomment-1116147117

The TL;DR is:
- Shuffle sharding lookback period uses the registration timestamp to check which instances should be queried
- If the ring is running on consul or etcd and the storage is reset/flushed, then each instance lifecycler will add itself back to the ring (this will happen at a slightly different time for each instance)
- Since the instance is adding back to the ring, it should set a registration timestamp set to "now" (like if they're joining the ring for the first time) otherwise the shuffle sharding lookback doesn't work as expected and could lead to partial query results (it's an issue we experienced in Mimir)

Part of https://github.com/grafana/mimir/issues/1766

**Checklist**
- [ ] Tests updated

### DIFF
--- a/ring/basic_lifecycler_test.go
+++ b/ring/basic_lifecycler_test.go
@@ -292,14 +292,14 @@ func TestBasicLifecycler_HeartbeatWhileStopping(t *testing.T) {
 	assert.True(t, onStoppingCalled)
 }
 
-func TestBasicLifecycler_HeartbeatAfterBackendRest(t *testing.T) {
+func TestBasicLifecycler_HeartbeatAfterBackendReset(t *testing.T) {
 	ctx := context.Background()
 	cfg := prepareBasicLifecyclerConfig()
 	cfg.HeartbeatPeriod = 10 * time.Millisecond
 
 	lifecycler, delegate, store, err := prepareBasicLifecycler(t, cfg)
 	require.NoError(t, err)
-	defer services.StopAndAwaitTerminated(ctx, lifecycler) //nolint:errcheck
+	t.Cleanup(func() { require.NoError(t, services.StopAndAwaitTerminated(ctx, lifecycler)) })
 
 	registerTokens := Tokens{1, 2, 3, 4, 5}
 	delegate.onRegister = func(_ *BasicLifecycler, _ Desc, _ bool, _ string, _ InstanceDesc) (state InstanceState, tokens Tokens) {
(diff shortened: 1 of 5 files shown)


ring/basic_lifecycler_test.go
```diff
@@ -323,9 +327,12 @@ func TestBasicLifecycler_HeartbeatAfterBackendRest(t *testing.T) {
...
 			desc.GetState() == ACTIVE &&
 			Tokens(desc.GetTokens()).Equals(registerTokens) &&
-			desc.GetAddr() == cfg.Addr &&
-			desc.GetRegisteredAt().Unix() == expectedRegisteredAt.Unix()
+			desc.GetAddr() == cfg.Addr
 	})
+
+	// Ensure the registration timestamp has been updated.
+	desc, _ := getInstanceFromStore(t, store, testInstanceID)
+	assert.Greater(t, desc.GetRegisteredTimestamp(), prevRegisteredAt.Unix())
```
pstibrany, line 335: We should check that `lifecycler.GetRegisteredAt()` is also correct.

ring/lifecycler_test.go
```diff
@@ -254,6 +254,66 @@ func TestLifecycler_ShouldHandleInstanceAbruptlyRestarted(t *testing.T) {
...
+	}))
+
+	test.Poll(t, time.Second, true, func() interface{} {
+		_, ok := getInstanceFromStore(t, store, testInstanceID)
+		return ok
+	})
+
+	// Ensure the registration timestamp has been updated.
+	desc, _ := getInstanceFromStore(t, store, testInstanceID)
+	assert.Greater(t, desc.GetRegisteredTimestamp(), prevRegisteredAt.Unix())
```
pstibrany, line 307: We should also check `lifecycler.getRegisteredAt()` here.

pstibrany (approved):
LGTM, fix makes sense, and makes no difference if shuffle sharding is not used.

colega (approved):

ring/lifecycler_test.go
pracucci: Done!

ring/basic_lifecycler_test.go
pracucci: Done!

PR #167, opened 2022-05-17
Add functionality to clear tokens on lifecycler shutdown

chaudum:
**What this PR does**:

This PR adds a flag to the ingester lifecycler that, when set to `true`,
allows the component to delete the ring token file if the
`-<prefix>.tokens-file-path` is set to persist tokens across restarts.

This enables components that use the lifecycler to implement a handler
that deletes the ring token file to "clear" tokens from the instance and
force a re-distribution of them when the instance is restarted.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -520,7 +520,7 @@ heartbeatLoop:
 		if err := os.Remove(i.cfg.TokensFilePath); err != nil {
 			return errors.Wrapf(err, "failed to delete tokens file %s", i.cfg.TokensFilePath)
 		}
-		level.Info(i.logger).Log("msg", "removed tokens file from disk", "filename", i.cfg.TokensFilePath)
+		level.Info(i.logger).Log("msg", "removed tokens file from disk", "path", i.cfg.TokensFilePath)
 	}
 
 	return nil
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -110,8 +110,9 @@ type Lifecycler struct {
 	Zone     string
 
 	// Whether to flush if transfer fails on shutdown.
-	flushOnShutdown      *atomic.Bool
-	unregisterOnShutdown *atomic.Bool
+	flushOnShutdown       *atomic.Bool
+	unregisterOnShutdown  *atomic.Bool
+	clearTokensOnShutdown *atomic.Bool
 
 	// We need to remember the ingester state, tokens and registered timestamp just in case the KV store
 	// goes away and comes back empty. The state changes during lifecycle of instance.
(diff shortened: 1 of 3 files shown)


pracucci:
Have you considered moving to `BasicLifecycler` and use an hook to do it?

ring/lifecycler.go
```diff
@@ -510,11 +511,18 @@ heartbeatLoop:
...
+			return errors.Wrapf(err, "failed to unregister from the KV store, ring: %s", i.RingName)
 		}
 		level.Info(i.logger).Log("msg", "instance removed from the KV store", "ring", i.RingName)
 	}
 
+	if i.cfg.TokensFilePath != "" && i.ClearTokensOnShutdown() {
+		if err := os.Remove(i.cfg.TokensFilePath); err != nil {
+			return errors.Wrapf(err, "failed to delete tokens file %s", i.cfg.TokensFilePath)
+		}
+		level.Info(i.logger).Log("msg", "removed tokens file from disk", "filename", i.cfg.TokensFilePath)
```
pracucci, line 523: To keep it consistent with existing logs (and also because "filename" typically refers to the name of the file, not the whole path).

```suggestion
		level.Info(i.logger).Log("msg", "removed tokens file from disk", "path", i.cfg.TokensFilePath)
```

pracucci (commented):
> This enables components that use the lifecycler to implement a handler
> that deletes the ring token file to "clear" tokens from the instance and
> force a re-distribution of them when the instance is restarted.

I don't fully understand the use case. If you need to clear tokens, then why using the tokens on disk at all? Couldn't you just not use it (set to empty path)?

chaudum:
> If you're using it in a new project, have you considered moving to `BasicLifecycler` and use an hook to do it? (otherwise ignore this comment, cause migrating from `Lifecycler` to `BasicLifecycler` may be non trivial)

It's for the Loki ingester, so unfortunately, as you said, migrating to `BasicLifecycler` is not that easy. But I will keep that in mind :)

chaudum:
> > This enables components that use the lifecycler to implement a handler
> > that deletes the ring token file to "clear" tokens from the instance and
> > force a re-distribution of them when the instance is restarted.
> 
> I don't fully understand the use case. If you need to clear tokens, then why using the tokens on disk at all? Couldn't you just not use it (set to empty path)?

This would be for a one-off clearance of tokens of an instance, when you generally run the component with persisted tokens.

I agree, you can do that manually as well by just deleting the file by hand, gracefully shutting down the service and restart the process (Pod), but we would like to do that in a handler of an HTTP endpoint, see https://github.com/grafana/loki/pull/6179
You might have a better idea to achieve this in the lifecycler :-)

pracucci (approved):
LGTM

> This would be for a one-off clearance of tokens of an instance, when you generally run the component with persisted tokens.

Ok, I see. Thanks for the details!

chaudum:
@pracucci Changed log line as suggested and added changelog entry.

PR #172, opened 2022-05-25
make ingesters use heartbeat timeout instead of period to fix the bug…

zenador:
… where they sometimes appear as unhealthy

**What this PR does**:

Same as title

Fixes [#1917](https://github.com/grafana/mimir/issues/1917)

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,3 +56,4 @@
 * [BUGFIX] Modules: fix a module waiting for other modules depending on it before stopping. #141
 * [BUGFIX] Multi KV: fix panic when using function to modify primary KV store in runtime. #153
 * [BUGFIX] Lifecycler: if the ring backend storage is reset, the instance adds itself back to the ring with an updated registration timestamp set to current time. #165
+* [BUGFIX] Ring: fix bug where ingesters may appear unhealthy even though they are not #172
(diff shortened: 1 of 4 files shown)


zenador:
Also, can someone familiar with this code confirm that this line is meant to use HeartbeatPeriod not HeartbeatTimeout and does not need to be changed? https://github.com/grafana/dskit/blob/ed0225d378876928b44b94e1034d533275abb57f/ring/lifecycler.go#L613

CHANGELOG.md
```diff
@@ -56,3 +56,4 @@
 * [BUGFIX] Modules: fix a module waiting for other modules depending on it before stopping. #141
 * [BUGFIX] Multi KV: fix panic when using function to modify primary KV store in runtime. #153
 * [BUGFIX] Lifecycler: if the ring backend storage is reset, the instance adds itself back to the ring with an updated registration timestamp set to current time. #165
+* [BUGFIX] Ring: fix bug where ingesters may appear unhealthy even though they are not #172
```
pracucci, line 59: This bug was reported for ingesters, but could potentially affect any hash ring, so I would generalise it. Moreover, dskit is a general purpose library and it has no notion of "ingesters" (unfortunately, due to some legacy code there are still references to it in dskit, but that's something we should cleanup).

```suggestion
* [BUGFIX] Ring: fix bug where hash ring instances may appear unhealthy in the web UI even though they are not. #172
```

pracucci:
> Also, can someone familiar with this code confirm that this line is meant to use HeartbeatPeriod not HeartbeatTimeout and does not need to be changed?

This is correct. In `lifecycler.go:613` we check if heartbeating has been disabled (it's disabled when `HeartbeatPeriod == 0`). Running the hash ring with heartbeating disabled is currently an experimental feature.

pracucci (approved):
Good job, LGTM!

pstibrany:
Heads up @grafana/loki-team and @grafana/tempo teams using dskit. Make sure to add `HeartbeatTimeout` to your lifecycler config, otherwise ring status page exposed via lifecycler handler may show "Unhealthy" ingesters.

PR #173, opened 2022-06-08
Change moduleService to implement NamedService

dimitarvdimitrov:
**What this PR does**:

The moduleService has a name, and it's trivial to implement ServiceName method.

I came across this while trying to add debug information about running services in mimir

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/modules/module_service.go
+++ b/modules/module_service.go
@@ -27,6 +27,10 @@ type moduleService struct {
 	startDeps, stopDeps func(string) map[string]services.Service
 }
 
+func (w moduleService) ServiceName() string {
+	return w.name
+}
+
 // NewModuleService wraps a module service, and makes sure that dependencies are started/stopped before module service starts or stops.
 // If any dependency fails to start, this service fails as well.
 // On stop, errors from failed dependencies are ignored.
--- a/modules/module_service.go
+++ b/modules/module_service.go
@@ -28,10 +28,11 @@ type moduleService struct {
 }
 
 func (w moduleService) ServiceName() string {
+	msg := fmt.Sprintf("module service %s", w.name)
 	if namedSvc, isNamed := w.Service.(services.NamedService); isNamed {
-		return namedSvc.ServiceName()
+		msg += fmt.Sprintf(" for %s", namedSvc.ServiceName())
 	}
-	return w.name
+	return msg
 }
 
 // NewModuleService wraps a module service, and makes sure that dependencies are started/stopped before module service starts or stops.
--- a/modules/module_service.go
+++ b/modules/module_service.go
@@ -28,6 +28,9 @@ type moduleService struct {
 }
 
 func (w moduleService) ServiceName() string {
+	if namedSvc, isNamed := w.Service.(services.NamedService); isNamed {
+		return namedSvc.ServiceName()
+	}
 	return w.name
 }
 
--- a/modules/module_service.go
+++ b/modules/module_service.go
@@ -27,6 +27,16 @@ type moduleService struct {
 	startDeps, stopDeps func(string) map[string]services.Service
 }
 
+type delegatedNamedService struct {
+	services.Service
+
+	delegate services.NamedService
+}
+
+func (n delegatedNamedService) ServiceName() string {
+	return n.delegate.ServiceName()
+}
+
 // NewModuleService wraps a module service, and makes sure that dependencies are started/stopped before module service starts or stops.
 // If any dependency fails to start, this service fails as well.
 // On stop, errors from failed dependencies are ignored.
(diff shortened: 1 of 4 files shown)


pstibrany:
Module service is a wrapper around another service. `name` field is name of the module. If anything, module service wrapper should expose name of wrapped service, instead of the module name.

dimitarvdimitrov:
thanks for the review. I pushed a commit to try delegating ServiceName to wrapped Service, ptal

pstibrany (commented):
If underlying service doesn't implement `NamedService`, I don't think `moduleService` should implement it or expose module name as "ServiceName". That implies that moduleService needs different implementation based on wrapped service implementing NamedService or not.

dimitarvdimitrov:
If `moduleSerice` implements `Service` with its own implementation (albeit some of it delegated), wouldn't it also make sense to expose a best-effort name? I am thinking of `ServiceName()` as existing to primarily aid debugging and make services human-readable. Isn't more information better in this case or will it be confusing to clients?

pstibrany:
> If `moduleSerice` implements `Service` with its own implementation (albeit some of it delegated), wouldn't it also make sense to expose a best-effort name? I am thinking of `ServiceName()` as existing to primarily aid debugging and make services human-readable. Isn't more information better in this case or will it be confusing to clients?

[In general "service" here is an object with observable state and a lifecycle.](https://github.com/google/guava/wiki/ServiceExplained) Not all services have or need a name.

To your question, if we `moduleService` wrapper a name, I think something like "service for module <name>" would be more appropriate (or "service for module <name>: <originalServiceName>") than just a module name.

dimitarvdimitrov:
I pushed a commit that changes it to `module service <name>` and `module service <name> for <originalServiceName>`. This is becoming slightly more verbose than I hoped for initially, but I don't have better suggestions.

pstibrany (dismissed):

modules/module_service.go
```diff
@@ -27,6 +27,14 @@ type moduleService struct {
 	startDeps, stopDeps func(string) map[string]services.Service
 }
 
+func (w moduleService) ServiceName() string {
+	msg := fmt.Sprintf("module service %s", w.name)
```
pstibrany, line 31: I think it's "service" for "module XXX". Not sure I would understand "module service XXX" as having the same meaning.

dimitarvdimitrov: yeah, `service` and `module` aren't very clear terms anyways. 

should we just drop `service`? the struct is an implementation of `Service`. None of the other services mention that they are a service.

I feel like we're spending time without too much added benefit, if we can't reach consensus, I'm happy to close this PR

dimitarvdimitrov: so I added a decorator that adds the `NamedService` implementation. The decorator is used when the wrapped service already implements `NamedService`. I changed the constructor function of `moduleService`. Doesn't look too bad.


on another note I noticed the `services.DescribeService` function which does kinda what I needed in the first place - a way to describe a service.

https://github.com/grafana/dskit/blob/0e4b5def51c060fc498de65ee5b0a20563891159/services/services.go#L135-L145

modules/module_service.go
```diff
@@ -40,6 +40,13 @@ func NewModuleService(name string, logger log.Logger, service services.Service,
 	}
 
 	w.Service = services.NewBasicService(w.start, w.run, w.stop)
+
+	if namedService, isNamed := service.(services.NamedService); isNamed {
+		return services.NamedServiceDecorator{
+			Service: w,
+			Name:    namedService.ServiceName(),
```
pstibrany, line 47: I don't think we should cache `namedService.ServiceName()` into a field, but call it each time. Service may update its name dynamically (eg. to include state or some other value).

pstibrany (commented):
Thanks. I like where this is going.

modules/module_service.go
pstibrany: > Service may update its name dynamically (eg. to include state or some other value).

This is actually quite common in Guava services.

dimitarvdimitrov: we should probably not cache them in mimir either.. https://github.com/grafana/mimir/blob/7aa6015704d6b5c080f5fd66dff9303df122563e/pkg/mimir/mimir.go#L438

but got your point - will refactor this

pstibrany: > we should probably not cache them in mimir either.. https://github.com/grafana/mimir/blob/7aa6015704d6b5c080f5fd66dff9303df122563e/pkg/mimir/mimir.go#L438

I'm not sure I understand what you're saying. I don't see any place in Mimir where we would call `ServiceName` and cache it. In fact the only place where it is used is in failure watcher (indirectly, via `DescribeService`).

services/service.go
```diff
@@ -99,7 +99,7 @@ type Service interface {
 type NamedService interface {
 	Service
 
-	// ServiceName returns name of the service, if it has one.
+	// ServiceName returns name of the service, if it has one. Subsequent calls to ServiceName can return different values.
```
pstibrany, line 102: I would include example to explain why:

```suggestion
	// ServiceName returns name of the service, if it has one.
	// Subsequent calls to ServiceName can return different values,
	// for example service may update its name based on its state.
```

pstibrany (approved):
Thank you!

modules/module_service.go
dimitarvdimitrov: you're right. what I linked is caching module names. seem like I'm prone to confusing the two

PR #175, opened 2022-06-10
Don't wait until next heartbeat to update ring on initialization.

### LINKED ISSUE #174: When lifecycler reuses existing LEAVING entry, it waits for another heartbeat period to write ACTIVE state to ring
When lifecycler finds existing entry in the ring, it will reuse state and tokens from it. If the state was LEAVING and entry has enough tokens, lifecycler immediately switches its internal state to ACTIVE, but it doesn't update the ring with this state, and instead waits for next heartbeat.

With long heartbeats (eg. 2 minutes), this makes the time until ACTIVE state of lifecycler is visible to other components longer.

This also affects how soon lifecycler reports "ready" in its `CheckReady` [function](https://github.com/grafana/dskit/blob/789ec0ca4a3b372335fad1c73147366a23ddc8b6/ring/lifecycler.go#L192), because it actually checks the state in the ring. (This is correct behaviour, if it only consulted its internal state, we could report "ready" before other components saw it as active, which could result in an outage in case of Mimir ingesters for example).

pstibrany:
**What this PR does**:
This PR changes lifecycler initialization such that it always updates ring immediately when it finds existing ring entry that it can reuse, instead of waiting for next heartbeat, since heartbeat period can be configured with long time.

Fixes #174

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,3 +57,4 @@
 * [BUGFIX] Multi KV: fix panic when using function to modify primary KV store in runtime. #153
 * [BUGFIX] Lifecycler: if the ring backend storage is reset, the instance adds itself back to the ring with an updated registration timestamp set to current time. #165
 * [BUGFIX] Ring: fix bug where hash ring instances may appear unhealthy in the web UI even though they are not. #172
+* [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #174
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,4 +57,4 @@
 * [BUGFIX] Multi KV: fix panic when using function to modify primary KV store in runtime. #153
 * [BUGFIX] Lifecycler: if the ring backend storage is reset, the instance adds itself back to the ring with an updated registration timestamp set to current time. #165
 * [BUGFIX] Ring: fix bug where hash ring instances may appear unhealthy in the web UI even though they are not. #172
-* [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #174
+* [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #175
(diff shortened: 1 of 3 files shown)


ring/lifecycler.go
```diff
@@ -600,17 +600,20 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
...
 
 		level.Info(i.logger).Log("msg", "existing entry found in ring", "state", i.GetState(), "tokens", len(tokens), "ring", i.RingName)
 
-		// Update the ring if the instance has been changed and the heartbeat is disabled.
-		// We dont need to update KV here when heartbeat is enabled as this info will eventually be update on KV
-		// on the next heartbeat
-		if i.cfg.HeartbeatPeriod == 0 && !instanceDesc.Equal(ringDesc.Ingesters[i.ID]) {
+		// Update the ring if the instance has been changed. We don't want to rely on heartbeat update, as heartbeat
+		// can be configured to long time, and until then lifecycler would not report this instance as ready in CheckReady.
+		if !instanceDesc.Equal(ringDesc.Ingesters[i.ID]) {
```
pracucci, line 616: It was also part of a change I tried to do in Cortex long time ago, but was never merged:
https://github.com/cortexproject/cortex/pull/3774/files

Just saying that it did look a good idea to me back in time, and it still does.

CHANGELOG.md
```diff
@@ -57,3 +57,4 @@
 * [BUGFIX] Multi KV: fix panic when using function to modify primary KV store in runtime. #153
 * [BUGFIX] Lifecycler: if the ring backend storage is reset, the instance adds itself back to the ring with an updated registration timestamp set to current time. #165
 * [BUGFIX] Ring: fix bug where hash ring instances may appear unhealthy in the web UI even though they are not. #172
+* [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #174
```
pracucci, line 60:
```suggestion
* [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #175
```

pracucci (approved):

PR #176, opened 2022-06-21
Make -memberlist.abort-if-join-fails default to false.

pstibrany:
**What this PR does**:

This PR changes default of `-memberlist.abort-if-join-fails` option to `false`, because in Grafana Mimir we recommend this value for most setups (when running in K8s).

**Checklist**
- [na] Tests updated

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -180,7 +180,7 @@ func (cfg *KVConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.DurationVar(&cfg.MinJoinBackoff, prefix+"memberlist.min-join-backoff", 1*time.Second, "Min backoff duration to join other cluster members.")
 	f.DurationVar(&cfg.MaxJoinBackoff, prefix+"memberlist.max-join-backoff", 1*time.Minute, "Max backoff duration to join other cluster members.")
 	f.IntVar(&cfg.MaxJoinRetries, prefix+"memberlist.max-join-retries", 10, "Max number of retries to join other cluster members.")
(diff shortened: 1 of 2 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -180,7 +180,7 @@ func (cfg *KVConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.DurationVar(&cfg.MinJoinBackoff, prefix+"memberlist.min-join-backoff", 1*time.Second, "Min backoff duration to join other cluster members.")
 	f.DurationVar(&cfg.MaxJoinBackoff, prefix+"memberlist.max-join-backoff", 1*time.Minute, "Max backoff duration to join other cluster members.")
 	f.IntVar(&cfg.MaxJoinRetries, prefix+"memberlist.max-join-retries", 10, "Max number of retries to join other cluster members.")
-	f.BoolVar(&cfg.AbortIfJoinFails, prefix+"memberlist.abort-if-join-fails", true, "If this node fails to join memberlist cluster, abort.")
+	f.BoolVar(&cfg.AbortIfJoinFails, prefix+"memberlist.abort-if-join-fails", cfg.AbortIfJoinFails, "If this node fails to join memberlist cluster, abort.")
```
pracucci, line 183: This trick is used to let downstream projects to change the default value, right?

pracucci (approved):

kv/memberlist/memberlist_client.go
pstibrany: Yes, if they prefer to.

I think we should split setting default values and configuring CLI flags into two methods, something like:
1. `(cfg Config).DefaultValues()`
2. `(cfg Config).RegisterFlags(...)` -- this would use values from `cfg` as default values.

But that's a separate discussion and PR. Here I do it (second step) for single field only.

PR #178, opened 2022-06-22
Make Consul CasRetryDelay configurable

DylanGuedes:
**What this PR does**:
Adds a new flag `consul.cas-retry-delay` that allows the configuration of freezing period before retrying CAS operations for Consul.
Without this, the CAS just retries too quickly what in some occasions is the same as not retrying.

N/A

Fixes #<issue number>
N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] Added new `consul.cas-retry-delay` flag. #178
 * [CHANGE] Flagext: `DayValue` now always uses UTC when parsing or displaying dates. #71
 * [CHANGE] Closer: remove the closer package since it's trivial to just copy/paste. #70
 * [CHANGE] Memberlist: allow specifying address and port advertised to the memberlist cluster members by setting the following configuration: #37
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 
 * [CHANGE] Added new `consul.cas-retry-delay` flag. #178
+  * It has a default of `1s`, different from the previous behavior, which is the same as `0`.
 * [CHANGE] Flagext: `DayValue` now always uses UTC when parsing or displaying dates. #71
 * [CHANGE] Closer: remove the closer package since it's trivial to just copy/paste. #70
 * [CHANGE] Memberlist: allow specifying address and port advertised to the memberlist cluster members by setting the following configuration: #37
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## Changelog
 
-* [CHANGE] Added new `consul.cas-retry-delay` flag. #178
-  * It has a default of `1s`, different from the previous behavior, which is the same as `0`.
+* [CHANGE] Added new `-consul.cas-retry-delay` flag. It has a default value of `1s`, while previously there was no delay between retries. #178
 * [CHANGE] Flagext: `DayValue` now always uses UTC when parsing or displaying dates. #71
 * [CHANGE] Closer: remove the closer package since it's trivial to just copy/paste. #70
 * [CHANGE] Memberlist: allow specifying address and port advertised to the memberlist cluster members by setting the following configuration: #37
--- a/kv/consul/client.go
+++ b/kv/consul/client.go
@@ -78,7 +78,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.BoolVar(&cfg.ConsistentReads, prefix+"consul.consistent-reads", false, "Enable consistent reads to Consul.")
 	f.Float64Var(&cfg.WatchKeyRateLimit, prefix+"consul.watch-rate-limit", 1, "Rate limit when watching key or prefix in Consul, in requests per second. 0 disables the rate limit.")
 	f.IntVar(&cfg.WatchKeyBurstSize, prefix+"consul.watch-burst-size", 1, "Burst size used in rate limit. Values less than 1 are treated as 1.")
-	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 1*time.Second, "Max duration to wait before retrying a Compare And Swap (CAS) operation.")
+	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 1*time.Second, "Maximum duration to wait before retrying a Compare And Swap (CAS) operation.")
 }
 
 // NewClient returns a new Client.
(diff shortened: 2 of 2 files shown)


kv/consul/client.go
```diff
@@ -49,7 +49,7 @@ type Config struct {
 
 	// Used in tests only.
 	MaxCasRetries int           `yaml:"-"`
-	CasRetryDelay time.Duration `yaml:"-"`
+	CasRetryDelay time.Duration `yaml:"cas_retry_delay" category:"advanced"`
```
56quarters, line 52: Should probably be moved to the group above since it's not just for tests now.

kv/consul/client.go
```diff
@@ -78,6 +78,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.BoolVar(&cfg.ConsistentReads, prefix+"consul.consistent-reads", false, "Enable consistent reads to Consul.")
 	f.Float64Var(&cfg.WatchKeyRateLimit, prefix+"consul.watch-rate-limit", 1, "Rate limit when watching key or prefix in Consul, in requests per second. 0 disables the rate limit.")
 	f.IntVar(&cfg.WatchKeyBurstSize, prefix+"consul.watch-burst-size", 1, "Burst size used in rate limit. Values less than 1 are treated as 1.")
+	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 0, "Duration to be waited before retrying a Compare And Swap (CAS) failure.")
```
chaudum, line 81: nit: 

```suggestion
	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 0, "Duration to be waited before retrying after a Compare And Swap (CAS) failure.")
```

chaudum (approved):
lgtm

kv/consul/client.go
pstibrany, line 81: also nit:

```suggestion
	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 0, "Max duration to wait before retrying a Compare And Swap (CAS) operation.")
```

pstibrany (approved):

kv/consul/client.go
DylanGuedes: hey your suggestion got out-of-date after applying another suggestion, lmk if you have other rewritten suggestion

kv/consul/client.go
```diff
@@ -78,6 +78,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.BoolVar(&cfg.ConsistentReads, prefix+"consul.consistent-reads", false, "Enable consistent reads to Consul.")
 	f.Float64Var(&cfg.WatchKeyRateLimit, prefix+"consul.watch-rate-limit", 1, "Rate limit when watching key or prefix in Consul, in requests per second. 0 disables the rate limit.")
 	f.IntVar(&cfg.WatchKeyBurstSize, prefix+"consul.watch-burst-size", 1, "Burst size used in rate limit. Values less than 1 are treated as 1.")
+	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 0, "Max duration to wait before retrying a Compare And Swap (CAS) operation.")
```
slim-bean, line 81: Can we make the default here non zero?

I think maybe `500ms` seems reasonable to me?

slim-bean, line 81:
```suggestion
	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 500*time.Milliseconds, "Duration to wait before retrying a Compare And Swap (CAS) operation.")
```

I think a reasonable non-zero default here would make sense, part of the current issue is we retry in a tight loop.

Also changed the comment a little because it's not a backoff, it's a fixed retry time.

slim-bean: I just saw @pstibrany and I are going back and forth a little one this.

I would prefer not to include `Max` because it implies the delay could be a variable amount less than the specified value when the code currently does a fixed delay of exactly this amount.

pstibrany: It's not backoff, but we use it as input into random function:

```
	sleepBeforeRetry := time.Duration(0)
	if c.cfg.CasRetryDelay > 0 {
		sleepBeforeRetry = time.Duration(rand.Int63n(c.cfg.CasRetryDelay.Nanoseconds()))
	}
```

So it's really a "maximum duration".

slim-bean: Oh i'm sorry, read the code again, should have done this first, there is a `rand` in it

pstibrany: Short non-zero default value makes sense to me. We should document it the change in the changelog.

slim-bean: Given this, I believe @pstibrany wording is correct.

I also updated the default to 1s

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] Added new `consul.cas-retry-delay` flag. #178
```
slim-bean, line 3:
```suggestion
* [CHANGE] Added new `consul.cas-retry-delay` flag which exposes an existing config and now sets a default value for the retry delay from previous 0s to 1s, the retry delay is a random number up to a maximum of this configured value. #178
```

slim-bean (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
...
+  * It has a default of `1s`, different from the previous behavior, which is the same as `0`.
```
pstibrany, line 4:
```suggestion
* [CHANGE] Added new `-consul.cas-retry-delay` flag. It has a default value of `1s`, while previously there was no delay between retries. #178
```

kv/consul/client.go
```diff
@@ -78,6 +78,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 	f.BoolVar(&cfg.ConsistentReads, prefix+"consul.consistent-reads", false, "Enable consistent reads to Consul.")
 	f.Float64Var(&cfg.WatchKeyRateLimit, prefix+"consul.watch-rate-limit", 1, "Rate limit when watching key or prefix in Consul, in requests per second. 0 disables the rate limit.")
 	f.IntVar(&cfg.WatchKeyBurstSize, prefix+"consul.watch-burst-size", 1, "Burst size used in rate limit. Values less than 1 are treated as 1.")
+	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 1*time.Second, "Max duration to wait before retrying a Compare And Swap (CAS) operation.")
```
pstibrany, line 81:
```suggestion
	f.DurationVar(&cfg.CasRetryDelay, prefix+"consul.cas-retry-delay", 1*time.Second, "Maximum duration to wait before retrying a Compare And Swap (CAS) operation.")
```

pstibrany (approved):

CHANGELOG.md
DylanGuedes: thanks, thats much better.

PR #180, opened 2022-06-28
Optimize ring computation when lookback period is so large that it selects all instances.

pstibrany:
**What this PR does**:
When using big lookback period that selects all instances, we don't need to recompute the ring from scratch, but can use original ring instead.

We have encountered this problem when running ring with almost 400 instances and large lookback period:

Screenshot 2022-06-28 at 15 59 35

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/ring_test.go
+++ b/ring/ring_test.go
@@ -2199,27 +2199,27 @@ func TestShuffleShardWithCaching(t *testing.T) {
 	// Change of shard size -> new subring needed.
 	subring = newSubring
 	newSubring = ring.ShuffleShard("user", 1)
-	require.False(t, subring == newSubring)
+	require.NotSame(t, subring, newSubring)
 	// Zone-aware shuffle-shard gives all zones the same number of instances (at least one).
 	require.Equal(t, zones, newSubring.InstancesCount())
 
 	// Verify that getting the same subring uses cached instance.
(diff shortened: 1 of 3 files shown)


ring/ring_test.go
```diff
@@ -2212,6 +2212,14 @@ func TestShuffleShardWithCaching(t *testing.T) {
 	ring.CleanupShuffleShardCache("user")
 	newSubring = ring.ShuffleShard("user", 1)
 	require.False(t, subring == newSubring)
+
+	// If we ask for ALL instances, we get original ring.
+	newSubring = ring.ShuffleShard("user", numLifecyclers)
+	require.True(t, ring == newSubring)
```
56quarters, line 2218: Can this be `require.Equals(t, ring, newSubring)` to provide more detail errors if the comparison fails?

56quarters (approved):
LGTM

ring/ring_test.go
pstibrany: The point is to verify that we do indeed get exact same pointer, and not just another ring that may be equal.

pstibrany: > In that case, `require.Same()`

Done.

pracucci (approved):
Thanks!

Issue #184, opened 2022-07-06
Empty ring right after startup when using memberlist


pracucci:
In Mimir, we're occasionally seeing "empty ring" ring right after a process startup (e.g. querier). It's an issue that has started after the migration to memberlist.

## Possible root cause

I think the issue is caused by the ring client implementation not guaranteeing to wait to get the initial ring state before switching to `Running` state. In the following I share some thoughts about the code.

The ring client service is expected to switch to `Running` state only after it initialized its internal state with the ring data structure. This is why it calls `r.KVClient.Get()` in the `Ring.starting()`:
https://github.com/grafana/dskit/blob/e441b77be7780e03f2c37659839bfe90dfde7dd3/ring/ring.go#L252-L256

When using Consul or etcd as backend, the `r.KVClient.Get()` guarantees to return the state of the ring, but I think this guarantee has been lost in the memberlist implementation and it could return a zero data structure.

The memberlist client `Get()` is implemented here:
https://github.com/grafana/dskit/blob/e441b77be7780e03f2c37659839bfe90dfde7dd3/kv/memberlist/memberlist_client.go#L63-L70

It waits until the backend KV client is running. But does waiting for it to be running guarantee the ring data structure to be populated before that? I don't think so.

The memberlist `KV.starting()` just initialise memberlist but doesn't join the cluster:
https://github.com/grafana/dskit/blob/e441b77be7780e03f2c37659839bfe90dfde7dd3/kv/memberlist/memberlist_client.go#L426-L453

The memberlist cluster is joined only in the `KV.running()`, but that's too late, because at that point our code assume the ring data structure to be already populated:
https://github.com/grafana/dskit/blob/e441b77be7780e03f2c37659839bfe90dfde7dd3/kv/memberlist/memberlist_client.go#L457-L472

pstibrany:
Joining the memberlist cluster or even getting full state from another node are no guarantees that KV store is complete or up to date. Shall we just introduce "wait" period into KV store operations and return error or block before this wait period has elapsed?

Note that lifecyclers already have option to "observe" the ring for similar reason (to avoid generating conflicting tokens).

bboreham:
Let's add some cases:
* 10 ingesters are running; this process is a querier that expects to learn about them via memberlist (I think this is the case you had in mind).
* This process is, and will be, the only one in the ring.
* 100 processes just started up; we have managed to join a memberlist with 2 of them, and 97 more will join shortly.

pracucci:
[...]

PR #195, opened 2022-07-20
Join memberlist cluster in Starting state.

pstibrany:
**What this PR does**:

This PR adds option (defaults to true) to make memberlist-based KV store join the memberlist cluster before serving any requests.

This prevents any clients from using KV store before node has fully joined the memberlist cluster, and received state from other nodes.

Fixes https://github.com/grafana/dskit/issues/184

**Checklist**
- [x] Tests updated


kv/memberlist/memberlist_client.go
```diff
@@ -185,6 +186,7 @@ func (cfg *KVConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.IntVar(&cfg.MaxJoinRetries, prefix+"memberlist.max-join-retries", 10, "Max number of retries to join other cluster members.")
 	f.BoolVar(&cfg.AbortIfJoinFails, prefix+"memberlist.abort-if-join-fails", cfg.AbortIfJoinFails, "If this node fails to join memberlist cluster, abort.")
 	f.DurationVar(&cfg.RejoinInterval, prefix+"memberlist.rejoin-interval", 0, "If not 0, how often to rejoin the cluster. Occasional rejoin can help to fix the cluster split issue, and is harmless otherwise. For example when using only few components as a seed nodes (via -memberlist.join), then it's recommended to use rejoin. If -memberlist.join points to dynamic service that resolves to all gossiping nodes (eg. Kubernetes headless service), then rejoin is not needed.")
+	f.BoolVar(&cfg.JoinInStarting, prefix+"memberlist.join-in-starting-state", true, "Join memberlist cluster before allowing clients to use KV store.")
```
pracucci, line 189: Do we really need yet another config option?

kv/memberlist/memberlist_client.go
```diff
@@ -463,15 +465,27 @@ func (m *KV) starting(_ context.Context) error {
 	}
 	m.initWG.Done()
 
+	if m.cfg.JoinInStarting && len(m.cfg.JoinMembers) > 0 {
+		// Lookup SRV records for given addresses to discover members.
+		members := m.discoverMembers(ctx, m.cfg.JoinMembers)
```
pracucci, line 470: I would change a couple of things:
1. Remove the new config option `m.cfg.JoinInStarting` (let's try to shrink the config options, not increase them)
2. Only join a subset of nodes in `starting()` and then join all nodes in `running()`
3. Move the call to `discoverMembers()` inside `joinMembersOnStartup()` and call it for each retry. Why? Because if `discoverMembers()` returns a node which is terminating, the join and all retries will fail, while it shouldn't

kv/memberlist/memberlist_client.go
pstibrany: It's removed now.

kv/memberlist/memberlist_client.go
pstibrany: Latest version of PR implements these suggestions.

kv/memberlist/memberlist_client.go
```diff
@@ -533,43 +529,83 @@ func (m *KV) JoinMembers(members []string) (int, error) {
...
-	reached, err := m.memberlist.Join(members)
-	if err == nil {
-		level.Info(m.logger).Log("msg", "joined memberlist cluster", "reached_nodes", reached)
-		return nil
+// fastJoinMembersOnStartup attempts to reach small subset of nodes (computed as RetransmitMult * log10(number of discovered members + 1)).
+func (m *KV) fastJoinMembersOnStartup(ctx context.Context) {
+	startTime := time.Now()
+
+	members := m.discoverMembers(ctx, m.cfg.JoinMembers)
+	toReach := m.cfg.RetransmitMult * int(math.Ceil(math.Log10(float64(len(members)+1))))
```
pracucci, line 537: Add a comment to explain this is the same algorithm used by memberlist to select the set of nodes to which a messages should be gossiped to.

kv/memberlist/memberlist_client.go
```diff
@@ -533,43 +529,83 @@ func (m *KV) JoinMembers(members []string) (int, error) {
...
+	if toReach > len(members) {
+		toReach = len(members)
 	}
 
-	if m.cfg.MaxJoinRetries <= 0 {
-		return err
+	level.Info(m.logger).Log("msg", "memberlist fast-join starting", "members_found", len(members), "to_reach", toReach)
```
pracucci, line 542: We're calling this feature "joining cluster", so why not being consistent here?

```suggestion
	level.Info(m.logger).Log("msg", "memberlist fast-join starting", "members_found", len(members), "to_join", toReach)
```

kv/memberlist/memberlist_client.go
```diff
@@ -533,43 +529,83 @@ func (m *KV) JoinMembers(members []string) (int, error) {
...
+			}
+		}
+
+		// If there are no more nodes to try, we give up.
+		if len(members) == 0 {
+			break
+		}
+
+		if len(members) > toReach {
+			members = members[:toReach]
```
pracucci, line 561: I would randomize the list of members before selecting the subset. If `discoverMembers()` reply is consistent over time, this way different nodes starting up at nearly the same time will try to join a different set of members.

kv/memberlist/memberlist_client.go
```diff
@@ -533,43 +529,83 @@ func (m *KV) JoinMembers(members []string) (int, error) {
...
+		}
+
+		for _, m := range members {
+			previouslyAttemptedMembers[m] = true
+		}
+
+		reached, err := m.memberlist.Join(members)
```
pracucci, line 568: I think this whole implementation is more complicated than it should be. Given `Join()` iterates over `members`, while not calling it 1-by-1 for each member until we reached the number of members to join?

kv/memberlist/memberlist_client.go
```diff
@@ -533,43 +529,83 @@ func (m *KV) JoinMembers(members []string) (int, error) {
...
-		backoff.Wait()
+	for boff.Ongoing() {
+		members := m.discoverMembers(ctx, m.cfg.JoinMembers)
 
-		reached, err := m.memberlist.Join(members)
-		if err != nil {
-			lastErr = err
-			level.Debug(m.logger).Log("msg", "attempt to join memberlist cluster failed", "retries", backoff.NumRetries(), "err", err)
-			continue
+		reached, err := m.memberlist.Join(members) // err is only returned if reached==0.
```
pracucci, line 595: If we have 100 nodes, but only 1 fail, we redo the pull-push against all 100. I think this is inefficient. Since the `Join()` internally iterates over the members, why not just calling it 1-by-1 so that we can known which member succeeded and which failed, and then we re-try only the failed ones?

pstibrany: > If we have 100 nodes, but only 1 fail, we redo the pull-push against all 100.

If only 1 node fails, we don't get error. We only get error if `m.memberlist.Join(members)` failed to join ALL nodes.

kv/memberlist/memberlist_client.go
colega: +1 please, I was going to ask where the log10 part comes from, and I guess that answers my question.

kv/memberlist/memberlist_client.go
pracucci: Right!

pstibrany:
Thanks for review @pracucci. I've addressed your feedback, please take a look again.

pracucci (approved):
LGTM! Took me a while to understand the test, but LGTM too 😉

pstibrany:
Thanks for review, I've added last touches based on your suggestions.

PR #202, opened 2022-07-28
Upgrade golang to 1.18

periklis:
**What this PR does**:
Upgrade golang to 1.18

Required to support #185

**Checklist**
- [ ] Tests updated

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -1,4 +1,4 @@
-local image = 'golang:1.17.6';
+local image = 'golang:1.18.4';
 
 local pipeline = {
   new(name):: {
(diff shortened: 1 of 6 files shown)


cyriltovena (approved):
There might be other place where we need to update the go version like in the Dockerfile no ?

Otherwise LGTM

periklis:
> There might be other place where we need to update the go version like in the Dockerfile no ?
> 
> Otherwise LGTM

There is no Dockerfile in this repo AFAICS.

zalegrala:
https://github.com/periklis/dskit/pull/1 for the signature update if needd.

pracucci (approved):
👍 Unblocking

PR #208, opened 2022-08-08
Add net.LookipIP DNS provider implementation

zserge:
As we are looking forward to using `dskit` inside Grafana for HA, we would require an implementation for the `kv/memberlist.DNSProvider`. Currently it looks like the interface closely matches the API from Thanos, but we would need something lighter, possibly built over `net.LookupIP`. This PR introduces the "builtin" DNSProvider implementation and a basic test to make sure it works.

### DIFF
--- a/kv/memberlist/dnsprovider.go
+++ b/kv/memberlist/dnsprovider.go
@@ -2,6 +2,8 @@ package memberlist
 
 import (
 	"context"
+	"net"
+	"sync"
 )
 
 // DNSProvider supports storing or resolving a list of addresses.
@@ -12,4 +14,32 @@ type DNSProvider interface {
 
 	// Addresses returns the latest addresses present in the DNSProvider.
 	Addresses() []string
+}
+
+type dnsProvider struct {
+	sync.Mutex
+	addr []string
+}
+
+func NewDNSProvider() DNSProvider { return &dnsProvider{} }
+
+func (d *dnsProvider) Resolve(ctx context.Context, addrs []string) error {
+	d.Lock()
+	defer d.Unlock()
+	for _, a := range addrs {
+		ips, err := net.LookupIP(a)
+		if err != nil {
+			return err
+		}
+		for _, ip := range ips {
+			d.addr = append(d.addr, ip.String())
+		}
+	}
+	return nil
+}
+
+func (d *dnsProvider) Addresses() []string {
+	d.Lock()
+	defer d.Unlock()
+	return d.addr
 }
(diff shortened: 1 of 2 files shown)


kv/memberlist/dnsprovider.go
```diff
@@ -13,3 +15,31 @@ type DNSProvider interface {
...
+			d.addr = append(d.addr, ip.String())
+		}
+	}
+	return nil
+}
+
+func (d *dnsProvider) Addresses() []string {
+	d.Lock()
+	defer d.Unlock()
+	return d.addr
```
56quarters, line 44: I think you need to make a copy of this otherwise the underlying storage of the slice can still be modified after the caller starts using it (I could be wrong, we _do_ run tests with `-race` but this code feels like a race condition).

kv/memberlist/dnsprovider.go
```diff
@@ -13,3 +15,31 @@ type DNSProvider interface {
...
+func (d *dnsProvider) Resolve(ctx context.Context, addrs []string) error {
+	d.Lock()
+	defer d.Unlock()
+	for _, a := range addrs {
+		ips, err := net.LookupIP(a)
+		if err != nil {
+			return err
+		}
+		for _, ip := range ips {
+			d.addr = append(d.addr, ip.String())
```
56quarters, line 35: Instead of doing an `append(d.addr)` in a loop, I'd rather see a new slice of addresses constructed and then swapped at the end. That prevents cases where an error is returned and `d.addr` is only half updated. This would also allow you to avoid holding the lock while doing network operations. Something like:

```go
var addrs []string
for _, a: range addrs {
    // do the lookup
    addrs = append(addrs, ip.String())
}
d.Lock()
d.addrs = addrs
d.Unlock()
return nil
```

kv/memberlist/dnsprovider.go
```diff
@@ -13,3 +15,31 @@ type DNSProvider interface {
...
+	addr []string
+}
+
+func NewDNSProvider() DNSProvider { return &dnsProvider{} }
+
+func (d *dnsProvider) Resolve(ctx context.Context, addrs []string) error {
+	d.Lock()
+	defer d.Unlock()
+	for _, a := range addrs {
+		ips, err := net.LookupIP(a)
```
56quarters, line 30: Should the context be checked for cancellation on each iteration? That seems like it'd make this more responsive.

56quarters (commented):
Seems like a good change to me. A few safety related comments.

kv/memberlist/dnsprovider.go
56quarters: Or perhaps just once before doing any DNS lookups?

pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

PR #209, opened 2022-08-08
Change memberlist leave and packet-write timeouts

pstibrany:
**What this PR does**:

This PR Increases the leave timeout to 10x the connection timeout, so that we can communicate the leave to at least 1 node, if the first 9 we try to contact times out.

**Checklist**
- [na] Tests updated


pracucci (approved):
Mimir is already using these settings in production and I think they're a way better default to increase the chances a leaving node message is correctly propagated during rollouts.

PR #210, opened 2022-08-09
runtutil: add ExhaustCloseWithErrCapture

bboreham:
Code from https://github.com/thanos-io/thanos/blob/6d1b98d49781/pkg/runutil/, slightly modified for consistency.

This enables grafana/e2e to build without a dependency on Thanos - https://github.com/grafana/e2e/pull/4

**Checklist**
- [x] Tests updated
- NA `CHANGELOG.md` updated

### DIFF
--- a/runutil/runutil_test.go
+++ b/runutil/runutil_test.go
@@ -1,11 +1,18 @@
+// Provenance-includes-location: https://github.com/thanos-io/thanos/blob/main/pkg/runutil/runutil_test.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Thanos Authors.
+
 package runutil
 
 import (
 	"fmt"
+	"io"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/go-kit/log/level"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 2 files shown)


runutil/runutil_test.go
```diff
@@ -56,3 +63,107 @@ func (l *fakeLogger) Log(keyvals ...interface{}) error {
...
+			closer:         testCloser{err: errors.New("test")},
+			expectedErrStr: "close: test",
+		},
+		{
+			err:            errors.New("test"),
+			closer:         testCloser{err: errors.New("test")},
+			expectedErrStr: "2 errors: test; close: test",
+		},
+	} {
+		if ok := t.Run("", func(t *testing.T) {
```
56quarters, line 97: This seems like a really odd way to structure this test (I assume this is a verbatim copy of the Thanos tests?)
* Conditionals in the test body for the various test cases instead of having three separate tests ala `t.Run("something", ...)`
* Using `t.Fail` and `t.Error` when we're already importing testify assert
* Using `t.Run()` with an empty test name?

runutil/runutil_test.go
```diff
@@ -56,3 +63,107 @@ func (l *fakeLogger) Log(keyvals ...interface{}) error {
...
+	return &emulatedCloser{Reader: r}
+}
+
+func TestCloseMoreThanOnce(t *testing.T) {
+	lc := &loggerCapturer{}
+	r := newEmulatedCloser(strings.NewReader("somestring"))
+
+	CloseWithLogOnErr(lc, r, "should not be called")
+	CloseWithLogOnErr(lc, r, "should not be called")
+	assert.Equal(t, false, lc.WasCalled)
```
56quarters, line 165: This can be replaced with `assert.False` or `assert.True`

56quarters (approved):
Code seems fine but the tests are a bit rough

pracucci (approved):

bboreham:
I changed the tests as suggested by @56quarters.

pracucci (approved):
Still LGTM :)

56quarters (approved):

PR #215, opened 2022-08-22
memberlist: allow retries when dns returns no peers

krajorama:
**What this PR does**:

When join-members is set, assume that the list of nodes from
discoverMembers should be non zero (eventually). Retry the
discoverMembers with backoff if the result is no nodes instead
of giving up right away and starting without peers.

This fixes the case when K8S DNS is slow to register multiple services,
which happens in low resource Kind cluster in CI.

N/A
CI fails in grafana/mimir

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,4 +70,4 @@
 * [BUGFIX] Ring: fix bug where hash ring instances may appear unhealthy in the web UI even though they are not. #172
 * [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #175
 * [BUGFIX] stringslicecsv: handle unmarshalling empty yaml string #206
-* [BUGFIX] Memberlist: retry discoverMembers is join-members is set, but discoverMembers return zero node. #215
+* [BUGFIX] Memberlist: retry joining memberlist cluster on startup when no nodes are resolved. #215
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,4 +70,4 @@
 * [BUGFIX] Ring: fix bug where hash ring instances may appear unhealthy in the web UI even though they are not. #172
 * [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #175
 * [BUGFIX] stringslicecsv: handle unmarshalling empty yaml string #206
-
+* [BUGFIX] Memberlist: retry discoverMembers is join-members is set, but discoverMembers return zero node. #215
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -565,7 +565,14 @@ func (m *KV) fastJoinMembersOnStartup(ctx context.Context) {
 	l.Log("msg", "memberlist fast-join finished", "joined_nodes", totalJoined, "elapsed_time", time.Since(startTime))
 }
 
+// The joinMembersOnStartup method resolves the addresses of the given join_members hosts and asks memberlist to join to them.
+// This method cannot be called before KV.running state as it may wait for K8S DNS to resolve the service addresses of members
+// running this very method. Which means the service needs to be READY for K8S to add it to DNS.
 func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
+	if len(m.cfg.JoinMembers) == 0 {
+		return true
+	}
+
 	startTime := time.Now()
 
 	level.Info(m.logger).Log("msg", "joining memberlist cluster", "join_members", strings.Join(m.cfg.JoinMembers, ","))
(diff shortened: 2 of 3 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -585,7 +585,7 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
 		nodes := m.discoverMembers(ctx, m.cfg.JoinMembers)
 
 		reached, err := m.memberlist.Join(nodes) // err is only returned if reached==0.
-		if err == nil {
+		if err == nil && !(len(m.cfg.JoinMembers) > 0 && len(nodes) == 0) { // fail on error or join members is set, but no members actually resolved
```
pstibrany, line 588: If `discoverMembers` returns 0 members, then we don't need to try `m.memberlist.Join` at all, but can back off and retry. WDYT?

pracucci (commented):
At first I thought it would have introduced a chicken-egg problem (at cluster creation, no pod is ready until discover other members, but if no pod is ready then no pod is returned by the service DNS), but then I realized joinMembersOnStartup() is called in KV.running() so not blocking the readiness. I think that's the reason why this change works. If so, I would document it somewhere, because it looks very subtle to me.

kv/memberlist/memberlist_client.go
krajorama: made the optimization and a bit more

krajorama:
> At first I thought it would have introduced a chicken-egg problem (at cluster creation, no pod is ready until discover other members, but if no pod is ready then no pod is returned by the service DNS), but then I realized joinMembersOnStartup() is called in KV.running() so not blocking the readiness. I think that's the reason why this change works. If so, I would document it somewhere, because it looks very subtle to me.

Added docstring to the method

kv/memberlist/memberlist_client.go
```diff
@@ -565,7 +565,14 @@ func (m *KV) fastJoinMembersOnStartup(ctx context.Context) {
...
 }
 
+// The joinMembersOnStartup method resolves the addresses of the given join_members hosts and asks memberlist to join to them.
+// This method cannot be called before KV.running state as it may wait for K8S DNS to resolve the service addresses of members
+// running this very method. Which means the service needs to be READY for K8S to add it to DNS.
 func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
+	if len(m.cfg.JoinMembers) == 0 {
+		return true
+	}
+
```
pstibrany, line 575: We don't need this check, it's already done before calling `joinMembersOnStartup`: https://github.com/grafana/dskit/blob/26c1b619d0599b44104131d1f3a61ddbec63e188/kv/memberlist/memberlist_client.go#L478

kv/memberlist/memberlist_client.go
```diff
@@ -584,15 +591,18 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
+		if len(nodes) > 0 {
+			reached, err := m.memberlist.Join(nodes) // err is only returned if reached==0.
+			if err == nil {
+				level.Info(m.logger).Log("msg", "joining memberlist cluster succeeded", "reached_nodes", reached, "elapsed_time", time.Since(startTime))
+				return true
+			}
+			level.Warn(m.logger).Log("msg", "joining memberlist cluster: failed to reach any nodes", "retries", boff.NumRetries(), "err", err)
+			lastErr = err
+		} else {
+			level.Warn(m.logger).Log("msg", "joining memberlist cluster: failed to resolve nodes", "retries", boff.NumRetries())
```
pstibrany, line 603:
```suggestion
			level.Warn(m.logger).Log("msg", "joining memberlist cluster: found no nodes to join", "retries", boff.NumRetries())
```

We don't know if it failed, or just nothing was found.

CHANGELOG.md
```diff
@@ -70,4 +70,4 @@
 * [BUGFIX] Ring: fix bug where hash ring instances may appear unhealthy in the web UI even though they are not. #172
 * [BUGFIX] Lifecycler: if existing ring entry is reused, ring is updated immediately, and not on next heartbeat. #175
 * [BUGFIX] stringslicecsv: handle unmarshalling empty yaml string #206
-
+* [BUGFIX] Memberlist: retry discoverMembers is join-members is set, but discoverMembers return zero node. #215
```
pstibrany, line 73: This entry speaks in terms of internal implementation. What about:

```suggestion
* [BUGFIX] Memberlist: retry joining memberlist cluster on startup when no nodes are resolved. #215
```

kv/memberlist/memberlist_client.go
krajorama: I'd rather remove the if on the outside, to establish an invariant at the top of the function

kv/memberlist/memberlist_client.go
krajorama: ack

kv/memberlist/memberlist_client.go
pstibrany: That works too.

pstibrany (approved):
Thank you!

PR #218, opened 2022-09-09
Don't unmarshal and merge runtime config files if they haven't changed.

pstibrany:
**What this PR does**:

This PR modifies `runtimeconfig.Manager` to avoid unmarshaling and merging config files that haven't changed since last check.

**Checklist**
- [x] Tests updated

### DIFF
--- a/runtimeconfig/manager.go
+++ b/runtimeconfig/manager.go
@@ -58,7 +58,7 @@ type Manager struct {
 	configLoadSuccess prometheus.Gauge
 	configHash        *prometheus.GaugeVec
 
-	// Maps path to hash. Only used by loadConfig in Starting and Running, so it doesn't need synchronization.
+	// Maps path to hash. Only used by loadConfig in Starting and Running states, so it doesn't need synchronization.
 	fileHashes map[string]string
 }
 
--- a/runtimeconfig/manager.go
+++ b/runtimeconfig/manager.go
@@ -57,6 +57,9 @@ type Manager struct {
 
 	configLoadSuccess prometheus.Gauge
 	configHash        *prometheus.GaugeVec
+
+	// Maps path to hash. Only used by loadConfig in Starting and Running, so it doesn't need synchronization.
+	fileHashes map[string]string
 }
 
 // New creates an instance of Manager and starts reload config loop based on config
@@ -73,7 +76,7 @@ func New(cfg Config, registerer prometheus.Registerer, logger log.Logger) (*Mana
 		}),
 		configHash: promauto.With(registerer).NewGaugeVec(prometheus.GaugeOpts{
 			Name: "runtime_config_hash",
-			Help: "Hash of the currently active runtime config file.",
+			Help: "Hash of the currently active runtime configuration, merged from all configured files.",
 		}, []string{"sha256"}),
 		logger: logger,
 	}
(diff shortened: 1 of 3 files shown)


runtimeconfig/manager.go
```diff
@@ -147,26 +150,51 @@ func (om *Manager) loop(ctx context.Context) error {
...
-		err = yaml.Unmarshal(buf, &yamlFile)
+
+		rawData[f] = buf
+		hashes[f] = fmt.Sprintf("%x", sha256.Sum256(buf))
+	}
+
+	// check if new hashes are the same as before
+	sameHashes := true
+	for f, h := range hashes {
+		if om.fileHashes[f] != h {
```
aknuds1, line 170: Isn't `om.fileHashes` initially `nil`?

aknuds1 (commented):
Generally looks like a nice performance improvement, but shouldn't `om.fileHashes` first be initialized (to an empty map)?

runtimeconfig/manager.go
pstibrany: Yes. Reading from nil map is fine.

aknuds1: @pstibrany you're right, I tested it in the meantime. I must be remembering it wrong, I thought reading from nil maps would crash.

runtimeconfig/manager_test.go
```diff
@@ -419,3 +436,37 @@ func TestManager_ShouldFastFailOnInvalidConfigAtStartup(t *testing.T) {
...
+	for i := 0; i < mods; i++ {
+		writeValueToFile(t, cfg.LoadPath[0], value{Value: i})
+		// wait before next rewrite, but also after last rewrite to give manager a chance to reload the file again
+		time.Sleep(modDelay)
+	}
+
+	require.NoError(t, services.StopAndAwaitTerminated(context.Background(), overridesManager))
+
+	assert.Equal(t, int(loadCounter.Load()), mods+1)  // + 1 for initial load, before modifications
+	assert.Equal(t, int(loadCounter.Load()), len(ch)) // Loaded values
```
aknuds1, line 471: The expectation should come before the tested value:
```suggestion
	assert.Equal(t, mods+1, int(loadCounter.Load()))  // + 1 for initial load, before modifications
	assert.Equal(t, len(ch), int(loadCounter.Load())) // Loaded values
```

runtimeconfig/manager_test.go
```diff
@@ -54,26 +55,48 @@ func testLoadOverrides(r io.Reader) (interface{}, error) {
...
+	err = yaml.Unmarshal(buf, &v)
+	if err != nil {
+		return nil, err
+	}
+
+	return v, nil
+}
+
+func writeValueToFile(t *testing.T, path string, v value) {
+	buf, err := yaml.Marshal(v)
```
aknuds1, line 78: [Nit] I think test utility functions should be marked with `t.Helper()`.
```suggestion
	t.Helper()

	buf, err := yaml.Marshal(v)
```

aknuds1 (approved):
LGTM, except there are a couple of `assert.Equal` calls where the arguments should be switch places (so expectations are called out correctly).

runtimeconfig/manager_test.go
pstibrany: 👍  Second line actually tests that `len(ch)` is correct.

aknuds1: @pstibrany aha, was difficult to tell what's the expectation on that line :D

aknuds1: I would then suggest to make it, since the number of loads could be wrong (in case of a bug):

```
assert.Equal(t, mods+1, len(ch)) // Loaded values
```

aknuds1: Updated my suggestion to reflect the above :point_up:

runtimeconfig/manager.go
```diff
@@ -180,7 +208,10 @@ func (om *Manager) loadConfig() error {
 
 	// expose hash of runtime config
 	om.configHash.Reset()
-	om.configHash.WithLabelValues(fmt.Sprintf("%x", hash[:])).Set(1)
+	om.configHash.WithLabelValues(fmt.Sprintf("%x", hash)).Set(1)
+
+	// preserve hashes for next loop
+	om.fileHashes = hashes
```
56quarters, line 214: The other updates here (`om.setConfig`) take the mutex to update. The hashes need to take the mutex to update and while being read in the loop above.

runtimeconfig/manager.go
```diff
@@ -147,26 +150,51 @@ func (om *Manager) loop(ctx context.Context) error {
 // loadConfig loads all configuration files using the loader function then merges the yaml configuration files into one yaml document.
 // and notifies listeners if successful.
 func (om *Manager) loadConfig() error {
-	mergedConfig := map[string]interface{}{}
+	rawData := map[string][]byte{}
+	hashes := map[string]string{}
```
56quarters, line 154: nit: these could be initialized to the size of `om.cfg.LoadPath`, no?

runtimeconfig/manager.go
pstibrany: `om.fileHashes` doesn't need extra synchronization, as commented on the field. It's only accessed from `loadConfig`, which is called from `Starting` or `Running` state (called `loop`) functions of the service, and these two states already form happens-before.

runtimeconfig/manager.go
pstibrany: Yes, but values below 8 are ignored in `make(map[]..., len)` call anyway, so why bother.

pstibrany: (I mean, most people will use handful of files, hopefully not 8+)

pracucci (approved):
:clap

PR #219, opened 2022-09-19
Make FailureWatcher.WatchService() and FailureWatcher.WatchManager() panic if FailureWatcher is nil

pracucci:
**What this PR does**:
Today I realized few places in Mimir where we were using `FailureWatcher` but it was `nil`, so it was doing nothing (fixed in https://github.com/grafana/mimir/pull/2978). Finding these bugs is very hard, so I propose to make `FailureWatcher.WatchService()` and `FailureWatcher.WatchManager()` panicking very loudly if that happens.

N/A

**Checklist**
- [x] Tests updated


pstibrany (approved):
Thanks!

PR #228, opened 2022-10-26
Import Bytes type, DeleteAll function and DNS package from Thanos

bboreham:
**What this PR does**:

Adds:
* `flagext.Bytes` type, so you can define a quantity like 10GB.  (Note this is only implemented for Yaml so far).
* `runutil.DeleteAll` to delete files.
* `dns` package for service discovery, with two implementations `miekgdns` and `godns`.

**Checklist**
- [x] Tests updated.
- [x] `CHANGELOG.md` updated.


pracucci (approved):
👍

PR #237, opened 2022-12-01
Execute health checks for ring clients in parallel

### LINKED ISSUE #236: Ring client pool should not execute health checks sequentially
The ring client pool executes health checks to its target addresses sequentially, see https://github.com/grafana/dskit/blob/0bff37e60ff46c6f7e634abd42366aeebe0784bf/ring/client/pool.go#L178-L188

This can be a serious problem, when there are hundreds of target addresses to check, and you still want to have a short health check interval and a reasonably long timeout. E.g. 10s interval, 1s timeout. In case of e.g. 500 target addresses, a single request may take 20ms so that all 500 requests can finish within the 10s interval. Under normal load that should not be a problem. However, such a configuration would only give time for 10 requests to time out until the next interval is missed (it would be skipped). That could delay the next health check significantly, which could lead to clients trying to connect to unhealthy targets.

I propose to parallelize the health check execution.

chaudum:
**What this PR does**:

The ring client pool executes the health checks for its client sequentially, which can lead to problems when there are a lot of clients to check, especially when the targets do not respond fast enough.

This PR changes the execution from sequential to parallel. If the new `MaxConcurrentHealthChecks` config setting is not set (`0` value), then all health checks of all clients are executed in parallel, otherwise the parallelism from the setting is used.

Signed-off-by: Christian Haudum <christian.haudum@gmail.com>

Fixes #236

**Checklist**
- [x] Tests updated


ring/client/pool.go
```diff
@@ -173,9 +175,18 @@ func (p *Pool) removeStaleClients() {
...
+// If this is not set (0), then all healthchecks of all clients are executed in parallel.
 func (p *Pool) cleanUnhealthy() {
-	for _, addr := range p.RegisteredAddresses() {
+	addresses := p.RegisteredAddresses()
+
+	maxConcurrent := p.cfg.MaxConcurrentHealthChecks
+	if p.cfg.MaxConcurrentHealthChecks == 0 {
+		maxConcurrent = len(addresses)
+	}
+	_ = concurrency.ForEachJob(context.Background(), len(addresses), maxConcurrent, func(_ context.Context, idx int) error {
```
dannykopping, line 188: If we're executing these checks concurrently, I wonder if we need to cancel the checks if they overrun the interval at which `cleanUnhealthy` is invoked - otherwise they'll start to run on divergent intervals. Is that even an issue?

dannykopping (approved):
LGTM, with a comment

I'll approve, but I think we need a maintainer's approval before merging.

ring/client/pool.go
bboreham: Create a context with timeout instead of passing `context.Background()` ?

ring/client/pool.go
```diff
@@ -173,9 +175,18 @@ func (p *Pool) removeStaleClients() {
 	}
 }
 
-// cleanUnhealthy loops through all servers and deletes any that fails a healthcheck.
+// cleanUnhealthy loops through all servers and deletes any that fails a healthcheck
```
bboreham, line 178: The change here was to remove a full stop?
Seems unnecessary, though you could s/fails/fail/.

ring/client/pool.go
```diff
@@ -173,9 +175,18 @@ func (p *Pool) removeStaleClients() {
...
+// The health checks are executed concurrently with p.cfg.MaxConcurrentHealthChecks.
+// If this is not set (0), then all healthchecks of all clients are executed in parallel.
 func (p *Pool) cleanUnhealthy() {
-	for _, addr := range p.RegisteredAddresses() {
+	addresses := p.RegisteredAddresses()
+
+	maxConcurrent := p.cfg.MaxConcurrentHealthChecks
+	if p.cfg.MaxConcurrentHealthChecks == 0 {
+		maxConcurrent = len(addresses)
+	}
```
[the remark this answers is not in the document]
bboreham: Something like 16 seems plausible to me.

bboreham (commented):
Broadly OK; I find "client" in the description confusing; logically we are removing servers.
(The code talks about both; it is holding a client to some server address).

ring/client/pool.go
```diff
@@ -173,9 +175,18 @@ func (p *Pool) removeStaleClients() {
...
+// cleanUnhealthy loops through all servers and deletes any that fails a healthcheck
+// The health checks are executed concurrently with p.cfg.MaxConcurrentHealthChecks.
+// If this is not set (0), then all healthchecks of all clients are executed in parallel.
 func (p *Pool) cleanUnhealthy() {
-	for _, addr := range p.RegisteredAddresses() {
+	addresses := p.RegisteredAddresses()
+
+	maxConcurrent := p.cfg.MaxConcurrentHealthChecks
+	if p.cfg.MaxConcurrentHealthChecks == 0 {
+		maxConcurrent = len(addresses)
```
pracucci, line 186: Typically we use a value of 0 to disable a feature. I'm wondering if 0 should be treated as 1. Default to maximum concurrency looks too aggressive to me. Think about the case you run 1k ingesters and you run 1k health checks concurrently.

ring/client/pool.go
pracucci: Which timeout should we pass?

pracucci, line 188: To keep design cleaner, `healthCheck()` could take this `ctx` instead of initializing a new `context.Background()`.

ring/client/pool.go
pracucci: Comment raised in another comment. Having a sane default (like 16) LGTM too.

ring/client/pool.go
chaudum: Maybe `CheckInterval`?

chaudum: So, when I pass a cancellable context or a context with timeout to the `healtCheck()` function and the context gets cancelled, it could be that not all health checks have been executed - which is ok. However, there is a race condition, where the context is cancelled and there are health check requests currently in flight. In that case, a context cancellation would cause the request to fail and the client to be removed, even though it may have been healthy, right?

chaudum: @bboreham @pracucci Since all child contexts have a timeout, do I really need to set a timeout on the parent context as well? The only case I could think of is when the health check timeout is greater than the check interval - which should never be the case and we could also add an assertion for that.

chaudum: Forget what I said. The function `concurrency.ForEachJob()` is waiting for all health checks to finish :man_facepalming: We don't need to set a timeout on the parent context.

pracucci (approved):
LGTM, thanks!

ring/client/pool_test.go
```diff
@@ -120,28 +121,43 @@ func TestPoolCache(t *testing.T) {
...
+			for _, addr := range badAddrs {
+				clients[addr] = mockClient{happy: false, status: grpc_health_v1.HealthCheckResponse_NOT_SERVING}
+			}
+
+			cfg := PoolConfig{
+				MaxConcurrentHealthChecks: tc.maxConcurrent,
+				CheckInterval:             1 * time.Second,
+				HealthCheckTimeout:        5 * time.Millisecond,
+			}
+			pool := NewPool("test", cfg, nil, nil, nil, log.NewLogfmtLogger(os.Stderr))
```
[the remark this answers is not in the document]
chaudum: Yes, I wanted more logs, but this should be reverted.

ring/client/pool.go
```diff
@@ -173,24 +179,28 @@ func (p *Pool) removeStaleClients() {
...
-	}
+		return nil
+	})
 }
 
 // healthCheck will check if the client is still healthy, returning an error if it is not
-func healthCheck(client PoolClient, timeout time.Duration) error {
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+func healthCheck(ctx context.Context, client PoolClient, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
```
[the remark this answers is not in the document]
chaudum: The context could not be cancelled elsewhere. But when looking at the `concurrency.ForEachJob` function again, I realised, that it would stop after the first error. This would prevent it from executing all health checks, and that's definitely not what we want.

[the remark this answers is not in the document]
chaudum: @bboreham I implemented the concurrency directly in the `cleanUnhealthy()` function.

dannykopping (approved):
LGTM

ring/client/pool.go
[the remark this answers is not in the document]
pracucci: > The context could not be cancelled elsewhere. But when looking at the concurrency.ForEachJob function again, I realised, that it would stop after the first error. This would prevent it from executing all health checks, and that's definitely not what we want.

Not sure reimplementing it from scratch (as done in the last commit) is a good approach. Why don't we keep using `concurrency.ForEachJob()` but we never return an error from the function? That's how we use it in Mimir when we don't want to stop on first error.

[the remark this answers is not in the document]
chaudum: Reverted the last commit.

pracucci (approved):
LGTM (modulo a nit)

PR #239, opened 2022-12-02
Add `cache`, `gate` and `stringsutil` packages

ortuman:

**What this PR does**:

This PR moves the `cache` and `cacheutil` packages from mimir to dskit, allowing to reuse this same logic across external components, such as GEx admin. 

Fixes https://github.com/grafana/mimir-squad/issues/991

**Checklist**
- [ ] Tests updated


cache/cache.go
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
56quarters, line 1: dskit is Apache 2

cache/lru.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	cache := &LRUCache{
+		c:          c,
+		lru:        lru,
+		name:       name,
+		defaultTTL: defaultTTL,
+
+		requests: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name:        "cortex_cache_memory_requests_total",
```
56quarters, line 52: I get not changing things when doing the initial import but we should probably change this at least, right?

util/strings.go
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
56quarters, line 1: Can these functions be added as private functions wherever they're used or copy/pasted? They're trivial and adding a `util` package to a util library is too much.

cache/lru.go
johannaratliff: For dskit, probably, yes. Do you think we should just do `lru_name_of_metric` instead, make it more configurable with `<your service>_name_of_metric` or just rely on labels etc for that? I lean toward `lru_metric_name` + labels if in dskit. What do you both think?

cache/cache.go
johannaratliff: It looks like the rest of dskit just doesn't have a license line at the top and is covered by the LICENSE file - so I think we can go ahead and delete these lines

cache/lru.go
56quarters: I think just using `lru_metric_name` + labels is good because people can wrap the `reg` being used via `WrapRegistererWithPrefix("cortex_", reg)` if needed

util/strings.go
ortuman: Sounds reasonable to me. 👍 

done @ 4fd8d8f9b13ca99da881ea1f1c5f2db463263c0d

cache/lru.go
ortuman: > I get not changing things when doing the initial import but we should probably change the cortex_ prefix at least, right?

I overlooked that indeed. 

Done at 6a40d0f3ceb7e7c418de4b22e3dd94472bf97c70. And for the same reason, also got rid of `thanos_` metrics prefix.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 * [ENHANCEMENT] Replace go-kit/kit/log with go-kit/log. #52
 * [ENHANCEMENT] Add spanlogger package. #42
 * [ENHANCEMENT] Add runutil.CloseWithLogOnErr function. #58
+* [ENHANCEMENT] Add cache and cacheutils packages. #239
```
pracucci, line 50: [nit] Mention `gate` package too.

pracucci (commented):
I'm up to it. Could we move the content of `cacheutil` to `cache` package, without renaming stuff? Would there be any naming conflict?

cache/compression.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	return decoded
+}
+
+func (s *snappyCache) Name() string {
+	return s.next.Name()
+}
+
+func strSliceContains(slice []string, s string) bool {
```
pracucci, line 97: We already have this utility in `ring/util/string_utils.go`. I think it's time to move it out of the ring package (e.g. `stringutil` package) and reuse it here too.

cache/memcached_client.go
```diff
@@ -0,0 +1,695 @@
...
+	Addresses() []string
+}
+
+type memcachedGetMultiResult struct {
+	items map[string]*memcache.Item
+	err   error
+}
+
+// NewMemcachedClient makes a new RemoteCacheClient.
+func NewMemcachedClient(logger log.Logger, name string, conf []byte, reg prometheus.Registerer) (RemoteCacheClient, error) {
```
pracucci, line 223: Delete this function. Was inherited from Thanos, but we don't want to pass the config as a marshalled `[]byte` slice.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 * [ENHANCEMENT] Replace go-kit/kit/log with go-kit/log. #52
 * [ENHANCEMENT] Add spanlogger package. #42
 * [ENHANCEMENT] Add runutil.CloseWithLogOnErr function. #58
+* [ENHANCEMENT] Add cache and gate packages. #239
```
pracucci, line 50: [nit] And stringutil too.

cache/memcached_client.go
```diff
@@ -219,16 +219,6 @@ type memcachedGetMultiResult struct {
 	err   error
 }
 
-// NewMemcachedClient makes a new RemoteCacheClient.
-func NewMemcachedClient(logger log.Logger, name string, conf []byte, reg prometheus.Registerer) (RemoteCacheClient, error) {
-	config, err := parseMemcachedClientConfig(conf)
```
pracucci: `parseMemcachedClientConfig()` should be unused now.

pracucci (approved):
Reviewed looking at the diff in https://github.com/grafana/mimir/pull/3705 and LGTM!

ring/client/pool.go
```diff
@@ -171,7 +171,7 @@ func (p *Pool) removeStaleClients() {
 	}
 
 	for _, addr := range p.RegisteredAddresses() {
-		if util.StringsContain(serviceAddrs, addr) {
+		if stringsutil.SliceContains(serviceAddrs, addr) {
```
56quarters, line 174: This can use `slices.Contains()` since we target go 1.18

pracucci: Nice! I didn't know.

PR #240, opened 2022-12-02
[modules]: add test for ordered shutdown

zalegrala:
**What this PR does**:

I had some questions about the shutdown process with regard to ordering.  I see some of the code is ordering by name, and so here I included a test that checks the shutdown order of the various services.

**Checklist**
- [x] Tests updated

### DIFF
--- a/modules/modules_test.go
+++ b/modules/modules_test.go
@@ -332,6 +332,8 @@ func TestModuleWaitsForAllDependencies(t *testing.T) {
 
 	expectedStopOrder := []string{"c", "a"}
 
+	close(stopChanOrder)
+
 	for i, stopID := range <-stopChanOrder {
 		assert.Equal(t, expectedStopOrder[i], string(stopID))
 	}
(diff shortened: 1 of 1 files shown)


modules/modules_test.go
```diff
@@ -320,6 +329,12 @@ func TestModuleWaitsForAllDependencies(t *testing.T) {
 	require.NoError(t, err)
 	assert.NoError(t, services.StartManagerAndAwaitHealthy(context.Background(), servManager))
 	assert.NoError(t, services.StopManagerAndAwaitStopped(context.Background(), servManager))
+
+	expectedStopOrder := []string{"c", "a"}
+
+	for i, stopID := range <-stopChanOrder {
```
pstibrany, line 335: Without closing `stopChanOrder` channel, this will be stuck forever.

pstibrany (commented):
In this case, since B depends on A, and C depends on B, we need to stop C first, then B and finally A. It's not based on naming, but dependencies ordering.

If A and B modules have no dependencies, their shutdown order is not defined and implementation dependant.

PR #242, opened 2022-12-14
Remove github.com/prometheus/prometheus dependency

ortuman:

After merging #239 we introduced a [github.com/prometheus/prometheus](https://github.com/prometheus/prometheus/)  dependency.

@ go.mod 👉  https://github.com/grafana/dskit/blob/main/go.mod#L28

Apart from being an old version, this is also causing some issues when validating CI in enterprise repositories. 

The purpose of this PR is to get rid of it.

### DIFF
--- a/gate/gate.go
+++ b/gate/gate.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	promgate "github.com/prometheus/prometheus/util/gate"
 )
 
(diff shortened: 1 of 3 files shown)


krajorama (approved):

gate/gate.go
```diff
@@ -159,3 +158,35 @@ func (g *blockingGate) Done() {
 		panic("gate.Done: more operations done than started")
 	}
 }
```
56quarters, line 160: Why was this introduced? This does exactly the same thing as the existing blocking gate?

PR #245, opened 2022-12-15
cache: added Redis support

ortuman:

**What this PR does**:

This PR adds support for Redis cache. Most of the original logic has been extracted from thanos repository, adapting it to be consistent with our `RemoteCacheClient` interface. 

Amongst most notable changes with respect to the original code, worth mentioning the following: 

* Got rid of the `SetMulti` method for Redis client.
* Adapted `SetAsync` to actually run asynchronously.
* Replaced `redis.Client` with `redis.UniversalClient` to support cluster, failover and single node configurations.
* Added missing `Delete` method.
* Adapted redis client reported metrics to be consistent with the ones reported from memcached client. 

Also, as part of the above mentioned changes some extra fields were added into `RedisClientConfig`. 

Fixes N/A

**Checklist**
- [X] Tests updated


cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+				"msg", "failed to delete redis item",
+				"key", key,
+				"err", err,
+			)
+			c.trackError(opDelete, err)
+		} else {
+			c.duration.WithLabelValues(opDelete).Observe(time.Since(start).Seconds())
+		}
+		errCh <- err
+	})
```
johannaratliff, line 396: I'm not positive enqueueErr actually get's called. The func itself get's returned and passed around but does it get invoked? let's add a test for the delete function in `redis_client_test.go`

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	// Set async only available when MaxAsyncConcurrency > 0.
+	if c.MaxAsyncConcurrency <= 0 {
+		return errRedisMaxAsyncConcurrencyNotPositive
+	}
+	return nil
+}
+
+func (c *RedisClientConfig) getAddresses() ([]string, error) {
+	addresses := strings.Split(c.Endpoint, ",")
```
pracucci, line 143: The CLI flag description doesn't explain these are comma separated. Should be mentioned.

pracucci: Also you should directly use `flagext.StringSliceCSV` in the config field, instead of a `string` which then is split.

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+	// Handle single configuration endpoint which resolves multiple nodes.
+	if len(addresses) == 1 {
```
pracucci, line 145: I'm not sure this is correct. This gets resolved only at the process startup. If the resolution of `addresses[0]` changes over time, it won't reflect the change.

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+	logger     log.Logger
+	operations *prometheus.CounterVec
+	failures   *prometheus.CounterVec
+	skipped    *prometheus.CounterVec
+	duration   *prometheus.HistogramVec
+	dataSize   *prometheus.HistogramVec
+}
+
+// NewRedisClientWithConfig makes a new RedisClient.
+func NewRedisClientWithConfig(logger log.Logger, name string, config RedisClientConfig, reg prometheus.Registerer) (RemoteCacheClient, error) {
```
pracucci, line 189: [nit] There's no other alternative than passing a config.

```suggestion
func NewRedisClient(logger log.Logger, name string, config RedisClientConfig, reg prometheus.Registerer) (RemoteCacheClient, error) {
```

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+	DialTimeout time.Duration `yaml:"dial_timeout"`
+
+	// ReadTimeout specifies the client read timeout.
+	ReadTimeout time.Duration `yaml:"read_timeout"`
+
+	// WriteTimeout specifies the client write timeout.
+	WriteTimeout time.Duration `yaml:"write_timeout"`
+
+	// Maximum number of socket connections.
+	PoolSize int `yaml:"pool_size"`
```
pracucci, line 91: This is never read.

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/grafana/dskit/flagext"
+	"github.com/grafana/dskit/gate"
+)
+
+var (
+	errRedisAsyncBufferFull                = errors.New("the async buffer is full")
+	errMemcachedConfigNoEndpoint           = errors.New("no redis endpoint provided")
```
pracucci, line 25:
```suggestion
	errRedisConfigNoEndpoint           = errors.New("no redis endpoint provided")
```

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+			)
+			c.trackError(opSet, err)
+			return
+		}
+
+		c.dataSize.WithLabelValues(opSet).Observe(float64(len(value)))
+		c.duration.WithLabelValues(opSet).Observe(time.Since(start).Seconds())
+	})
+
+	if errors.Is(err, errMemcachedAsyncBufferFull) {
```
pracucci, line 326: Should either be renamed to a generic error, or a Redis specific one.

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+		logger:          logger,
+		asyncQueue:      make(chan func(), config.MaxAsyncBufferSize),
+		stop:            make(chan struct{}, 1),
+		getMultiGate: gate.New(
+			prometheus.WrapRegistererWithPrefix("redis_getmulti_", reg),
+			config.MaxGetMultiConcurrency,
+		),
+	}
+
+	c.operations = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
```
pracucci, line 241: Most of these metrics are a copy of the memcached client. Could be worth moving them to a dedicated generic struct holding the cache client metrics, then used by both redis and memcached clients. The references to "redis" or "memcached" can be parametrized (pass the backend name in input to `NewCacheClientMetrics()`).

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+
+	// Wait for all workers to finish.
+	c.workers.Wait()
+
+	if err := c.Close(); err != nil {
+		level.Error(c.logger).Log("msg", "redis close err")
+	}
+}
+
+func (c *redisClient) enqueueAsync(op func()) error {
```
pracucci, line 419: I think all the logic around the async queue is the same with memcached. Can be moved to a dedicated struct (e.g. `clientAsyncQueue` and generalised there.

cache/redis_client.go
```diff
@@ -0,0 +1,463 @@
...
+	if errors.Is(err, errMemcachedAsyncBufferFull) {
+		c.skipped.WithLabelValues(opSet, reasonAsyncBufferFull).Inc()
+		level.Debug(c.logger).Log("msg", "failed to store item to redis because the async buffer is full", "err", err, "size", len(c.asyncQueue))
+		return nil
+	}
+	return err
+}
+
+// GetMulti implement RemoteCacheClient.
+func (c *redisClient) GetMulti(ctx context.Context, keys []string) map[string][]byte {
```
pracucci, line 335: The same logic in memcached client looks easier to follow. Two questions:
1. Any reason why logic here was written differently?
2. Can we generalize it and reuse the same logic between the two clients?

cache/redis.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// RedisCache is a redis cache.
+type RedisCache struct {
```
pracucci, line 14: Everything here looks the exact same of `MemcachedCache`. Can we generalize it?

pracucci (commented):
I did an initial review. My main feedback is that there's a bunch of duplicated code between Redis and Memcached clients, that I think we should generalize and reuse. If generalization succeed, then I would like to see it happen in a preliminary PR (just the refactoring on memcached client), then reused here.

cache/redis_client.go
ortuman: Not sure what was the original intention here (copied this directly from Thanos). 

On k8s this should point to some service DNS name that resolves in front of a dynamic set of pods, and no manual resolution would be needed on client side. 

I'll remove this check. 👍

go.mod
```diff
@@ -4,16 +4,19 @@ go 1.18
 
 require (
```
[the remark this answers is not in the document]
ortuman: That was unintentional.

Updated `go.mod` to only add strictly necessary updates.

cache/redis_client_test.go
```diff
@@ -0,0 +1,219 @@
...
+		},
+		{
+			name: "MaxConcurrency=0",
+			redisConfig: func() RedisClientConfig {
+				cfg := DefaultRedisClientConfig
+				cfg.Endpoint = s.Addr()
+				cfg.MaxGetMultiConcurrency = 0
+				cfg.GetMultiBatchSize = 0
+				return cfg
+			},
```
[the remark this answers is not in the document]
ortuman: in this scenario, 0 = no concurrency limitation when fetching keys in multiple batches.

cache/redis_client.go
ortuman: > 1. Any reason why logic here was written differently?

I'd say the main difference between the two implementations is the sharding logic. For memcached this is done on client side whereas for Redis this logic is applied on the cluster instance, when configured in cluster mode. 

> Can we generalize it and reuse the same logic between the two clients?

Perhaps the code contained within the `doWithBatch` function would be susceptible to reuse?

However, it makes use of client-specific interface method for multiple fetch on each implementation (`GetMulti` for memcached and `MGet` for redis), so don't think it's worth it.

cache/util_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"reflect"
+	"runtime"
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+// ok fails the test if an err is not nil.
+func ok(tb testing.TB, err error, v ...interface{}) {
```
pracucci, line 15: You shouldn't use any of those, but `testify` library instead.

cache/cache.go
```diff
@@ -26,6 +51,7 @@ type Cache interface {
 
 const (
 	BackendMemcached = "memcached"
+	BackendRedis     = "redis"
 )
 
 type BackendConfig struct {
```
pracucci, line 57: I would add Redis support here. We want Redis support in Mimir too (experimental at the beginning).

pracucci: Discussed on Slack. Miguel prefers to give it a try in GEM first. I'm OK.

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+
+		if err != nil {
+			return nil, err
+		}
+
+		opts.TLSConfig = tlsClientConfig
+	}
+
+	cli := redis.NewUniversalClient(opts)
+	if reg != nil {
```
pracucci, line 182: [nit] Not needed to check `if reg != nil`. The wrapper handles it.

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// getMultiGate used to enforce the max number of concurrent GetMulti() operations.
+	getMultiGate gate.Gate
+
+	logger  log.Logger
+	metrics *clientMetrics
+}
+
+// NewRedisClient makes a new RedisClient.
+func NewRedisClient(logger log.Logger, name string, config RedisClientConfig, reg prometheus.Registerer) (RemoteCacheClient, error) {
```
pracucci, line 152: Wrap the logger with `log.With(logger, "name", name)`

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+		prometheus.WrapRegistererWithPrefix("redis_", reg),
+	)
+
+	c.asyncQueue = newAsyncQueue(config.MaxAsyncBufferSize, config.MaxAsyncConcurrency)
+
+	return c, nil
+}
+
+// SetAsync implement RemoteCacheClient.
+func (c *redisClient) SetAsync(ctx context.Context, key string, value []byte, ttl time.Duration) error {
```
pracucci, line 208: [non blocking, but we should do further refactoring in a follow up PR] This function is almost the exact copy of `memcachedClient.SetAsync()`.

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+	// Skip hitting redis at all if the item is bigger than the max allowed size.
+	if c.config.MaxItemSize > 0 && uint64(len(value)) > uint64(c.config.MaxItemSize) {
+		c.metrics.skipped.WithLabelValues(opSet, reasonMaxItemSize).Inc()
+		return nil
+	}
+
+	err := c.asyncQueue.submit(func() {
+		start := time.Now()
+
```
pracucci, line 217: Shouldn't we do `c.metrics.operations.WithLabelValues(opSet).Inc()` here as well, like we do in the memcached client?

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+		level.Warn(c.logger).Log("msg", "failed to mget items from redis", "err", err, "items", len(keys))
+		return nil
+	}
+	c.metrics.dataSize.WithLabelValues(opGetMulti).Observe(float64(total))
+	c.metrics.duration.WithLabelValues(opGetMulti).Observe(time.Since(start).Seconds())
+	return results
+}
+
+// Delete implement RemoteCacheClient.
+func (c *redisClient) Delete(ctx context.Context, key string) error {
```
pracucci, line 285: [follow up refactoring] This is almost an identical copy of the memcached client Delete.

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+
+// GetMulti implement RemoteCacheClient.
+func (c *redisClient) GetMulti(ctx context.Context, keys []string) map[string][]byte {
+	if len(keys) == 0 {
+		return nil
+	}
+	start := time.Now()
+	results := make(map[string][]byte, len(keys))
+	var mu sync.Mutex
+	var total int
```
pracucci, line 250: [nit] Just for clarity.

```suggestion
	var cacheHitBytes int
```

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		level.Warn(c.logger).Log("msg", "failed to mget items from redis", "err", err, "items", len(keys))
+		return nil
+	}
+	c.metrics.dataSize.WithLabelValues(opGetMulti).Observe(float64(total))
+	c.metrics.duration.WithLabelValues(opGetMulti).Observe(time.Since(start).Seconds())
```
pracucci, line 280: In memcached there are tracked for each "get multi" request. I would move it inside the `doWithBatch()` callback function to keep the same behaviour

cache/redis_client.go
```diff
@@ -0,0 +1,347 @@
...
+		PoolSize:     config.ConnectionPoolSize,
+		MinIdleConns: config.MinIdleConnections,
+		MaxConnAge:   config.MaxConnectionAge,
+		IdleTimeout:  config.IdleTimeout,
+	}
+
+	if config.TLSEnabled {
+		tlsConfig := config.TLS
+
+		tlsClientConfig, err := newClientConfig(logger, tlsConfig.CertPath, tlsConfig.KeyPath,
```
pracucci, line 171: I don't think we should use `newClientConfig()` at all. Since we use dskit TLS config struct, why don't we just use `config.TLS.GetTLSConfig()`?

cache/redis_client_test.go
```diff
@@ -0,0 +1,275 @@
...
+			redisConfig: func() RedisClientConfig {
+				cfg := defaultRedisClientConfig
+				cfg.Endpoint = flagext.StringSliceCSV{s.Addr()}
+				cfg.MaxGetMultiConcurrency = 2
+				cfg.GetMultiBatchSize = 2
+				return cfg
+			},
+		},
+		{
+			name: "MaxConcurrency=0",
```
pracucci, line 121:
```suggestion
			name: "MaxConcurrency and GetMultiBatchSize set to 0",
```

cache/redis_client_test.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		ok(t, err)
+	}
+	defer s.Close()
+	redisConfigs := []struct {
+		name        string
+		redisConfig func() RedisClientConfig
+	}{
+		{
+			name: "MaxConcurrency>0",
```
pracucci, line 111:
```suggestion
			name: "MaxConcurrency and GetMultiBatchSize set to a value > 0",
```

cache/redis_client_test.go
```diff
@@ -0,0 +1,219 @@
...
+					}
+					defer c.Stop()
+					defer s.FlushAll()
+					ctx := context.Background()
+					for k, v := range tt.args.data {
+						_ = c.SetAsync(ctx, k, v, time.Hour)
+					}
+
+					// Wait for SetAsync to finish.
+					time.Sleep(time.Millisecond * 250)
```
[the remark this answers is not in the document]
pracucci: This is fragile. Remove `time.Sleep()` and wrap the following assertion in a `test.Poll()`.

cache/redis_client_test.go
```diff
@@ -0,0 +1,275 @@
...
+	key1 := "key1"
+	value1 := []byte{1}
+
+	_ = c.SetAsync(context.Background(), key1, value1, time.Hour)
+
+	// Wait for SetAsync to finish.
+	time.Sleep(time.Millisecond * 250)
+
+	hits := c.GetMulti(context.Background(), []string{key1})
+	equals(t, map[string][]byte{key1: value1}, hits)
```
pracucci, line 185: Use `test.Pool()` instead of `time.Sleep()`.

cache/redis_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/grafana/dskit/flagext"
+
+	"github.com/alicebob/miniredis"
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	prom_testutil "github.com/prometheus/client_golang/prometheus/testutil"
+)
+
+func TestRedisCache(t *testing.T) {
```
pracucci, line 17: Isn't a duplicate of the test we have in `redis_client_test.go`?

ortuman: It is indeed, but using the Redis cache wrapper instead of Redis client interface. 

I don't see much value in having this duplicity, so I'll remove the one in `redis_client_test.go`.

cache/redis_client.go
```diff
@@ -0,0 +1,268 @@
...
+
+// GetMulti implement RemoteCacheClient.
+func (c *redisClient) GetMulti(ctx context.Context, keys []string) map[string][]byte {
+	if len(keys) == 0 {
+		return nil
+	}
+	start := time.Now()
+	results := make(map[string][]byte, len(keys))
+	var mu sync.Mutex
+	var cacheHitBytes int
```
pracucci, line 209: [nit] Move this inside the `doWithBatch()` callback function, please.

pracucci (approved):
3..2...1... LGTM! Thanks a lot for your patience! I just left a final nit 🙇

cache/redis_client.go
```diff
@@ -203,12 +203,15 @@ func (c *redisClient) GetMulti(ctx context.Context, keys []string) map[string][]
...
 	}
-	start := time.Now()
-	results := make(map[string][]byte, len(keys))
 	var mu sync.Mutex
-	var cacheHitBytes int
+	results := make(map[string][]byte, len(keys))
 
 	err := doWithBatch(ctx, len(keys), c.config.GetMultiBatchSize, c.getMultiGate, func(startIndex, endIndex int) error {
+		start := time.Now()
+		c.metrics.operations.WithLabelValues(opGetMulti).Inc()
```
pracucci, line 211: I didn't notice it in the previous last review. Thanks for fixing it too.

PR #246, opened 2022-12-19
cache: refactor memcached client code

ortuman:
**What this PR does**:

This PR is a follow-up of https://github.com/grafana/dskit/pull/245, based on suggested feedback. 

It aims to extract generic memcached behavior, so that it can be reused on alternate remote cache implementations such as Redis. 

Fixes N/A

**Checklist**
- [X] Tests updated


cache/memcached_client.go
```diff
@@ -587,47 +512,23 @@ func (c *memcachedClient) trackError(op string, err error) {
...
-			op()
-		case <-c.stop:
-			return
-		}
+		c.metrics.failures.WithLabelValues(op, reasonOther).Inc()
 	}
 }
 
 func (c *memcachedClient) resolveAddrsLoop() {
-	defer c.workers.Done()
```
pracucci, line 629: We lost this logic. We're not waiting for `resolveAddrsLoop()` go routine to end when we call `memcachedClient.Stop()`. However, I think it's not a problem.

cache/async_queue.go
```diff
@@ -0,0 +1,61 @@
...
+		return errAsyncQueueFull
+	}
+}
+
+func (q *asyncQueue) stop() {
+	close(q.stopCh)
+	q.workers.Wait()
+}
+
+func (q *asyncQueue) length() int {
```
pracucci, line 46: What is this used for?

cache/async_queue_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	var i atomic.Int32
+	_ = q.run(func() { i.Add(1) })
+	_ = q.run(func() { i.Add(-1) })
+	_ = q.run(func() { i.Add(1) })
+	_ = q.run(func() { i.Add(-1) })
+	_ = q.run(func() { i.Add(1) })
+
+	// Wait for all operations to finish.
+	time.Sleep(100 * time.Millisecond)
```
pracucci, line 24: Better to call `q.stop()` here. No need to `time.Sleep()` and we also test `stop()` waits until all workers have done.

cache/async_queue_test.go
```diff
@@ -0,0 +1,51 @@
...
+	doneCh := make(chan struct{})
+	defer func() {
+		close(doneCh)
+	}()
+
+	// Keep worker busy.
+	_ = q.run(func() {
+		<-doneCh
+	})
+	time.Sleep(100 * time.Millisecond)
```
pracucci, line 44: Why is this required? Shouldn't be.

pracucci (approved):

cache/remote_cache.go
```diff
@@ -12,51 +12,66 @@ import (
 
 // MemcachedCache is a memcached-based cache.
 type MemcachedCache struct {
```
56quarters, line 14: I'm unclear on why `MemcachedCache` exists now if it's exposing the same functionality as `remoteCache` after this refactor.

cache/async_queue.go
```diff
@@ -0,0 +1,61 @@
...
+	// Start a number of goroutines - processing async operations - equal
+	// to the max concurrency we have.
+	q.workers.Add(maxConcurrency)
+	for i := 0; i < maxConcurrency; i++ {
+		go q.asyncQueueProcessLoop()
+	}
+	return q
+}
+
+func (q *asyncQueue) run(op func()) error {
```
56quarters, line 32: nit: I would call this `submit()` since it doesn't run the provided closure immediately.

cache/async_queue_test.go
56quarters: I think your suggestion exposes a potential problem with the current implementation: not everything that has been queued will be executed if `.stop()` is called immediately due to the `select` statement in the worker loop.

cache/async_queue.go
pracucci: Good suggestion!

cache/remote_cache.go
pracucci: I agree we should simplify the design of the cache client, but I would suggest to postpone it to a follow up PR.

cache/async_queue_test.go
pracucci: Right, it's a valid concern.

cache/async_queue_test.go
ortuman: Here we try to simulate a long running task that prevents our only worker from pulling any enqueued operation. This way we can fill the queue until getting the expected error. 

The sleep before the loop is done in order to ensure that the worker is already running before enqueuing any extra ops. Otherwise we may end up getting a _queue full error_ in the case it haven't triggered yet.

cache/remote_cache.go
ortuman: See: https://github.com/grafana/dskit/pull/245#pullrequestreview-1220550728

PR #247, opened 2023-01-03
Remove trivial "util" exported package

56quarters:
**What this PR does**:

Replaced with an internal-only function. This package was never meant to be exported but the change got lost as part of [code review](https://github.com/grafana/dskit/pull/239#discussion_r1038312842).

See #239

**Checklist**
- [ ] Tests updated


internal/slices/slices.go
```diff
@@ -0,0 +1,11 @@
+package slices
+
+func Contains[T comparable](haystack []T, needle T) bool {
```
charleskorn, line 3: Like the variable names 😃 

Rather than implementing this ourselves, could we use `golang.org/x/exp/slices.Contains` instead?

charleskorn (approved):
LGTM, but we may be able to replace our `slices.Contains` with the implementation in `golang.org/x/exp/slices` instead.

internal/slices/slices.go
56quarters: Yes! But we aren't already including that package in dskit so I didn't want to add a dependency just for a few lines of code.

charleskorn: Makes sense, happy to leave it as-is.

pracucci (approved):
No objection 😉

PR #250, opened 2023-01-09
spanlogger: skip formatting if trace is not sampled

bboreham:
Motivation for this change is profiles like this from Mimir, where 4% of all memory allocation is in `log.InterleavedKVToFields`.

image

If the span was not sampled (and we sample something like 1 in 1,000 of those) then we don't need to format the output.
Add benchmark to demonstrate.
```
name          old time/op    new time/op    delta
SpanLogger-4     436ns ±26%     230ns ±20%  -47.19%  (p=0.008 n=5+5)

name          old alloc/op   new alloc/op   delta
SpanLogger-4      288B ± 0%      160B ± 0%  -44.44%  (p=0.008 n=5+5)

name          old allocs/op  new allocs/op  delta
SpanLogger-4      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)
```

I considered just setting `SpanLogger.Span` to nil if it's not sampled, rather than adding a boolean, but since the field is exported it could trip up someone using this library who expects there is always a span.

**Checklist**
- NA Tests updated
- NA `CHANGELOG.md` updated

### DIFF
--- a/spanlogger/spanlogger.go
+++ b/spanlogger/spanlogger.go
@@ -34,6 +34,7 @@ var (
 type SpanLogger struct {
 	log.Logger
 	opentracing.Span
+	sampled bool
 }
 
 // New makes a new SpanLogger with a log.Logger to send logs to. The provided context will have the logger attached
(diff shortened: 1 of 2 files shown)


pracucci (approved):
Nice work!

56quarters (approved):
LGTM, nice improvement.

PR #252, opened 2023-01-11
Ring: Add GetTokens to ReadRing interface

DylanGuedes:
**What this PR does**:
- Add a new `GetTokens` method to the `ReadRing` interface
- Add `ring.Ring` implementation of `GetTokens`

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -77,6 +77,9 @@ type ReadRing interface {
 
 	// CleanupShuffleShardCache should delete cached shuffle-shard subrings for given identifier.
 	CleanupShuffleShardCache(identifier string)
+
+	// GetTokens returns ring instance tokens.
+	GetTokens(ctx context.Context) Tokens
 }
 
 var (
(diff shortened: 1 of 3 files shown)


ring/ring.go
```diff
@@ -402,6 +405,14 @@ func (r *Ring) Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts,
 	}, nil
 }
 
+// GetTokens implement ReadRing.
```
pstibrany, line 408: I'm not sure if caller has any use for these tokens. Aren't `InstancesCount` or `GetAllHealthy` enough to check if ring is not empty?

(This method is also unsafe: it should create a copy and not expose internal data of `*Ring` to client, which can then modify it)

DylanGuedes: > I'm not sure if caller has any use for these tokens. Aren't InstancesCount or GetAllHealthy enough to check if ring is not empty?

They don't check for the number of tokens :(

pstibrany: > They don't check for the number of tokens :(

Do you want to protect against the situation when there are only instances with no tokens in the ring? Does that happen in your setup?

PR #255, opened 2023-01-20
Add Delete method to cache.Cache interface

lamida:
**What this PR does**:

- Refactor and simplify cache interfaces in dskit. We are adding the Delete method in the cache.Cache interface so that it can cover the use case on where cache items can be deleted.
- Implement backward compatible cache metrics exposition by moving backend type (e.g. redis/memcached) as label instead of part of metric name. Was part of #260

More context: ([internal link](https://github.com/grafana/backend-enterprise/pull/4030#issuecomment-1379944449)).

Dependency to be updated (will do once this PR merged):
* Backend Enterprise: https://github.com/grafana/backend-enterprise/pull/4113
* Mimir: https://github.com/grafana/mimir/pull/4078

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,26 @@
 * [ENHANCEMENT] Add the metrics package to support per tenant metric aggregation. #258
 * [ENHANCEMENT] Cache: Add Delete method to cache.Cache interface. #255
 * [ENHANCEMENT] Cache: Add backward compatible metrics exposition for cache. #255
+  * `<prefix>_cache_client_info{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_dns_failures_total{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_dns_lookups_total{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_dns_provider_results{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_getmulti_gate_duration_seconds_bucket{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_getmulti_gate_duration_seconds_count{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_getmulti_gate_duration_seconds_sum{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_getmulti_gate_queries_concurrent_max{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_getmulti_gate_queries_in_flight{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_hits_total{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_data_size_bytes_bucket{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_data_size_bytes_count{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_data_size_bytes_sum{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_duration_seconds_bucket{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_duration_seconds_count{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_duration_seconds_sum{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_failures_total{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operation_skipped_total{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_operations_total{backend="[memcached|redis]",...}`
+  * `<prefix>_cache_requests_total{backend="[memcached|redis]",...}`
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
 * [BUGFIX] Ring: `ring_member_ownership_percent` and `ring_tokens_owned` metrics are not updated on scale down. #109
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,7 @@
 * [ENHANCEMENT] Cache: Add the ability to use a custom memory allocator for cache results. #249
 * [ENHANCEMENT] Add the metrics package to support per tenant metric aggregation. #258
 * [ENHANCEMENT] Cache: Add Delete method to cache.Cache interface. #255
-* [ENHANCEMENT] Cache: Add backward compatible metrics exposition for cache. #260
+* [ENHANCEMENT] Cache: Add backward compatible metrics exposition for cache. #255
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
 * [BUGFIX] Ring: `ring_member_ownership_percent` and `ring_tokens_owned` metrics are not updated on scale down. #109
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 * [ENHANCEMENT] Cache: Add the ability to use a custom memory allocator for cache results. #249
 * [ENHANCEMENT] Add the metrics package to support per tenant metric aggregation. #258
 * [ENHANCEMENT] Cache: Add Delete method to cache.Cache interface. #255
+* [ENHANCEMENT] Cache: Add backward compatible metrics exposition for cache. #260
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
 * [BUGFIX] Ring: `ring_member_ownership_percent` and `ring_tokens_owned` metrics are not updated on scale down. #109
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@
 * [ENHANCEMENT] Execute health checks in ring client pool concurrently. #237
 * [ENHANCEMENT] Cache: Add the ability to use a custom memory allocator for cache results. #249
 * [ENHANCEMENT] Add the metrics package to support per tenant metric aggregation. #258
+* [ENHANCEMENT] Cache: Add Delete method to cache.Cache interface. #255
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
 * [BUGFIX] Ring: `ring_member_ownership_percent` and `ring_tokens_owned` metrics are not updated on scale down. #109
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -43,14 +43,10 @@ type Cache interface {
 	// Option instances may be passed to modify the behavior of this Fetch call.
 	Fetch(ctx context.Context, keys []string, opts ...Option) map[string][]byte
 
-	Name() string
-}
-
-// DeletableCache is cache that can delete its key.
-type DeletableCache interface {
-	Cache
 	// Delete cache entry with the given key if it exists.
 	Delete(ctx context.Context, key string) error
+
+	Name() string
 }
 
 // Options are used to modify the behavior of an individual call to get results
@@ -132,17 +128,4 @@ func CreateClient(cacheName string, cfg BackendConfig, logger log.Logger, reg pr
 	default:
 		return nil, errors.Errorf("unsupported cache type for cache %s: %s", cacheName, cfg.Backend)
 	}
-}
-
-func CreateClientWithDelete(cacheName string, cfg BackendConfig, logger log.Logger, reg prometheus.Registerer) (DeletableCache, error) {
-	client, err := CreateClient(cacheName, cfg, logger, reg)
-	if err != nil {
-		return nil, err
-	}
-
-	clientCacheWithDelete, ok := client.(DeletableCache)
-	if !ok {
-		return nil, fmt.Errorf("client is not DeletableCache type")
-	}
-	return clientCacheWithDelete, nil
 }
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -46,6 +46,13 @@ type Cache interface {
 	Name() string
 }
 
+// DeletableCache is cache that can delete its key.
+type DeletableCache interface {
+	Cache
+	// Delete cache entry with the given key if it exists.
+	Delete(ctx context.Context, key string) error
+}
+
 // Options are used to modify the behavior of an individual call to get results
 // from a cache backend. They are constructed by applying Option callbacks passed
 // to a client method to a default Options instance.
@@ -84,8 +91,9 @@ const (
 )
 
 type BackendConfig struct {
-	Backend   string          `yaml:"backend"`
-	Memcached MemcachedConfig `yaml:"memcached"`
+	Backend   string            `yaml:"backend"`
+	Memcached MemcachedConfig   `yaml:"memcached"`
+	Redis     RedisClientConfig `yaml:"redis" category:"experimental"`
 }
 
 // Validate the config.
@@ -101,7 +109,7 @@ func (cfg *BackendConfig) Validate() error {
 	return nil
 }
 
-func CreateClient(cacheName string, cfg BackendConfig, logger log.Logger, reg prometheus.Registerer) (Cache, error) {
+func CreateClient(cacheName string, cfg BackendConfig, logger log.Logger, reg prometheus.Registerer) (DeletableCache, error) {
 	switch cfg.Backend {
 	case "":
 		// No caching.
@@ -113,6 +121,13 @@ func CreateClient(cacheName string, cfg BackendConfig, logger log.Logger, reg pr
 			return nil, errors.Wrapf(err, "failed to create memcached client")
 		}
 		return NewMemcachedCache(cacheName, logger, client, reg), nil
+
+	case BackendRedis:
+		client, err := NewRedisClient(logger, cacheName, cfg.Redis, reg)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create redis client")
+		}
+		return NewRedisCache(cacheName, logger, client, reg), nil
 
 	default:
 		return nil, errors.Errorf("unsupported cache type for cache %s: %s", cacheName, cfg.Backend)
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -109,7 +109,7 @@ func (cfg *BackendConfig) Validate() error {
 	return nil
 }
 
-func CreateClient(cacheName string, cfg BackendConfig, logger log.Logger, reg prometheus.Registerer) (DeletableCache, error) {
+func CreateClient(cacheName string, cfg BackendConfig, logger log.Logger, reg prometheus.Registerer) (Cache, error) {
 	switch cfg.Backend {
 	case "":
 		// No caching.
@@ -132,4 +132,17 @@ func CreateClient(cacheName string, cfg BackendConfig, logger log.Logger, reg pr
 	default:
 		return nil, errors.Errorf("unsupported cache type for cache %s: %s", cacheName, cfg.Backend)
 	}
+}
+
+func CreateClientWithDelete(cacheName string, cfg BackendConfig, logger log.Logger, reg prometheus.Registerer) (DeletableCache, error) {
+	client, err := CreateClient(cacheName, cfg, logger, reg)
+	if err != nil {
+		return nil, err
+	}
+
+	clientCacheWithDelete, ok := client.(DeletableCache)
+	if !ok {
+		return nil, fmt.Errorf("client is not DeletableCache type")
+	}
+	return clientCacheWithDelete, nil
 }
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -43,10 +43,10 @@ type Cache interface {
 	// Option instances may be passed to modify the behavior of this Fetch call.
 	Fetch(ctx context.Context, keys []string, opts ...Option) map[string][]byte
 
+	Name() string
+
 	// Delete cache entry with the given key if it exists.
 	Delete(ctx context.Context, key string) error
-
-	Name() string
 }
 
 // Options are used to modify the behavior of an individual call to get results
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -87,9 +87,8 @@ const (
 )
 
 type BackendConfig struct {
-	Backend   string            `yaml:"backend"`
-	Memcached MemcachedConfig   `yaml:"memcached"`
-	Redis     RedisClientConfig `yaml:"redis" category:"experimental"`
+	Backend   string          `yaml:"backend"`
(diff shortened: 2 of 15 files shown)


cache/cache.go
```diff
@@ -46,6 +46,13 @@ type Cache interface {
 	Name() string
 }
 
+// DeletableCache is cache that can delete its key.
+type DeletableCache interface {
```
pracucci, line 50: Do we need this interface? Can't we just add `Delete()` to the `Cache` interface?

pracucci (approved):
LGTM (modulo a comment on the interface). This means we'll implicitly add Redis support to Mimir, which is OK. After this PR will be merged, can you open a PR to Mimir to update dskit?

cache/cache.go
pstibrany: While I agree, I'd point out that non-compatible API changes are quite painful for updates.

pracucci: Assuming this interface is used with dskit cache client implementations, is it really an issue?

pstibrany: Yes, changing `Cache` interface will require a cascading update of multiple dependencies in order to update our internal enterprise product. It is doable (we've done it many times, last time earlier this week), just pointing out the pain.

pracucci: I've seen you introduced `CreateClientWithDelete()` in the last commit. Is that necessary?

56quarters: I agree that we should just add `Delete` to the `Cache` interface. It will require cascading changes but I think they'll be straightforward to fix: the dependent projects will either compile or not, no subtle errors to debug.

lamida: > I've seen you introduced CreateClientWithDelete() in the last commit. Is that necessary?
It was intended to create DeletableCache without affecting existing Cache interface useage.

Nevertheless I will just update the implementation by adding Delete method to the cache interface.

cache/compression.go
```diff
@@ -98,3 +98,8 @@ func (s *snappyCache) Fetch(ctx context.Context, keys []string, opts ...Option)
 func (s *snappyCache) Name() string {
 	return s.next.Name()
 }
+
+func (s *snappyCache) Delete(_ context.Context, _ string) error {
+	// no-op
```
Logiraptor, line 103: Given most of these are wrappers around another Cache implementation, I think it would be better to at least delegate the deletion through to the underlying cache. It would be confusing to use snappyCache for example and suddenly deletes stop working.

lamida: Fixed.

cache/mock.go
```diff
@@ -110,6 +111,11 @@ func (m *InstrumentedMockCache) Name() string {
 	return m.cache.Name()
 }
 
+func (m *InstrumentedMockCache) Delete(_ context.Context, _ string) error {
+	// no-op
+	return nil
```
pracucci, line 116: Should call the wrapped one.

cache/remote_cache.go
```diff
@@ -86,14 +86,14 @@ func newRemoteCache(name string, logger log.Logger, remoteClient RemoteCacheClie
...
-func (c *remoteCache) Store(ctx context.Context, data map[string][]byte, ttl time.Duration) {
+func (c *remoteCache) Store(_ context.Context, data map[string][]byte, ttl time.Duration) {
 	var (
 		firstErr error
 		failed   int
 	)
 
 	for key, val := range data {
-		if err := c.remoteClient.SetAsync(ctx, key, val, ttl); err != nil {
+		if err := c.remoteClient.SetAsync(context.Background(), key, val, ttl); err != nil {
```
pracucci, line 96: Should use the input context.

cache/tracing.go
```diff
@@ -45,3 +45,8 @@ func (t SpanlessTracingCache) Fetch(ctx context.Context, keys []string, opts ...
 func (t SpanlessTracingCache) Name() string {
 	return t.c.Name()
 }
+
+func (t SpanlessTracingCache) Delete(_ context.Context, _ string) error {
+	// no-op
+	return nil
```
pracucci, line 51: Should call the wrapped one.

cache/versioned.go
```diff
@@ -49,6 +49,11 @@ func (c Versioned) Name() string {
 	return c.cache.Name()
 }
 
+func (c Versioned) Delete(_ context.Context, _ string) error {
+	// no-op
+	return nil
```
pracucci, line 54: Why not implemented here?

cache/lru.go
```diff
@@ -134,3 +134,8 @@ func (l *LRUCache) Fetch(ctx context.Context, keys []string, opts ...Option) (re
 func (l *LRUCache) Name() string {
 	return "in-memory-" + l.name
 }
+
+func (l *LRUCache) Delete(_ context.Context, _ string) error {
+	// no-op
+	return nil
```
pracucci, line 140: We should implement it here.

pracucci (commented):
Jon, I think you should implement `Delete()` in the various caches. I left comments.

cache/versioned.go
lamida: Added

cache/lru.go
lamida: Added

Logiraptor: I think this also needs to call delete on the underlying cache `l.c.Delete`, right?

lamida: That's true :palm-face:. Added.

cache/mock.go
```diff
@@ -110,6 +111,10 @@ func (m *InstrumentedMockCache) Name() string {
 	return m.cache.Name()
 }
 
+func (m *InstrumentedMockCache) Delete(ctx context.Context, key string) error {
```
56quarters, line 114: Maybe add a counter for deletes here?

56quarters (approved):
LGTM with one suggestion.

Logiraptor (approved):
LGTM modulo CI passing

lamida:
Please help do one (hopefully) final review @pracucci . I will update backend-enterprise and mimir dependency after this merged. I replied to the comment about context.Background and for now will stop cascading the context from the caller. Let me know what do you think.

cache/mock.go
lamida: Added

CHANGELOG.md
```diff
@@ -78,6 +78,8 @@
 * [ENHANCEMENT] Execute health checks in ring client pool concurrently. #237
 * [ENHANCEMENT] Cache: Add the ability to use a custom memory allocator for cache results. #249
 * [ENHANCEMENT] Add the metrics package to support per tenant metric aggregation. #258
+* [ENHANCEMENT] Cache: Add Delete method to cache.Cache interface. #255
+* [ENHANCEMENT] Cache: Add backward compatible metrics exposition for cache. #260
```
pracucci, line 82:
```suggestion
* [ENHANCEMENT] Cache: Add backward compatible metrics exposition for cache. #255
```

cache/lru.go
```diff
@@ -134,3 +134,8 @@ func (l *LRUCache) Fetch(ctx context.Context, keys []string, opts ...Option) (re
 func (l *LRUCache) Name() string {
 	return "in-memory-" + l.name
 }
+
+func (l *LRUCache) Delete(ctx context.Context, key string) error {
+	l.lru.Remove(key)
```
pracucci, line 139: You need to take the `mtx` while accessing the `lru`. Release it before calling `l.c.Delete()`.

cache/lru_test.go
```diff
@@ -56,6 +56,11 @@ func TestLRUCache_StoreFetch(t *testing.T) {
 		# TYPE cache_memory_requests_total counter
 		cache_memory_requests_total{name="test"} 4
 	`)))
+
+	err = lru.Delete(context.Background(), "buzz")
```
pracucci, line 60: [nit] Please rename the test to `TestLRUCache_StoreFetchDelete`. Please also add an assertion on metrics at the end of this test. It should be something like this:

```
	require.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(`
		# HELP cache_memory_items_count Total number of items currently in the in-memory cache.
		# TYPE cache_memory_items_count gauge
		cache_memory_items_count{name="test"} 2
		# HELP cache_memory_hits_total Total number of requests to the in-memory cache that were a hit.
		# TYPE cache_memory_hits_total counter
		cache_memory_hits_total{name="test"} 2
		# HELP cache_memory_requests_total Total number of requests to the in-memory cache.
		# TYPE cache_memory_requests_total counter
		cache_memory_requests_total{name="test"} 4
	`)))
```

cache/redis_client.go
```diff
@@ -89,8 +91,8 @@ type RedisClientConfig struct {
 	// If set to 0, concurrency is unlimited.
 	MaxGetMultiConcurrency int `yaml:"max_get_multi_concurrency" category:"advanced"`
 
-	// GetMultiBatchSize specifies the maximum size per batch for mget.
-	GetMultiBatchSize int `yaml:"get_multi_batch_size" category:"advanced"`
+	// MaxGetMultiBatchSize specifies the maximum size per batch for mget.
+	MaxGetMultiBatchSize int `yaml:"max_get_multi_batch_size" category:"advanced"`
```
pracucci, line 95: This change is fine (actually wanted to keep consistency with the memcached client) but please mention it in the CHANGELOG.

cache/remote_cache.go
```diff
@@ -8,13 +8,25 @@ import (
...
 )
 
 var (
 	_ Cache = (*MemcachedCache)(nil)
 	_ Cache = (*RedisCache)(nil)
 )
 
+const (
+	labelName             = "name"
+	labelBackend          = "backend"
```
pracucci, line 22: [nit] I would clarify it. Especially `labelName` is very generic.

```suggestion
	labelCacheName             = "name"
	labelBackendName          = "backend"
```

CHANGELOG.md
pracucci, line 82: Clarify the impact of this change please.

cache/versioned_test.go
```diff
@@ -31,5 +37,14 @@ func TestVersioned(t *testing.T) {
 		assert.Equal(t, v1Data, resV1)
 		resV2 := v2.Fetch(context.Background(), []string{"hit", "miss"})
 		assert.Equal(t, v2Data, resV2)
+
+		err := v1.Delete(context.Background(), "hit")
+		require.NoError(t, err)
+		resV1 = v1.Fetch(context.Background(), []string{"hit", "miss"})
+		assert.Equal(t, map[string][]uint8{}, resV1)
+		err = v2.Delete(context.Background(), "hit")
```
pracucci, line 45: I think here you want to test that V1 doesn't delete V2 too. So I think it's more interesting calling `v2.Fetch()` and ensuring it still returns a value. Then you can try to delete V2 too.

cache/remote_cache.go
```diff
@@ -8,13 +8,25 @@ import (
...
+	backendRedis          = "redis"
+	backendMemcached      = "memcached"
+	cachePrefix           = "cache_"
```
pracucci, line 25: [nit] I would clarify which prefix it is.

```suggestion
	cacheMetricNamePrefix           = "cache_"
```

cache/remote_cache.go
```diff
@@ -8,13 +8,25 @@ import (
...
+	getMultiPrefix        = "getMulti_"
```
pracucci, line 26: Pay attention that the "m" should be lowercase. I would also clarify which prefix it is.

```suggestion
	getMultiMetricNamePrefix        = "getmulti_"
```

cache/memcached_client.go
```diff
@@ -178,15 +179,21 @@ func newMemcachedClient(
 	reg prometheus.Registerer,
 	name string,
 ) (*memcachedClient, error) {
+	newRegisterer := prometheus.WrapRegistererWith(
```
pracucci, line 182: I would do the naming the other way round. I would call `newRegisterer` just `reg` and the other `legacyReg` to clarify which one is the legacy one.

cache/memcached_client.go
```diff
@@ -196,13 +203,16 @@ func newMemcachedClient(
 		addressProvider: addressProvider,
 		stop:            make(chan struct{}, 1),
 		getMultiGate: gate.New(
-			prometheus.WrapRegistererWithPrefix("memcached_getmulti_", reg),
+			promregistry.TeeRegisterer{
```
pracucci, line 206: I don't think this is correct. The final metric name will start with `getmulti_` while it should be `(cache|memcached)_getmulti_*`.

cache/memcached_client.go
```diff
@@ -196,13 +203,16 @@ func newMemcachedClient(
...
+				prometheus.WrapRegistererWithPrefix(getMultiPrefix, reg),
+				prometheus.WrapRegistererWithPrefix(getMultiPrefix, newRegisterer),
+			},
 			config.MaxGetMultiConcurrency,
 		),
 	}
 
 	c.clientInfo = promauto.With(reg).NewGaugeFunc(prometheus.GaugeOpts{
```
pracucci, line 214: Here you should use `backwardCompatibleRegs` otherwise it's a breaking change.

cache/redis_client.go
```diff
@@ -172,11 +177,14 @@ func NewRedisClient(logger log.Logger, name string, config RedisClientConfig, re
 		opts.TLSConfig = tlsClientConfig
 	}
 
-	reg = prometheus.WrapRegistererWith(prometheus.Labels{"name": name}, reg)
+	if reg != nil {
```
pracucci, line 180: [nit] No need for this `if`. The wrappers can handle a nil registry.

cache/redis_client.go
```diff
@@ -185,10 +193,28 @@ func NewRedisClient(logger log.Logger, name string, config RedisClientConfig, re
 	}
 	if config.MaxGetMultiConcurrency > 0 {
 		c.getMultiGate = gate.New(
-			prometheus.WrapRegistererWithPrefix("redis_getmulti_", reg),
+			prometheus.WrapRegistererWithPrefix(getMultiPrefix, reg),
```
pracucci, line 196: Same problem here. The final one will be `getmulti_cache_*` instead of `cache_getmulti_*`.

cache/redis_client.go
```diff
@@ -185,10 +193,28 @@ func NewRedisClient(logger log.Logger, name string, config RedisClientConfig, re
...
 			config.MaxGetMultiConcurrency,
 		)
 	}
+
+	c.clientInfo = promauto.With(reg).NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "redis_client_info",
```
pracucci, line 202: This metric name is not consistent with the memcached client.

pracucci (commented):
Thanks Jon for working on this. I left some comments. Generally speaking I don't like very much bundling together two unrelated changes: adding `Delete()` and changing metrics. I think it adds confusion and makes the review harder. Now, given you've already done this work let's keep it bundled this time, but next time they should be two different PRs.

cache/remote_cache.go
lamida: Thank you so much for catching this. I did multiple manual tests and this was missed from my visual observation. I will see if I can test to catch this.

cache/memcached_client.go
lamida: In my test, it gives me the correct result. Examples:
```
versioned_client_cache_getmulti_gate_queries_concurrent_max{backend="redis",name="admin-cache"} 100
thanos_cache_getmulti_gate_queries_in_flight{backend="memcached",component="query-frontend",name="frontend-cache"} 0
```

So from what I can see, `prometheus.WrapRegisterWithPrefix` will prepend passed registery and add prefix for subsequent wrap or for metrics which will use the Registerer.

I am building a simple program to verify:
```
package main

import (
	"net/http"
	"time"

	"github.com/prometheus/client_golang/prometheus"
	"github.com/prometheus/client_golang/prometheus/promauto"
	"github.com/prometheus/client_golang/prometheus/promhttp"
)

var (
	opsProcessed prometheus.Counter
)

func main() {
	reg := prometheus.NewRegistry()
	reg1 := prometheus.WrapRegistererWithPrefix("one_", reg)
	reg2 := prometheus.WrapRegistererWithPrefix("two_", reg1)
	reg3 := prometheus.WrapRegistererWithPrefix("three_", reg2)

	opsProcessed = promauto.With(reg3).NewCounter(prometheus.CounterOpts{
		Name: "myapp_processed_ops_total",
		Help: "The total number of processed events",
	})
	go func() {
		for {
			opsProcessed.Inc()
			time.Sleep(2 * time.Second)
		}
	}()
	http.Handle("/metrics2", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg3}))
	http.ListenAndServe(":2112", nil)
}
```

It gives me the following metrics:
```
# HELP one_two_three_myapp_processed_ops_total The total number of processed events
# TYPE one_two_three_myapp_processed_ops_total counter
one_two_three_myapp_processed_ops_total 2
# HELP one_two_three_promhttp_metric_handler_errors_total Total number of internal errors encountered by the promhttp metric handler.
# TYPE one_two_three_promhttp_metric_handler_errors_total counter
one_two_three_promhttp_metric_handler_errors_total{cause="encoding"} 0
one_two_three_promhttp_metric_handler_errors_total{cause="gathering"} 0
```

Please let me know if I might still misunderstand something. :)

cache/redis_client.go
lamida: Same reply on above https://github.com/grafana/dskit/pull/255#discussion_r1098156614

cache/memcached_client.go
lamida: Actually, I am not adding `backend="memcached"` label intentionally to keep this metric in the old format because `client_info` labels are really client specific and removing memcached from metric name to label might make it confusing. Both redis and memcached will have different set of labels. In this line, `reg` (now renamed to `legacyReg` already has `memcached_` prefix, hence this line will produce the following as in the old format:

```
thanos_memcached_client_info{component="query-frontend",dns_provider_update_interval="30s",max_async_buffer_size="25000",max_async_concurrency="50",max_get_multi_batch_size="100",max_get_multi_concurrency="100",max_idle_connections="100",max_item_size="1048576",name="frontend-cache",timeout="500ms"} 1
versioned_client_memcached_client_info{dns_provider_update_interval="30s",max_async_buffer_size="25000",max_async_concurrency="50",max_get_multi_batch_size="100",max_get_multi_concurrency="100",max_idle_connections="100",max_item_size="1048576",name="admin-cache",timeout="200ms"} 1
```
For redis, I want to have the following format (in this commit still inconsistently produce `versioned_client_cache_redis_client_info`):
```
versioned_client_redis_client_info{backend="redis",connection_pool_size="100",dial_timeout="5s",max_async_buffer_size="25000",max_async_concurrency="50",max_get_multi_batch_size="100",max_get_multi_concurrency="100",max_item_size="1048576",name="admin-cache",read_timeout="3s",write_timeout="3s"} 1
```

Definitely we will maintain backward compatible change for memcached client_info, but do you think we still need to remove cache from metric name and move it to label?

cache/redis_client.go
lamida: As mentioned in https://github.com/grafana/dskit/pull/255#discussion_r1098161871 memcached will produce `<prefix>_memcached_client_info` as this redis format. The reason to keep cache name in the metric name because different cache will expose different label.

Same question do you think we still need to remove cache from metric name and move it to `backend` label with the above reason?

cache/redis_client.go
lamida: The first time this is added, the config name and variable name missing max prefix. Ok I will add this new config name in the dskit changelog.

lamida: Also, the renaming above is to make the config name consistent with Memcached.

CHANGELOG.md
lamida: Changelog is updated with list of metrics with new format.

lamida:
Hi @pracucci thank you for the review yesterday. Yes this PR contains merge from #260. Sorry to make this PR bigger and contain more than one topic.

I have applied all of your PR suggestions and replied to some comments. The main points to summarise here again:
* `prometheus.WrapRegistererWithPrefix` behaviour is replied here: https://github.com/grafana/dskit/pull/255#discussion_r1098156614
* I reply a comment on `client_info` metric name (https://github.com/grafana/dskit/pull/255#discussion_r1098161871) asking whether we can have cache name in the metric name. Nevertheless I still followed your suggestion. But let me know if you think my reply there is reasonable.

cache/memcached_client.go
pracucci: You're right. What you did is correct. Thanks!

cache/redis_client.go
pracucci: What you did is correct. I would just add a comment both here and in the memcached one to explain why the metric names here are specific.

cache/redis_client.go
```diff
@@ -185,10 +191,28 @@ func NewRedisClient(logger log.Logger, name string, config RedisClientConfig, re
...
 	if config.MaxGetMultiConcurrency > 0 {
 		c.getMultiGate = gate.New(
-			prometheus.WrapRegistererWithPrefix("redis_getmulti_", reg),
+			prometheus.WrapRegistererWithPrefix(getMultiMetricNamePrefix, reg),
 			config.MaxGetMultiConcurrency,
 		)
 	}
+
+	c.clientInfo = promauto.With(reg).NewGaugeFunc(prometheus.GaugeOpts{
+		Name: clientInfoMetricName,
```
pracucci, line 200: You were convincing about using per-backend specific metric names. Now we're using a common metric name. What has changed?

cache/memcached_client.go
```diff
@@ -175,18 +176,24 @@ func newMemcachedClient(
...
 	selector updatableServerSelector,
 	config MemcachedClientConfig,
-	reg prometheus.Registerer,
+	legacyRegister prometheus.Registerer,
 	name string,
 ) (*memcachedClient, error) {
+	reg := prometheus.WrapRegistererWith(
+		prometheus.Labels{labelCacheBackend: backendMemcached},
+		prometheus.WrapRegistererWithPrefix(cacheMetricNamePrefix, legacyRegister))
+	legacyRegister = prometheus.WrapRegistererWithPrefix(legacyMemcachedPrefix, legacyRegister)
```
pracucci, line 185: Ops. Some renaming was applied to the input param too. The input registerer is not legacy, so should just be named `reg`.

I'm suggesting something like this:

```suggestion
	reg prometheus.Registerer,
	name string,
) (*memcachedClient, error) {
	legacyRegister := prometheus.WrapRegistererWithPrefix(legacyMemcachedPrefix, reg)
	reg = prometheus.WrapRegistererWith(
		prometheus.Labels{labelCacheBackend: backendMemcached},
		prometheus.WrapRegistererWithPrefix(cacheMetricNamePrefix, reg))
```

cache/redis_client.go
lamida: Yes nothing changes. Initially I wanted to keep cache name in the metric name for client_info but now decided to just use backend label as everything else (backward compatible behaviour is kept for memcached_client_info). I just moved "client_info" to common variable name because it is used in two places now.

cache/memcached_client.go
lamida: I will check and apply the suggestion once I got back in front of my laptop.

cache/redis_client.go
pracucci: Ok. Sounds good. Don't have a big preference between the two options.

pracucci (approved):
Thanks Jon for your patience! LGTM modulo [this comment](https://github.com/grafana/dskit/pull/255#discussion_r1098426160).

PR #258, opened 2023-01-26
Add metrics helper with history

gotjosh:
**What this PR does**:

Copies `pkg/util/metrics_helper.go` as `metrics/user_registries.go` from Mimir directly while preserving history.

This probably doesn't work as is but let's use this as the base for fixing whatever needs fixing - are these names OK?

https://github.com/grafana/mimir/issues/4082

**Checklist**
- [x] Tests updated
- [x] Rename files to remove traces of the word `user`


metrics/user_registries.go
```diff
@@ -0,0 +1,811 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
56quarters, line 1: dskit is Apache 2. I'm also not sure if we should be using these `Provenance` headers since we don't elsewhere.

metrics/user_registries.go
```diff
@@ -0,0 +1,811 @@
...
+			}
+
+			aggregateFn(key, mlv.labelValues, val)
+		}
+	}
+}
+
+// MetricFamiliesPerUser is a collection of metrics gathered via calling Gatherer.Gather() method on different
+// gatherers, one per user.
+type MetricFamiliesPerUser []struct {
```
56quarters, line 154: Could we use `Tenant` in place of `User` here since we have a `tenant/` package in dskit and use that term elsewhere?

gotjosh:
@56quarters, thanks for the feedback - at the moment, I'm looking for feedback around package name, file name and approach to copying history. 

With that set, I'll do a round of changes and incorporate your feedback.

metrics/user_registries.go
```diff
@@ -0,0 +1,809 @@
...
+	builder := labels.NewBuilder(nil)
+	for _, pair := range pairs {
+		builder.Set(pair.GetName(), pair.GetValue())
+	}
+	return builder.Labels(nil)
+}
+
+// GetSumOfHistogramSampleCount returns the sum of samples count of histograms matching the provided metric name
+// and optional label matchers. Returns 0 if no metric matches.
+func GetSumOfHistogramSampleCount(families []*dto.MetricFamily, metricName string, matchers labels.Selector) uint64 {
```
gotjosh, line 756: There's a problem with this function.

For context, it is only used in these two places https://github.com/grafana/mimir/blob/eb812fd67b21e89b055e88946e03b321455551bc/pkg/distributor/ha_tracker_test.go#L255-L263

But it imports the whole of Prometheus because it needs the `labels` package to use `matchers`, I took a look at https://github.com/prometheus/common, and there's no such struct (that allows us to do regex matching on labels).

The way I see it we have a couple of options:

- Import Prometheus as part of dskit (there's no precedent for this it seems)
- Take `common.Matcher` from the silence package in prometheus/common and attempt to do the comparison that way. 
- Rewrite the test that uses this function in Mimir

gotjosh: I've removed it for now - think it's the safest option.

gotjosh:
I'm gonna leave it here for now - I have removed all traces of "user" from the main file, but left the "user" label in the tests as that's what is used in Mimir.

We should probably also rename the file to `tenant_registries.go` but I'll leave that when I get some confirmation that this is the right direction.

pracucci (approved):
I'm also happy with the changes. I agree on CHANGELOG. Would be great if you could upgrade dskit in Mimir after that, but please wait until https://github.com/grafana/mimir/pull/4078 is merged, otherwise we'll duplicate the effort.

gotjosh:
@pstibrany @pracucci I can't merge via merge commit it says, "this is not enabled for this repository", I'm not an admin for this repository can you please take a look?

PR #260, opened 2023-01-30
Make metrics cache exposition consistent

lamida:
**What this PR does**:

Make cache metrics exposition to be consistent across different cache backends. This PR is to be merged to https://github.com/grafana/dskit/pull/255 because they are in the related work. Reviewers could review the change in #255 too.

As we are working [on adding Redis cache](https://github.com/grafana/backend-enterprise/pull/4030) for the admin bucket in GEM, we see that the metrics exposition for the cache component can be improved.

The existing metrics format is having metrics backend type in the name. As an example:

```
versioned_client_memcached_hits_total
```

We want to improve it by moving the metric type to the label.
```
versioned_client_hits_total{backend: "memcached"}
# hence can support new backend type
versioned_client_hits_total{backend: "redis"}
```

Since we don't want to break existing metrics, we will support both metric expositions. We are doing so by adding additional `prometheus.Registry` to support the old and new formats in the Memcache metric exposition. To avoid breaking the API, we are adding a new contructor to the affected components to accept slice of `prometheus.Registerer`.

Note that, since Prometheus seems doesn't really support multiple registries out of the box (at least from what I know), we have to create the metrics using `prometheus` package instead of `promauto` because we will register the metrics at once to the multiple registries. Because of this, we need to add `faillint` ignore in all codes that call this "blacklisted" package. If there is a better suggestion, we can improve this.

As mentioned in #255 following are dependency to be updated (will do once this and #255 merged):
* Backend Enterprise: https://github.com/grafana/backend-enterprise/pull/4113
* Mimir: https://github.com/grafana/mimir/pull/4078

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/cache/client.go
+++ b/cache/client.go
@@ -35,6 +35,7 @@ type clientMetrics struct {
 }
 
 func newClientMetrics(regs []prometheus.Registerer) *clientMetrics {
+	//lint:ignore faillint need to apply the metric to multiple registerer
 	operations := prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "operations_total",
 		Help: "Total number of operations against cache.",
@@ -43,6 +44,7 @@ func newClientMetrics(regs []prometheus.Registerer) *clientMetrics {
 	operations.WithLabelValues(opSet)
 	operations.WithLabelValues(opDelete)
 
+	//lint:ignore faillint need to apply the metric to multiple registerer
 	failures := prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "operation_failures_total",
 		Help: "Total number of operations against cache that failed.",
@@ -55,6 +57,7 @@ func newClientMetrics(regs []prometheus.Registerer) *clientMetrics {
 		failures.WithLabelValues(op, reasonOther)
 	}
 
+	//lint:ignore faillint need to apply the metric to multiple registerer
 	skipped := prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "operation_skipped_total",
 		Help: "Total number of operations against cache that have been skipped.",
@@ -63,6 +66,7 @@ func newClientMetrics(regs []prometheus.Registerer) *clientMetrics {
 	skipped.WithLabelValues(opSet, reasonMaxItemSize)
 	skipped.WithLabelValues(opSet, reasonAsyncBufferFull)
 
+	//lint:ignore faillint need to apply the metric to multiple registerer
 	duration := prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Name:    "operation_duration_seconds",
 		Help:    "Duration of operations against cache.",
@@ -72,6 +76,7 @@ func newClientMetrics(regs []prometheus.Registerer) *clientMetrics {
 	duration.WithLabelValues(opSet)
 	duration.WithLabelValues(opDelete)
 
+	//lint:ignore faillint need to apply the metric to multiple registerer
 	dataSize := prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Name: "operation_data_size_bytes",
 		Help: "Tracks the size of the data stored in and fetched from cache.",
--- a/cache/client.go
+++ b/cache/client.go
@@ -11,7 +11,6 @@ import (
 	"github.com/grafana/gomemcache/memcache"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


cache/client.go
```diff
@@ -35,47 +34,50 @@ type clientMetrics struct {
 	dataSize   *prometheus.HistogramVec
 }
 
-func newClientMetrics(reg prometheus.Registerer) *clientMetrics {
-	cm := &clientMetrics{}
-
-	cm.operations = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+func newClientMetrics(regs []prometheus.Registerer) *clientMetrics {
+	//lint:ignore faillint need to apply the metric to multiple registerer
```
beorn7, line 38: I think it is OK to ignore the lint rule in this case. `promauto` is meant for "simple" cases. And registering the same metric with different registerers is a prime example for those rare "involved" cases for which `promauto` isn't enough.

cache/client.go
```diff
@@ -35,47 +34,50 @@ type clientMetrics struct {
 	dataSize   *prometheus.HistogramVec
 }
 
-func newClientMetrics(reg prometheus.Registerer) *clientMetrics {
-	cm := &clientMetrics{}
-
-	cm.operations = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+func newClientMetrics(regs []prometheus.Registerer) *clientMetrics {
```
beorn7, line 37: Idea to avoid changing the argument to a slice of Registerers here:

Would it work to implement a "tee registerer" in the following way and use it to bundle up all the registerer before passing them to `newClientMetrics`? This would probably also avoid the lint warning below.

```Go
type TeeRegisterer struct {
	Regs []prometheus.Registerer
}

func (t TeeRegisterer) Register(c prometheus.Collector) error {
	for _, reg := range t.Regs {
		if err := reg.Register(c); err != nil {
			return err
		}
	}
	return nil
}

func (t TeeRegisterer) MustRegister(cs ...prometheus.Collector) {
	for _, reg := range t.Regs {
		reg.MustRegister(cs...)
	}
}

func (t TeeRegisterer) Unregister(c prometheus.Collector) bool {
	result := false
	for _, reg := range t.Regs {
		if reg.Unregister(c) {
			result = true
		}
	}
	return result
}
```

Logiraptor: I agree with @beorn7 's idea. I think that should result in a much smaller diff as well. Can you give that a try @lamida?

cache/client.go
beorn7: Now you can use `promauto` again, can't you?

promregistry/registry.go
```diff
@@ -0,0 +1,33 @@
+package promregistry
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// TeeRegisterer supports MultipleRegisterer.
+type TeeRegisterer struct {
+	Regs []prometheus.Registerer
+}
```
beorn7, line 8: I think we can simplify this even more by defining 
```Go
type TeeRegisterer []prometheus.Registerer
```

And then adjust the methods below a tiny bit.

cache/client.go
lamida: The PR is updated with that great suggestion. Now there is no more need to update the constructor.

beorn7:
@lamida have you noticed my two latest comments? (The arrived at about the same time that you requested a review.)

cache/client.go
lamida: I should have realized this. Updated, now everything use promauto again.

promregistry/registry.go
lamida: Yes, I should used type alias in the first place. Thanks.

lamida:
@beorn7 sorry, I missed your last two comments earlier. I updated everything as advised. Now the diffs are even much smaller. Thanks for the feedback. Please help to do one final review. Thank you very much.

cache/remote_cache.go
```diff
@@ -44,7 +63,13 @@ func NewRedisCache(name string, logger log.Logger, redisClient RemoteCacheClient
 			name,
 			logger,
 			redisClient,
-			prometheus.WrapRegistererWithPrefix("cache_redis_", reg),
+			promregistry.TeeRegisterer(
```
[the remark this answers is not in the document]
lamida: That's really true. I couldn't recall why this use TeeRegisterer in the first place. 🤔

I will fix it once I return from PTO.

cache/remote_cache.go
```diff
@@ -27,7 +39,14 @@ func NewMemcachedCache(name string, logger log.Logger, memcachedClient RemoteCac
...
 			memcachedClient,
-			prometheus.WrapRegistererWithPrefix("cache_memcached_", reg),
+			promregistry.TeeRegisterer(
+				[]prometheus.Registerer{
+					prometheus.WrapRegistererWithPrefix(cachePrefix+legacyMemcachedPrefix, reg),
+					prometheus.WrapRegistererWith(
+						prometheus.Labels{labelBackend: backendMemcached},
+						prometheus.WrapRegistererWithPrefix(cachePrefix, reg)),
+				},
+			),
```
beorn7, line 49: Same comment as above.

promregistry/registry.go
```diff
@@ -0,0 +1,31 @@
+package promregistry
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// TeeRegisterer supports MultipleRegisterer.
```
beorn7, line 5: Nit: `MultipleRegisterer` looks like there is a type with that name. I would just write plain English: "TeeRegesterer supports multiple Registerers."

Or explain the functionality even a bit more: "TeeRegisterer is a slice of Registerers. It implements Registerer itself and forwards all registrations to all Registerers in the slice."

beorn7 (commented):
This looks good it terms of usage of prometheus/client_golang. I leave the rest to @Logiraptor .

I have added a few comments about a more compact syntax when creating TeeRegisterers.

lamida:
@beorn7 thank you for the reviews. I have addressed the remaining comments. 

@Logiraptor I will merge this first to #255 and do a bit more testings. Please help to do one more review on #255 before that merged to master.

PR #261, opened 2023-01-30
Use ChainInterceptor from upstream grpc

bboreham:
**What this PR does**:

Use ChainInterceptor from upstream grpc; also update reference to weaveworks/common so we eliminate the dependency on `grpc-ecosystem/go-grpc-middleware`. ChainInterceptor functionality is identical; the feature was added upstream in v1.21 for client side.

**Checklist**
- NA Tests updated
- NA `CHANGELOG.md` updated - not user-visible.


pstibrany (approved):

pracucci:
Thanks Bryan!

PR #262, opened 2023-01-31
runtimeconfig: Fix `runtime_config_last_reload_successful` metric after recovery.

kavirajk:
Currently, we have some optimization of not actually reloading the config if the new config matches the `hash` with old config. 
But that's not properly handled for metric `runtime_config_last_reload_successful` exposing subtle bug during following scenario.

1. Correct config - hash x, sucess_metric 1
2. Bad config - hash x (NOTE: we just keep last successful config hash), success_metric 0
3. Correct config (exactly as 1) - hash x, success_metric 0

Problem with (3) is, since hash never changed, success_metric is never set back to 1.

This causes problem say if alert is triggered for bad config (during (2)), And even after fixing the bad config, if config hash turns out to be exactly same as before, then alert will never resolve.

Signed-off-by: Kaviraj <kavirajkanagaraj@gmail.com>

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [BUGFIX] runtimeconfig: Fix `runtime_config_last_reload_successful` metric after recovery from bad config with exact same config hash as before. #262
 * [CHANGE] Change `WaitRingStability` and `WaitInstanceState` methods signature to rely on `ReadRing` instead. #251
 * [CHANGE] Added new `-consul.cas-retry-delay` flag. It has a default value of `1s`, while previously there was no delay between retries. #178
 * [CHANGE] Flagext: `DayValue` now always uses UTC when parsing or displaying dates. #71
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## Changelog
 
-* [BUGFIX] runtimeconfig: Fix `runtime_config_last_reload_successful` metric after recovery from bad config with exact same config hash as before. #262
 * [CHANGE] Change `WaitRingStability` and `WaitInstanceState` methods signature to rely on `ReadRing` instead. #251
 * [CHANGE] Added new `-consul.cas-retry-delay` flag. It has a default value of `1s`, while previously there was no delay between retries. #178
 * [CHANGE] Flagext: `DayValue` now always uses UTC when parsing or displaying dates. #71
@@ -91,3 +90,4 @@
 * [BUGFIX] Memberlist: retry joining memberlist cluster on startup when no nodes are resolved. #215
 * [BUGFIX] Ring status page: display 100% ownership as "100%", rather than "1e+02%". #231
 * [BUGFIX] Memberlist: fix crash when methods from `memberlist.Delegate` interface are called on `*memberlist.KV` before the service is fully started. #244
+* [BUGFIX] runtimeconfig: Fix `runtime_config_last_reload_successful` metric after recovery from bad config with exact same config hash as before. #262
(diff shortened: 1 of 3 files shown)


runtimeconfig/manager.go
```diff
@@ -175,6 +175,11 @@ func (om *Manager) loadConfig() error {
 
 	if sameHashes {
 		// No need to rebuild runtime config.
+
+		// NOTE: Need to set load success metric to 1 because there can be the case where
+		// success metric was set to 0 during bad config and may never set back to 1 if bad config
+		// was fixed with good config with same hash as before.
```
bboreham, line 181: I don't think this length of comment is necessary.  Put it in the git commit description.

runtimeconfig/manager_test.go
```diff
@@ -438,6 +438,83 @@ func TestManager_ShouldFastFailOnInvalidConfigAtStartup(t *testing.T) {
...
+					# HELP runtime_config_last_reload_successful Whether the last runtime-config reload attempt was successful.
+					# TYPE runtime_config_last_reload_successful gauge
+					runtime_config_last_reload_successful 0
+				`, fmt.Sprintf("%x", validConfigSha256)))))
+
+	// Revert config to good state. Make sure it has same hash as before.
+	err = os.WriteFile(tempFile.Name(), validConfig, 0600)
+	require.NoError(t, err)
+
+	time.Sleep(1 * time.Second)
```
bboreham, line 505: Do we really need to sleep this long?  Unit tests should complete in total in a second or so; taking at least two for one test is relatively expensive.

runtimeconfig/manager_test.go
```diff
@@ -438,6 +438,83 @@ func TestManager_ShouldFastFailOnInvalidConfigAtStartup(t *testing.T) {
...
+
+	assert.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(fmt.Sprintf(`
+					# HELP runtime_config_hash Hash of the currently active runtime configuration, merged from all configured files.
+					# TYPE runtime_config_hash gauge
+					runtime_config_hash{sha256="%s"} 1
+					# HELP runtime_config_last_reload_successful Whether the last runtime-config reload attempt was successful.
+					# TYPE runtime_config_last_reload_successful gauge
+					runtime_config_last_reload_successful 0
+				`, fmt.Sprintf("%x", validConfigSha256)))))
```
bboreham, line 499: Suggest factoring out the repetition here compared to similar blocks above and below; it will make maintenance easier.

kavirajk:
@bboreham thanks for the review. Addressed your comments. ready for another look!

runtimeconfig/manager_test.go
```diff
@@ -438,6 +438,72 @@ func TestManager_ShouldFastFailOnInvalidConfigAtStartup(t *testing.T) {
...
+
+	assertHashAndSuccessMetric := func(config []byte, successCount int) {
+		assert.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(fmt.Sprintf(`
+					# HELP runtime_config_hash Hash of the currently active runtime configuration, merged from all configured files.
+					# TYPE runtime_config_hash gauge
+					runtime_config_hash{sha256="%s"} 1
+					# HELP runtime_config_last_reload_successful Whether the last runtime-config reload attempt was successful.
+					# TYPE runtime_config_last_reload_successful gauge
+					runtime_config_last_reload_successful %d
+				`, fmt.Sprintf("%x", sha256.Sum256(config)), successCount))))
```
pstibrany, line 482: `successCount` -> `lastSuccessful` (it's not a counter, as "count" suggests)

pstibrany (approved):
thank you!

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [BUGFIX] runtimeconfig: Fix `runtime_config_last_reload_successful` metric after recovery from bad config with exact same config hash as before. #262
```
pstibrany, line 3: Bugfixes go at the end of changelog.

PR #264, opened 2023-02-01
Reduce level of log on nil check in consul WatchKey()

joe-elliott:
**What this PR does**:
Consul is the only client that will log anything if the watched for key does not exist. This is causing some [log spam](https://github.com/grafana/tempo/issues/2048) in Tempo under certain configurations. Fixing it on Tempo's side cleanly is surprisingly difficult due to the way config is loaded independently by various services.

Given that this is unlogged by other clients I'm hoping this change is acceptable. I'd also be willing to remove the line entirely.

### DIFF
--- a/kv/consul/client.go
+++ b/kv/consul/client.go
@@ -252,7 +252,6 @@ func (c *Client) WatchKey(ctx context.Context, key string, f func(interface{}) b
 		}
 
 		if kvp == nil {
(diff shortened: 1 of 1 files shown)


56quarters (approved):
LGTM. I'm fine with a change to `debug`, this doesn't seem particularly useful or informative.  If others are OK with removing it entirely, that's good too.

pstibrany (approved):

PR #265, opened 2023-02-07
flagext: Add new type `URLEscaped`

kavirajk:
**What this PR does**:
Sometimes we need to store just escaped URL in the config like S3 URL that contains credentials which can include characters like `:` and `/`. Currently we cannot use normal `flagext.URLValue` as it just uses unescaped URL.

Please check issue
https://github.com/grafana/loki/issues/1607

Fixes https://github.com/grafana/loki/issues/1607

**Checklist**
- [ ] Tests updated

### DIFF
--- a/flagext/urlescaped.go
+++ b/flagext/urlescaped.go
@@ -1,0 +1,63 @@
+package flagext
+
+import "net/url"
+
+// URLEscaped is a url.URL that can be used as a flag.
+// URL value it contains will always be URL escaped and safe.
+type URLEscaped struct {
+	*url.URL
+}
+
+// String implements flag.Value
+func (v URLEscaped) String() string {
+	if v.URL == nil {
+		return ""
+	}
+	return v.URL.String()
+}
+
+// Set implements flag.Value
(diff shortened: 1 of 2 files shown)


pstibrany:
Is this work around for properly formatting strings in YAML file?

kavirajk:
> Is this work around for properly formatting strings in YAML file?

Kinda. So basically user should be able to give unescaped S3 bucket URL in both  yaml config and as a CLI flag.

Our internal dependency `aws.ConfigFromURL` [expects escaped URL always](https://github.com/grafana/loki/blob/main/vendor/github.com/weaveworks/common/aws/config.go#L16-L21) (from `weaveworks/common`). 

Trying to transparenly handle this case. But realized it's hard to do it in backward compatible way. e.g: If people already working with escaped URL. Tried few things but looks bit tricky. Converting to draft PR till I find right way to do it.

pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

PR #266, opened 2023-02-11
Lifecycler: return the number of instances in lifecycler's zone

duricanikolic:
**What this PR does**:
This PR enriches `Lifecycler` with the `InstancesInZoneCount()` method, that returns the number of instances in ring belonging to the lifecycler's zone, updated during the last heartbeat period. 
A test (`TestLifecycler_InstancesInZoneCount`) showing the expected behaviour of `InstancesInZoneCount()` has been added.

This PR is a pre-requisite for fixing the issue [4208](https://github.com/grafana/mimir/issues/4208).

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,7 +100,7 @@
   * `<prefix>_cache_operation_skipped_total{backend="[memcached|redis]",...}`
   * `<prefix>_cache_operations_total{backend="[memcached|redis]",...}`
   * `<prefix>_cache_requests_total{backend="[memcached|redis]",...}`
-* [ENHANCEMENT] Lifecycler: Added `InstancesInZoneCount` method returning the number of instances in the ring that are registered in lifecycler's zone, updated during the last heartbeat period. #266
+* [ENHANCEMENT] Lifecycler: Added `HealthyInstancesInZoneCount` method returning the number of healthy instances in the ring that are registered in lifecycler's zone, updated during the last heartbeat period. #266
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
 * [BUGFIX] Ring: `ring_member_ownership_percent` and `ring_tokens_owned` metrics are not updated on scale down. #109
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@
   * `<prefix>_cache_operation_skipped_total{backend="[memcached|redis]",...}`
   * `<prefix>_cache_operations_total{backend="[memcached|redis]",...}`
   * `<prefix>_cache_requests_total{backend="[memcached|redis]",...}`
+* [ENHANCEMENT] Lifecycler: Added `InstancesInZoneCount` method returning the number of instances in the ring that are registered in lifecycler's zone, updated during the last heartbeat period. #266
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
 * [BUGFIX] Ring: `ring_member_ownership_percent` and `ring_tokens_owned` metrics are not updated on scale down. #109
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -130,6 +130,7 @@ type Lifecycler struct {
 	// Keeps stats updated at every heartbeat period
 	countersLock          sync.RWMutex
 	healthyInstancesCount int
+	instancesInZoneCount  int
 	zonesCount            int
 
 	lifecyclerMetrics *LifecyclerMetrics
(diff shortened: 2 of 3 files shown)


ring/lifecycler.go
```diff
@@ -795,13 +805,13 @@ func (i *Lifecycler) changeState(ctx context.Context, state InstanceState) error
...
 	healthyInstancesCount := 0
-	zones := map[string]struct{}{}
+	zones := map[string]int{}
 
 	if ringDesc != nil {
 		now := time.Now()
 
 		for _, ingester := range ringDesc.Ingesters {
-			zones[ingester.Zone] = struct{}{}
+			zones[ingester.Zone]++
```
pracucci, line 814: I think for your purpose you should also count the healthy ones, like we do for `healthyInstancesCount` and then I would rename `InstancesInZoneCount()` to `HealthyInstancesInZoneCount()`. Pay attention that `zonesCount` count logic shouldn't change.

duricanikolic: Hi @pracucci, 
Actually I was not sure whether we were supposed to count only healthy or all instances in each zone. Thank you for this remark. I have fixed this, and I have also added a test that tests both `HealthyInstancesInZoneCount()` and `ZonesCount()` according to whether the added instance is healthy or not.

CHANGELOG.md
```diff
@@ -100,6 +100,7 @@
   * `<prefix>_cache_operation_skipped_total{backend="[memcached|redis]",...}`
   * `<prefix>_cache_operations_total{backend="[memcached|redis]",...}`
   * `<prefix>_cache_requests_total{backend="[memcached|redis]",...}`
+* [ENHANCEMENT] Lifecycler: Added `InstancesInZoneCount` method returning the number of instances in the ring that are registered in lifecycler's zone, updated during the last heartbeat period. #266
```
pracucci, line 103:
```suggestion
* [ENHANCEMENT] Lifecycler: Added `HealthyInstancesInZoneCount ` method returning the number of healthy instances in the ring that are registered in lifecycler's zone, updated during the last heartbeat period. #266
```

ring/lifecycler_test.go
```diff
@@ -98,6 +98,113 @@ func TestLifecycler_HealthyInstancesCount(t *testing.T) {
...
+
+	var ringConfig Config
+	flagext.DefaultValues(&ringConfig)
+	ringConfig.KVStore.Mock = ringStore
+
+	instances := []struct {
+		zone                                string
+		healthy                             bool
+		expectedHealthyInstancesInZoneCount int
+		expectedHealthyInstancesCount       int
```
pracucci, line 113: It's the number of expected zones, not healthy instances count.

```suggestion
		expectedZonesCount       int
```

ring/lifecycler_test.go
```diff
@@ -98,6 +98,113 @@ func TestLifecycler_HealthyInstancesCount(t *testing.T) {
...
+	expectedHealthInstancesCounter := 0
+	for idx, instance := range instances {
+		ctx := context.Background()
+
+		// Register an instance to the ring.
+		cfg := testLifecyclerConfig(ringConfig, fmt.Sprintf("instance-%d", idx))
+		cfg.HeartbeatPeriod = 100 * time.Millisecond
+		joinTimeoutMs := 1000
+		// unhealthy instances join after 2000ms, which exceeds the 1000ms timeout
+		joinAfterMs := 2000
```
pracucci, line 182: If I understsand this test corretly, I think the reason why this test is not failing is just beacuse it takes less than 2s to run. These "unhealthy" instances will actually join the ring at some point, after 2s. So to make it more resilient I would set this to 1 minute instead of 2 seconds, so that we're sure to not hit it.

I would also update the message above to clarify that we're doing here.

ring/lifecycler_test.go
```diff
@@ -98,6 +98,113 @@ func TestLifecycler_HealthyInstancesCount(t *testing.T) {
...
+	}
+
+	expectedHealthInstancesCounter := 0
+	for idx, instance := range instances {
+		ctx := context.Background()
+
+		// Register an instance to the ring.
+		cfg := testLifecyclerConfig(ringConfig, fmt.Sprintf("instance-%d", idx))
+		cfg.HeartbeatPeriod = 100 * time.Millisecond
+		joinTimeoutMs := 1000
```
pracucci, line 180: [nit] It's not really a timeout, in the sense the instance will actually join after `joinAfterMs` but were not waiting for that in the test because this value is lower than 2 seconds.

```suggestion
		joinWaitMs := 1000
```

pracucci (approved):

ring/lifecycler.go
```diff
@@ -795,24 +805,28 @@ func (i *Lifecycler) changeState(ctx context.Context, state InstanceState) error
...
+	zones := map[string]int{}
 
 	if ringDesc != nil {
 		now := time.Now()
 
 		for _, ingester := range ringDesc.Ingesters {
-			zones[ingester.Zone] = struct{}{}
+			if _, ok := zones[ingester.Zone]; !ok {
+				zones[ingester.Zone] = 0
+			}
```
pstibrany, line 816: Nit: Setting default value `0` to `zones[ingester.Zone]` is unnecessary. `zones[ingester.Zone]++` below will work just fine without this.

duricanikolic: Hi @pstibrany, 
This assignment is needed in order to count the number of zones. If we remove the block setting `0` to `zones[ingester.Zone]`, we will count only the number of zones with healthy ingesters.

pstibrany: Good point that I missed! That's worth adding a comment in my opinion.

PR #268, opened 2023-02-17
cache: Enable Redis support and simplify Memcached config

56quarters:
**What this PR does**:

Include Redis configuration as part of the `cache.BackendConfig` struct and support creating Redis clients based on it.

This change also removes the `cache.MemcachedConfig` struct in favor of `cache.MemcachedClientConfig` which offers identical functionality. This module is only used by Mimir and requires minimal changes.

See grafana/mimir#4236

**Checklist**
- [X] Tests updated


pracucci (approved):
LGTM! I'm very glad you got rid of `MemcachedConfig`. I just realized that we didn't get rid of it when we imported the memcached client into dskit.

PR #269, opened 2023-02-18
Add MinIdleConnectionsHeadroom support to memcached client config

pracucci:
**What this PR does**:
In this PR I'm upgrading gomemcache library and add `MinIdleConnectionsHeadroom` support to configuration. 

See https://github.com/grafana/gomemcache/pull/10 for more details.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/cache/memcache_config.go
+++ b/cache/memcache_config.go
@@ -14,19 +14,21 @@ var (
 )
 
 type MemcachedConfig struct {
-	Addresses              string        `yaml:"addresses"`
-	Timeout                time.Duration `yaml:"timeout"`
-	MaxIdleConnections     int           `yaml:"max_idle_connections" category:"advanced"`
-	MaxAsyncConcurrency    int           `yaml:"max_async_concurrency" category:"advanced"`
-	MaxAsyncBufferSize     int           `yaml:"max_async_buffer_size" category:"advanced"`
-	MaxGetMultiConcurrency int           `yaml:"max_get_multi_concurrency" category:"advanced"`
-	MaxGetMultiBatchSize   int           `yaml:"max_get_multi_batch_size" category:"advanced"`
-	MaxItemSize            int           `yaml:"max_item_size" category:"advanced"`
+	Addresses                            string        `yaml:"addresses"`
+	Timeout                              time.Duration `yaml:"timeout"`
+	MinIdleConnectionsHeadroomPercentage float64       `yaml:"min_idle_connections_headroom_percentage" category:"advanced"`
+	MaxIdleConnections                   int           `yaml:"max_idle_connections" category:"advanced"`
+	MaxAsyncConcurrency                  int           `yaml:"max_async_concurrency" category:"advanced"`
+	MaxAsyncBufferSize                   int           `yaml:"max_async_buffer_size" category:"advanced"`
+	MaxGetMultiConcurrency               int           `yaml:"max_get_multi_concurrency" category:"advanced"`
+	MaxGetMultiBatchSize                 int           `yaml:"max_get_multi_batch_size" category:"advanced"`
+	MaxItemSize                          int           `yaml:"max_item_size" category:"advanced"`
 }
 
 func (cfg *MemcachedConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.StringVar(&cfg.Addresses, prefix+"addresses", "", "Comma-separated list of memcached addresses. Each address can be an IP address, hostname, or an entry specified in the DNS Service Discovery format.")
 	f.DurationVar(&cfg.Timeout, prefix+"timeout", 200*time.Millisecond, "The socket read/write timeout.")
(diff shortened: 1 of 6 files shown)


cache/memcache_config.go
```diff
@@ -10,23 +10,26 @@ import (
...
+	MaxAsyncBufferSize         int           `yaml:"max_async_buffer_size" category:"advanced"`
+	MaxGetMultiConcurrency     int           `yaml:"max_get_multi_concurrency" category:"advanced"`
+	MaxGetMultiBatchSize       int           `yaml:"max_get_multi_batch_size" category:"advanced"`
+	MaxItemSize                int           `yaml:"max_item_size" category:"advanced"`
 }
 
 func (cfg *MemcachedConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.StringVar(&cfg.Addresses, prefix+"addresses", "", "Comma-separated list of memcached addresses. Each address can be an IP address, hostname, or an entry specified in the DNS Service Discovery format.")
 	f.DurationVar(&cfg.Timeout, prefix+"timeout", 200*time.Millisecond, "The socket read/write timeout.")
+	f.Float64Var(&cfg.MinIdleConnectionsHeadroom, prefix+"min-idle-connections-headroom", 0, "The minimum number of idle connections to keep open as a percentage of the number of recently used idle connections. The percentage should be in the range [0, 1]. If zero, idle connections are kept open indefinitely.")
```
pstibrany, line 32: "Percentage" should be number between 0 and 100. You mean "fraction" here.

pstibrany (approved):

cache/memcache_config.go
pracucci: How do you suggest to rephrase it? Just trying to understand how the final sentence will look like.

pstibrany: We have two options: keep the text-as is (although... see below), and accept value between 0 and 100.

Or keep accepting value between 0 and 1, and then I would rephrase it as:

> The fraction of recently used idle connections that should be kept open. The value should be in the range [0, 1]. If zero, all idle connections are kept open indefinitely.

(Why does value of 0 act as 1?)

pracucci: > (Why does value of 0 act as 1?)

To not introduce any backward compatibility.

pstibrany: > > (Why does value of 0 act as 1?)
> 
> To not introduce any breaking change.

Why not make the default value 1 then?

colega: This conversation made me think twice about https://github.com/grafana/gomemcache/pull/10 and I think that:
0 should be valid: we may not want to keep any connections on top of the recently valid ones.
>1 should be valid: we may want to keep double idle conns compared to recently used ones.
Disabled should be a negative value, with a default negative value as @pstibrany pointed out.

pracucci: Ok. I've updated https://github.com/grafana/gomemcache/pull/10 to both specify the headroom percentage as 0-100+ percentage and support 0 value too. Negative value means disabled.

cache/memcache_config.go
```diff
@@ -14,19 +14,21 @@ var (
...
+	MaxAsyncBufferSize                   int           `yaml:"max_async_buffer_size" category:"advanced"`
+	MaxGetMultiConcurrency               int           `yaml:"max_get_multi_concurrency" category:"advanced"`
+	MaxGetMultiBatchSize                 int           `yaml:"max_get_multi_batch_size" category:"advanced"`
+	MaxItemSize                          int           `yaml:"max_item_size" category:"advanced"`
 }
 
 func (cfg *MemcachedConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.StringVar(&cfg.Addresses, prefix+"addresses", "", "Comma-separated list of memcached addresses. Each address can be an IP address, hostname, or an entry specified in the DNS Service Discovery format.")
 	f.DurationVar(&cfg.Timeout, prefix+"timeout", 200*time.Millisecond, "The socket read/write timeout.")
+	f.Float64Var(&cfg.MinIdleConnectionsHeadroomPercentage, prefix+"min-idle-connections-headroom-percentage", -1, "The minimum number of idle connections to keep open as a percentage of the number of recently used idle connections. If negative, idle connections are kept open indefinitely.")
```
colega, line 31: Percentage is 0-100, but as a user of this library, I wouldn't assume that and I'd still have to verify. We can save that dive into the code to the user by being explicit with something like this:

```suggestion
	f.Float64Var(&cfg.MinIdleConnectionsHeadroomPercentage, prefix+"min-idle-connections-headroom-percentage", -1, "The minimum number of idle connections to keep open as a percentage (0-100) of the number of recently used idle connections. If negative, idle connections are kept open indefinitely.")
```

colega (approved):
LGTM, please consider being explicit on the fact that this percentage is 0-100 valued (not 0-1).

cache/memcache_config.go
pracucci: Good point. Accepted.

PR #273, opened 2023-03-02
Remove context argument from cache.Cache.Store method

56quarters:
**What this PR does**:

The Redis and Memcached implementations of the `Cache` interface use a job queue to perform writes in the background after the `Store` method has already returned. Including a context argument creates the possibility of the writes being cancelled after the originating request finishes while the writes are still being run in the background. Removing the context makes it harder to misuse the cache like this.

This change also renames the `Store` method to `StoreAsync` to make this behavior more obvious.

See grafana/mimir#4341
See grafana/mimir#4342

**Checklist**
- [X] Tests updated

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -20,7 +20,7 @@ type RemoteCacheClient interface {
 	// SetAsync enqueues an asynchronous operation to store a key into memcached.
 	// Returns an error in case it fails to enqueue the operation. In case the
 	// underlying async operation will fail, the error will be tracked/logged.
-	SetAsync(ctx context.Context, key string, value []byte, ttl time.Duration) error
+	SetAsync(key string, value []byte, ttl time.Duration) error
 
 	// Delete deletes a key from memcached.
 	// This is a synchronous operation. If an asynchronous set operation for key is still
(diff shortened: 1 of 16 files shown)


pracucci (approved):
As discussed internally, LGTM. Thanks!

pstibrany:
What about `Delete`? Is that async operation too?

56quarters:
> What about `Delete`? Is that async operation too?

No, it's blocking. The underlying Redis/Memcached implementations use the job queue but block until the task completes.

cache/cache.go
```diff
@@ -20,7 +20,7 @@ type RemoteCacheClient interface {
 	// SetAsync enqueues an asynchronous operation to store a key into memcached.
 	// Returns an error in case it fails to enqueue the operation. In case the
 	// underlying async operation will fail, the error will be tracked/logged.
-	SetAsync(ctx context.Context, key string, value []byte, ttl time.Duration) error
+	SetAsync(key string, value []byte, ttl time.Duration) error
```
charleskorn, line 23: Out of the scope of this PR, but something that might be good to look at while we're making breaking changes: is there a reason why the methods on `Cache` and `RemoteCacheClient` are named differently? eg. `Fetch` / `GetMulti` and `StoreAsync` / `SetAsync`

56quarters: Yes! Longer term I want to remove `RemoteCacheClient` entirely. It's an abstraction around Redis/Memcached which is then used by our even more general `cache.Cache` abstraction.

PR #274, opened 2023-03-02
Use Reader interface to fetch TLS certs and keys

fayzal-g:
**What this PR does**:
Uses a Reader interface to fetch secrets (certs/keys) used to configure TLS. Now allows reading not just from a file, but for anything that implements `ReadSecret`

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/crypto/tls/tls.go
+++ b/crypto/tls/tls.go
@@ -54,9 +54,9 @@ func (cfg *ClientConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet)
 	// Trim any trailing "." since we include our own here
 	prefix = strings.TrimRight(prefix, ".")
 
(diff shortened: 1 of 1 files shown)


crypto/tls/tls.go
```diff
@@ -79,10 +90,15 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
 		ServerName:         cfg.ServerName,
 	}
 
-	// read ca certificates
+	// If Reader interface not provided, default to reading from File
+	if cfg.Reader == nil {
+		cfg.Reader = &fileReader{}
```
pracucci, line 95: I suggest to not overwrite `cfg.Reader`. You can just store the reader in a local variable.

PR #279, opened 2023-03-16
Add WaitN to the multi-tenant RateLimiter

npazosmendez:
**What this PR does**:

Adds `WaitN` to the multitenant RateLimiter's interface. This is provided by the underlying `rate.Limiter`, so this is as simple as exposing it. See https://pkg.go.dev/golang.org/x/time/rate#Limiter.WaitN . 

Copying the docs from there:

> WaitN blocks until lim permits n events to happen. It returns an error if n exceeds the Limiter's burst size, the Context is canceled, or the expected wait time exceeds the Context's Deadline. The burst limit is ignored if the rate limit is Inf.

So this is useful when we don't want to directly error out and discard the task, but wait to adjust the pace.

n/a

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/limiter/rate_limiter_test.go
+++ b/limiter/rate_limiter_test.go
@@ -1,6 +1,7 @@
 package limiter
 
 import (
+	"context"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


limiter/rate_limiter_test.go
```diff
@@ -65,6 +66,48 @@ func TestRateLimiter_AllowN(t *testing.T) {
 	assert.Equal(t, true, limiter.AllowN(now.Add(time.Second), "tenant-2", 2))
 }
 
+func TestRateLimiter_WaitN(t *testing.T) {
```
npazosmendez, line 69: Note for reviewer: this test takes about 0.7s on my machine

pstibrany (approved):

pstibrany:
Can you please add a changelog entry before merging?

PR #282, opened 2023-04-13
Ring status page: fix the owned tokens percentage value displayed

pracucci:
**What this PR does**:
I'm unbalanced series between Mimir ingesters. While investigating, I noticed the owned tokens percentage displayed in the ring status page is wrong. This PR fixed it and also expose `ring.Desc.CountTokens()` so that I can re-use it in a ring analysis tool I'm building (instead of copying the code).

What's the bug in `countTokens()` and why this PR fixes it? The tokens in the ring are the end of the range, while `countTokens()` was computing the owned tokens range as if the token is the start of the range.

N/A

**Checklist**
- [ ] Tests updated


treid314 (approved):
Great Catch!

duricanikolic (approved):
LGTM

PR #286, opened 2023-04-27
Memcached: preallocate slice used for grouping cache keys by server

56quarters:
**What this PR does**:

We know the number of keys and so should pre-allocate the slice used. Runtime of both the new and old version varies by a lot however, the new version tends to be faster (in addition to consistently allocating less memory).

```
$ benchstat -delta-test=none old.txt new.txt 
name                                 old time/op    new time/op    delta
MemcachedClient_sortKeysByServer-16    2.53ms ± 0%    2.11ms ± 0%  -16.60%

name                                 old alloc/op   new alloc/op   delta
MemcachedClient_sortKeysByServer-16    1.31MB ± 0%    0.89MB ± 0%  -32.23%

name                                 old allocs/op  new allocs/op  delta
MemcachedClient_sortKeysByServer-16     20.1k ± 0%     20.1k ± 0%   -0.02%
```

N/A

**Checklist**
- [X] Tests updated

### DIFF
--- a/cache/memcached_client.go
+++ b/cache/memcached_client.go
@@ -488,7 +488,7 @@ func (c *memcachedClient) sortKeysByServer(keys []string) []string {
 		bucketed[addrString] = append(bucketed[addrString], key)
 	}
 
-	var out []string
+	out := make([]string, 0, len(keys))
 	for srv := range bucketed {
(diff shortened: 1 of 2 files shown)


pracucci (approved):
Nice work!

PR #295, opened 2023-05-11
Fix race condition in Ring.ShuffleShardWithLookback()

pracucci:
**What this PR does**:
This PR fix race condition in `Ring.getCachedShuffledSubringWithLookback()`, specifically the access to `r.shuffledSubringWithLookbackCache` which was not protected by `r.mtx`.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,7 +109,7 @@
 * [ENHANCEMENT] Memcached: Add support for using TLS or mTLS with Memcached based caching. #278
 * [ENHANCEMENT] Ring: improve performance of shuffle sharding computation. #281
 * [ENHANCEMENT] Add option to enable IPv6 address detection in ring and memberlist handling. #185
-* [ENHANCEMENT] Ring: cache results of shuffle sharding with lookback where possible. #283 
+* [ENHANCEMENT] Ring: cache results of shuffle sharding with lookback where possible. #283 295
 * [ENHANCEMENT] BasicLifecycler: functions `ShouldKeepInstanceInTheRingOnShutdown` and `SetKeepInstanceInTheRingOnShutdown` for checking and setting whether instances should be kept in the ring or unregistered on shutdown have been added. #290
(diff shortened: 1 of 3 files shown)


ring/ring.go
```diff
@@ -184,6 +184,7 @@ type Ring struct {
 
 	// Cache of shuffle-sharded subrings per identifier. Invalidated when topology changes.
 	// If set to nil, no caching is done (used by tests, and subrings).
+	// Both maps are protected by mtx.
 	shuffledSubringCache             map[subringCacheKey]*Ring
 	shuffledSubringWithLookbackCache map[subringCacheKey]cachedSubringWithLookback
```
colega, line 189: I'd move this [under the mutex-hat](https://dmitri.shuralyov.com/idiomatic-go/entries/2)

CHANGELOG.md
```diff
@@ -109,7 +109,7 @@
 * [ENHANCEMENT] Memcached: Add support for using TLS or mTLS with Memcached based caching. #278
 * [ENHANCEMENT] Ring: improve performance of shuffle sharding computation. #281
 * [ENHANCEMENT] Add option to enable IPv6 address detection in ring and memberlist handling. #185
-* [ENHANCEMENT] Ring: cache results of shuffle sharding with lookback where possible. #283 
+* [ENHANCEMENT] Ring: cache results of shuffle sharding with lookback where possible. #283 295
```
colega, line 112:
```suggestion
* [ENHANCEMENT] Ring: cache results of shuffle sharding with lookback where possible. #283 #295
```

ring/ring.go
pracucci: Well, it's already under the mutex hat. Everything above these lines is also protected by mtx. Going to remove the comment I just added.

colega: Oh okay, then yes, I think removing it makes it less ambiguous (otherwise it feels like the stuff above isn't protected anymore)

PR #296, opened 2023-05-12
Add config for connect params for the gRPC client

chaudum:
**What this PR does**:

This PR adds configuration options for the gRPC client, namely the connect connect timeout, and backoff delays.

In Loki we observed that the default gRPC connect timeout caused tail latency when the client is used in a client pool for the ring, e.g. index-gateway clients. In case a server is unexpectedly restarted (e.g. due to OOMing) the server still remains in the ring even though it is may be unavailable or not ready (during startup). However, the client pool still create the client for the address of the server and tries to connect. A long timeout when establishing a connection introduces the latency until the next server is tried.

https://github.com/grafana/loki/blob/main/pkg/storage/stores/indexshipper/gatewayclient/gateway_client.go#L327-L357

**Checklist**
- [ ] Tests updated

### DIFF
--- a/grpcclient/grpcclient.go
+++ b/grpcclient/grpcclient.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
+	grpcbackoff "google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/encoding/gzip"
 	"google.golang.org/grpc/keepalive"
 
@@ -28,6 +29,10 @@ type Config struct {
 
 	TLSEnabled bool             `yaml:"tls_enabled" category:"advanced"`
 	TLS        tls.ClientConfig `yaml:",inline"`
+
+	ConnectTimeout          time.Duration `yaml:"connect_timeout" category:"advanced"`
+	ConnectBackoffBaseDelay time.Duration `yaml:"connect_backoff_base_delay" category:"advanced"`
+	ConnectBackoffMaxDelay  time.Duration `yaml:"connect_backoff_max_delay" category:"advanced"`
 }
 
 // RegisterFlags registers flags.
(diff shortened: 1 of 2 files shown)


grpcclient/grpcclient.go
```diff
@@ -88,6 +98,19 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{NewRateLimiter(cfg)}, unaryClientInterceptors...)
 	}
 
+	if cfg.ConnectTimeout > 0 {
+		opts = append(
+			opts,
+			grpc.WithConnectParams(grpc.ConnectParams{
+				Backoff: grpcbackoff.Config{
```
charleskorn, line 105: Should this be built on top of `DefaultConfig`, so that we use the default values for the other properties like `Jitter` and `Multiplier`?

If not, adding a comment that we want to set these to 0 might be a good idea.

DylanGuedes: Great find. I did the following:
- Reused the `grpcbackoff.DefaultConfig`
- If BaseDelay or MaxDelay were explicitly set, use them. Otherwise, use the default.

I decided to do it this way to make this retrocompatible. WDYT?

charleskorn: Sounds good to me. Might be worth adding a test or two as well to make sure this doesn't break in the future.

DylanGuedes: I also sent on your DM, but sharing here for visibility :smile: 
I found it quite hard to testing this in a decent way because the []Opts doesn't expose a way for me to recover the final backoff params. Do you have a suggestion on how I can test this? I could write a test just to check if `config.Timeout` was the expected value but tbh that doesn't seem like a valuable test. However, If I could recover the gRPC opt that will be used by the connection the test would be pretty good.

DylanGuedes (approved):

PR #302, opened 2023-05-18
flagext: Add support for using Bytes as a CLI flag

56quarters:
**What this PR does**:

Adds support for use as a CLI flag in addition to the existing YAML serialization.

N/A

**Checklist**
- [X] Tests updated

### DIFF
--- a/flagext/bytes_test.go
+++ b/flagext/bytes_test.go
@@ -1,0 +1,127 @@
+package flagext
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
+)
+
+func Test_Bytes_String(t *testing.T) {
+	for _, tcase := range []struct {
+		value uint64
+		str   string
+	}{
+		{
+			value: 1024 * 1024 * 1024,
+			str:   "1GiB",
+		},
+		{
+			value: 1024 * 1024,
+			str:   "1MiB",
+		},
+		{
+			value: 1024,
+			str:   "1KiB",
+		},
+	} {
+		b := Bytes(tcase.value)
+		require.Equal(t, tcase.str, b.String())
+	}
+}
+
+func Test_Bytes_Set(t *testing.T) {
+	for _, tcase := range []struct {
+		value uint64
+		str   string
+	}{
+		{
+			value: 1024 * 1024 * 1024,
+			str:   "1GiB",
+		},
+		{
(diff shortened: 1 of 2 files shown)


flagext/bytes_test.go
```diff
@@ -0,0 +1,127 @@
...
+			str:   "1KiB",
+		},
+	} {
+		b := Bytes(tcase.value)
+		require.Equal(t, tcase.str, b.String())
+	}
+}
+
+func Test_Bytes_Set(t *testing.T) {
+	for _, tcase := range []struct {
```
charleskorn, line 34: [nit] Given `str` is the input value, I'd put it first, as then the test cases read "this is the input, and this is the expected output"

flagext/bytes_test.go
```diff
@@ -0,0 +1,127 @@
...
+		b := Bytes(tcase.value)
+		y, err := yaml.Marshal(&b)
+		require.NoError(t, err)
+		require.Equal(t, []byte(tcase.str), y)
+	}
+}
+
+func Test_Bytes_UnmarshalYAML(t *testing.T) {
+	for _, tcase := range []struct {
+		value uint64
```
charleskorn, line 95: [nit] Same as above - given `str` is the input value, I'd put it first, as then the test cases read "this is the input, and this is the expected output"

flagext/bytes_test.go
charleskorn, line 1: Not sure if this is really necessary given we're using a library for the conversions, but it might be good to include some test cases that aren't exactly 1 GB / MB / KB (eg. 1500 bytes or 2.3 GB)

pracucci (approved):
LGTM (modulo Charles' comments). Thanks!

flagext/bytes_test.go
56quarters: Will do :+1:

56quarters: Turns out the library will format things like that in a very precise yet ugly way: `1536 * 1024 * 1024` becomes `1GiB512MiB`

56quarters: I guess the good news is that we'll almost never go from `int` -> `string`. This would mostly be used for going the other direction, taking user input of `string` and turning it into an `int`.

charleskorn: Yeah I think that's fine, as long as it can handle a user providing something like `2.3 GiB`.

PR #303, opened 2023-05-31
Fix marshalling of flagext bytes

jhesketh:


charleskorn (commented):
Could you please also update the test case in `bytes_test.go` to match?

charleskorn (approved):

PR #306, opened 2023-06-02
ring: modify `DoUntilQuorum` to optionally only send the minimum number of requests required to reach quorum

charleskorn:
**What this PR does**:

This PR modifies `DoUntilQuorum` to optionally only send the minimum number of requests required to reach quorum. 

For example, assume zone-awareness is enabled, the replication set contains three zones and one zone failure is tolerated:

* With request minimization disabled, `DoUntilQuorum` behaves as it does currently: it calls `f` for all instances in the replication set immediately, but only returns the results from the calls required to reach quorum and cancels any outstanding calls once quorum is reached.

* With request minimization enabled, `DoUntilQuorum` would initially only call `f` for instances in two of the three zones. If those calls all succeed, then `DoUntilQuorum` returns those results and ignores the remaining zone. If any call fails, then `DoUntilQuorum` calls `f` for the instances in the remaining zone. The set of results returned only includes results from instances in successful zones, as before.

Enabling request minimization is expected to reduce the number of calls made in the happy case (all requests succeed) at the cost of additional latency in the event that one or more of the initial calls fails.

This is a breaking change to the signature for `DoUntilQuorum`. As far as I can see, `DoUntilQuorum` is only used in two places, once in Mimir ([source](https://github.com/grafana/mimir/blob/bea4bd616c42c5ea42b3cbc7e350e3ebf2a291f0/pkg/distributor/query.go#L306)) and once in Phlare ([source](https://github.com/grafana/phlare/blob/d9eb816015a8d255c8faa4c58a4f4f169a3cabc1/pkg/querier/replication.go#L29)), so the small impact seems worthwhile over introducing a second method to maintain backwards compatibility.

(none)

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/replication_set.go
+++ b/ring/replication_set.go
@@ -2,7 +2,6 @@ package ring
 
 import (
 	"context"
-	"errors"
 	"sort"
 	"time"
 )
@@ -90,8 +89,6 @@ func (r ReplicationSet) Do(ctx context.Context, delay time.Duration, f func(cont
 	return results, nil
 }
 
-var errResultNotNeeded = errors.New("result from this instance is not needed and so f has not been called")
-
 // DoUntilQuorum runs function f in parallel for all replicas in r.
 //
 // If r.MaxUnavailableZones is greater than zero:
@@ -162,20 +159,20 @@ func DoUntilQuorum[T any](ctx context.Context, r ReplicationSet, minimizeRequest
 
 	for i := range r.Instances {
 		instance := &r.Instances[i]
-		instanceCtx := contextTracker.contextFor(instance)
+		ctx := contextTracker.contextFor(instance)
 
 		go func(desc *InstanceDesc) {
-			if !resultTracker.awaitStart(desc) {
+			if err := resultTracker.awaitStart(ctx, desc); err != nil {
 				// Post to resultsChan so that the deferred cleanup handler above eventually terminates.
 				resultsChan <- instanceResult[T]{
-					err:      errResultNotNeeded, // This will never be returned from this method.
+					err:      err,
 					instance: desc,
 				}
 
 				return
 			}
 
-			result, err := f(instanceCtx, desc)
+			result, err := f(ctx, desc)
 			resultsChan <- instanceResult[T]{
 				result:   result,
 				err:      err,
--- a/ring/replication_set.go
+++ b/ring/replication_set.go
@@ -2,6 +2,7 @@ package ring
 
 import (
 	"context"
+	"errors"
 	"sort"
 	"time"
 )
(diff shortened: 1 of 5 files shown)


ring/replication_set.go
```diff
@@ -140,11 +154,27 @@ func DoUntilQuorum[T any](ctx context.Context, r ReplicationSet, f func(context.
...
+	} else {
+		resultTracker.releaseAllRequests()
+	}
+
 	for i := range r.Instances {
 		instance := &r.Instances[i]
 		instanceCtx := contextTracker.contextFor(instance)
 
 		go func(desc *InstanceDesc) {
+			if !resultTracker.awaitRelease(desc) {
```
pracucci, line 168: Below, when the context is canceled we call `cleanupResultsAlreadyReceived()` and return. I've the feeling we may have a condition where this goroutine is never terminated because `awaitRelease()` waits indefinitely in that case.

[non blocking] Generally speaking, I've found quite hard to follow the "failure" case to make sure `awaitRelease()` always return in every case. I'm wondering if there's anything we can do to make the code easier to follow.

pracucci (commented):
Good job! I find the overall logic non trivial to follow, but I also don't have a good suggestion to improve it, so I wouldn't block on it. The reason why I'm not approving yet is because I have a doubt about the context cancellation case (see related comment) I would like you to double check. Thanks!

ring/replication_set_tracker.go
```diff
@@ -20,6 +25,25 @@ type replicationSetResultTracker interface {
 	// This method should only be called after succeeded returns true for the first time and before
 	// calling done any further times.
 	shouldIncludeResultFrom(instance *InstanceDesc) bool
+
+	// Releases an initial set of requests sufficient to meet the quorum requirements of this tracker.
+	// Further requests will be released if necessary when done is called with a non-nil error.
+	// Calling this method multiple times may lead to unpredictable behaviour.
+	// Calling both this method and releaseAllRequests may lead to unpredictable behaviour.
+	// This method must only be called before calling done.
+	releaseMinimumRequests()
```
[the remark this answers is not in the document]
charleskorn: If "start" is clearer to you, then I'm happy to change it.

I've done this in b9a4db9.

ring/replication_set.go
charleskorn: > Below, when the context is canceled we call `cleanupResultsAlreadyReceived()` and return. I've the feeling we may have a condition where this goroutine is never terminated because `awaitRelease()` waits indefinitely in that case.

Good catch, yep, that could block forever. I'll add a test for this case and then fix it.

> [non blocking] Generally speaking, I've found quite hard to follow the "failure" case to make sure `awaitRelease()` always return in every case. I'm wondering if there's anything we can do to make the code easier to follow.

We might be able to solve both of these problems in one go: I'll modify `awaitRelease()` to take the request context as well, so if that is cancelled, it'll return immediately. This should make it easier to confirm that `awaitRelease()` will always return.

charleskorn: I've implemented this (f272c76), however it introduces additional complexity for the tests (93656f2):

Because `awaitStart()` (formerly `awaitRelease()`) now returns if the instance's context has been cancelled, there's a chance that `f` is not always called and therefore won't need to be cleaned up. The tests previously assumed this was more deterministic, and so the assertions around cleanup behaviour are now more complex.

PR #312, opened 2023-06-05
Add flags to set HTTP flow control options for gRPC clients, and fix some typos in flag descriptions

charleskorn:
**What this PR does**:

This PR adds additional configuration options to `grpcclient.Config` that allow modifying HTTP flow control options.

It also fixes some minor typos in the descriptions for other `grpcclient.Config` configuration options.

(none)

**Checklist**
- [n/a] Tests updated

### DIFF
--- a/grpcclient/grpcclient.go
+++ b/grpcclient/grpcclient.go
@@ -47,7 +47,7 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.StringVar(&cfg.GRPCCompression, prefix+".grpc-compression", "", "Use compression when sending messages. Supported values are: 'gzip', 'snappy' and '' (disable compression)")
 	f.Float64Var(&cfg.RateLimit, prefix+".grpc-client-rate-limit", 0., "Rate limit for gRPC client; 0 means disabled.")
 	f.IntVar(&cfg.RateLimitBurst, prefix+".grpc-client-rate-limit-burst", 0, "Rate limit burst for gRPC client.")
-	f.BoolVar(&cfg.BackoffOnRatelimits, prefix+".backoff-on-ratelimits", false, "Enable backoff and retry when we hit ratelimits.")
+	f.BoolVar(&cfg.BackoffOnRatelimits, prefix+".backoff-on-ratelimits", false, "Enable backoff and retry when we hit rate limits.")
 	f.BoolVar(&cfg.TLSEnabled, prefix+".tls-enabled", cfg.TLSEnabled, "Enable TLS in the gRPC client. This flag needs to be enabled when any other TLS flag is set. If set to false, insecure connection to gRPC server will be used.")
 
 	f.DurationVar(&cfg.ConnectTimeout, prefix+".connect-timeout", 0, "The maximum amount of time to establish a connection. A value of 0 means default gRPC connect timeout and backoff.")
--- a/grpcclient/grpcclient.go
+++ b/grpcclient/grpcclient.go
@@ -150,9 +150,3 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 		}),
 	), nil
 }
-
-func must(err error) {
-	if err != nil {
-		panic(err)
-	}
-}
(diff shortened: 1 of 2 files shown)


grpcclient/grpcclient.go
```diff
@@ -27,6 +28,9 @@ type Config struct {
 	BackoffOnRatelimits bool           `yaml:"backoff_on_ratelimits" category:"advanced"`
 	BackoffConfig       backoff.Config `yaml:"backoff_config"`
 
+	InitialStreamWindowSize     flagext.Bytes `yaml:"initial_stream_window_size" category:"advanced"`
+	InitialConnectionWindowSize flagext.Bytes `yaml:"initial_connection_window_size" category:"advanced"`
```
pracucci, line 32: These are very advanced settings, which I'm not 💯 sure we should even expose to the final user. I'm OK having us experimenting with it, but I'm wondering if it's better to categorise them as "experimental" so that we can eventually remove them anytime with no notice.

grpcclient/grpcclient.go
```diff
@@ -128,5 +137,13 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 			Timeout:             time.Second * 10,
 			PermitWithoutStream: true,
 		}),
+		grpc.WithInitialConnWindowSize(int32(cfg.InitialConnectionWindowSize)),
+		grpc.WithInitialWindowSize(int32(cfg.InitialStreamWindowSize)),
```
pracucci, line 141: See here:
https://github.com/grafana/mimir/issues/4010#issuecomment-1396936200

I think that once you set it, you automatically disable the BDP estimator, which is something we don't want.

grpcclient/grpcclient.go
```diff
@@ -118,6 +129,14 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 		)
 	}
 
+	if cfg.InitialStreamWindowSize > defaultInitialWindowSize {
```
pracucci, line 132: [nit] I would add a comment to explain **why** (BDP estimator being turned off when it's explicitely set, even at the same value of the default one).

pracucci (approved):
LGTM (modulo a nit). Thanks!

grpcclient/grpcclient.go
charleskorn: Good point, added in 003f044.

grpcclient/grpcclient.go
```diff
@@ -41,16 +45,23 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 	f.IntVar(&cfg.MaxSendMsgSize, prefix+".grpc-max-send-msg-size", 100<<20, "gRPC client max send message size (bytes).")
 	f.StringVar(&cfg.GRPCCompression, prefix+".grpc-compression", "", "Use compression when sending messages. Supported values are: 'gzip', 'snappy' and '' (disable compression)")
 	f.Float64Var(&cfg.RateLimit, prefix+".grpc-client-rate-limit", 0., "Rate limit for gRPC client; 0 means disabled.")
 	f.IntVar(&cfg.RateLimitBurst, prefix+".grpc-client-rate-limit-burst", 0, "Rate limit burst for gRPC client.")
-	f.BoolVar(&cfg.BackoffOnRatelimits, prefix+".backoff-on-ratelimits", false, "Enable backoff and retry when we hit ratelimits.")
-	f.BoolVar(&cfg.TLSEnabled, prefix+".tls-enabled", cfg.TLSEnabled, "Enable TLS in the GRPC client. This flag needs to be enabled when any other TLS flag is set. If set to false, insecure connection to gRPC server will be used.")
-
+	f.BoolVar(&cfg.BackoffOnRatelimits, prefix+".backoff-on-ratelimits", false, "Enable backoff and retry when we hit rate limits.")
+	f.Var(&cfg.InitialStreamWindowSize, prefix+".initial-stream-window-size", "Initial stream window size. Values less than the default are not supported and are ignored.")
```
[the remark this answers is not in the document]
charleskorn: Good point, added in 003f044.

PR #319, opened 2023-06-13
ring: make it harder to leak contexts when using DoUntilQuorum

charleskorn:
**What this PR does**:

This PR makes two changes to `DoUntilQuorum` to make it harder to accidentally leak `context.Context`s:

* The behaviour of `DoUntilQuorum` has been changed to always cancel all contexts passed to `f`. Previously, if the result of a call to `f` was returned from `DoUntilQuorum`, then the corresponding context was not cancelled.

* A new method has been introduced that behaves the same as `DoUntilQuorum` used to, `DoUntilQuorumWithoutSuccessfulContextCancellation`. In addition to the existing parameters, this method also passes a `context.CancelFunc` to `f` that must be called to avoid leaking any contexts.

Notes to reviewers:

* I'm not wedded to the name `DoUntilQuorumWithoutSuccessfulContextCancellation`, open to other suggestions. I wanted a name that conveyed you were doing something a little unusual and pushes people towards `DoUntilQuorum` by default.

* In the interests of simplicity and ease of maintenance, I haven't duplicated all of the previous tests for both `DoUntilQuorum` and `DoUntilQuorumWithoutSuccessfulContextCancellation`. Open to other opinions though.

Related: https://github.com/grafana/mimir/pull/5199

(none)

**Checklist**
- [x] Tests updated


pracucci (approved):
Good job! I agree having an explicit (even if verbose) naming.

PR #318, opened 2023-06-12
Introduce TokenGenerator interface and randomTokenGenerator type

duricanikolic:
**What this PR does**:
This PR introduces:
- `TokenGenerator` interface, which exposes the `GenerateTokens()` function for generating unique and sorted tokens. 
- `randomTokenGenerator` type implementing the `TokenGenerator` interface, and which generates unique, random and sorted tokens.

Implementation of `randomTokenGenerator.GenerateTokens()` replaces `util.GenerateTokens()`.

In the following PRs related to this [issue](https://github.com/grafana/mimir-squad/issues/1042), an additional implementation of `TokenGenerator` interface will be provided.

**Which issue(s) this PR fixes**:
Part of https://github.com/grafana/mimir-squad/issues/1042

**Checklist**
- [x] Tests updated
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #321, opened 2023-06-13
Introducing SpreadMinimizingTokenGenerator

duricanikolic:
**What this PR does**:
This PR introduces `SpreadMinimizingTokenGenerator`, another implementation of the `TokenGenerator` interface (added in https://github.com/grafana/dskit/pull/318). For the given instance id and zone, `SpreadMinimizingTokenGenerator` generates required number of tokens for that instance in such a way that the _spread_, i.e., the ratio between the minimal and maximal token range owned by the instances from the ring, is minimal.

In the executed tests, when shuffle sharding is disabled, the spread is shown to be below 0.2% (have a look at the example provided below). 

mimir-dedicated-17

This PR just introduces `SpreadMinimizingTokenGenerator` and tests that show its functionality. It is not yet used by any other component of the system. This will be addressed in the following PRs.

**Checklist**
- [x] Tests updated


ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,200 @@
...
+
+// ownershipPriorityQueue is a max-heap, i.e., a priority queue
+// where items with a higher priority will be extracted first.
+// Namely, items with a higher ownership have a higher priority.
+// In order to guarantee that 2 instances of ownershipPriorityQueue
+// with the same items always assign equal priorities to equal items,
+// in the case of items with equal ownership, we rely on the
+// order of item ids.
+type ownershipPriorityQueue struct {
+	items []*ownershipInfo
```
colega, line 99: Given that one `ownershipPriorityQueue` should contain homogeneous elements (it can't contain some instances and some tokens), I think this should be:
```go
type ringItem[T any] interface {
	GetID() int
	Less(other T) bool
}

type ownershipInfo[T ringItem[T]] struct {
	ringItem  T
	ownership float64
	index     int
}

type ownershipPriorityQueue[T ringItem[T]] struct {
	items []*ownershipInfo[T]
}

func newPriorityQueue[T ringItem[T]](len int) *ownershipPriorityQueue[T] {
	return &ownershipPriorityQueue[T]{
		items: make([]*ownershipInfo[T], 0, len),
	}
}
```

This way the specific implementations don't need to type assert, and they can be just:

```go
type ringInstance struct{ instanceID int }

func (ri *ringInstance) GetID() int                    { return ri.instanceID }
func (ri *ringInstance) Less(other *ringInstance) bool { return ri.instanceID < other.instanceID }

type ringToken struct{ token, prevToken uint32 }

func (rt *ringToken) GetID() int                 { return int(rt.token) }
func (rt *ringToken) Less(other *ringToken) bool { return rt.token < other.token }
```

ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,200 @@
...
+	token     uint32
+	prevToken uint32
+}
+
+func newRingToken(token, prevToken uint32) *ringToken {
+	return &ringToken{
+		token:     token,
+		prevToken: prevToken,
+	}
+}
```
colega, line 50: These constructors don't add any value to the code, and they're not idiomatic in golang. 
These are private types and we build them with all fields and we don't do any kind of logic (not even a single `make()` call that could justify them). I'd remove them.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,281 @@
...
+	tokensQueues[0] = tokensQueue
+
+	// instanceQueue is a priority queue of instances such that instances with higher ownership have a higher priority
+	instanceQueue := newPriorityQueue(t.instanceID)
+	instanceQueue.Add(newRingInstanceOwnershipInfo(0, instanceOwnership))
+	heap.Init(instanceQueue)
+	defer t.cleanUp(instanceQueue, tokensQueues)
+
+	for i := 1; i <= t.instanceID; i++ {
+		optimalInstanceOwnership := float64(1<<32) / float64(i+1)
```
colega, line 205: I'd suggest extracting `1<<32` into:
```go
const totalTokensCount = 1<<32
```
To reduce the cognitive load here.

ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,200 @@
...
+func (pq *ownershipPriorityQueue) Peek() *ownershipInfo {
+	return (pq.items)[0]
+}
+
+// Update updates the element ownershipInfo passed as parameter by applying to it the updating function update
+// passed as parameter, and propagates this modification to ownershipPriorityQueue. Element ownershipInfo must
+// be already present on ownershipPriorityQueue. Time complexity is O(log n), where n = Len().
+func (pq *ownershipPriorityQueue) Update(ownershipInfo *ownershipInfo, update func(*ownershipInfo)) {
+	update(ownershipInfo)
+	heap.Fix(pq, ownershipInfo.index)
```
colega, line 170: The fact that `ownershipInfo` had an `index` was bothering me, so I followed the code and I don't think that it's needed, neither this `Update` method.
The only two usages of this method use an info that came from peek, so it's index is always 0. I tried adding this here:

```
	if ownershipInfo.index != 0 {
		panic("HERE IT IS NOT 0" + strconv.Itoa(ownershipInfo.index))
	}
```

And the only test that failed was the `Update` one (because it violates that assumption). 

I would remove `index` from `ownershipPriorityQueue` and all the logic related to keeping track of it. Then instead of calling `Update` with a lambda, I'd just update the element and call `heap.Fix(pq, 0)`.

ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	// We are implementing a max-heap, so we are using > here.
+	// Since we compare float64, NaN values must be placed at the end.
+	return pq.items[i].ownership > pq.items[j].ownership || (math.IsNaN(pq.items[j].ownership) && !math.IsNaN(pq.items[i].ownership))
+}
+
+func (pq *ownershipPriorityQueue) compareRingItems(ri1, ri2 ringItem) bool {
+	if ri1 == nil || ri2 == nil {
+		return true
+	}
```
colega, line 133: When can these be nil?

ring/ownership_priority_queue.go
colega: While reviewing the code further, I realized that then we also don't need the `GetID()` method. And we also don't need any of those methods exporter. All we need need is to compare the values. However that can be made even easier.

```go
// ordered is like comparable constraint but it returns an int key to compare instead of being comparable itself.
type ordered interface{ key() int }

type ringInstance struct{ instanceID int }

func (ri ringInstance) key() int       { return ri.instanceID }

type ringToken struct{ token, prevToken uint32 }

func (rt ringToken) key() int { return int(rt.token) }

// ...


type ownershipInfo[T ordered] struct {
	ringItem  T
	ownership float64
	index     int
}


// ...

type ownershipPriorityQueue[T ordered] struct {
	items []*ownershipInfo[T]
}

func newPriorityQueue[T ordered](len int) *ownershipPriorityQueue[T] {
	return &ownershipPriorityQueue[T]{
		items: make([]*ownershipInfo[T], 0, len),
	}
}

```

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,281 @@
...
+	maxTokenValue := (math.MaxUint32/zonesCount - 1) * zonesCount
+	offset := maxTokenValue - token.prevToken
+	if offset < optimalTokenOwnership {
+		newToken := optimalTokenOwnership - offset
+		return newToken, nil
+	}
+	return token.prevToken + optimalTokenOwnership, nil
+}
+
+// getOptimalTokenOwnership calculates the optimal ownership of the remaining currTokensCount tokens of an instance
```
colega, line 158: I find confusing that `currTokensCount` isn't the _current tokens count_ but the _remaining tokens count_.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,281 @@
...
+			if t.cfg.tokenUniquenessCheck {
+				for _, token := range tokens {
+					if slices.Contains(takenTokens, token) {
+						level.Error(t.logger).Log("msg", "token uniqueness check has failed", "token", token, "instanceID", t.instanceID, "zone", t.cfg.zone)
+						return nil, errorTokenNotUnique(i, t.cfg.zone, token)
+					}
+				}
+				level.Info(t.logger).Log("msg", "token uniqueness has been verified", "instanceID", t.instanceID, "zone", t.cfg.zone)
+			}
+			return tokens, nil
```
colega, line 262: Let's make this complex method shorter but extracting this distracting part into a separate method:
```suggestion
			return t.sortAndMaybeCheckUniqueness(tokens)
```

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,281 @@
...
+				tokenOwnershipInfo.ownership = newTokenOwnership
+			})
+			instanceQueue.Update(highestOwnershipInstance, func(instanceOwnershipInfo *ownershipInfo) {
+				instanceOwnershipInfo.ownership = highestOwnershipInstance.ownership - oldTokenOwnership + newTokenOwnership
+			})
+			addedTokens++
+		}
+		if addedTokens != tokensCount {
+			return nil, errorNotAllTokenCreated(i, t.cfg.zone, tokensCount)
+		}
```
colega, line 250: As far as I understand, this can only happen if we perform the `break` above. So I'd move this there and remove the condition.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,281 @@
...
+	for i := 1; i <= t.instanceID; i++ {
+		optimalInstanceOwnership := float64(1<<32) / float64(i+1)
+		currInstanceOwnership := 0.0
+		addedTokens := 0
+		tokens := make(Tokens, 0, tokensCount)
+		// currInstanceTokenQueue is the priority queue of tokens of newInstance
+		currInstanceTokenQueue := newPriorityQueue(tokensCount)
+		for addedTokens < tokensCount {
+			optimalTokenOwnership := t.getOptimalTokenOwnership(optimalInstanceOwnership, currInstanceOwnership, uint32(tokensCount-addedTokens))
+			highestOwnershipInstance := instanceQueue.Peek()
```
colega, line 213: How can we be sure that there are enough instances in the queue to peek here?

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, errorZoneSetTooBig(len(t.zones), tokensPerInstance)
+	}
+	tokens := make(Tokens, 0, tokensPerInstance)
+	for i := 0; i < tokensPerInstance; i++ {
+		token := uint32(i << tokenDistanceLen)
+		token = (token/zonesCount)*zonesCount + uint32(zoneID)
+		tokens = append(tokens, token)
+	}
+	return tokens, nil
+}
```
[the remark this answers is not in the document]
duricanikolic: I have addressed this hint, and I have removed the modulo `len(zones)` constraint.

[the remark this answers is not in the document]
duricanikolic: In this [commit](https://github.com/grafana/dskit/pull/321/commits/52689ba9693cdf1c18ec4e0b3a27849f930a9b4b) I simplify the calculation of first instance tokens.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,287 @@
...
+	tokensPerInstance int
+}
+
+func NewSpreadMinimizingConfig(instanceID, zone string, tokensPerInstance int) *SpreadMinimizingConfig {
+	return &SpreadMinimizingConfig{
+		instanceID:        instanceID,
+		zone:              zone,
+		tokensPerInstance: tokensPerInstance,
+	}
+}
```
colega, line 49: `Config` structs are usually just created, without a constructor. Just export the fields.

OTOH, no need to use a pointer for a config.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,287 @@
...
+	instanceID int
+	zoneID     int
+	zones      []string
+	logger     log.Logger
+}
+
+func NewSpreadMinimizingTokenGenerator(cfg *SpreadMinimizingConfig, zones []string, logger log.Logger) (*SpreadMinimizingTokenGenerator, error) {
+	if !slices.IsSorted(zones) {
+		sort.Strings(zones)
+	}
```
colega, line 62: If we're going to modify the `zones`, we should copy them first.

ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,145 @@
...
+// with the same items always assign equal priorities to equal items,
+// in the case of items with equal ownership, we rely on the
+// order of item ids.
+type ownershipPriorityQueue[T ringItem] struct {
+	items []ownershipInfo[T]
+}
+
+func newPriorityQueue[T ringItem](len int) ownershipPriorityQueue[T] {
+	return ownershipPriorityQueue[T]{
+		items: make([]ownershipInfo[T], 0, len),
```
colega, line 77: Let's reduce possible confusions, since we're not building a slice of that length, but that capacity.
```suggestion
func newPriorityQueue[T ringItem](capacity int) ownershipPriorityQueue[T] {
	return ownershipPriorityQueue[T]{
		items: make([]ownershipInfo[T], 0, capacity),
```

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+		tokensQueue.Add(newRingTokenOwnershipInfo(token, firstInstanceTokens[prev]))
+		prev = tk
+	}
+	heap.Init(&tokensQueue)
+	tokensQueues[0] = tokensQueue
+
+	// instanceQueue is a priority queue of instances such that instances with higher ownership have a higher priority
+	instanceQueue := newPriorityQueue[ringInstance](t.instanceID)
+	instanceQueue.Add(newRingInstanceOwnershipInfo(0, firstInstanceOwnership))
+	heap.Init(&instanceQueue)
```
colega, line 205: Note that `heap.Init` call is useless for a single element, but if we really want to use heap, we can just `heap.Push(&instanceQueue, newRingInstanceOwnershipInfo(0, firstInstanceOwnership))`

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+		tokens := make(Tokens, 0, tokensCount)
+		// currInstanceTokenQueue is the priority queue of tokens of newInstance
+		currInstanceTokenQueue := newPriorityQueue[ringToken](tokensCount)
+		for addedTokens < tokensCount {
+			optimalTokenOwnership := t.getOptimalTokenOwnership(optimalInstanceOwnership, currInstanceOwnership, uint32(tokensCount-addedTokens))
+			highestOwnershipInstance := instanceQueue.Peek()
+			if highestOwnershipInstance == nil || highestOwnershipInstance.ownership <= float64(optimalTokenOwnership) {
+				level.Warn(t.logger).Log("msg", "it was impossible to add a token because the instance with the highest ownership cannot satisfy the request", "added tokens", addedTokens+1, "highest ownership", highestOwnershipInstance.ownership, "requested ownership", optimalTokenOwnership)
+				// if this happens, we return the current set of tokens, although it is less than t.cfg.tokensPerInstance
+				break
```
colega, line 223: Now we need to panic here. We do have a test that checks that this doesn't happen up to an established number of instances, right? If so, let's just check that number of instances when we create this strategy, and panic here as it's unexpected.

Same applies to the `break` below.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+	instanceQueue.Add(newRingInstanceOwnershipInfo(0, firstInstanceOwnership))
+	heap.Init(&instanceQueue)
+
+	for i := 1; i <= t.instanceID; i++ {
+		optimalInstanceOwnership := float64(totalTokensCount) / float64(i+1)
+		currInstanceOwnership := 0.0
+		addedTokens := 0
+		// ignoredInstances is a slice of the current instances whose tokens
+		// don't have enough space to accommodate new tokens.
+		ignoredInstances := make([]ownershipInfo[ringInstance], 0, i)
```
colega, line 213: Please initialize this slice once before the loop for the max capacity, and reset it (`ignoredInstances = ignoredInstances[:0]`) on each loop iteration, otherwise we're allocating a slice for each loop iteration (and we know that we won't even need it until we reach hundreds of instances.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// getOptimalTokenOwnership calculates the optimal ownership of the remaining currTokensCount tokens of an instance
+// having the given current instances ownership currInstanceOwnership and the given optimal instance ownership
+// optimalInstanceOwnership. The resulting token ownership must be a multiple of the number of zones.
+func (t *SpreadMinimizingTokenGenerator) getOptimalTokenOwnership(optimalInstanceOwnership, currInstanceOwnership float64, currTokensCount uint32) uint32 {
+	optimalTokenOwnership := uint32(optimalInstanceOwnership-currInstanceOwnership) / currTokensCount
+	zonesCount := uint32(len(t.zones))
+	optimalTokenOwnership = (optimalTokenOwnership / zonesCount) * zonesCount
+	return optimalTokenOwnership
```
colega, line 144: Nit: 
```suggestion
	return (optimalTokenOwnership / zonesCount) * zonesCount
```

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+		newToken := optimalTokenOwnership - offset
+		return newToken, nil
+	}
+	return token.prevToken + optimalTokenOwnership, nil
+}
+
+// getOptimalTokenOwnership calculates the optimal ownership of the remaining currTokensCount tokens of an instance
+// having the given current instances ownership currInstanceOwnership and the given optimal instance ownership
+// optimalInstanceOwnership. The resulting token ownership must be a multiple of the number of zones.
+func (t *SpreadMinimizingTokenGenerator) getOptimalTokenOwnership(optimalInstanceOwnership, currInstanceOwnership float64, currTokensCount uint32) uint32 {
```
colega, line 140: Nit, but still _important_ to keep the style consistent:
```suggestion
func (t *SpreadMinimizingTokenGenerator) optimalTokenOwnership(optimalInstanceOwnership, currInstanceOwnership float64, currTokensCount uint32) uint32 {
```

[Effective Go](https://go.dev/doc/effective_go):
> it's neither idiomatic nor necessary to put Get into the getter's name

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+	parts := instanceIDRegex.FindStringSubmatch(instanceID)
+	if len(parts) != 4 {
+		return 0, errorBadInstanceIDFormat(instanceID)
+	}
+	return strconv.Atoi(parts[3])
+}
+
+// getZoneID gets a zone name and a slice of sorted zones,
+// and return the index of the zone in the slice.
+func getZoneID(zone string, sortedZones []string) (int, error) {
```
colega, line 86: Nit, and slightly optionated, but there aren't really _getters_ (and as mentioned in previous comment, getters shouldn't have a `get` prefix), so I'd call these methods:

```go 
func parseInstanceID(instanceID string) (int, error)

func findZoneID(zone string, sortedZones []string) (int, error)
```

Because the first one does the parsing of the string, while the second one tries to find it in the list of provided zones.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	return index, nil
+}
+
+// generateFirstInstanceTokens calculates a set of tokens that should be assigned to the first instance (with id 0)
+// of the zone of the underlying instance.
+func (t *SpreadMinimizingTokenGenerator) generateFirstInstanceTokens() Tokens {
+	zonesCount := len(t.zones)
+	tokensPerInstance := t.cfg.TokensPerInstance
+	tokenDistance := (totalTokensCount / tokensPerInstance / zonesCount) * zonesCount
```
colega, line 102: As discussed offline, relying on `len(t.zones)` makes it impossible to change the number of zones. Can we remove this calculation and rely on `const maxZonesCount = 5` everywhere? That would allow us scaling up to 5 zones as needed, which is a reasonable tradeoff.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	errorDistanceBetweenTokensNotBigEnough = func(optimalTokenOwnership, ownership int, token ringToken) error {
+		return fmt.Errorf("calculation of a new token between %d and %d with optimal token ownership %d was impossible: distance between lower and upper bound %d is not big enough", token.prevToken, token.token, optimalTokenOwnership, ownership)
+	}
+)
+
+type SpreadMinimizingConfig struct {
+	InstanceID        string
+	Zone              string
+	TokensPerInstance int
```
colega, line 40: As discussed: changing the number of tokens per instance will not really work for this strategy (because the ownership decisions will most likely lead to different set of tokens, especially when not powers of 2). 

So we should hard code this to 512, as that's the number we know works and we recommend to the users
Verify in GenerateTokens that we're requested that number of tokens, and panic otherwise.

ring/spread_minimizing_token_generator.go
colega: Also, as an extension to this: we're doing `/ zonesCount) * zonesCount` everywhere, with zonesCount being usually 3, sometimes 5. That's an ugly math for computers and it will always lead to strange numbers.

What if we set `const maxZonesCount = 8`, so `/ zonesCount) * zonesCount` can be translated by the compiler into `>> 3) << 3`, and also the modulo checks below would become just `& 0b111`.

This will also make it easier to understand what's assigned to each zone: each zone has $2^{32-3}$ tokens space, rather than $2^{32}/5$ (which isn't even divisible so it's harder to reason about, IMO).

I've checked few of your tests and they run fine with:
```
	zones = func() (zones []string) {
		for i := 0; i < 8; i++ {
			zones = append(zones, fmt.Sprintf("zone-%s", string([]byte{'a' + byte(i)})))
		}
		return zones
	}()
```

WDYT?

ring/spread_minimizing_token_generator_test.go
```diff
@@ -0,0 +1,553 @@
...
+	tokensPerInstance = 512
+)
+
+func TestSpreadMinimizingTokenGenerator_GetInstanceID(t *testing.T) {
+	tests := map[string]struct {
+		instanceID    string
+		expectedID    int
+		expectedError error
+	}{
+		"instance-zone-a-10 is correct": {
```
colega, line 28: I know it works, but can you add this test case that includes a dash in the instance name?
```
		"store-gateway-zone-a-10 is correct": {
			instanceID: "store-gateway-zone-a-10",
			expectedID: 10,
		},
```

ring/spread_minimizing_token_generator_test.go
```diff
@@ -0,0 +1,553 @@
...
+				if _, found := allTokens[token]; found {
+					err := fmt.Errorf("token %d been found more than once", token)
+					panic(err)
+				}
+				allTokens[token] = true
+			}
+		}
+	}
+}
+
```
colega, line 313:
```suggestion
func BenchmarkSpreadMinimizingTokenGenerator_GenerateTokens(b *testing.B) {
	tokensPerInstance := 512
	instanceID := 1000
	zone := zones[0]
	instance := fmt.Sprintf("instance-%s-%d", zone, instanceID)
	cfg := &SpreadMinimizingConfig{instance, zone, tokensPerInstance}
	tokenGenerator := createSpreadMinimizingTokenGenerator(b, cfg, zones)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		tokenGenerator.GenerateTokens(512, nil)
	}
}
```

You'll also have to change `createSpreadMinimizingTokenGenerator` to accept `testing.TB` instead of `*testing.T`.

ring/spread_minimizing_token_generator_test.go
```diff
@@ -0,0 +1,553 @@
...
+			heap.Fix(&instanceQueue, 0)
+
+			addedTokens++
+		}
+		slices.Sort(tokens)
+		allTokens[i] = tokens
+		if i == t.instanceID {
+			return allTokens
+		}
+		if len(ignoredInstances) != 0 {
```
colega, line 540: This condition isn't needed, if it's empty, the for loop will be skipped.

ring/spread_minimizing_token_generator.go
duricanikolic: Thank you for this hint. I thought it was mandatory to call `heap.Init` before doing any other operation.

colega: `heap.Init()` will just ensure that it's a valid heap, but a slice of one element is already a max heap :D (and a min heap!)

ring/token_generator.go
```diff
@@ -7,8 +7,9 @@ import (
 )
 
 type TokenGenerator interface {
-	// GenerateTokens generates unique tokensCount tokens, none of which clash
-	// with the given takenTokens. Generated tokens are sorted.
+	// GenerateTokens generates tokensCount unique tokens, none of which clash with the given takenTokens.
+	// Generated tokens are sorted.
+	// If it is not possible to generate the requested tokensCount tokens, GenerateTokens panics.
```
colega, line 12: There's no need to specify this. All implementations should be able to generate the requested count of tokens. 

The implementation you're introducing also does that, we have a test that ensures that it doesn't panic.

ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,139 @@
...
+		// higher priority is determined by the order of ids.
+		return pq.items[i].item.key() > pq.items[j].item.key()
+	}
+	// We are implementing a max-heap, so we are using > here.
+	// Since we compare float64, NaN values must be placed at the end.
+	return pq.items[i].ownership > pq.items[j].ownership || (math.IsNaN(pq.items[j].ownership) && !math.IsNaN(pq.items[i].ownership))
+}
+
+// Push implements heap.Push(any). It pushes the element item onto ownershipPriorityQueue.
+// Time complexity is O(log n), where n = Len().
```
pracucci, line 102: [nit] This is not the complexity of this `Push()` implementation, but the "push" operation when issued by the `heap`, right? If my understanding is correct, then I think it's a bit misleading mentioning it here.

_Same comment applies below._

ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Pop implements heap.Pop(). It removes and returns the element with the highest priority from ownershipPriorityQueue.
+// Time complexity is O(log n), where n = Len().
+func (pq *ownershipPriorityQueue[T]) Pop() any {
+	old := pq.items
+	n := len(old)
+	item := old[n-1]
+	pq.items = old[0 : n-1]
+	return item
```
pracucci, line 115: [nit] Up to you, but you don't need `old`. You can rewrite it as:

```suggestion
	n := len(pq.items)
	item := pq.items[n-1]
	pq.items = pq.items[0 : n-1]
	return item
```

ring/ownership_priority_queue.go
```diff
@@ -0,0 +1,139 @@
...
+	ownershipInfo := item.(ownershipInfo[T])
+	pq.items = append(pq.items, ownershipInfo)
+}
+
+// Pop implements heap.Pop(). It removes and returns the element with the highest priority from ownershipPriorityQueue.
+// Time complexity is O(log n), where n = Len().
+func (pq *ownershipPriorityQueue[T]) Pop() any {
+	old := pq.items
+	n := len(old)
+	item := old[n-1]
```
pracucci, line 113: I'm missing something. The heap keeps the minimum element in the tree is the root, so at index 0. Since Less() is implemented in "inverse order" ("less" returns true for the item with the highest priority), at index 0 we expect the item with the highest priority. Why do we look at the last one instead?

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,284 @@
...
+)
+
+const (
+	totalTokensCount         = 1 << 32
+	optimalTokensPerInstance = 1 << 9
+	maxZonesCount            = 1 << 3
+)
+
+var (
+	instanceIDRegex          = regexp.MustCompile(`^(.*)-(zone-[a-z])-(\d+)$`)
```
pracucci, line 23: Why do we require "-zone-X-" in the ID at all? We just need to parse the ID cause the zone will be an information in input. If not strictly required, I would relax this, so that there's no assumption on the format of the zone in the instance ID.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,284 @@
...
+type SpreadMinimizingTokenGenerator struct {
+	cfg        SpreadMinimizingConfig
+	instanceID int
+	zoneID     int
+	zones      []string
+	logger     log.Logger
+}
+
+func NewSpreadMinimizingTokenGenerator(cfg SpreadMinimizingConfig, zones []string, logger log.Logger) (*SpreadMinimizingTokenGenerator, error) {
+	sortedZones := make([]string, len(zones))
```
pracucci, line 55: Shouldn't we add a check to ensure `len(zones) < maxZonesCount`?

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,284 @@
...
+	copy(sortedZones, zones)
+	if !slices.IsSorted(sortedZones) {
+		sort.Strings(sortedZones)
+	}
+	instanceID, err := parseInstanceID(cfg.InstanceID)
+	if err != nil {
+		return nil, err
+	}
+	zoneID, err := findZoneID(cfg.Zone, sortedZones)
```
pracucci, line 64: [nit] `findZoneID` looks a complication to me. You can just use `slices.Index()`. You don't need any binary search or so here, the number of zones is small.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	return strconv.Atoi(parts[3])
+}
+
+// findZoneID gets a zone name and a slice of sorted zones,
+// and return the index of the zone in the slice.
+func findZoneID(zone string, sortedZones []string) (int, error) {
+	if !slices.IsSorted(sortedZones) {
+		slices.Sort(sortedZones)
+	}
```
pracucci, line 92: [nit] `sortedZones` is already sorted. But I think you don't need this function at all.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,284 @@
...
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"golang.org/x/exp/slices"
+)
+
+const (
+	totalTokensCount         = 1 << 32
+	optimalTokensPerInstance = 1 << 9
+	maxZonesCount            = 1 << 3
```
pracucci, line 19: [nit] Do we really need bit wise operators here? Can't we just put the actual number to help anyone else different than chatgpt to read it? 😉

ring/spread_minimizing_token_generator.go
pracucci, line 64: I'm a bit confused by the concept of zone ID here, because I don't know which actual value will be passed as `zones`. Let's say you're creating a new cluster and the first instance registers in the ring, and it's from `zone-b`. What will be the value of `zones` passed to `NewSpreadMinimizingTokenGenerator()`?

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,284 @@
...
+	if index >= len(sortedZones) {
+		return 0, errorZoneNotValid(zone)
+	}
+	return index, nil
+}
+
+// generateFirstInstanceTokens calculates a set of tokens that should be assigned to the first instance (with id 0)
+// of the zone of the underlying instance.
+func (t *SpreadMinimizingTokenGenerator) generateFirstInstanceTokens() Tokens {
+	tokenDistance := (totalTokensCount / optimalTokensPerInstance / maxZonesCount) * maxZonesCount
```
pracucci, line 103: I think you will have to explain me why we need to do `/ maxZonesCount) * maxZonesCount` 🤔

ring/ownership_priority_queue_test.go
```diff
@@ -0,0 +1,130 @@
...
+func getRandomTokenWeight() float64 {
+	return (maxTokenWeight-minTokenWeight)*rand.Float64() + minTokenWeight
+}
+
+func createPriorityQueue(size int) (ownershipPriorityQueue[ringInstance], float64, float64) {
+	pq := newPriorityQueue[ringInstance](size)
+
+	minWeight := float64(math.MaxUint32)
+	maxWeight := 0.0
+	for i := 1; i <= size-1; i++ {
```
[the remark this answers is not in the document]
duricanikolic: This is a leftover from a previous implementation. Will fix this.

ring/ownership_priority_queue.go
duricanikolic: Here is the [example](https://pkg.go.dev/container/heap#:~:text=Example,-(PriorityQueue)%20%C2%B6) provided by container/heap that I used as a model. Basically every custom implementation is used [at the end](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/container/heap/heap.go;l=63) of [heap.Pop](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/container/heap/heap.go;l=59) implementation that I call when I want to remove the element with the highest ownership. At line [61](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/container/heap/heap.go;l=61) the highest element and the element at index len()-1 are swapped. So it means that the highest element is now at index len()-1 . Then at line 62, the down method is called, which “heapifies” elements [0-len()-1) (I used “)” to show that the last one is excluded). And that’s why the custom Pop implementation needs to return element len()-1. Basically, this doesn’t change either when you implement min or max heap. The only difference should be the Less method, which in the case of max heap should invert the order.

ring/ownership_priority_queue_test.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func TestOwnershipPriorityQueue_EqualOwnershipOfRingTokens(t *testing.T) {
+	pq := newPriorityQueue[ringToken](3)
+	// ownership of first is 20, and its token is 60
+	first := newRingTokenOwnershipInfo(60, 40)
+
+	// ownership of second is 20, and its token is 40, so second > first
+	// (when ownership is equal, priority is decided by the order of ids)
+	second := newRingTokenOwnershipInfo(40, 40)
```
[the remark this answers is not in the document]
duricanikolic: Good catch, what the test was showing and what it was intended to show were 2 different things. I have fixed it now.

ring/spread_minimizing_token_generator.go
duricanikolic: I need an identifier that will distinguish different zones. 
If we have zones `zone-a`, `zone-b` and `zone-c`, I would need to assign them numbers `0`,`1` and `2` (in any order, but the same zone must always be assigned to the same number), because in the approach I am introducing tokens from the same zone must be equal modulo `maxZoneCount` to each other.

ring/spread_minimizing_token_generator.go
duricanikolic: Similarly to the previous comment, in this approach tokens from one zone are always equal to each other modulo `maxZonesCount`. Therefore, I calculated tokenDistance as a multiple of `maxZonesCount`, so we are sure that the loop that calculates the actual token will satisfy that pattern.
With the current values of constants `totalTokensCount`, `optimalTokensPerInstance` and `maxZonesCount` (being $2^{32}$, $2^{9}$ and $2^{3}$) it is not necessary to do `/ maxZonesCount) * maxZonesCount`, but the check is there to guarante that if those values are changed, the property of multilpe of `maxZonesCount` is still guaranteed.

colega (approved):
:rocket:

ring/spread_minimizing_token_generator.go
pracucci: Ok, so in practice we expect input `zones` to always be `[]string{"zone-a", "zone-b", "zone-c"}` even when the 1st instance is registered to the ring and the ring is empty.

ring/spread_minimizing_token_generator.go
```diff
@@ -0,0 +1,287 @@
...
+		newToken := optimalTokenOwnership - offset
+		return newToken, nil
+	}
+	return token.prevToken + optimalTokenOwnership, nil
+}
+
+// optimalTokenOwnership calculates the optimal ownership of the remaining currTokensCount tokens of an instance
+// having the given current instances ownership currInstanceOwnership and the given optimal instance ownership
+// optimalInstanceOwnership. The resulting token ownership must be a multiple of the number of zones.
+func (t *SpreadMinimizingTokenGenerator) optimalTokenOwnership(optimalInstanceOwnership, currInstanceOwnership float64, currTokensCount uint32) uint32 {
```
pracucci, line 143: [nit] Have you considered rename "currTokensCount" to "remainingTokensCount"? I think may help to read it.

ring/spread_minimizing_token_generator.go
pracucci: [nit] Can you add a comment to explain it briefly, please?

ring/spread_minimizing_token_generator_test.go
```diff
@@ -0,0 +1,558 @@
...
+			prev = tk
+		}
+	}
+	return ownershipByInstanceByZone
+}
+
+// generateTokensByInstanceID is a modified implementation of SpreadMinimizingTokenGenerator.generateAllTokens that
+// is used in the tests to speed up the generation of tokens, since the generation of tokens for instance with id
+// instanceID already generates the tokens of all other instances with lower ids.
+func (t *SpreadMinimizingTokenGenerator) generateTokensByInstanceID() map[int]Tokens {
```
pracucci, line 467: Unfortunately I don't think this is a good idea. In the tests using this, we're not testing the real implementation but a mocked version, so a change in the real implementation which introduces a bug may not being noticed. I strongly recommend to get rid of this.

What can we do instead? We could slightly modify `SpreadMinimizingTokenGenerator.generateAllTokens()` to also return the tokens for all instances (since they were computed). Then this value is ignored by `SpreadMinimizingTokenGenerator.GenerateTokens()`, but in the tests you call `generateAllTokens()` instead of `GenerateTokens()`.

pracucci (approved):
Good job! The code is not trivial to understand, not because you coded it poorly but because of its intrinsic complexity. I think I've got a decent understanding of it, and I haven't noticed any issue. However, I feel quite strong about not mocking the logic in tests: https://github.com/grafana/dskit/pull/321#discussion_r1232329274

Approving, but I would like to see the comment about the tests addressed.

duricanikolic:
> Good job! The code is not trivial to understand, not because you coded it poorly but because of its intrinsic complexity. I think I've got a decent understanding of it, and I haven't noticed any issue. However, I feel quite strong about not mocking the logic in tests: [#321 (comment)](https://github.com/grafana/dskit/pull/321#discussion_r1232329274)
> 
> Approving, but I would like to see the comment about the tests addressed.

Hi @pracucci, 
I have implemented your requests about the mocked logic. I am going to merge, and if you have additional requests I can prepare another PR.

PR #322, opened 2023-06-14
Allow TokenGenerator.GenerateTokens() to return error

duricanikolic:
**What this PR does**:
This PR modifies the signature of `TokenGenerator.GenerateTokens()` method in such a way that it, beside a set of tokens, returns also an error in the case something went wrong during the token generation.

Additionally, the signature of `BasicLifecyclerDelegate.OnRingInstanceRegister` was modified in a similar way: beside the instance state and the set of tokens assigned to the registered instance, it also returns an error in the case it was not possible to obtain the set of tokens to assign to the instance.
 
**Checklist**
- [ ] Tests updated

### DIFF
--- a/ring/basic_lifecycler.go
+++ b/ring/basic_lifecycler.go
@@ -21,8 +21,9 @@ import (
 type BasicLifecyclerDelegate interface {
 	// OnRingInstanceRegister is called while the lifecycler is registering the
 	// instance within the ring and should return the state and set of tokens to
-	// use for the instance itself.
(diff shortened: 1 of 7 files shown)


pracucci (commented):
In which case an implementation will return an error, and what behaviour do we expect from the lifecycler when an error is returned?

duricanikolic:
> In which case an implementation will return an error, and what behaviour do we expect from the lifecycler when an error is returned?

In the implementation of the spread minimizing approach there might be some (theoretical) errors, when there is so many tokens in the ring that it is impossible to build new ones and assign them to the new instance. If the token generator cannot generate new tokens for an instance, then the lifecycler should also not be able to register that instance, and hence should return the error.

PR #323, opened 2023-06-19
Introduce RingTokenGenerator configuration in LifecyclerConfig and BasicLifecyclerConfig

duricanikolic:
**What this PR does**:
This PR introduces a `RingTokenGenerator` configuration in `LifecyclerConfig` and `BasicLifecyclerConfig`. `RingTokenGenerator`, if set, specifies the `TokenGenerator` implementation that will be used for token generation. Default value is `nil`, meaning that `RandomTokenGenerator` is used.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/spread_minimizing_token_generator.go
+++ b/ring/spread_minimizing_token_generator.go
@@ -11,8 +11,6 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 
-	"github.com/grafana/dskit/flagext"
-
 	"golang.org/x/exp/slices"
 )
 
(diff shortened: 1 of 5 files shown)


ring/lifecycler.go
```diff
@@ -41,11 +41,12 @@ type LifecyclerConfig struct {
...
-	TokensFilePath           string        `yaml:"tokens_file_path"`
-	Zone                     string        `yaml:"availability_zone"`
-	UnregisterOnShutdown     bool          `yaml:"unregister_on_shutdown" category:"advanced"`
-	ReadinessCheckRingHealth bool          `yaml:"readiness_check_ring_health" category:"advanced"`
+	FinalSleep               time.Duration          `yaml:"final_sleep" category:"advanced"`
+	TokensFilePath           string                 `yaml:"tokens_file_path"`
+	Zone                     string                 `yaml:"availability_zone"`
+	UnregisterOnShutdown     bool                   `yaml:"unregister_on_shutdown" category:"advanced"`
+	ReadinessCheckRingHealth bool                   `yaml:"readiness_check_ring_health" category:"advanced"`
+	SpreadMinimizingZones    flagext.StringSliceCSV `yaml:"spread_minimizing_zones" category:"advanced"`
```
pracucci, line 49: This should be `experimental` instead of `advanced`.

ring/spread_minimizing_token_generator.go
```diff
@@ -42,32 +45,36 @@ var (
 )
 
 type SpreadMinimizingConfig struct {
```
pracucci, line 47: Do we need this config at all? I would just pass everything to `NewSpreadMinimizingTokenGenerator()` for now. In practice it doesn't change the number of parameters to pass to `NewSpreadMinimizingTokenGenerator()` given this config struct has only 1 field.

pracucci (approved):
LGTM!

ring/spread_minimizing_token_generator.go
duricanikolic: You are right, we don't need this config.
I have removed it, and used a slice of zones instead.

PR #325, opened 2023-06-22
Make ring.CountTokens() take into account instances zones

duricanikolic:
**What this PR does**:
This PR fixes a bug that was present in the calculation of instance ownership. `ring.CountTokens()`.
Namely, the previous implementation did not take into account the zone different instances belong to, and was calculating, for each instance, the sum of distances between its token and the first token preceding the former. That is wrong, because only the tokens belonging to the instances of the same zone should be taken into account.

For example, suppose we have the following tokens in 2 zones belonging to 2 instances:
- `instance-zone-a-1` with tokens `1000`, `20000`, `300000` 
- `instance-zone-b-1` with tokens `1001`, `20001`, `300001`

It is clear that the ownership of both instances is $2^{32}$, since they belong to different zones. Nevertheless, the previous implementation would return ownership $2^{32}-3$ for `instance-zone-a-1` and $3$ for `instance-zone-b-1`, which is wrong.

**Checklist**
- [x] Tests updated


pracucci (approved):
The change to `CountTokens()` makes sense to me. However, after this change, the % computed by `ringPageHandler.handle()` (the called of `CountTokens()`) will change because it will be the % of tokens owned within the zone of the instance. It's fine to me, but I think this deserves a `[CHANGE]` changelog entry.

PR #326, opened 2023-06-23
Impose SpreadMinimizingTokenGenerator to register instances in ascending order

duricanikolic:
**What this PR does**:
This PR proposes a way of how to impose a Lifecycler to accept registering instances with ids in ascending order in case of spread minimizing token generation strategy.
It also imposes Lifecycler not to load or store tokens to a file in the case of the above mentioned strategy.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -102,6 +102,19 @@ func (cfg *LifecyclerConfig) RegisterFlagsWithPrefix(prefix string, f *flag.Flag
 	f.BoolVar(&cfg.EnableInet6, prefix+"enable-inet6", false, "Enable IPv6 support. Required to make use of IP addresses from IPv6 interfaces.")
 }
 
+// Validate checks the consistency of LifecyclerConfig, and fails if this cannot be achieved.
+func (cfg *LifecyclerConfig) Validate() error {
+	_, ok := cfg.RingTokenGenerator.(*SpreadMinimizingTokenGenerator)
+	if ok {
+		// If cfg.RingTokenGenerator is a SpreadMinimizingTokenGenerator, we must ensure that
+		// the tokens are not loaded from file.
+		if cfg.TokensFilePath != "" {
+			return errors.New("you can't configure the tokens file path when using the spread minimizing token strategy. Please set the tokens file path to an empty string")
+		}
+	}
+	return nil
+}
+
 /*
 Lifecycler is a Service that is responsible for publishing changes to a ring for a single instance.
 
@@ -156,19 +169,6 @@ type Lifecycler struct {
 	logger            log.Logger
 }
 
-// Validate checks the consistency of LifecyclerConfig, and fails if this cannot be achieved.
-func (cfg *LifecyclerConfig) Validate(logger log.Logger) error {
-	_, ok := cfg.RingTokenGenerator.(*SpreadMinimizingTokenGenerator)
-	if ok {
-		// If cfg.RingTokenGenerator is a SpreadMinimizingTokenGenerator, we must ensure that
-		// the tokens are not loaded from file.
-		warn := fmt.Sprintf("spread minimizing token strategy does not support loading and storing tokens to a file. Configuration %q set to %q will be ignored", "tokens-file-path", cfg.TokensFilePath)
-		level.Warn(logger).Log("warn", warn)
-		cfg.TokensFilePath = ""
-	}
-	return nil
-}
-
 // NewLifecycler creates new Lifecycler. It must be started via StartAsync.
 func NewLifecycler(cfg LifecyclerConfig, flushTransferer FlushTransferer, ringName, ringKey string, flushOnShutdown bool, logger log.Logger, reg prometheus.Registerer) (*Lifecycler, error) {
 	addr, err := GetInstanceAddr(cfg.Addr, cfg.InfNames, logger, cfg.EnableInet6)
@@ -200,7 +200,7 @@ func NewLifecycler(cfg LifecyclerConfig, flushTransferer FlushTransferer, ringNa
 	}
 
 	// We validate cfg before we create a Lifecycler.
-	err = cfg.Validate(logger)
+	err = cfg.Validate()
 	if err != nil {
 		return nil, err
 	}
@@ -782,19 +782,16 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
 func (i *Lifecycler) autoJoin(ctx context.Context, targetState InstanceState) error {
 	var ringDesc *Desc
 
-	err := i.KVStore.CAS(ctx, i.RingKey, func(in interface{}) (out interface{}, retry bool, err error) {
+	err := i.tokenGenerator.CanJoin(ringDesc.GetIngesters())
+	if err != nil {
+		level.Error(i.logger).Log("msg", "there was a problem while checking whether this instance could join the ring - will continue anyway", "err", err)
+	}
+
+	err = i.KVStore.CAS(ctx, i.RingKey, func(in interface{}) (out interface{}, retry bool, err error) {
 		if in == nil {
 			ringDesc = NewDesc()
 		} else {
 			ringDesc = in.(*Desc)
-		}
-
-		conditionalTokenGenerator, ok := i.tokenGenerator.(ConditionalTokenGenerator)
-		if ok {
-			conditionErr := conditionalTokenGenerator.CheckConditions(ringDesc.GetIngesters())
-			if conditionErr != nil {
-				return nil, true, conditionErr
-			}
 		}
 
 		// At this point, we should not have any tokens, and we should be in PENDING state.
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -779,17 +779,7 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
 	return true
 }
 
-func (i *Lifecycler) canJoin(ctx context.Context) error {
-	desc, err := i.KVStore.Get(ctx, i.RingKey)
-	if err != nil {
-		return fmt.Errorf("error getting the ring from the KV store: %s", err)
-	}
-
-	ringDesc, ok := desc.(*Desc)
-	if !ok || ringDesc == nil {
-		return fmt.Errorf("no ring returned from the KV store")
-	}
-
+func (i *Lifecycler) waitBeforeJoining(ctx context.Context) error {
 	retries := backoff.New(ctx, backoff.Config{
 		MinBackoff: 10 * time.Second,
 		MaxBackoff: 1 * time.Minute,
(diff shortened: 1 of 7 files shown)


ring/lifecycler.go
```diff
@@ -763,6 +782,14 @@ func (i *Lifecycler) autoJoin(ctx context.Context, targetState InstanceState) er
 			ringDesc = in.(*Desc)
 		}
 
+		conditionalTokenGenerator, ok := i.tokenGenerator.(ConditionalTokenGenerator)
```
pracucci, line 785: Some random thoughts:
1. When using memberlist, you don't have the guarantee that all previous instances registration has been propagated yet. Handling this case is complicated and possibly error prone. I'm OK not checking it, but you should add a comment about it, saying that we're aware of it and what we do here is just a best effort.
2. What happens if, for any reason, the previous instances are never registered? This will loop forever. We should probably have a timeout (e.g. if the condition is not honored within 5m, then move on and register anyway).
3. Each time this check fails, we return error to `CAS()` and the KVStore client will retry. However, specific implementations may or may not have a backup (e.g. could retry continuously in a loop, with no throttling) and also specific implementations may have a max number of retries after which the CAS() will just stop and fail. For example, in case of memberlist client, we retry up to 10 times and them we give up. As you can see, this behaviour is incompatible with the new logic you're introducing.

For the reasons above, I believe you should move the logic out of `CAS()`. Before the `CAS()` you could have a loop (with backoff and max retries) to wait until the condition is satified. Then, once the condition is satisfied, we call CAS() as of today.

ring/token_generator.go
```diff
@@ -6,6 +6,14 @@ import (
 	"time"
 )
 
+type ConditionalTokenGenerator interface {
+	TokenGenerator
+
+	// CheckConditions checks whether the given set of instances satisfies a condition provided by a specific
+	// implementation, and fails if the latter is not satisfied.
+	CheckConditions(instances map[string]InstanceDesc) error
```
pracucci, line 14: I would be more specific with the name. What's about `CanJoin()`?

ring/token_generator.go
```diff
@@ -6,6 +6,14 @@ import (
 	"time"
 )
 
+type ConditionalTokenGenerator interface {
```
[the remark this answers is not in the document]
duricanikolic: Removed

ring/spread_minimizing_token_generator.go
```diff
@@ -303,3 +332,18 @@ func (t *SpreadMinimizingTokenGenerator) generateTokensByInstanceID() map[int]To
...
+func (t *SpreadMinimizingTokenGenerator) CheckConditions(instances map[string]InstanceDesc) error {
+	prevInstance, err := previousInstance(t.instance)
+	if err != nil {
+		if errors.Is(err, errorNoPreviousInstance(t.instance)) {
+			return nil
+		}
+		return err
+	}
+	instanceDesc, ok := instances[prevInstance]
+	if !ok || instanceDesc.State == PENDING {
```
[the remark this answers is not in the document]
duricanikolic: Correct. I've modified the implementation accordingly.

ring/lifecycler.go
duricanikolic: I have added 2 flags: `canJoinEnabled` and `canJoinTimeout` that will be passed to `NewSpreadMinimizingTokenGenerator()` directly from `mimir`.

I have placed the loop checking the previous instances inside `SpreadMinimizingTokenGenerator.CanJoin()` because this way both config flags remain on `SpreadMinimizingTokenGenerator` and other `dskit` sources will not need to know about their existence.

ring/lifecycler.go
```diff
@@ -763,7 +782,12 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
 func (i *Lifecycler) autoJoin(ctx context.Context, targetState InstanceState) error {
 	var ringDesc *Desc
 
-	err := i.KVStore.CAS(ctx, i.RingKey, func(in interface{}) (out interface{}, retry bool, err error) {
+	err := i.tokenGenerator.CanJoin(ringDesc.GetIngesters())
```
pracucci, line 785: You have to pass `ctx` to the function. Then, inside the function, you have to check the context and return `ctx.Err()` if the context is canceled.

ring/spread_minimizing_token_generator.go
```diff
@@ -303,3 +337,27 @@ func (t *SpreadMinimizingTokenGenerator) generateTokensByInstanceID() map[int]To
 
 	return tokensByInstanceID
 }
+
+func (t *SpreadMinimizingTokenGenerator) CanJoin(instances map[string]InstanceDesc) error {
```
pracucci, line 341: I can't understand how this could work. The input `instances` is just a one time snapshot of the ring. You need to move the `for` loop outside (back into `Lifecycler`) and call it each time with the updated map of `instances`.

You also need to handle the `ctx` cancellation in the for loop and you need some throttling. In the for loop below you're looping with no throttling, while we should check the ring every X time.

In dskit there's `backoff`. Use it: it will handle throttling, backoff and context cancellation for you.

ring/lifecycler.go
```diff
@@ -759,11 +779,43 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
 	return true
 }
 
+func (i *Lifecycler) canJoin(ctx context.Context) error {
+	desc, err := i.KVStore.Get(ctx, i.RingKey)
```
pracucci, line 783: This needs to be called before each call to `CanJoin()`. Otherwide `desc` will never change and then how can the state change when you call `CanJoin()` in the `for` loop?

duricanikolic: Hi @pracucci,
This is probably due to my misunderstanding of the result returned by `i.KVStore.Get()`. My interpretation was:
- `desc` of type `*Desc` is returned by `i.KVStore.Get()`
- inside the loop we call `desc.GetIngesters()`, which gives us the latest map of ingesters
- during the waiting times, somebody calls `desc.AddIngester(newIngesterDesc)` and this way the next call to `desc.GetIngesters()` will include `newIngesterDesc` as well.

I would like to understand what is wrong with this interpretation. Anyway, I moved the call to `i.KVStore.Get()` inside the loop.

ring/lifecycler.go
```diff
@@ -759,11 +779,42 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
 	return true
 }
 
+func (i *Lifecycler) waitBeforeJoining(ctx context.Context) error {
+	retries := backoff.New(ctx, backoff.Config{
+		MinBackoff: 10 * time.Second,
+		MaxBackoff: 1 * time.Minute,
+		MaxRetries: 5,
```
pracucci, line 786: We agreed that timeout should be based on "time" and not "max retries", so MaxRetries should be set to 0. Moreover I think we can be a more aggressive on the speed of retries.

```suggestion
		MinBackoff: 1 * time.Second,
		MaxBackoff: 1 * time.Second,
		MaxRetries: 0,
```

ring/lifecycler.go
```diff
@@ -759,11 +779,42 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
 	return true
 }
 
+func (i *Lifecycler) waitBeforeJoining(ctx context.Context) error {
+	retries := backoff.New(ctx, backoff.Config{
```
pracucci, line 783: Since I suggested to set MaxRetries = 0 you should wrap context with a `context.WithTimeout()`.

ring/lifecycler.go
```diff
@@ -759,11 +779,42 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
...
+		MinBackoff: 10 * time.Second,
+		MaxBackoff: 1 * time.Minute,
+		MaxRetries: 5,
+	})
+
+	for retries.Ongoing() {
+		desc, err := i.KVStore.Get(ctx, i.RingKey)
+		if err != nil {
+			return fmt.Errorf("error getting the ring from the KV store: %s", err)
```
pracucci, line 792: You should just retry in this case, so `continue`.

ring/lifecycler.go
```diff
@@ -759,11 +779,42 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
...
+		}
+
+		ringDesc, ok := desc.(*Desc)
+		if !ok || ringDesc == nil {
+			return fmt.Errorf("no ring returned from the KV store")
```
pracucci, line 797: You should just retry in this case, so `continue`.

ring/lifecycler.go
```diff
@@ -759,11 +779,42 @@ func (i *Lifecycler) compareTokens(fromRing Tokens) bool {
 	return true
 }
 
+func (i *Lifecycler) waitBeforeJoining(ctx context.Context) error {
```
pracucci, line 782: I suggest to add a log (info level) at the beginning of the waiting period and one at the end.

pracucci, line 782: I think we should skip this entire logic if a feature flag is not enabled.

ring/lifecycler_test.go
```diff
@@ -1268,6 +1292,173 @@ func TestJoinInJoiningState(t *testing.T) {
...
+		return tokenGenerator
+	}
+
+	tests := map[string]struct {
+		targetInstanceID string
+		tokenGenerator   TokenGenerator
+		timeoutRequired  bool
+		ctx              context.Context
+	}{
+		"RandomTokenGenerator never returns errors": {
```
pracucci, line 1341: [nit] For all these test cases, the test doesn't check whether returns error or not (it never returns errors), so I think we can improve a bit the description of the test cases. What we're really testing is whether it returns immediately or after the timeout.

pracucci (approved):
LGTM (modulo few nits)

ring/lifecycler_test.go
duricanikolic: I forgot to add a test case in which the context gets cancelled, that I have now corrected. 
If you are still not happy with it, I will modify it in a follow-up PR.

ring/lifecycler_test.go
```diff
@@ -1268,6 +1292,173 @@ func TestJoinInJoiningState(t *testing.T) {
...
+		// Check that the lifecycler was able to join
+		test.Poll(t, 1000*time.Millisecond, true, func() interface{} {
+			d, err := r.KVClient.Get(context.Background(), ringKey)
+			require.NoError(t, err)
+
+			desc, ok := d.(*Desc)
+			if !ok {
+				return false
+			}
+			_, ok = desc.Ingesters[targetInstanceID]
```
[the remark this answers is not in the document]
duricanikolic: I have added additional conditions that check both the state and the tokens that the new registered ingesters have.
The test has been re-organized. If you are not happy with it, I will change it in a follow-up PR.

PR #327, opened 2023-06-23
memberlist metrics

### LINKED ISSUE #305: Potential memberlist metrics issues.
Hey all,

We are using Tempo, and trying to make use of the Tempo mixin Grafana dashboards... https://github.com/grafana/tempo/tree/main/operations/tempo-mixin

However I've noticed that some of the metrics that the dashboard expects that come from dskit/kv/memberlist/metrics might not be coming out correctly.

1. `tempo_memberlist_client_kv_store_count`
This chart https://github.com/grafana/tempo/blob/main/operations/tempo-mixin/dashboards/tempo-operational.json#L4585
is looking for the metric `tempo_memberlist_client_kv_store_count`. I couldn't find it exposed on any of the `/metrics` endpoints I looked at.

I went looking for where this metric is configured and found this: https://github.com/grafana/dskit/blob/main/kv/memberlist/metrics.go#L123

```
	m.storeValuesDesc = prometheus.NewDesc(
		prometheus.BuildFQName(m.cfg.MetricsNamespace, subsystem, "kv_store_count"), // gauge
		"Number of values in KV Store",
		nil, nil)
```

I'm not super familiar with Prometheus, but the rest of the metric definitions in that file have `NewGaugeVec` or `NewCounterVec`. From what I can tell `NewDesc` is used for metric metadata?
But the metric name "kv_store_count" sounds like it should be a numerical value?

2. `gauge_memberlist_health_score`
This chart https://github.com/grafana/tempo/blob/main/operations/tempo-mixin/dashboards/tempo-operational.json#LL4319C1-L4319C1 is looking for a metric `gauge_memberlist_health_score`. Again I couldn't find it exposed in the `/metrics` endpoint.

I went looking for where this metric is configured and I couldn't find it!

But I did find this: https://github.com/grafana/dskit/blob/main/kv/memberlist/metrics.go#L154

```
	m.memberlistHealthScore = promauto.With(m.registerer).NewGaugeFunc(prometheus.GaugeOpts{
		Namespace: m.cfg.MetricsNamespace,
		Subsystem: subsystem,
		Name:      "cluster_node_health_score",
		Help:      "Health score of this cluster. Lower value is better. 0 = healthy",
	}, func() float64 {
		// m.memberlist is not set before Starting state
		if m.State() == services.Running || m.State() == services.Stopping {
			return float64(m.memberlist.GetHealthScore())
		}
		return 0
	})
```

The Go struct member is `memberlistHealthScore` similar to the metric I'm looking for, but the actual metric name is `cluster_node_health_score`.

Could this be a metric that has been changed in code, but the metric name hasn't been updated yet?

Also I think the Tempo mixin might be wrong... `gauge_memberlist_health_score` seems like it should be `tempo_memberlist_client_cluster_node_health_score` or something?

pstibrany:
**What this PR does**:

This PR fixes registration of `memberlist_client_kv_store_count` metric (this broke in #22) and also disables expiration of internal memberlist-library metrics.

This PR also removes `MetricsRegisterer` field from `memberlist.KVConfig` struct in favor of registrerer passed to `memberlist.NewKV` function.

Fixes #305

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@
 * [CHANGE] Multierror: Implement `Is(error) bool`. This allows to use `multierror.MultiError` with `errors.Is()`. `MultiError` will in turn call `errors.Is()` on every error value. #254
 * [CHANGE] Cache: Remove the `context.Context` argument from the `Cache.Store` method and rename the method to `Cache.StoreAsync`. #273
 * [CHANGE] ring: make it harder to leak contexts when using `DoUntilQuorum`. #319
-* [CHANGE] memberlist: MetricsRegisterer field has been removed in favor of registrerer passed via into `NewKV` function.
+* [CHANGE] memberlist: `MetricsRegisterer` field has been removed from `memberlist.KVConfig` in favor of registrerer passed via into `NewKV` function. #327
 * [FEATURE] Cache: Add support for configuring a Redis cache backend. #268 #271 #276
 * [FEATURE] Add support for waiting on the rate limiter using the new `WaitN` method. #279
 * [ENHANCEMENT] Add configuration to customize backoff for the gRPC clients.
@@ -139,3 +139,4 @@
 * [BUGFIX] Ring: prevent iterating the whole ring when using `ExcludedZones`. #285
 * [BUGFIX] grpcclient: fix missing `.` in flag name for initial connection window size flag. #314
 * [BUGFIX] ring.Lifecycler: Handle when previous ring state is leaving and the number of tokens has changed. #79
+* [BUGFIX] memberlist metrics: fix registration of `memberlist_client_kv_store_count` metric and disable expiration of metrics exposed by memberlist library. #327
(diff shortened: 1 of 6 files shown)


pracucci (approved):
LGTM. I checked if in Mimir we passed a different registerer (e.g. maybe with different const labels, or so...) between `NewKV()` and `KVConfig.MetricsRegisterer` but looks like we don't. I suggest you to do a quick check to Loki and Tempo too before merging tho.

pstibrany:
> LGTM. I checked if in Mimir we passed a different registerer (e.g. maybe with different const labels, or so...) between `NewKV()` and `KVConfig.MetricsRegisterer` but looks like we don't. I suggest you to do a quick check to Loki and Tempo too before merging tho.

Loki uses the same `reg` for both: [`NewKVInitService`](https://github.com/grafana/loki/blob/d4cfbaac4f04507cb92776d1b408d5a51f290bbe/pkg/loki/modules.go#L1077) and [`MetricsRegisterer`](https://github.com/grafana/loki/blob/d4cfbaac4f04507cb92776d1b408d5a51f290bbe/pkg/loki/modules.go#L1062).

pstibrany:
> LGTM. I checked if in Mimir we passed a different registerer (e.g. maybe with different const labels, or so...) between `NewKV()` and `KVConfig.MetricsRegisterer` but looks like we don't. I suggest you to do a quick check to Loki and Tempo too before merging tho.

Same is true in Tempo: [`NewKVInitService`](https://github.com/grafana/tempo/blob/ed0665dad15d6fd8b1e7e1b61e11a3543e85d65b/cmd/tempo/app/modules.go#L394) and [`MetricsRegisterer`](https://github.com/grafana/tempo/blob/ed0665dad15d6fd8b1e7e1b61e11a3543e85d65b/cmd/tempo/app/modules.go#L389) both use the same `reg` (`prometheus.DefaultRegisterer`).

PR #329, opened 2023-06-24
Add a basic help target to the Makefile

jhalterman:
**What this PR does**:

Adds a basic help target to the makefile.

Screenshot 2023-06-23 at 7 37 53 PM

Fixes #<issue number>

**Checklist**
- [ ] Tests updated


replay (approved):
nice idea with the `help` target!

pstibrany (approved):
One would hope there's support for this already in `make`, so that we don't need to do hacks like this. But it's a nice trick.

PR #330, opened 2023-06-27
ring: add support for hedging to `DoUntilQuorum` when request minimisation is enabled

charleskorn:
**What this PR does**:

This PR introduces support for a form of request hedging to `DoUntilQuorum`.

This is designed to allow consumers of `DoUntilQuorum` to mitigate against slow or unresponsive instances when request minimisation is enabled.

This is best illustrated with an example. Say we have instances in four zones, tolerate two unavailable zones, enable request minimisation and set a hedging delay of 1 second. `DoUntilQuorum` will behave as follows:

* immediately calls `f` for all instances in two of the zones
* if `DoUntilQuorum`'s termination conditions are not met after a second has elapsed, it calls `f` for all instances in a third zone
* if `DoUntilQuorum`'s termination conditions are still not met after another second, it calls `f` for all instances in the remaining zone

(I'm not sure if calling this hedging is confusing given it doesn't retry requests to the same endpoints - open to feedback on the name.)

(none)

**Checklist**
- [x] Tests updated


pracucci (approved):
Nice work! I'm surprised by how small the code change is (symptom of a good design ;))

PR #332, opened 2023-06-30
Set a default gRPC client connection timeout of 5s

charleskorn:
**What this PR does**:

This PR sets a default gRPC client connection timeout of 5s.

The previous default value of 0 meant we used the default value from the gRPC client library, which is currently 20s.

Note that explicitly setting the default client connection timeout like this also means that our default values for `connect-backoff-base-delay` and `connect-backoff-max-delay` will now be used everywhere by default as well. The defaults we have here (1s and 5s, respectively) are not the same as the gRPC client defaults (1s and 120s), but given our defaults seem reasonable for our use cases, I'm not too concerned about this.

(none)

**Checklist**
- [n/a] Tests updated


charleskorn:
I'm going to leave this PR as a draft until we've tested the impact of this more widely.

pracucci (approved):
I agree with this change. If the connection creation takes longer than 5s, you already have serious issues and I can't see a real difference between 5s and 20s (it's very high in both cases).

PR #339, opened 2023-07-25
Fix BufferedLogger's threadsafeBuffer locking

pracucci:
**What this PR does**:
I was reviewing https://github.com/grafana/mimir/pull/5506 post merge and I noticed the locking done by `threadsafeBuffer` is wrong: we can't safely concurrently read from `bytes.Buffer`. The lock must be an exclusive one.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/log/buffered.go
+++ b/log/buffered.go
@@ -113,31 +113,30 @@ func NewBufferedLogger(w io.Writer, cap uint32, opts ...BufferedLoggerOption) *B
 
 // threadsafeBuffer wraps the non-threadsafe bytes.Buffer.
 type threadsafeBuffer struct {
-	sync.Mutex
-
+	mx  sync.Mutex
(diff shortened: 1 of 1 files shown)


pstibrany (approved):
While at it, it would be nice to not use anonymous field for lock -- that way Lock/Unlock methods are callable directly on `threadsafeBuffer`, which is very weird.

pracucci:
> While at it, it would be nice to not use anonymous field for lock -- that way Lock/Unlock methods are callable directly on `threadsafeBuffer`, which is very weird.

Agree. Done in https://github.com/grafana/dskit/pull/339/commits/d4423299b3d02b9e072f47244ebcd6f941b587fa

56quarters (approved):

PR #345, opened 2023-08-04
[Chore]Opentracing to Opentelemetry migration with bridge

ying-jeanne:
**What this PR does**:
Simplified OpenTracing to OpenTelemetry Migration with Bridge

Note: this is a breaking change, since opentelemetry 1.14.0 require go 1.19

Our goal is to gradually shift from OpenTracing to OpenTelemetry while ensuring compatibility. We're using a bridge solution to make this transition smooth. Once our database team completes the OpenTelemetry migration (including underlying libraries), we'll phase out the bridge.

This approach allows us to migrate step by step, minimising disruptions. Performance tests show the bridge would increase cpu/memory consumption by more than 50%. 

Given our complex dependencies, the bridge remains useful during migration. We should start with the write path to minimize impact.

See the bridge benchmark [here](https://github.com/ying-jeanne/bridge_benchmark) and the load impact results to be added.

Something good to know, jaeger-debug-id header is not supported by opentelemetry, I have seen some e2e test failed because of this header, see [thread](https://github.com/open-telemetry/opentelemetry-specification/issues/231)

Fixes #<issue number>

**Checklist**
- [x] Tests updated


tracing/tracing.go
```diff
@@ -6,51 +6,185 @@ package tracing
...
+	}
+	if cfg.samplerRemoteURL == "" && cfg.host == "" && cfg.collectorURL == "" {
+		return nil, ErrBlankTraceConfiguration
+	}
+	return cfg.initJaegerTracerProvider(serviceName)
+}
 
-	// put the metricsFactory earlier so provided options can override it
-	opts := append([]jaegercfg.Option{jaegercfg.Metrics(metricsFactory)}, options...)
+// Parse Jeager tags from env var JAEGER_TAGS, example of TAGs format: key1=value1,key2=value2
```
bboreham, line 54: s/Jeager/Jaeger/
Probably worth mentioning that `${FOO}` sequences are expanded as environment variables.

tracing/tracing.go
```diff
@@ -6,51 +6,185 @@ package tracing
...
+			}
+			res = append(res, attribute.String(k, v))
+		} else if p != "" {
+			return nil, errors.Errorf("invalid tag %q, expected key=value", p)
+		}
+	}
+	return res, nil
+}
+
+type Config struct {
```
bboreham, line 78: Could this be non-exported?  Seems it is only used within the package.

tracing/tracing.go
```diff
@@ -6,51 +6,185 @@ package tracing
...
+}
 
-	// put the metricsFactory earlier so provided options can override it
-	opts := append([]jaegercfg.Option{jaegercfg.Metrics(metricsFactory)}, options...)
+// Parse Jeager tags from env var JAEGER_TAGS, example of TAGs format: key1=value1,key2=value2
+func parseAttributes(sTags string) ([]attribute.KeyValue, error) {
+	pairs := strings.Split(sTags, ",")
+	res := []attribute.KeyValue{}
+	for _, p := range pairs {
+		kv := strings.SplitN(p, "=", 2)
```
bboreham, line 59: I think this could be simplified using `strings.Cut`.

bboreham (commented):
A bit more explanation of what is going on at high level would make this code easier to follow.

httpgrpc/server/server.go
```diff
@@ -154,7 +154,7 @@ func NewClient(address string) (*Client, error) {
 
 // HTTPRequest wraps an ordinary HTTPRequest with a gRPC one
 func HTTPRequest(r *http.Request) (*httpgrpc.HTTPRequest, error) {
-	body, err := ioutil.ReadAll(r.Body)
+	body, err := io.ReadAll(r.Body)
```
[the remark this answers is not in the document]
ying-jeanne: this is needed for make linter happy, since ioutil.ReadAll is deprecated in go 1.19

CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## Changelog
-
+* [CHANGE] Upgraded Tracer: Transitioned from OpenTracing to OpenTelemetry using a seamless bridge integration. #345
```
dimitarvdimitrov, line 2:
```suggestion

* [CHANGE] Upgraded Tracer: Transitioned from OpenTracing to OpenTelemetry using a seamless bridge integration. #345
```

tracing/tracing.go
```diff
@@ -6,51 +6,185 @@ package tracing
...
+	envSamplerParam           = "JAEGER_SAMPLER_PARAM"
+	envEndpoint               = "JAEGER_ENDPOINT"
+	envJaegerAgentHost        = "JAEGER_AGENT_HOST"
+	envJaegerAgentPort        = "JAEGER_AGENT_PORT"
+	envSamplerType            = "JAEGER_SAMPLER_TYPE"
+	envSamplerRemoteURL       = "JAEGER_SAMPLING_ENDPOINT"
+	DefaultSamplingServerPort = 5778
+)
+
+func NewFromEnv(serviceName string) (*tracesdk.TracerProvider, error) {
```
dimitarvdimitrov, line 43: Can you add a godoc comment on this function? I think all exported members should get a godoc in a library like dskit

tracing/tracing.go
dimitarvdimitrov, line 54: nitpick, but the comment should start with `parseAttributes`

tracing/tracing.go
```diff
@@ -6,51 +6,185 @@ package tracing
...
+	}
+	return cfg.initJaegerTracerProvider(serviceName)
+}
 
-	// put the metricsFactory earlier so provided options can override it
-	opts := append([]jaegercfg.Option{jaegercfg.Metrics(metricsFactory)}, options...)
+// Parse Jeager tags from env var JAEGER_TAGS, example of TAGs format: key1=value1,key2=value2
+func parseAttributes(sTags string) ([]attribute.KeyValue, error) {
+	pairs := strings.Split(sTags, ",")
+	res := []attribute.KeyValue{}
```
dimitarvdimitrov, line 57: This can be preallocated to the number of pairs

tracing/tracing.go
```diff
@@ -6,51 +6,185 @@ package tracing
...
+	host             string
+	collectorURL     string
+	port             int
+	samplerType      string
+	samplerRemoteURL string
+	samplerParam     float64
+	customAttributes []attribute.KeyValue
+}
+
+func parseTracingConfig() (*Config, error) {
```
dimitarvdimitrov, line 88: a nitpick: this doesn't need to return a pointer. Could save an allocation 🤷

tracing/tracing.go
```diff
@@ -6,51 +6,185 @@ package tracing
...
+		return nil, errors.Errorf("unknown sampler type %q", cfg.samplerType)
 	}
+	tp := tracesdk.NewTracerProvider(
+		tracesdk.WithBatcher(exp),
+		tracesdk.WithResource(res),
+		tracesdk.WithSampler(sampler),
+	)
+
+	otel.SetTracerProvider(tp)
+	propagators := []propagation.TextMapPropagator{}
```
dimitarvdimitrov, line 175: nitpick: this can be preallocated or you can add all the propagators in the slice declaration 

```go
	propagators := []propagation.TextMapPropagator{
    propagation.TraceContext{}, propagation.Baggage{},
		jaegerpropagator.Jaeger{},
	}
```

tracing/tracing.go
dimitarvdimitrov, line 43: We spoke about this offline, but also leaving a comment here. This PR changes the public interface for NewFromEnv. From a quick look it seems that the removed variadic options weren't ever used, but the return value that used to be io.Closer is no longer that. I think we should return an implementation of io.Closer which is backed by `*tracesdk.TracerProvider`

dimitarvdimitrov (commented):
Mostly nitpicks. I think the most important one is for the changed interface of NewFromEnv

tracing/tracing.go
ying-jeanne: thanks for the review, addressed in the latest commit.

jpkrohling:
@sh0rez, you are generally interested in the performance differences, perhaps you it would be worth tracking this data point somewhere?

tracing/tracing.go
```diff
@@ -6,51 +6,194 @@ package tracing
...
+	if err != nil {
+		return nil, errors.Wrap(err, "could not load jaeger tracer configuration")
+	}
+	if cfg.samplerRemoteURL == "" && cfg.host == "" && cfg.collectorURL == "" {
+		return nil, ErrBlankTraceConfiguration
+	}
+	return cfg.initJaegerTracerProvider(serviceName)
+}
+
+// Parse Jaeger tags from env var JAEGER_TAGS, example of TAGs format: key1=value1,key2=${value2:value3} where value2 is an env var
```
jpkrohling, line 55: > key2=${value2:value3}

OTel supports OTEL_RESOURCE_ATTRIBUTES, but I'm not sure it supports this notation (value, default value). Is this why you have this custom handling?

tracing/tracing.go
```diff
@@ -6,51 +6,194 @@ package tracing
...
-		return nil, ErrBlankTraceConfiguration
+	res, err := NewResource(serviceName, cfg.customAttributes)
+	if err != nil {
+		return nil, err
+	}
+
+	// Configure sampling strategy
+	sampler := tracesdk.AlwaysSample()
+	if cfg.samplerType == "const" || cfg.samplerType == "probabilistic" {
+		sampler = tracesdk.TraceIDRatioBased(cfg.samplerParam)
```
jpkrohling, line 163: Idea: how about adding this as a resource attribute? I don't think we store it by default in OTel, but we did have it in the Jaeger client.

tracing/tracing.go
```diff
@@ -6,51 +6,194 @@ package tracing
...
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator([]propagation.TextMapPropagator{
+		// w3c Propagator is the default propagator for opentelemetry
+		propagation.TraceContext{}, propagation.Baggage{},
+		// jaeger Propagator is for opentracing backwards compatibility
+		jaegerpropagator.Jaeger{},
+	}...))
+
+	// OpenTracing <=> OpenTelemetry bridge
+	// The tracer name is empty so that the bridge uses the default tracer.
+	otTracer, _ := bridge.NewTracerPair(tp.Tracer(""))
```
jpkrohling, line 186: Does it *have* to be empty? If possible at all, I'd use this repo's name in the tracer.

tracing/tracing.go
```diff
@@ -6,51 +6,194 @@ package tracing
...
+	}
+	if cfg.samplerRemoteURL == "" && cfg.host == "" && cfg.collectorURL == "" {
+		return nil, ErrBlankTraceConfiguration
+	}
+	return cfg.initJaegerTracerProvider(serviceName)
+}
+
+// Parse Jaeger tags from env var JAEGER_TAGS, example of TAGs format: key1=value1,key2=${value2:value3} where value2 is an env var
+// and value3 is the default value, which is optional.
+func parseAttributes(sTags string) ([]attribute.KeyValue, error) {
```
charleskorn, line 57: If we do need to keep this method (see @jpkrohling's comment above), would be good to add some tests for it.

tracing/tracing.go
```diff
@@ -6,51 +6,194 @@ package tracing
...
+	collectorURL     string
+	port             int
+	samplerType      string
+	samplerRemoteURL string
+	samplerParam     float64
+	customAttributes []attribute.KeyValue
+}
+
+// This function facilitates initialization that is compatible with Jaeger's InitGlobalTracer method.
+func parseTracingConfig() (*config, error) {
```
charleskorn, line 90: Rather than doing all this parsing ourselves, could we call `jaegercfg.FromEnv()` and translate the returned config to what the OTel libraries require?

charleskorn (commented):
Seems like there are a couple of unrelated changes in this PR:

* removal of Go 1.18 as a supported Go version and linting fixes
* upgrade of `github.com/stretchr/testify`

Could they be moved to a separate PR to help keep this one focused?

ying-jeanne:
> Seems like there are a couple of unrelated changes in this PR:
> 
> * removal of Go 1.18 as a supported Go version and linting fixes
> * upgrade of `github.com/stretchr/testify`
> 
> Could they be moved to a separate PR to help keep this one focused?

It is related since opentelemetry 1.14.0 require go 1.19, but I can move it to a separate PR if it is preferred.

tracing/tracing.go
ying-jeanne: this is to keep backward compatible for JAEGER_TAGS settings, we don't use environment variable in Mimir, but other team who use the lib might needed it.

tracing/tracing.go
```diff
@@ -6,51 +6,204 @@ package tracing
...
+	}
+
+	if e := os.Getenv(envSamplerRemoteURL); e != "" {
+		cfg.samplerRemoteURL = e
+	} else if e := os.Getenv(envJaegerAgentHost); e != "" {
+		// Fallback if we know the agent host - try the sampling endpoint there
+		cfg.samplerRemoteURL = fmt.Sprintf("http://%s:%d/sampling", e, DefaultSamplingServerPort)
+	}
+
+	cfg.customAttributes, err = parseAttributes(os.Getenv("JAEGER_TAGS"))
```
pracucci, line 143: [nit] For consistency, I would define a const for `JAEGER_TAGS` too.

tracing/tracing.go
```diff
@@ -6,51 +6,204 @@ package tracing
...
+	customAttributes []attribute.KeyValue
+}
+
+// parseTracingConfig facilitates initialization that is compatible with Jaeger's InitGlobalTracer method.
+func parseTracingConfig() (config, error) {
+	cfg := config{}
+	var err error
+
+	// We first parse the collector endpoint, if it is not exist then we parse the agent endpoint, if it is not exist, we default to localhost:6831
+	if e := os.Getenv(envEndpoint); e != "" {
```
pracucci, line 104: I was expecting `parseTracingConfig()` to be close to `samplerConfigFromEnv()` from `github.com/uber/jaeger-client-go/config/config_env.go`, but in this function the sampling server URL setup logic looks more complicated. Can you just help me understand why we need a different logic here, please?

tracing/tracing.go
```diff
@@ -6,51 +6,204 @@ package tracing
...
-	if cfg.Sampler.SamplingServerURL == "" && cfg.Reporter.LocalAgentHostPort == "" && cfg.Reporter.CollectorEndpoint == "" {
+	if cfg.samplerRemoteURL == "" && cfg.host == "" && cfg.collectorURL == "" {
 		return nil, ErrBlankTraceConfiguration
 	}
+	return cfg.initJaegerTracerProvider(serviceName)
+}
+
+// parseAttributes Parse Jaeger tags from env var JAEGER_TAGS, example of TAGs format: key1=value1,key2=${value2:value3} where value2 is an env var
+// and value3 is the default value, which is optional.
+func parseAttributes(sTags string) ([]attribute.KeyValue, error) {
```
pracucci, line 63: [nit] I would name it for what it does.

```suggestion
func parseJaegerTags(sTags string) ([]attribute.KeyValue, error) {
```

pracucci (commented):
I'm not very familiar with tracing and OTel, so take my review with a grain of salt. I haven't noticed any blocker. I left few nits that should be easy to address. I just would like to better understand the different AGENT_HOST parsing logic, and then I'm good.

tracing/tracing.go
ying-jeanne: @gouthamve Could I set Tracer name to dskit, and do tp.GetTracer("") to have the correct tracer cfg with otel sdk?

tracing/tracing.go
```diff
@@ -6,78 +6,260 @@ package tracing
...
+		port := envJaegerDefaultUDPSpanServerPort
+		if e := os.Getenv(envJaegerAgentPort); e != "" {
+			port = e
+			useEnv = true
+		}
+
+		if useEnv {
+			cfg.agentHost = host
+			cfg.agentPort = port
+			cfg.agentHostPort = fmt.Sprintf("%s:%s", host, port)
```
zalegrala, line 133: Consider using `net.JoinHostPort` here for IPv6 support, in case a hostname is not used.

tracing/tracing.go
ying-jeanne: Hello @jpkrohling, thanks for the input and would like to engage in a discussion to gain a deeper understanding of your perspective. However you are off this week, I reached out to @gouthamve instead. Upon conversing with him, we concluded that the current situation lacks an optimal solution. Setting it as a resource attribute would lead to a static value if not restart, out of sync with any updates to the remote sampling setting.

Given these circumstances, it appears prudent to put this matter on hold for now. I suggest revisiting it when the community arrives at a more suitable solution. Your insights remain valuable, and I look forward to further discussions.

PR #347, opened 2023-08-04
server: correctly route PATCH requests when using RouteHTTPToGRPC

kvrhdn:
**What this PR does**:

Tempo relies on the `RouteHTTPToGRPC` option in `server` to handle HTTP and gRPC requests from a single endpoint. This options configures a multiplexer ([soheilhy/cmux](https://github.com/soheilhy/cmux)) to match and route HTTP and gRPC requests to their respective listener.

For HTTP matching it uses `cmux.HTTP1Fast` which matches the method of the request with a predefined set of methods. This change adds `PATCH` to the list to ensure PATCH requests are also sent to the HTTP listener.

More details here: https://github.com/grafana/tempo/issues/2756
For context, this configuration option was added a couple of months ago to support Tempo: https://github.com/weaveworks/common/pull/288

Fixes https://github.com/grafana/tempo/issues/2756

**Checklist**
- [ ] Tests updated

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -259,7 +259,7 @@ func newServer(cfg Config, metrics *Metrics) (*Server, error) {
 	if cfg.RouteHTTPToGRPC {
 		grpchttpmux = cmux.New(httpListener)
 
-		httpListener = grpchttpmux.Match(cmux.HTTP1Fast())
+		httpListener = grpchttpmux.Match(cmux.HTTP1Fast("PATCH"))
 		grpcOnHTTPListener = grpchttpmux.Match(cmux.HTTP2())
 	}
 


kvrhdn:
FYI draft PR to show this change works on Tempo: all tests succeed except vendor-check.
- https://github.com/grafana/tempo/pull/2770

joe-elliott:
Now that this in dskit we may want to restructure this code, but I think this is fine for now. At the time I was trying to write in the style of the library, but I think we should adjust this to just allow for some functions to be passed in as hooks where the caller can just inject the behavior necessary.

mdisibio (approved):

PR #348, opened 2023-08-04
Change init of consul client to respect TLS

### LINKED ISSUE #346: consider consul with TLS
Hi,

I'm currently trying to setup loki with consul.

As you use the `consul.NewClient` but pass in a [custom httpclient](https://github.com/grafana/dskit/blob/main/kv/consul/client.go#L91)

Consul ignores its environment variables (`CONSUL_CACERT`, `CONSUL_CLIENT_CERT`, `CONSUL_CLIENT_KEY`) that point to the TLS parts https://github.com/hashicorp/consul/blob/main/api/api.go#L706

It looks like the only reason that you pass in a custom client is to use the clean http transport and the timeout.

the `transport` could be passed in seperatly https://github.com/hashicorp/consul/blob/main/api/api.go#L678

for the timeout you could do this afterwards

So instead of
```go
client, err := consul.NewClient(&consul.Config{
		Address: cfg.Host,
		Token:   cfg.ACLToken.String(),
		Scheme:  "http",
		HttpClient: &http.Client{
			Transport: cleanhttp.DefaultPooledTransport(),
			// See https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/
			Timeout: cfg.HTTPClientTimeout,
		},
	})
```

Something like this:
```go

config := &consul.Config{
		Address: cfg.Host,
		Token:   cfg.ACLToken.String(),
		Scheme:  "http", 
                Transport: cleanhttp.DefaultPooledTransport(),
	}
client, err := consul.NewClient(config)

// not nice, but should work I guess :)
config.HttpClient.Timeout =  cfg.HTTPClientTimeout
```

mxab:
This PR changes the initialisation of the consul client.

As during the creation of a new consul client a lot of environment variables are taken as fallback default configs, passing in a own HttpClient prevents on picking the consul specific TLS environment variables.

Instead of passing in the complete new HttpClient it only passes the transport and sets the timeout on the HttpClient afterwards

Fixes #346

**Checklist**
- [ ] Tests updated

### DIFF
--- a/kv/consul/client.go
+++ b/kv/consul/client.go
@@ -6,7 +6,6 @@ import (
 	"flag"
 	"fmt"
 	"math/rand"
-	"net/http"
 	"time"
 
 	"github.com/go-kit/log"
@@ -84,16 +83,16 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, prefix string) {
 
 // NewClient returns a new Client.
 func NewClient(cfg Config, codec codec.Codec, logger log.Logger, registerer prometheus.Registerer) (*Client, error) {
-	client, err := consul.NewClient(&consul.Config{
-		Address: cfg.Host,
-		Token:   cfg.ACLToken.String(),
-		Scheme:  "http",
-		HttpClient: &http.Client{
-			Transport: cleanhttp.DefaultPooledTransport(),
-			// See https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/
-			Timeout: cfg.HTTPClientTimeout,
-		},
-	})
+	config := &consul.Config{
+		Address:   cfg.Host,
+		Token:     cfg.ACLToken.String(),
+		Scheme:    "http",
+		Transport: cleanhttp.DefaultPooledTransport(),
+	}
+	client, err := consul.NewClient(config)
+	// See https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/
+	config.HttpClient.Timeout = cfg.HTTPClientTimeout
+
 	if err != nil {
 		return nil, err
 	}


bboreham:
Can you be a little more specific about "a lot of environment variables"?

After this PR the only comment is pointing to a justification for something rather different; can you add one on why the timeout is post-initialization?

mxab:
Sorry for the confusing PR comment and I can also understand that the timeout parameters post-init looks strange :)

I'm trying to use loki with our consul which uses mTLS with self signed certificates.

I saw that you are using the official consul client to work with the KV store.

When using consul's  `NewClient(...`

```go
client, err := consul.NewClient(&consul.Config{
  Address: cfg.Host,
  Token:   cfg.ACLToken.String(),
  Scheme:  "http",
  HttpClient: &http.Client{
	  Transport: cleanhttp.DefaultPooledTransport(),
	  // See https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/
	  Timeout: cfg.HTTPClientTimeout,
  },
})

```


[the init method](https://github.com/hashicorp/consul/blob/main/api/api.go#L666C6-L666C15
) sets already a lot of defaults it detects from [environment variables](https://github.com/hashicorp/consul/blob/main/api/api.go#L27) if not explicitly provided as init parameters.

Some environment variables that would take already affect would be e.g. `CONSUL_NAMESPACE` , `CONSUL_HTTP_TOKEN_FILE` or even `CONSUL_HTTP_ADDR`, `CONSUL_HTTP_TOKEN` if the passed arguments would be empty (`""`)

The TLS related variables (`CONSUL_CACERT`, `CONSUL_CLIENT_CERT`, ...) are unfortunatelly not taken into account as [they're skipped](https://github.com/hashicorp/consul/blob/main/api/api.go#L706) when a custom http client is passed in.

If you pass in the transport instead of the client, they would be used as far as I see it, hence the PR to align the usage of env vars.

The only downside is that that you would have to "sneak" in the timeout parameter after the init, which I fully understand looks very wired.
Preferably I would like to change consul clients parameters to pass in the timeout directly, but I thought I try this one first :)

pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

PR #353, opened 2023-08-09
Francoposa/httpgrpc handle http opentracing span enhancements

francoposa:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated


CHANGELOG.md
```diff
@@ -54,6 +54,7 @@
 * [FEATURE] Add `log.BufferedLogger` type. #338
 * [FEATURE] Add `flagext.ParseFlagsAndArguments()` and `flagext.ParseFlagsWithoutArguments()` utilities. #341
 * [FEATURE] Add `log.RateLimitedLogger` for limiting the rate of logging. The `logger_rate_limit_discarded_log_lines_total` metrics traces the total number of discarded log lines per level. #352
+* [FEATURE] Add `middleware.HTTPGRPCTracer` for more detailed server-side tracing spans and tags on `httpgrpc.HTTP/Handle` requests>>>>>>> 5cb64fe (changelog)
```
charleskorn, line 57:
```suggestion
* [FEATURE] Add `middleware.HTTPGRPCTracer` for more detailed server-side tracing spans and tags on `httpgrpc.HTTP/Handle` requests
```

middleware/http_tracing.go
```diff
@@ -50,3 +46,111 @@ func (t Tracer) Wrap(next http.Handler) http.Handler {
...
+			opentracing.Tag{Key: string(ext.HTTPUrl), Value: url},
+			opentracing.Tag{Key: string(ext.HTTPMethod), Value: method},
+			opentracing.Tag{Key: "http.route", Value: matchedRoute},
+			opentracing.Tag{Key: "http.user_agent", Value: userAgent},
+		}
+		if parentSpan != nil {
+			startSpanOpts = append(
+				startSpanOpts,
+				opentracing.SpanReference{
+					Type:              opentracing.SpanReferenceType(1),
```
charleskorn, line 133: Should we use one of the `SpanReferenceType` constants here? And should it be `ChildOfRef` (which is 0), rather than `FollowsFromRef` (1)?

server/server_test.go
```diff
@@ -770,6 +774,138 @@ func TestLogSourceIPs(t *testing.T) {
...
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(4*1024*1024)),
+	)
+	require.NoError(t, err)
+	defer conn.Close()
+	client := httpgrpc.NewHTTPClient(conn)
+
+	// emulateHTTPGRPCPRoxy mimics the usage of the Server type as a load balancing proxy,
+	// wrapping http requests into gRPC requests to utilize gRPC load balancing features
+	emulateHTTPGRPCPRoxy := func(
```
charleskorn, line 869: [nit] Is this really emulating something? Or is it just a convenience function to reduce the duplication in calling the httpgrpc server for the two test cases?

server/server_test.go
```diff
@@ -770,6 +774,138 @@ func TestLogSourceIPs(t *testing.T) {
...
+		)
+		require.NoError(t, err)
+		resp, err := emulateHTTPGRPCPRoxy(client, req)
+		require.NoError(t, err)
+		var bodyJSON tracingSpanJSON
+		err = json.NewDecoder(bytes.NewReader(resp.Body)).Decode(&bodyJSON)
+		require.NoError(t, err)
+		require.Equal(t, "HTTP GET - "+expectedHelloPathParamRouteLabel, bodyJSON.OperationName)
+		require.Equal(t, expectedTagsHelloPathParamRoute, bodyJSON.Tags)
+	}
```
charleskorn, line 904: Might be clearer to wrap these two test cases in a call to `t.Run(...)` each.

These two cases could also be set up as data-driven test cases, which would help reduce the amount of duplication for each case.

middleware/http_tracing.go
```diff
@@ -50,3 +46,111 @@ func (t Tracer) Wrap(next http.Handler) http.Handler {
...
+		// skip spans which were not forwarded from non-httpgrpc.HTTP/Handle spans;
+		// standard http spans started directly from the HTTP server are presumed to
+		// already be instrumented by Tracer.Wrap
+		parentSpan := opentracing.SpanFromContext(ctx)
+		if parentSpan, ok := parentSpan.(*jaeger.Span); ok {
+			if parentSpan.OperationName() != httpGRPCHandleMethod {
+				next.ServeHTTP(w, r)
+				return
+			}
+		}
```
charleskorn, line 102: Could you please add a test that covers this case? (standard HTTP requests not from httpgrpc)

middleware/http_tracing.go
```diff
@@ -50,3 +46,111 @@ func (t Tracer) Wrap(next http.Handler) http.Handler {
...
+		if parentSpan != nil {
+			parentSpan.SetTag(string(ext.HTTPUrl), url)
+			parentSpan.SetTag(string(ext.HTTPMethod), method)
+			parentSpan.SetTag("http.route", matchedRoute)
+			parentSpan.SetTag("http.user_agent", userAgent)
+		}
+
+		// create and start child HTTP span
+		// mirroring opentracing-contrib/go-stdlib/nethttp.Middleware span name and tags
+		childSpanName := makeHTTPOperationNameFunc(hgt.RouteMatcher)(r)
```
charleskorn, line 120: Should we store the result of `makeHTTPOperationNameFunc` on `HTTPGRPCTracer` once (eg. in `InitHTTPGRPCMiddleware`)? Then we could do something like this:

```suggestion
		childSpanName := hgt.makeHTTPOperationName(r)
```

server/server_test.go
```diff
@@ -770,6 +774,138 @@ func TestLogSourceIPs(t *testing.T) {
...
+			"GET", helloRouteURL, bytes.NewReader([]byte{}),
+		)
+		require.NoError(t, err)
+		resp, err := emulateHTTPGRPCPRoxy(client, req)
+		require.NoError(t, err)
+		var bodyJSON tracingSpanJSON
+		err = json.NewDecoder(bytes.NewReader(resp.Body)).Decode(&bodyJSON)
+		require.NoError(t, err)
+		require.Equal(t, "HTTP GET - "+expectedHelloRouteLabel, bodyJSON.OperationName)
+		require.Equal(t, expectedTagsHelloRoute, bodyJSON.Tags)
```
charleskorn, line 890: Do we also need to check that the parent span has the expected tags?

server/server_test.go
```diff
@@ -770,6 +774,138 @@ func TestLogSourceIPs(t *testing.T) {
...
+	// writeSpanToBodyHandleFunc provides a way to test tracing by dumping the data
+	// from the requests' currently-active span data into a structured JSON body.
+	//
+	// HTTPGRPCTracer's parent Span tagging cannot be tested with current tracer implementations;
+	// tagging the parentSpan depends on methods only available on the Jaeger tracer, which limits
+	// introspection in tests as jaeger.Tracer can only access the current span, not the parent.
+	//
+	// MockTracer from opentracing/opentracing-go provides visibility into all completed spans for
+	// testing, but does not implement the jaeger operations needed to tag the parent span
+	writeSpanToBodyHandleFunc := func(w http.ResponseWriter, r *http.Request) {
```
charleskorn, line 812: Can't we inspect the created spans on `jaeger.InMemoryReporter` with `GetSpans()`?

middleware/http_tracing.go
francoposa: I would lean towards *not* exposing the span naming scheme that as something a user could change, because then the two opentracing middlewares that are both active could be instantiated with different patterns of forming the operation name.

As long as the span naming is standards compliant, I think it's alright for the library to make that decision for the user, or at least to not allow the user to make two different decisions for two middlewares that are supposed to behave identically.

server/server_test.go
francoposa: I mean you could say it's more general, but it is emulating how a proxy using this library works.

Otherwise you would not bother creating a normal HTTP request object -
I think when I add the test cases for standard http requests it will become more clear, as then we can show that the same HTTP request object behaves the same whether it's used directly or "proxied" into a gRPC request

server/server_test.go
francoposa: after a ton of digging, I found that you can inject an `Observer` which uses `SpanObserver`s to collect all this - latest approach is up now

middleware/http_tracing.go
charleskorn: Sorry, I should have been clearer in my comment: I don't think we should expose this to the consumer. I'm wondering if we can just call `makeHTTPOperationNameFunc` once in `InitHTTPGRPCMiddleware`, rather than calling it every time in this method.

server/server_test.go
charleskorn: Does inspecting the `InMemoryReporter` we instantiate when creating the `Tracer` not work? Seems like it would be simpler than creating our own `Observer` type.

francoposa: No, the Reporter interface just produces `opentracing.Span`s which don't expose getter methods for anything about the span except `BaggageItems` - so you cannot ask an `opentracing.Span` implementation for the operation name or the tags.

So you have to mock and inject something, the observer is most straightforward it seems. You could mock the spans themselves, but that seems sketchy when you are trying to test their behavior in the first place.

middleware/http_tracing.go
francoposa: just updated - I inlined it in the outer `Wrap` method, rather than the returned middleware method, so it doesn't get called for every request. It now mirrors how the standard HTTP tracer above it works.

`InitHTTPGRPCMiddleware` is just a convenience func that doesn't necessarily need to be used to apply the middleware - putting a call to a private `makeHTTPOperationNameFunc` in there would make `InitHTTPGRPCMiddleware` instead of optional.

server/server_test.go
charleskorn: Right, makes sense, thanks for clarifying.

middleware/http_tracing.go
charleskorn: Even better, thanks for addressing this.

middleware/http_tracing.go
francoposa: added - it falls through to the other middleware, so it's a bit of an awkward case for a unit test, but the test now does both asserts that they both create the same span operation name and tags

charleskorn (approved):
LGTM, thanks @francoposa!

PR #355, opened 2023-08-10
fix: timer leak in backoff

rfyiamcool:
### change

Fix the timer leak problem caused in some cases.

### DIFF
--- a/backoff/backoff.go
+++ b/backoff/backoff.go
@@ -77,9 +77,12 @@ func (b *Backoff) Wait() {
 	sleepTime := b.NextDelay()
 
 	if b.Ongoing() {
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):
I don't think there will be a leak per se, as the timer will be garbage collected once it fires, but explicitly stopping the timer should at least expedite the cleanup.

kscooo:
> I don't think there will be a leak per se, as the timer will be garbage collected once it fires, but explicitly stopping the timer should at least expedite the cleanup. Thanks.

https://github.com/golang/go/commit/508bb17edd04479622fad263cd702deac1c49157
rsc has fixed this problem.

PR #358, opened 2023-08-11
Test against Go 1.21 on CI

charleskorn:
**What this PR does**:

This PR adds Go 1.21 to the list of versions used on CI.

(none)

**Checklist**
- [n/a] Tests updated
- [n/a] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -1,6 +1,7 @@
 local supported_golang_versions = [
   '1.19.3',
   '1.20.4',
+  '1.21.0',
 ];
 
 local images = {
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -7,7 +7,7 @@
          "commands": [
             "make mod-check"
          ],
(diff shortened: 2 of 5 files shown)


.drone/drone.jsonnet
```diff
@@ -1,6 +1,7 @@
 local supported_golang_versions = [
   '1.19.3',
```
aknuds1, line 2: How many versions back should we test? Just wondering where to draw the line :)

aknuds1 (approved):
LGTM!

.drone/drone.jsonnet
charleskorn: I think following [Go's release policy](https://go.dev/doc/devel/release#policy) makes sense in general. I've kept 1.19 for now as that's the minimum version `go.mod` declares.

PR #359, opened 2023-08-13
Getting rid of sirupsen/logrus and log.Interface

duricanikolic:
**What this PR does**:
Recently some parts of `weaveworks/common` have been migrated to `dskit` ([PR](https://github.com/grafana/dskit/pull/342)). That migration included into `dskit` also the `log.Interface` interface that 'unifies' `gokit` logging and `logrus` logging.
On the other hand, all `Loki`, `Tempo`, `Mimir` and `Pyroscope` use only `gokit`, and neither from them uses `logrus` directly. Therefore, this PR removes the usage of `lrogus` and `log.Interface` from `dskit`.

**Checklist**
- [x] Tests updated


middleware/logging.go
```diff
@@ -78,7 +82,7 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		headers, err := dumpRequest(r, l.HTTPHeadersToExclude)
 		if err != nil {
 			headers = nil
-			requestLog.Errorf("Could not dump request headers: %v", err)
+			level.Error(requestLog).Log("err", fmt.Sprintf("could not dump request headers: %v", err))
```
pracucci, line 85:
```suggestion
			level.Error(requestLog).Log("msg", "could not dump request headers", "err", err))
```

user/logging.go
```diff
@@ -7,20 +7,21 @@ package user
...
 
-	"github.com/grafana/dskit/log"
+	"github.com/go-kit/log"
 )
 
 // LogWith returns user and org information from the context as log fields.
-func LogWith(ctx context.Context, log log.Interface) log.Interface {
+func LogWith(ctx context.Context, logger log.Logger) log.Logger {
 	userID, err := ExtractUserID(ctx)
+	logergWithFields := logger
```
pracucci, line 16: [nit] There's a typo "logerg" instead of "logger", but I don't think you need this variable at all. You can just overwrite `logger`.

middleware/logging.go
```diff
@@ -89,12 +93,12 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		if writeErr != nil {
 			if errors.Is(writeErr, context.Canceled) {
 				if l.LogRequestAtInfoLevel {
-					requestLog.Infof("%s %s %s, request cancelled: %s ws: %v; %s", r.Method, uri, time.Since(begin), writeErr, IsWSHandshakeRequest(r), headers)
+					level.Info(requestLog).Log("msg", fmt.Sprintf("%s %s %s, request cancelled: %s ws: %v; %s", r.Method, uri, time.Since(begin), writeErr, IsWSHandshakeRequest(r), headers))
```
pracucci, line 96: I'm not a big fan of using `fmt.Sprintf()` here. Also, the main difference is that previously the formatting was deferred until the real logging (which was skipped if the configured log level was higher than the log level) while now we always format it. This is particularly impactful for debug lines. I _think_ you can define a lazy evaluated sprintf, make it implement `Stringer` and pass it here, to get the lazy evaluation (to be tested).

log/gokit.go
```diff
@@ -25,82 +24,18 @@ func NewGoKitFormat(l Level, f Format) Interface {
...
-func GoKit(logger log.Logger) Interface {
-	return gokit{logger}
-}
-
-type gokit struct {
-	log.Logger
-}
-
-// Helper to defer sprintf until it is needed.
-type sprintf struct {
```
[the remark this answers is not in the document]
pracucci: See also my comment in `middleware/logging.go`

pracucci (commented):
I just did a very high level pass.

log/global.go
```diff
@@ -4,59 +4,22 @@
...
-func Errorln(args ...interface{}) {
-	global.Errorln(args...)
+func SetGlobal(logger log.Logger) {
+	global = logger
 }
 
-// WithField convenience function calls the global logger.
-func WithField(key string, value interface{}) Interface {
-	return global.WithField(key, value)
+func Log(keyvals ...interface{}) error {
```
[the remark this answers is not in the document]
duricanikolic: I agree with this. I just wanted this PR to maintain the features that were present in the current `main`, that's why I added this method. But I am happy to remove it.

log/format.go
```diff
@@ -6,15 +6,11 @@ package log
 
 import (
 	"flag"
-
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 )
 
 // Format is a settable identifier for the output format of logs
 type Format struct {
```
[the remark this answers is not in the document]
duricanikolic: This was another place where I was tempted to do what you are suggesting. But I thought it was better to keep as much from the current `main` as possible. Again, happy to remove this.

middleware/logging.go
duricanikolic: Thank you for the hint.

zalegrala:
Are users meant to use `go-kit` instead of the `log.Interface` here?  I was looking in this area the other day and wondering which pattern users were meant to use.

middleware/logging.go
```diff
@@ -107,29 +113,33 @@ func (l Log) Wrap(next http.Handler) http.Handler {
...
-				r.Method, uri, statusCode, time.Since(begin), buf.Bytes(), IsWSHandshakeRequest(r), headers)
+			lazySprintf := dskit_log.NewGokitLazySprintf("%s %s (%d) %s Response: %q ws: %v; %s", []interface{}{r.Method, uri, statusCode, time.Since(begin), buf.Bytes(), IsWSHandshakeRequest(r), headers})
+			level.Warn(requestLog).Log("msg", lazySprintf)
 		}
 	})
 }
 
 // Logging middleware logs each HTTP request method, path, response code and
 // duration for all HTTP requests.
 var Logging = Log{
```
pracucci, line 141: Who uses it? I can't see any usage a part from tests. Can you check if Mimir, Loki, Tempo or Pyroscope uses it? Otherwise we should remove it.

signals/signals.go
```diff
@@ -51,19 +52,19 @@ func (h *Handler) Loop() {
...
-				h.log.Infof("=== received SIGINT/SIGTERM ===\n*** exiting")
+				level.Info(h.log).Log("msg", "=== received SIGINT/SIGTERM ===\n*** exiting")
 				for _, subsystem := range h.receivers {
 					_ = subsystem.Stop()
 				}
 				return
 			case syscall.SIGQUIT:
 				stacklen := runtime.Stack(buf, true)
-				h.log.Infof("=== received SIGQUIT ===\n*** goroutine dump...\n%s\n*** end", buf[:stacklen])
+				level.Info(h.log).Log("msg", "=== received SIGQUIT ===\n*** goroutine dump...\n%s\n*** end", buf[:stacklen])
```
pracucci, line 67: This is not correct. It's a sprintf (see `%s` in the message). You the lazy one here too.

server/server.go
```diff
@@ -331,7 +333,8 @@ func newServer(cfg Config, metrics *Metrics) (*Server, error) {
 		}
 	}
 
-	logger.WithField("http", httpListener.Addr()).WithField("grpc", grpcListener.Addr()).Infof("server listening on addresses")
+	loggerWithFields := gokit_log.With(logger, "http", httpListener.Addr(), "grpc", grpcListener.Addr())
+	level.Info(loggerWithFields).Log("msg", "server listening on addresses")
```
pracucci, line 337: Looks a very convoluted way to do it. Do this instead:

```suggestion
	level.Info(loggerWithFields).Log("msg", "server listening on addresses", "http", httpListener.Addr(), "grpc", grpcListener.Addr())
```

log/logging.go
```diff
@@ -7,26 +7,24 @@ package log
...
 func Setup(logLevel string) error {
-	level, err := logrus.ParseLevel(logLevel)
+	var level Level
+	err := level.Set(logLevel)
 	if err != nil {
 		return fmt.Errorf("error parsing log level: %v", err)
 	}
-	hook, err := promrus.NewPrometheusHook() // Expose number of log messages as Prometheus metrics.
+
+	// TODO understand whether this is needed
```
pracucci, line 21: Can you investigate this, please?

log/logging.go
```diff
@@ -7,26 +7,24 @@ package log
...
-	"github.com/sirupsen/logrus"
-
-	"github.com/weaveworks/promrus"
 )
 
-// Setup configures a global logrus logger to output to stderr.
-// It populates the standard logrus logger as well as the global logging instance.
+// Setup configures a global Gokit logger to output to stderr.
+// It populates the Gokit logger as well as the global logging instance.
 func Setup(logLevel string) error {
```
pracucci, line 14: Can you check among the projects who use this? I can't see it being called neither in dskit, neither in Mimir.

log/gokit_test.go
```diff
@@ -5,14 +5,18 @@
...
+	"fmt"
 	"testing"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/stretchr/testify/require"
 )
 
 func BenchmarkDebugf(b *testing.B) {
```
pracucci, line 18: [nit] Rename this.

log/gokit.go
```diff
@@ -6,101 +6,68 @@ package log
...
 
-func (g gokit) Infof(format string, args ...interface{}) {
-	level.Info(g.Logger).Log("msg", &sprintf{format: format, args: args})
-}
-func (g gokit) Infoln(args ...interface{}) {
-	level.Info(g.Logger).Log("msg", &sprint{args: args})
+// stand-alone for test purposes
+func addStandardFields(logger log.Logger, l Level) log.Logger {
+	logger = log.With(logger, "ts", log.DefaultTimestampUTC, "caller", log.Caller(5))
+	return level.NewFilter(logger, l.Option)
```
pracucci, line 50: [nit] It does more than adding fields. I would move the filter to the caller, for clarity.

log/gokit.go
```diff
@@ -6,101 +6,68 @@ package log
...
+)
 
-// Helper to defer sprint until it is needed.
-// Note we don't use Sprintln because the output is passed to go-kit as one value among many on a line
-type sprint struct {
-	args []interface{}
+// NewGoKit creates a new GoKit logger with the given level, format and writer.
+// If the given writer is nil, os.Stderr is used.
+// If the given format is nil, logfmt is used.
+func NewGoKit(lvl Level, format string, writer io.Writer) log.Logger {
```
pracucci, line 23: Who needs this? I'm wondering why we need a logger without the "standard fields".

log/logging.go
duricanikolic: I have checked `dskit`, `Mimir`, `Loki`, `Tempo` and `Pyroscope` and none of them uses this method.
I will delete it.

middleware/logging.go
duricanikolic: I have checked `Mimir`, `Loki`, `Tempo` and `Pyroscope` and none of them uses this method.
The only usage is in a test in `dskit`.
Therefore, I will delete it and use the global logger in the tests.

log/logging.go
duricanikolic: This method is not used by any of the projects (`dskit`, `Mimir`, `Loki`, `Tempo`, `Pyroscope`), so I decided to delete it.

log/gokit_test.go
duricanikolic: I have renamed the benchmark, and also changed its implementation: I replaced the `fmt.Sprintf` usage with `LazySprintf`.

duricanikolic:
> Are users meant to use `go-kit` instead of the `log.Interface` here? I was looking in this area the other day and wondering which pattern users were meant to use.

Hi @zalegrala, 
The main goal of this PR is to remove the `log.Interface` interface, and the usage of `logrus`, so the only supported logger would be `go-kit` Logger. 

I will ask all the teams that use `dskit` what they think about this idea, and if they all agree, we can merge this PR.

simonswine (approved):
I am okay with the proposed change, but in case we wanted to appeal to projects (external/internal) which don't use log-kit, this interface is useful and in theory could be developed into compatibility with [`slog`](https://github.com/golang/go/issues/56345), which is in the stdlib from 1.21.

In any way I am not seeing any blockers from a consumer perspective of `grafana/pyroscope`

duricanikolic:
> I am okay with the proposed change, but in case we wanted to appeal to projects (external/internal) which don't use log-kit, this interface is useful and in theory could be developed into compatibility with [`slog`](https://github.com/golang/go/issues/56345), which is in the stdlib from 1.21.
> 
> In any way I am not seeing any blockers from a consumer perspective of `grafana/pyroscope`

Hi @simonswine, 
Thank you for the feedback. Your remark about keeping `log.Interface` because it is compatible with other loggers sounds interesting. For the time being no project actually uses it, so I thought about removing it. To be more clear, since all `Loki`, `Tempo`, `Pyroscope` and `Mimir` use `go-kit' Logger, instead of doing
```
logger := GoKit(log) // this wraps the go-kit logger into log.Interface
logger.Infof("There are %d projects", 4)
```
they simply do (not exactly, this is just an example)
```
level.Info(log).Log("msg", fmt.Sprintf("There are %d projects", 4))
```

So, we can, for the moment, remove the interface, and if at a point there will be a need to move to another logger (e.g., `slog`) we will revert the change and put `log.Interface` back. That change will not be breaking, and projects will not need to change anything, unless they want to migrate to `log.Interface`. WDYT?

simonswine:
@duricanikolic the original ambition for dskit was to be generic and adopted inside and outside of Grafana Labs by people building distributed services. This is obviously not the reality today (and might never be). This is in no way blocking or a strongly held opinion, but I think this gets us further from that vision.

duricanikolic:
This [branch](https://github.com/grafana/loki/tree/yuri/dskit-log) show the changes needed to be done in `Loki` if this draft PR gets merged.

This [branch](https://github.com/grafana/tempo/tree/yuri/dskit-log) show the changes needed to be done in `Tempo` if this draft PR gets merged.

duricanikolic:
> @duricanikolic the original ambition for dskit was to be generic and adopted inside and outside of Grafana Labs by people building distributed services. This is obviously not the reality today (and might never be). This is in no way blocking or a strongly held opinion, but I think this gets us further from that vision.

@simonswine if we decide to keep `log.Interface` how feasible do you think it is that the teams will actually decide to use it and to switch from `gokit.Logger` to `log.Interface`?

For example, all occurrences of 
```
level.Info(logger).Log(...)
```
will have to be replaced with
```
wrappedLog := GoKit(logger) // wraps gokit.Logger into log.Interface
wrappedLog.Infoln(...)
```

Moreover, the interface has some disadvantages: for example, if we want to log 
```
level.Info(logger).Log("key1", "value1", "key2", "value2", "msg", "abc")
```
we need to do
```
wrappedLog.WithField("key1", "value1").WithField("key2", "value2").Infoln("abc") // Infoln and Infof implicitly add the "msg" key and print the argument as its value
```
but the calls to `WithField` are expensive.

pracucci:
I see @simonswine comment about the `log.Interface`, however I would like to share different perspective.

This PR is not a regression in that regards: dskit doesn't use `log.Interface`. The interface was imported 2 weeks ago when copying `weaveworks/common` into `dskit` ([PR](https://github.com/grafana/dskit/pull/342/files#diff-a2e9795e00e837c49599b9de41664c913be5cc571f79e9efa4e575908a8470c9)). Dskit itself wasn't using the interface before and neither after, given the interface is only used by the code copied from `weaveworks/common`.

Do we want to be less opinionated and support a generic log interface for all the logging done by dskit? Maybe. Is this the main blocker to dskit wide adoption? I doubt (e.g. the lack of any doc may be a bigger blocker). Personally, I would defer the discussion about a generic logging interface until we'll have a proof that's the blocker for an adoption outside Grafana Labs.

middleware/path_rewrite.go
```diff
@@ -31,7 +33,7 @@ func (p pathRewrite) Wrap(next http.Handler) http.Handler {
 		r.URL.RawPath = p.regexp.ReplaceAllString(r.URL.EscapedPath(), p.replacement)
 		path, err := url.PathUnescape(r.URL.RawPath)
 		if err != nil {
-			log.Errorf("Got invalid url-encoded path %v after applying path rewrite %v: %v", r.URL.RawPath, p, err)
+			level.Error(log.Global()).Log("msg", log.LazySprintf("got invalid url-encoded path %v after applying path rewrite %v: %v", r.URL.RawPath, p), "err", err)
```
pracucci, line 36: This doesn't look right. There are 3 substituions in the printf format but only 2 passed args.


```suggestion
			level.Error(log.Global()).Log("msg", log.LazySprintf("got invalid url-encoded path %v after applying path rewrite %v", r.URL.RawPath, p), "err", err)
```

log/gokit.go
```diff
@@ -6,101 +6,93 @@ package log
...
 }
 
-func (s *sprint) String() string {
-	return fmt.Sprint(s.args...)
+// NewGoKitLogFmt creates a new GoKit logger with the given level and writer,
+// enriched with the standard "ts" and "caller" fields.
+// The "logfmt" format is used.
+// If the given writer is nil, os.Stderr is used.
+func NewGoKitLogFmt(l Level, writer io.Writer) log.Logger {
+	return NewGoKit(l, "logfmt", writer)
```
pracucci, line 45: [nit] I doubt we need this function at all. I think the caller would just call `NewGoKit()` passing `LogfmtFormat`. If you decide to keep it, then see the following suggestion

```suggestion
	return NewGoKit(l, LogfmtFormat, writer)
```

log/gokit.go
```diff
@@ -6,101 +6,93 @@ package log
...
-}
-func (g gokit) Debugln(args ...interface{}) {
-	level.Debug(g.Logger).Log("msg", &sprint{args: args})
+// NewGoKitWithFields creates a new GoKit logger configured with the given level, format, writer and
+// rate limit-related configuration, enriched with the given fields.
+// If the given format is nil, logfmt is used.
+// If the given writer is nil, os.Stderr is used.
+// If the given rate limit configuration is nil, no rate limited logger is created.
+// If the given fields are empty, no fields are added.
+func NewGoKitWithFields(lvl Level, format string, writer io.Writer, rateLimitedCfg *RateLimitedLoggerCfg, fields Fields) log.Logger {
```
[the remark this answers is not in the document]
duricanikolic: It depends on what we want to achieve. I wanted to create a function that simplifies the job for the caller, and that calls `log.With`, `log.NewFilter` and `NewRateLimitedLogger` in the right order. But I can revert this change, and leave the responsibility to call the functions in the right order to the caller.

log/gokit.go
duricanikolic: Removed

log/ratelimit.go
```diff
@@ -23,106 +24,53 @@ type RateLimitedLogger struct {
 	discardedErrorLogLinesCounter prometheus.Counter
 }
 
-// NewRateLimitedLogger returns a logger.Interface that is limited to the given number of logs per second,
+// NewRateLimitedLogger returns a log.Logger that is limited to the given number of logs per second,
 // with the given burst size.
-func NewRateLimitedLogger(logger Interface, logsPerSecond rate.Limit, burstSize int, reg prometheus.Registerer) Interface {
```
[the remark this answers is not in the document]
duricanikolic: The main reason for this change was the introduction of `NewGoKitWithFields`, because with the `RateLimitedLoggerConfig` struct it was easier to pass the parameters.
But since we are going to remove `NewGoKitWithFields`, I will revert this change.

log/logging.go
duricanikolic: It is now deleted. I was sure I did it, but somehow I missed it.

server/server.go
```diff
@@ -230,11 +233,10 @@ func NewWithMetrics(cfg Config, metrics *Metrics) (*Server, error) {
 }
 
 func newServer(cfg Config, metrics *Metrics) (*Server, error) {
-	// If user doesn't supply a logging implementation, by default instantiate
-	// logrus.
+	// If user doesn't supply a logging implementation, by default instantiate go-kit.
 	logger := cfg.Log
 	if logger == nil {
-		logger = log.NewLogrus(cfg.LogLevel)
+		logger = log.NewGoKit(log.LogfmtFormat, os.Stderr)
```
pracucci, line 239: This is a regression. I think you misunderstood my previous comment. You should also have a `log. NewGoKitWithLevel()` where you pass the level and the level filter it added to the logger. Then you call it here, passing `cfg.LogLevel`.

pracucci (approved):
LGTM, modulo one last fix

zalegrala (commented):
This looks reasonable to me.

duricanikolic:
> This looks reasonable to me.

@zalegrala could you please approve the PR?

trevorwhitney (approved):
Looks good from Loki team. Thanks!

log/gokit.go
```diff
@@ -6,101 +6,58 @@ package log
...
-func NewGoKit(l Level) Interface {
-	return NewGoKitFormat(l, Format{s: "logfmt"})
+// NewGoKit creates a new GoKit logger with the given format and writer.
+// If the given writer is nil, os.Stderr is used.
+// If the given format is nil, logfmt is used.
+// No additional fields nor filters are added to the created logger, and
+// if they are required, the caller is expected to add them.
+func NewGoKit(format string, writer io.Writer) log.Logger {
+	if writer == nil {
+		writer = log.NewSyncWriter(os.Stderr)
```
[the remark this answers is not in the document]
duricanikolic: I have implemented the new constructor.

log/gokit.go
```diff
@@ -6,101 +6,68 @@ package log
...
-		logger = log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
-	}
-	return addStandardFields(logger, l)
+const (
+	LogfmtFormat = "logfmt"
+	JSONFormat   = "json"
+)
+
+// NewGoKit creates a new GoKit logger with the given format.
+// If the given format is nil, logfmt is used.
```
pracucci, line 22: [nit] Technically a string can't be nil.

```suggestion
// If the given format is empty or unknow, logfmt is used.
```

log/gokit.go
```diff
@@ -6,101 +6,68 @@ package log
...
+	return newGoKit(format, writer)
 }
 
-// stand-alone for test purposes
-func addStandardFields(logger log.Logger, l Level) Interface {
-	logger = log.With(logger, "ts", log.DefaultTimestampUTC, "caller", log.Caller(5))
-	logger = level.NewFilter(logger, l.Gokit)
-	return gokit{logger}
+// NewGoKitWithLevel creates a new GoKit logger with the given level and format.
+// If the given format is nil, logfmt is used.
```
pracucci, line 31:
```suggestion
// If the given format is empty or unknown, logfmt is used.
```

log/gokit.go
```diff
@@ -6,101 +6,68 @@ package log
...
+	return level.NewFilter(logger, lvl.Option)
 }
 
-// NewGoKit creates a new Interface backed by a GoKit logger
-func NewGoKit(l Level) Interface {
-	return NewGoKitFormat(l, Format{s: "logfmt"})
+// NewGoKitWithWriter creates a new GoKit logger with the given format and writer.
+// The input writer must be provided, must be thread-safe, and the caller is
+// expected to guarantee these requirements.
+// If the given format is nil, logfmt is used.
```
pracucci, line 40:
```suggestion
// If the given format is empty or unknown, logfmt is used.
```

pracucci (approved):
I hope I haven't missed anything. LGTM! (I just left a last nit)

PR #364, opened 2023-08-15
ring: DoUntilQuorum: cancel context for zone as soon as first failure is encountered

charleskorn:
**What this PR does**:

This PR improves the behaviour of `DoUntilQuorum` (and `DoUntilQuorumWithoutSuccessfulContextCancellation`) when a zone fails.

Previously, if a request to an instance failed and zone awareness was enabled, requests to the remaining instances in the same zone weren't cancelled until quorum was reached.

With this change, requests to all instances in the failing zone will be cancelled as soon as the first error is returned.

(none)

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/replication_set_test.go
+++ b/ring/replication_set_test.go
@@ -564,6 +564,79 @@ func TestDoUntilQuorumWithoutSuccessfulContextCancellation_PartialZoneFailure(t 
 	}
 }
 
+func TestDoUntilQuorumWithoutSuccessfulContextCancellation_CancelsEntireZoneImmediatelyOnSingleFailure(t *testing.T) {
+	defer goleak.VerifyNone(t)
+
+	replicationSet := ReplicationSet{
+		Instances: []InstanceDesc{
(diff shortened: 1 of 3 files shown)


ring/replication_set_test.go
```diff
@@ -564,6 +564,78 @@ func TestDoUntilQuorumWithoutSuccessfulContextCancellation_PartialZoneFailure(t
 	}
 }
 
+func TestDoUntilQuorumWithoutSuccessfulContextCancellation_CancelsEntireZoneImmediatelyOnSingleFailure(t *testing.T) {
```
jhesketh, line 567: that name tho :-o

jhesketh (approved):
lgtm

ring/replication_set_test.go
charleskorn: `DoUntilQuorumWithoutSuccessfulContextCancellation` is deliberately named to be awkward 🙂 Most consumers should use the far more friendly-sounding `DoUntilQuorum` unless they need the different behaviour the unwieldy-named method provides.

PR #365, opened 2023-08-21
Migrate metric test helpers from Mimir 

### LINKED ISSUE #351: Migrate metric test helpers from Mimir
Mimir has a couple of test helper methods for testing emitted metrics that would be good to migrate to dskit:

> I was wondering if we could reuse the utilities from the `metrics` package, but I don't think we can out of the box. Not high priority, but I think in Mimir we have most of the missing utilities ready and we could move them to dskit:
> - `findHistogramMatchingLabels()` from `pkg/frontend/querymiddleware/codec_json_test.go`
> - `getMetricsMatchingLabels()` from `pkg/storegateway/bucket_e2e_test.go`
> - `MatchesSelectors()` from `pkg/util`

From https://github.com/grafana/dskit/pull/344/files#r1285937024

charleskorn:
**What this PR does**:

This PR migrates three test helpers we use frequently in Mimir to test metrics emitted during a test.

Fixes #351.

**Checklist**
- [x] Tests updated


pstibrany:
> They were required to add the dependency on `github.com/prometheus/prometheus`.

Oh I see. I'm not sure if we actually want to add such dependency. Perhaps it may be better to keep this only in Mimir after all?

pracucci:
I suggested to do #351 but I haven't realised we would have vendored `prometheus`. I agree with Peter that dskit shouldn't do it.

charleskorn:
I've reworked `MatchesLabels` to no longer depend on `github.com/prometheus/prometheus` - let me know what you think @pstibrany @pracucci.

metrics/helpers.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"errors"
+	"fmt"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// MatchesLabels returns true if m has all the labels in l.
+func MatchesLabels(m *dto.Metric, labelNamesAndValues ...string) bool {
+	for _, l := range makeLabels(labelNamesAndValues...) {
```
pstibrany, line 12: nit: do we need to create `[]*dto.LabelPair` first? If not, we can save some work and allocations here.

pstibrany (approved):
Thanks, lgtm.

metrics/helpers.go
charleskorn: We don't have to, but it makes the code below clearer as we can refer to things as `Name` and `Value` rather than indexing into an array. Given this method is only intended to be used in tests, and with a small number of labels, I don't think the allocations are an issue, but happy to revisit this if you disagree.

(Given your approval, I'm going to merge the PR, but happy to address this in a follow-up PR.)

pstibrany: > Given this method is only intended to be used in tests, and with a small number of labels, I don't think the allocations are an issue, but happy to revisit this if you disagree.

I see. We can improve it once it becomes part of critical path :)

PR #367, opened 2023-08-21
spanlogger: much more efficient debug logging

bboreham:
**What this PR does**:

Defer creating the logger until we need it, and add a `DebugLog` entrypoint so callers that can say whether debug logging is enabled can avoid it entirely.

Previously `SpanLogger` exposed the `Log` member, so I hope nobody was relying on that.

```
cpu: Intel(R) Core(TM) i5-9400F CPU @ 2.90GHz
BenchmarkSpanLogger/log-4                5158831               290.5 ns/op           160 B/op          2 allocs/op
BenchmarkSpanLogger/level.debug-4        1927003               550.0 ns/op           400 B/op          5 allocs/op
BenchmarkSpanLogger/debuglog-4          125852071                9.472 ns/op           0 B/op          0 allocs/op
```

**Checklist**
- [x] Tests updated

### DIFF
--- a/spanlogger/spanlogger.go
+++ b/spanlogger/spanlogger.go
@@ -2,6 +2,7 @@ package spanlogger
 
 import (
 	"context"
+	"sync/atomic"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -33,10 +34,10 @@ var (
 
 // SpanLogger unifies tracing and logging, to reduce repetition.
 type SpanLogger struct {
-	ctx        context.Context // context passed in, with logger
-	resolver   TenantResolver  // passed in
-	baseLogger log.Logger      // passed in
-	logger     log.Logger      // initialized on first use
+	ctx        context.Context            // context passed in, with logger
+	resolver   TenantResolver             // passed in
+	baseLogger log.Logger                 // passed in
+	logger     atomic.Pointer[log.Logger] // initialized on first use
 	opentracing.Span
 	sampled      bool
 	debugEnabled bool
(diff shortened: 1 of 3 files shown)


spanlogger/spanlogger.go
```diff
@@ -68,22 +75,57 @@ func FromContext(ctx context.Context, fallback log.Logger, resolver TenantResolv
...
+	if s.debugEnabled {
+		if s.logger == nil {
+			s.logger = s.makeLogger()
+		}
+		// The call to Log() through an interface makes its argument escape, so make a copy here,
+		// in the debug-only path, so the function is faster for the non-debug path.
+		localCopy := append([]any{}, kvps...)
+		level.Debug(s.logger).Log(localCopy...)
+	}
+	return s.spanLog(kvps...)
```
charleskorn, line 125: Do we need to add a `level=debug` pair here in the call to `spanLog`? Otherwise events that previously included the level when using `level.Debug().Log()` will not have it when `DebugLog()` is used.

bboreham: I can't say I've ever needed that. Is it something you rely on?

dimitarvdimitrov (approved):
LGTM

spanlogger/spanlogger.go
charleskorn: Not something I rely on, but it is a change. I'm happy to go ahead without it and see if anyone complains / anything breaks.

pracucci (approved):
Nice optimization. I left a non-blocking comment, which I would consider tho. Would make the caller life way easier (because no change would be required).

PR #368, opened 2023-08-21
Fix computing of min/max of gauges.

pstibrany:
**What this PR does**:

This PR fixes sending of min/max of gauges:

- `MetricFamilyMap.MinGauges`, `MaxGauges` now return `NaN` if no gauge was found. Previously `0` was returned, but that made it difficult to distinguish between missing gauge and gauge with 0 value, and this distinction is important when computing min/max across tenants.
- `MetricFamiliesPerTenant.SendMaxOfGaugesPerTenant` no longer includes tenants, which do not have specified gauge.

- `MetricFamiliesPerTenant.SendMinOfGauges` will no longer return 0 if some tenant doesn't have the gauge in their registry, and other tenants have positive values only.

- `MetricFamiliesPerTenant.SendMaxOfGauges` will no longer return 0 if some tenant doesn't have the gauge in their registry, and other tenants have negative values only.

**Checklist**
- [x] Tests updated

### DIFF
--- a/metrics/tenant_registries.go
+++ b/metrics/tenant_registries.go
@@ -103,7 +103,7 @@ func (mfm MetricFamilyMap) SumGauges(name string) float64 {
 
 // MaxGauges returns max of gauges or NaN, if no gauge was found.
 func (mfm MetricFamilyMap) MaxGauges(name string) float64 {
-	return fold(mfm[name], gaugeValueNaN, func(val, res float64) float64 {
+	return fold(mfm[name], gaugeValueOrNaN, func(val, res float64) float64 {
 		if val > res {
 			return val
 		}
@@ -113,7 +113,7 @@ func (mfm MetricFamilyMap) MaxGauges(name string) float64 {
 
 // MinGauges returns minimum of gauges or NaN if no gauge was found.
 func (mfm MetricFamilyMap) MinGauges(name string) float64 {
-	return fold(mfm[name], gaugeValueNaN, func(val, res float64) float64 {
+	return fold(mfm[name], gaugeValueOrNaN, func(val, res float64) float64 {
 		if val < res {
 			return val
 		}
@@ -461,7 +461,7 @@ func sum(mf *dto.MetricFamily, fn func(*dto.Metric) float64) float64 {
 	return result
 }
 
-// fold returns value computed from multiple metrics, using folding function. if there are no metrics, it returns 0.
+// fold returns value computed from multiple metrics, using folding function. if there are no metrics, it returns NaN.
 func fold(mf *dto.MetricFamily, fn func(*dto.Metric) float64, foldFn func(val, res float64) float64) float64 {
 	result := math.NaN()
 
@@ -484,8 +484,8 @@ func fold(mf *dto.MetricFamily, fn func(*dto.Metric) float64, foldFn func(val, r
 func counterValue(m *dto.Metric) float64 { return m.GetCounter().GetValue() }
 func gaugeValue(m *dto.Metric) float64   { return m.GetGauge().GetValue() }
 
-// gaugeValueNaN returns Gauge value or math.NaN.
-func gaugeValueNaN(m *dto.Metric) float64 {
+// gaugeValueOrNaN returns Gauge value or math.NaN.
+func gaugeValueOrNaN(m *dto.Metric) float64 {
 	if m == nil || m.Gauge == nil || m.Gauge.Value == nil {
 		return math.NaN()
 	}
--- a/metrics/tenant_registries.go
+++ b/metrics/tenant_registries.go
@@ -91,20 +91,34 @@ func NewMetricFamilyMap(metrics []*dto.MetricFamily) (MetricFamilyMap, error) {
 	return perMetricName, nil
 }
 
+// SumCounters returns sum of gauges or 0, if no gauge was found.
 func (mfm MetricFamilyMap) SumCounters(name string) float64 {
 	return sum(mfm[name], counterValue)
 }
 
+// SumGauges returns sum of gauges or 0, if no gauge was found.
 func (mfm MetricFamilyMap) SumGauges(name string) float64 {
 	return sum(mfm[name], gaugeValue)
 }
 
+// MaxGauges returns max of gauges or NaN, if no gauge was found.
 func (mfm MetricFamilyMap) MaxGauges(name string) float64 {
-	return max(mfm[name], gaugeValue)
+	return fold(mfm[name], gaugeValueNaN, func(val, res float64) float64 {
+		if val > res {
+			return val
+		}
+		return res
+	})
 }
 
+// MinGauges returns minimum of gauges or NaN if no gauge was found.
 func (mfm MetricFamilyMap) MinGauges(name string) float64 {
-	return min(mfm[name], gaugeValue)
+	return fold(mfm[name], gaugeValueNaN, func(val, res float64) float64 {
+		if val < res {
+			return val
+		}
+		return res
+	})
 }
 
 func (mfm MetricFamilyMap) SumHistograms(name string) HistogramData {
@@ -235,6 +249,9 @@ func (d MetricFamiliesPerTenant) foldGauges(out chan<- prometheus.Metric, desc *
 	result := math.NaN()
 	for _, tenantEntry := range d {
 		value := valFn(tenantEntry.metrics)
+		if math.IsNaN(value) {
+			continue
+		}
 		if math.IsNaN(result) {
 			result = value
 		} else {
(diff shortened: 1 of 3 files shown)


metrics/tenant_registries.go
```diff
@@ -91,20 +91,34 @@ func NewMetricFamilyMap(metrics []*dto.MetricFamily) (MetricFamilyMap, error) {
 	return perMetricName, nil
 }
 
+// SumCounters returns sum of gauges or 0, if no gauge was found.
```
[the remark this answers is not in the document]
pstibrany: Thanks, fixed in https://github.com/grafana/dskit/pull/368/commits/64f8954c2c92302b4993a81dc9d527df3c2f555c

metrics/tenant_registries.go
```diff
@@ -445,51 +463,41 @@ func sum(mf *dto.MetricFamily, fn func(*dto.Metric) float64) float64 {
 
 // max returns the maximum value from all metrics from same metric family (= series with the same metric name, but different labels)
 // Supplied function extracts value.
```
krajorama, line 465: Extra docstring

metrics/tenant_registries.go
```diff
@@ -445,51 +463,41 @@ func sum(mf *dto.MetricFamily, fn func(*dto.Metric) float64) float64 {
...
-	return fold(mf, fn, func(val, res float64) float64 {
-		if val > res {
-			return val
-		}
-		return res
-	})
-}
 
 // min returns the minimum value from all metrics from same metric family (= series with the same metric name, but different labels)
 // Supplied function extracts value.
```
krajorama, line 468: Extra docstring

pstibrany: Fixed in https://github.com/grafana/dskit/pull/368/commits/e3957808e55fc03b571625e853d6814cd996fa31.

metrics/tenant_registries.go
```diff
@@ -248,7 +265,6 @@ func (d MetricFamiliesPerTenant) foldGauges(out chan<- prometheus.Metric, desc *
 	}
 
 	out <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, result)
-
 }
 
 func (d MetricFamiliesPerTenant) SendMinOfGauges(out chan<- prometheus.Metric, desc *prometheus.Desc, gauge string) {
```
krajorama, line 270: This calls d.foldGauges above, which returns 0 as default, so I don't think the statement in the PR description holds: 
> MetricFamiliesPerTenant.SendMinOfGauges will no longer return 0 if any tenant doesn't have the gauge.

pstibrany: I think I should fix the message to say "MetricFamiliesPerTenant.SendMinOfGauges will no longer return 0 if **some** tenant doesn't have the gauge."

krajorama: > > MetricFamiliesPerTenant.SendMinOfGauges will no longer return 0 if any tenant doesn't have the gauge.
> 
> This message from changelog describes fixed bug that exists in current version of dskit (before this PR): if there's a tenant registry without a gauge with given name, then `SendMinOfGauges` for given gauge always sends value 0 (assuming there's no value below 0). This PR fixes it.

The test bellow has:
```
	// No matching metric.
	verifySendMinOfGauges(t, 0)
```
It explicitly verifies the opposite, which is the behavior before this PR as far as I can tell. Confused...

pstibrany: The test relevant for this changelog entry is on line 326: https://github.com/grafana/dskit/blob/62aaeef8ecad2bb120a5c4c696ccf439c85a7c96/metrics/tenant_registries_test.go#L326

At that point in the test, there are 2 user-registries in `regs`, but only one of them has `test_metric` gauge registered. Now the code correctly sends `100`, before the fix it would send `0`.

In this changelog entry I'm trying to explain this bugfix, but I'm failing :(

metrics/tenant_registries.go
```diff
@@ -443,53 +461,37 @@ func sum(mf *dto.MetricFamily, fn func(*dto.Metric) float64) float64 {
...
-func min(mf *dto.MetricFamily, fn func(*dto.Metric) float64) float64 {
-	return fold(mf, fn, func(val, res float64) float64 {
-		if val < res {
-			return val
-		}
-		return res
-	})
-}
-
 // fold returns value computed from multiple metrics, using folding function. if there are no metrics, it returns 0.
```
pracucci, line 464: The part of the comment which says "if there are no metrics, it returns 0" doesn't look correct anymore. Now it returns NaN.

metrics/tenant_registries.go
```diff
@@ -443,53 +461,37 @@ func sum(mf *dto.MetricFamily, fn func(*dto.Metric) float64) float64 {
...
-
 	return result
 }
 
 // This works even if m is nil, m.Counter is nil or m.Counter.Value is nil (it returns 0 in those cases)
 func counterValue(m *dto.Metric) float64 { return m.GetCounter().GetValue() }
 func gaugeValue(m *dto.Metric) float64   { return m.GetGauge().GetValue() }
 
+// gaugeValueNaN returns Gauge value or math.NaN.
+func gaugeValueNaN(m *dto.Metric) float64 {
```
pracucci, line 488: [nit] I would clarify it a bit.

```suggestion
func gaugeValueOrNaN(m *dto.Metric) float64 {
```

pracucci (commented):
[not a blocker] I'm not super convinced about stop returning a metric once there are no more tenants with that gauge. After this change, when we'll query that metric we'll keep getting the last value for the lookback period (5m) which may be annoying. I'm wondering if we could return the StaleNaN instead.

> MetricFamiliesPerTenant.SendMinOfGauges will no longer return 0 if any tenant doesn't have the gauge.

A couple of things:
- The PR description should mention SendMaxOfGauges() too
- Is it true it doesn't return 0 anymore? It uses `foldGauges()` which sets the result to 0 when it's NaN.

metrics/tenant_registries.go
krajorama: ok, now I get it. That makes sense. Although that means that to test the same behavior with MAX, you need to set one user's gauge to -100 and not set a gauge on the other , otherwise the bug with the 0 does not show up.

pstibrany:
> [not a blocker] I'm not super convinced about stop returning a metric once there are no more tenants with that gauge. After this change, when we'll query that metric we'll keep getting the last value for the lookback period (5m) which may be annoying. I'm wondering if we could return the StaleNaN instead.

That's not what's happening :( I see that my changelog entry is super-confusing (but also the code is quite confusing, there are many different layers of it)

The test `TestSendMinMaxOfGauges` is showing current behaviour. It has two users with their metric registries registered in the `TenantRegistries`.

Scenario 1: At first both users don't have the gauge.
Expected results:
- `SendMinOfGauges` returns 0
- `SendMaxOfGauges` returns 0
- `SendMaxOfGaugesPerTenant` returns nothing. **This is a change: previously, `SendMaxOfGaugesPerTenant` would report 0 for the missing gauge.**

Scenario 2: First user gets the gauge registered into their registry, with value 100.
- `SendMinOfGauges` returns 100. **This is a change, before this PR, `SendMinOfGauges` would report 0.**
- `SendMaxOfGauges` returns 100.
- `SendMaxOfGaugesPerTenant` now returns gauge only for user1. **This is a change, previously user2 would also get the gauge reported with value 0.**

Scenario 3: Second user gets the gauge registered, with value 80.
- `SendMinOfGauges` now returns 80.
- `SendMaxOfGauges` still returns 100.
- `SendMaxOfGaugesPerTenant` returns gauge for both users, with their respective values.

Test continues with removal of user1 and then user2.

> > MetricFamiliesPerTenant.SendMinOfGauges will no longer return 0 if any tenant doesn't have the gauge.

This changelog entry is trying to explain change in `SendMinOfGauges` tested in scenario 2.

> 
> A couple of things:
> 
> * The PR description should mention SendMaxOfGauges() too

`SendMaxOfGauges` isn't really changed, it works as before. Update: well, in fact if per-tenant gauge values are negative, then there's a change. Previously we would return 0 if some tenant didn't have the gauge. Now we will correctly report negative value, and ignore tenant without the gauge. I'll add test for this and include it in the changelog.

> * Is it true it doesn't return 0 anymore? It uses `foldGauges()` which sets the result to 0 when it's NaN.

- single-tenant version (`MetricFamilyMap.MinGauges` and `MetricFamilyMap.MaxGauges`) return NaN
- while multi-tenant version (`MetricFamiliesPerTenant.SendMinOfGauges` and `MetricFamiliesPerTenant.SendMaxOfGauges`) send gauge with 0 value in that case.

metrics/tenant_registries.go
pstibrany: Done in https://github.com/grafana/dskit/pull/368/commits/661d25d7e6718d7ce9fa62e69c82b6b8c6957cb7.

metrics/tenant_registries.go
pstibrany: Done in https://github.com/grafana/dskit/pull/368/commits/661d25d7e6718d7ce9fa62e69c82b6b8c6957cb7.

pracucci (approved):
Thanks for the detailed explanation. LGTM!

pstibrany:
Thank you for reviews!

PR #377, opened 2023-09-14
gRPC server: add ability to reject methods early, before reading the request into memory

pstibrany:
**What this PR does**:

This PR introduces adds a way to reject calls to gRPC methods early, right after gRPC server receives request headers, but before reading the request body and starting a goroutine.

This PR is generalization of Mimir's https://github.com/grafana/mimir/pull/5976. Description in that PR explains this approach and why it works.

**Checklist**
- [x] Tests updated

### DIFF
--- a/server/limits.go
+++ b/server/limits.go
@@ -14,12 +14,12 @@ import (
 )
 
 type GrpcMethodLimiter interface {
-	// RpcCallStarting is called before request has been read into memory.
+	// RPCCallStarting is called before request has been read into memory.
 	// All that's known about the request at this point is grpc method name.
 	// Returned error should be convertible to gRPC Status via status.FromError,
 	// otherwise gRPC-server implementation-specific error will be returned to the client (codes.PermissionDenied in grpc@v1.55.0).
-	RpcCallStarting(methodName string) error
-	RpcCallFinished(methodName string)
+	RPCCallStarting(methodName string) error
+	RPCCallFinished(methodName string)
 }
 
 // Custom type to hide it from other packages.
@@ -92,7 +92,7 @@ func (g *grpcLimitCheck) TapHandle(ctx context.Context, info *tap.Info) (context
 	}
 
 	if g.methodLimiter != nil {
-		if err := g.methodLimiter.RpcCallStarting(info.FullMethodName); err != nil {
+		if err := g.methodLimiter.RPCCallStarting(info.FullMethodName); err != nil {
 			return ctx, err
 		}
 	}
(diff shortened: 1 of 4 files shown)


server/limits.go
```diff
@@ -0,0 +1,139 @@
...
+// Presence of this key in the context indicates that inflight request counter was increased for this request, and needs to be decreased when request ends.
+const (
+	requestFullMethod grpcLimitCheckContextKey = 1
+
+	errTooManyInflightRequestsMsg = "too many inflight requests"
+)
+
+var errTooManyInflightRequests = status.Error(codes.Unavailable, errTooManyInflightRequestsMsg)
+
+func newGrpcLimitCheck(maxInflight int, methodLimiter GrpcMethodLimiter, namespace string, reg prometheus.Registerer) *grpcLimitCheck {
```
pracucci, line 37: [nit] Have you considered calling it "... inflight limit" instead of just "limit"?

server/server.go
```diff
@@ -118,6 +118,7 @@ type Config struct {
 	GRPCServerTimeout                  time.Duration `yaml:"grpc_server_keepalive_timeout"`
 	GRPCServerMinTimeBetweenPings      time.Duration `yaml:"grpc_server_min_time_between_pings"`
 	GRPCServerPingWithoutStreamAllowed bool          `yaml:"grpc_server_ping_without_stream_allowed"`
+	GRPCServerMaxInflightRequests      int           `yaml:"grpc_server_max_inflight_requests"`
```
pracucci, line 121: I'm not super convinced of having a global config. I think in practice it's more useful to let the upstream project to have control over it (e.g. limit only specific methods). Have you considered designing so that we just configure a set of 0+ limiters, each limiter being able to limit a configurable set of methods? I'm asking because in Mimir ingester I would like to limit only the push gRPC and not every gRPC call with a "global" limiter.

pstibrany: > Have you considered designing so that we just configure a set of 0+ limiters, each limiter being able to limit a configurable set of methods? I'm asking because in Mimir ingester I would like to limit only the push gRPC and not every gRPC call with a "global" limiter.

This PR allows you to do exactly that already. There's global inflight limit, which I think can be useful but can also be disabled, and there's also a possibility to limit specific methods by using `GrpcMethodLimiter`. I imagine that Mimir would install its own `GrpcMethodLimiter` to implement ingesters's inflight limit.

pracucci: I would just remove the global one and let the upstream project to control it via `GrpcMethodLimiter`. WDYT?

pstibrany: > I would just remove the global one and let the upstream project to control it via `GrpcMethodLimiter`. WDYT?

OK, let's do that.

pstibrany: I've removed support for max inflight requests in gRPC server, and only added limiting via `GrpcInflightMethodLimiter`. Please take a look again.

pracucci (approved):
Thanks for addressing my feedback. LGTM!

PR #387, opened 2023-09-25
Include ring instance ID in InstanceDesc

56quarters:
**What this PR does**:

This change adds the ID on an instance in a ring to its InstanceDesc object. This allows callers to access the ID of a member of a ring.

This change also modifies the PoolFactory type to accept an InstanceDesc object when creating a new client instead of only the address. This allows PoolFactory implementations to attach any extra metadata about the instance to metrics/logs/traces for the client.

See grafana/mimir#6128

**Checklist**
- [X] Tests updated


pstibrany:
This looks like a good start, but the change will only apply to new entries that get added to the ring. What about existing entries?

I wonder if we should add some "fixup" method to the `*ring.Desc` that would fill all `ID` fields based on instances map, and call that `fixup` method after reading ring from the KV store (in `(*Ring).starting`, `(*Ring).loop`, or `GetOrCreateRingDesc`). This would also fix writes back to KV, because writes always read from ring first, and write back afterwards.

56quarters:
> This looks like a good start, but the change will only apply to new entries that get added to the ring. What about existing entries?

Actually from what I can tell looking at the ring code, it seems like pods will add themselves to the ring on start up, updating their existing entry if required. If we make sure ID is set when adding themselves, do we still need to go through and update the `ring.Desc`? IDs for entries in the ring would become available after the first time they restart with this new code if I'm understanding correctly.

56quarters:
I guess explicitly fixing up the `ring.Desc` means that we have a shorter window of time where `InstanceDesc.Id` is empty during the first rollout.

pstibrany:
> I guess explicitly fixing up the `ring.Desc` means that we have a shorter window of time where `InstanceDesc.Id` is empty during the first rollout.

It would mean that any code using updated dskit can already start using `ID` field, without relying on other components to re-register in the ring.

ring/ring.go
```diff
@@ -307,6 +307,11 @@ func (r *Ring) updateRingState(ringDesc *Desc) {
 		}
 	}
 
+	// Ensure the ID of each InstanceDesc is set based on the map of instances. This
+	// handles the case where some components are running older lifecyclers which do
+	// not set the instance ID when registering in the ring.
+	ringDesc.setInstanceIDs()
```
pstibrany, line 313: This requires updated dskit on *writer* side, instead of *reader* side. This can confuse readers expecting instance ID already.

56quarters:
I've tested setting instance IDs based on map keys locally with Mimir:

* Ingesters ran Mimir 2.10 which does not have this change
* All other components ran local image that included these changes

Everything worked as expected, distributors had access to the ingester IDs when creating clients for the ingesters.

I'm going to test in a dev cluster, only using this new code in distributors.

ring/ring.go
56quarters: I think that's backwards. This struct (`Ring`) is what is in-memory in distributors for the ingester ring. The previous iteration (https://github.com/grafana/dskit/pull/387/commits/821bbdd99927ea0727682a5144d16399d93d2ab8) was wrong and did require updated code on the writer side but this version is the reader side unless I'm missing something.

pstibrany: You're correct. I've got confused by commit message ("Set instance IDs when updating Ring, not when loading from KV store") and then haven't checked the actual code change. My bad.

PR #393, opened 2023-10-06
Randomize ports when running server tests.

pstibrany:
Randomize ports when running server tests. This will hopefully fix some flaky tests.

```
--- FAIL: TestStopWithDisabledSignalHandling (0.00s)
    --- FAIL: TestStopWithDisabledSignalHandling/signals_disabled (0.00s)
        server_test.go:800: 
            	Error Trace:	/drone/src/server/server_test.go:800
            	            				/drone/src/server/server_test.go:819
            	            	listen tcp 127.0.0.1:9198: bind: address already in use
            	Test:       	TestStopWithDisabledSignalHandling/signals_disabled
```

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -11,6 +11,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"flag"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


server/server_test.go
```diff
@@ -831,3 +801,20 @@ func (dh dummyHandler) Loop() {
 func (dh dummyHandler) Stop() {
 	close(dh.quit)
 }
+
+func setRandomPorts(t *testing.T, network string, cfg *Config) {
```
pracucci, line 805: [nit] I would clarify it.

```suggestion
func setAutoAssignedPorts(t *testing.T, network string, cfg *Config) {
```

pracucci (approved):
Nice job, LGTM!

server/server_test.go
pstibrany: Thanks, that's a better name.

Issue #381, opened 2023-09-15
Lots of tests listen on 0.0.0.0


kevinburkesegment:
Many of the tests in the test suite attempt to bind to a random port on 0.0.0.0, which means anything can make a connection to the server, including external devices. 

On Mac, this is very annoying because each invocation triggers a dialog box asking to allow connections to the test binary. I counted about ten or so of these dialog boxes in one run of the test suite.

To track down this issue, edit `$GOROOT/src/net/dial.go`, and add the following patch to `ListenConfig.Listen`:

```patch
 // See func Listen for a description of the network and address
 // parameters.
 func (lc *ListenConfig) Listen(ctx context.Context, network, address string) (Listener, error) {
+	if address == ":0" {
+		fmt.Println("ListenConfig.Listen", network, address)
+		debug.PrintStack()
+	}
 	addrs, err := DefaultResolver.resolveAddrList(ctx, "listen", network, address, nil)
```

Then rerun the tests. This will likely show you which test file is responsible for opening a socket to 0.0.0.0.

pracucci:
Would listening on `127.0.0.1` instead of `0.0.0.0` fix the issue?

kevinburkesegment:
Yes - here's an example CL submitting a patch: https://github.com/grafana/loki/pull/10411

Some machines map "localhost" to a different IP than 127.0.0.1 but it's sufficient for most machines to dial this host.

aknuds1:
Yep, please do open a PR :+1:

PR #396, opened 2023-10-09
kv/memberlist: fix incorrect TCP transport host parsing

kevinburkesegment:
Previously, if you passed "localhost" as the bind address for a TCPTransport, we would attempt to parse this as an IP address, fail, and then begin listening on 0.0.0.0.

This is a security issue since 0.0.0.0 binds to all interfaces, including public ones, while "localhost" is a shortcut for an IP address that is typically is only accessible from the local machine, not the wider Internet. A user who intended to bind only to the loopback interface might inadvertently expose a server to the public Internet.

Fix this by returning an error if you attempt to pass a BindAddr to TCPTransport that is not actually an IP address. Also, fix the tests to resolve "localhost" to an IP address before proceeding - typically, but not always, this is 127.0.0.1, which is why we try to parse the loopback address instead of hardcoding it.

I can confirm that on a Mac, with this patch applied I no longer get dialog boxes warning me that the tests are attempting to listen on 0.0.0.0.

Updates #381.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,3 +172,4 @@
 * [BUGFIX] `MetricFamiliesPerTenant.SendMaxOfGauges` will no longer return 0 if some tenant doesn't have the gauge in their registry, and other tenants have negative values only. #368
 * [BUGFIX] `MetricFamiliesPerTenant.SendMaxOfGaugesPerTenant` no longer includes tenants, which do not have specified gauge. #368
 * [BUGFIX] Correctly format `host:port` addresses when using IPv6. #388
+* [BUGFIX] TCPTransport will reject BindAddr entries that are not an IP address, like "localhost" (previously, these would listen on 0.0.0.0) #396
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,4 +172,4 @@
 * [BUGFIX] `MetricFamiliesPerTenant.SendMaxOfGauges` will no longer return 0 if some tenant doesn't have the gauge in their registry, and other tenants have negative values only. #368
 * [BUGFIX] `MetricFamiliesPerTenant.SendMaxOfGaugesPerTenant` no longer includes tenants, which do not have specified gauge. #368
 * [BUGFIX] Correctly format `host:port` addresses when using IPv6. #388
-* [BUGFIX] TCPTransport will reject BindAddr entries that are not an IP address, like "localhost" (previously, these would listen on 0.0.0.0) #396
+* [BUGFIX] Memberlist's TCP transport will now reject bind addresses that are not IP addresses, such as "localhost", rather than silently converting these to 0.0.0.0 and therefore listening on all addresses. #396
--- a/kv/memberlist/memberlist_client_test.go
+++ b/kv/memberlist/memberlist_client_test.go
@@ -234,17 +234,22 @@ func casWithErr(ctx context.Context, kv *Client, key string, updateFn func(*data
 	return kv.CAS(ctx, key, fn)
 }
 
-var localhostIP *net.IPAddr
-var localhostBindAddrs []string
-var localhostBindAddrsOnce sync.Once
+func getLocalhostAddr() string {
+	return getLocalhostAddrs()[0]
+}
 
-func setLocalhostBindAddrs() {
-	var err error
-	localhostIP, err = net.ResolveIPAddr("ip4", "localhost")
-	if err != nil {
-		localhostBindAddrs = []string{"127.0.0.1"}
-	}
-	localhostBindAddrs = []string{localhostIP.String()}
+var addrsOnce sync.Once
+var localhostIP string
+
+func getLocalhostAddrs() []string {
+	addrsOnce.Do(func() {
+		ip, err := net.ResolveIPAddr("ip4", "localhost")
+		if err != nil {
+			localhostIP = "127.0.0.1" // this is the most common answer, try it
+		}
+		localhostIP = ip.String()
+	})
+	return []string{localhostIP}
 }
 
 func TestBasicGetAndCas(t *testing.T) {
@@ -253,9 +258,8 @@ func TestBasicGetAndCas(t *testing.T) {
 	name := "Ing 1"
 	var cfg KVConfig
 	flagext.DefaultValues(&cfg)
-	localhostBindAddrsOnce.Do(setLocalhostBindAddrs)
 	cfg.TCPTransport = TCPTransportConfig{
-		BindAddrs: localhostBindAddrs,
+		BindAddrs: getLocalhostAddrs(),
 	}
 	cfg.Codecs = []codec.Codec{c}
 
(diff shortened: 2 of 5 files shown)


kv/memberlist/tcp_transport.go
```diff
@@ -147,6 +147,9 @@ func NewTCPTransport(config TCPTransportConfig, logger log.Logger, registerer pr
 	port := config.BindPort
 	for _, addr := range config.BindAddrs {
 		ip := net.ParseIP(addr)
+		if ip == nil {
```
kevinburkesegment, line 150: Another option would be to special case "localhost" here, and resolve that IP address specifically but no others.

kv/memberlist/memberlist_client_test.go
```diff
@@ -989,14 +1007,15 @@ func runClient(kv *Client, name string, ringKey string, portToConnect int, casIn
...
+	localhostBindAddrsOnce.Do(setLocalhostBindAddrs)
 	for {
 		select {
 		case <-start:
 			start = nil
 
 			// let's join the first member
 			if portToConnect > 0 {
-				_, err := kv.kv.JoinMembers([]string{fmt.Sprintf("127.0.0.1:%d", portToConnect)})
+				_, err := kv.kv.JoinMembers([]string{localhostIP.String() + fmt.Sprintf(":%d", portToConnect)})
```
kevinburkesegment, line 1018: Likely going to be a conflict between this and the "join host port" PR, I will resolve once that is merged.

kv/memberlist/tcp_transport.go
charleskorn: I think having no special cases here is fine.

kv/memberlist/memberlist_client_test.go
```diff
@@ -234,14 +234,28 @@ func casWithErr(ctx context.Context, kv *Client, key string, updateFn func(*data
...
+var localhostBindAddrsOnce sync.Once
+
+func setLocalhostBindAddrs() {
+	var err error
+	localhostIP, err = net.ResolveIPAddr("ip4", "localhost")
+	if err != nil {
+		localhostBindAddrs = []string{"127.0.0.1"}
+	}
+	localhostBindAddrs = []string{localhostIP.String()}
+}
```
charleskorn, line 248: Is this a particularly expensive operation? I'm wondering if we can simplify the code a little and just have a `getLocalhostAddr()` method that resolves the address each time.

kv/memberlist/tcp_transport.go
```diff
@@ -147,6 +147,9 @@ func NewTCPTransport(config TCPTransportConfig, logger log.Logger, registerer pr
 	port := config.BindPort
 	for _, addr := range config.BindAddrs {
 		ip := net.ParseIP(addr)
```
charleskorn, line 149: Could you please add a test case for the scenario where `addr` is not a valid IP address, to ensure we don't re-introduce this regression in the future?

CHANGELOG.md
```diff
@@ -172,3 +172,4 @@
 * [BUGFIX] `MetricFamiliesPerTenant.SendMaxOfGauges` will no longer return 0 if some tenant doesn't have the gauge in their registry, and other tenants have negative values only. #368
 * [BUGFIX] `MetricFamiliesPerTenant.SendMaxOfGaugesPerTenant` no longer includes tenants, which do not have specified gauge. #368
 * [BUGFIX] Correctly format `host:port` addresses when using IPv6. #388
+* [BUGFIX] TCPTransport will reject BindAddr entries that are not an IP address, like "localhost" (previously, these would listen on 0.0.0.0) #396
```
charleskorn, line 175: I think we should clarify the impact of this a bit more: 
```suggestion
* [BUGFIX] Memberlist's TCP transport will now reject bind addresses that are not IP addresses, such as "localhost", rather than silently converting these to 0.0.0.0 and therefore listening on all addresses. #396
```

charleskorn (commented):
Thanks for spotting and fixing this @kevinburkesegment!

kv/memberlist/memberlist_client_test.go
kevinburkesegment: I mean it's a DNS lookup so guessing about a millisecond. Let me see.

kv/memberlist/tcp_transport.go
kevinburkesegment: Done

kv/memberlist/tcp_transport_test.go
```diff
@@ -95,3 +95,15 @@ func TestFinalAdvertiseAddr(t *testing.T) {
...
+func TestNonIPsAreRejected(t *testing.T) {
+	cfg := TCPTransportConfig{BindAddrs: flagext.StringSlice{"localhost"}}
+	_, err := NewTCPTransport(cfg, nil, nil)
+	if err == nil {
+		t.Fatal("expected non-nil err, got nil")
+	}
+	want := `could not parse bind addr "localhost" as IP address`
+	if err.Error() != want {
+		t.Fatalf("NewTCPTransport with non-IP address BindAddr: got %v, want %v", err, want)
+	}
```
charleskorn, line 108: We use the `require` package in other tests to make assertions like these.

kv/memberlist/memberlist_client_test.go
charleskorn: I think let's simplify this for now and if we notice it being a problem in the future, we can re-introduce this more optimised way of doing things.

kevinburkesegment: Just realized I can make it substantially simpler with a helper method, I can also back out the sync.Once, let me know what you think

kv/memberlist/tcp_transport_test.go
```diff
@@ -95,3 +95,10 @@ func TestFinalAdvertiseAddr(t *testing.T) {
 		})
 	}
 }
+
+func TestNonIPsAreRejected(t *testing.T) {
+	cfg := TCPTransportConfig{BindAddrs: flagext.StringSlice{"localhost"}}
+	_, err := NewTCPTransport(cfg, nil, nil)
+	require.Error(t, err)
+	require.Equal(t, err.Error(), `could not parse bind addr "localhost" as IP address`)
```
charleskorn, line 103: You should be able to simplify this even further:

```suggestion
	require.EqualError(t, err, `could not parse bind addr "localhost" as IP address`)
```

kv/memberlist/memberlist_client_test.go
charleskorn: Looks good to me, thanks

kv/memberlist/tcp_transport_test.go
kevinburkesegment: Done

kv/memberlist/tcp_transport_test.go
kevinburkesegment: Done

kv/memberlist/tcp_transport_test.go
```diff
@@ -95,3 +95,10 @@ func TestFinalAdvertiseAddr(t *testing.T) {
 		})
 	}
 }
+
+func TestNonIPsAreRejected(t *testing.T) {
+	cfg := TCPTransportConfig{BindAddrs: flagext.StringSlice{"localhost"}}
+	_, err := NewTCPTransport(cfg, nil, nil)
+	require.Error(t, err)
```
charleskorn, line 102: This `require.Error()` call is no longer necessary - `EqualError` will fail if `err` is `nil`.

kevinburkesegment: Ahhh, yes. This should be fixed now

charleskorn (approved):

charleskorn:
Thanks @kevinburkesegment!

PR #397, opened 2023-10-09
crypto/tls/test: server listens on localhost

kevinburkesegment:
Previously, the test server would bind to all interfaces, which leads to a very annoying error message on Mac computers, and may expose the server to public access which can only be a bad thing, in a test context.

Updates #381.

### DIFF
--- a/crypto/tls/test/tls_integration_test.go
+++ b/crypto/tls/test/tls_integration_test.go
@@ -56,21 +56,21 @@ func (h *grpcHealthCheck) Watch(_ *grpc_health_v1.HealthCheckRequest, _ grpc_hea
 	return status.Error(codes.Unimplemented, "Watching is not supported")
 }
 
-func getLocalHostPort() (int, error) {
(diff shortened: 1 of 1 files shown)


kevinburkesegment:
I believe the build failure is a flake: https://github.com/grafana/dskit/issues/398

crypto/tls/test/tls_integration_test.go
```diff
@@ -55,21 +55,21 @@ func (h *grpcHealthCheck) Watch(_ *grpc_health_v1.HealthCheckRequest, _ grpc_hea
 	return status.Error(codes.Unimplemented, "Watching is not supported")
 }
 
-func getLocalHostPort() (int, error) {
+func getLocalHostPort() (*net.TCPAddr, error) {
```
charleskorn, line 58: Should we rename this method now that it's no longer returning only the port? `getLocalHostAddr` perhaps?

charleskorn (approved):

PR #403, opened 2023-10-13
`DoUntilQuorum`: don't use non-zone-aware logging when all zones are required for quorum

charleskorn:
**What this PR does**:

This PR fixes an issue where `DoUntilQuorum` may generate more log messages and span events than expected when zone-awareness is enabled but all healthy zones are required for quorum. (For example, zone-awareness is enabled, three zones are running with 100 instances each, but one instance in the first zone is unavailable, so all instances in the other two zones are required for quorum.)

Previously, if zone-awareness was enabled and all healthy zones were required for quorum, both `ReplicationSet.MaxErrors` and `ReplicationSet.MaxUnavailableZones` would be 0. `DoUntilQuorum` would then default to non-zone-aware mode. In non-zone-aware mode, `DoUntilQuorum` logs a `starting request to instance` message for every instance.

In contrast, in zone-aware mode, `DoUntilQuorum` logs a `starting requests to zone` message for each unique zone.

In the example above, with all instances healthy, `DoUntilQuorum` would log two or three `starting requests to zone` messages, but as soon as one instance becomes unhealthy, a subsequent `DoUntilQuorum` call would log 200 `starting request to instance` messages. 

This PR changes the behaviour of `DoUntilQuorum` to always run in zone-aware mode if zone-awareness is enabled, even if both `ReplicationSet.MaxErrors` and `ReplicationSet.MaxUnavailableZones` are 0.

(none)

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/replication_set.go
+++ b/ring/replication_set.go
@@ -208,7 +208,7 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
 	}
 
 	if r.ZoneAwarenessEnabled && r.MaxErrors > 0 {
-		return nil, fmt.Errorf("invalid ReplicationSet: MaxErrors is non-zero (is %v) but ZoneAwarenessEnabled is true", r.MaxErrors)
+		return nil, fmt.Errorf("invalid ReplicationSet: MaxErrors is non-zero (is %v) and ZoneAwarenessEnabled is true", r.MaxErrors)
 	}
 
 	var logger kitlog.Logger = cfg.Logger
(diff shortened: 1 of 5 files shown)


ring/replication_set.go
```diff
@@ -204,6 +207,10 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
 		return nil, err
 	}
 
+	if r.ZoneAwarenessEnabled && r.MaxErrors > 0 {
+		return nil, fmt.Errorf("invalid ReplicationSet: MaxErrors is non-zero (is %v) but ZoneAwarenessEnabled is true", r.MaxErrors)
```
jhesketh, line 211: (nit)

```suggestion
		return nil, fmt.Errorf("invalid ReplicationSet: MaxErrors is non-zero (is %v) and ZoneAwarenessEnabled is true", r.MaxErrors)
```

jhesketh (approved):
lgtm

PR #404, opened 2023-10-16
DoUntilQuorum: terminate early if a terminal error is received

charleskorn:
**What this PR does**:

This PR adds support for the concept of "terminal errors" to `DoUntilQuorum`.

This idea is best illustrated with an example. In Mimir, queriers use `DoUntilQuorum` to query ingesters for the chunks needed to evaluate a query. Queriers implement a limit on the maximum number of chunks a query can fetch. If a query fetches more chunks than is allowed by the limit, the query fails.

At present, if request minimisation is enabled and the chunks limit is reached as a result of the responses to requests to the first two of three zones, `DoUntilQuorum` will treat the limit error as a failure of just that zone and initiate requests to the third zone. This is pointless: making requests to the third zone will not resolve the issue, and these requests waste time and compute resources for a response that will not be used.

With this PR, queriers could set `IsTerminalError` to a function that returns `true` whenever a limit error is returned, which will cause `DoUntilQuorum` to abort and return immediately in this case.

(none)

**Checklist**
- [x] Tests updated


ring/replication_set.go
```diff
@@ -109,6 +109,15 @@ type DoUntilQuorumConfig struct {
 
 	// If non-nil, DoUntilQuorum will emit log lines and span events during the call.
 	Logger *spanlogger.SpanLogger
+
+	// If non-nil and a request returns an error for which IsTerminalError returns true,
```
lamida, line 113: Does non-nil refer to error or to request?

ring/replication_set.go
```diff
@@ -300,6 +319,13 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
 			resultTracker.startAdditionalRequests()
 		case result := <-resultsChan:
 			resultsRemaining--
+
+			if result.err != nil && cfg.IsTerminalError != nil && cfg.IsTerminalError(result.err) {
```
lamida, line 323: Wouldn't that be better if `IsTerminalError` also handles `err != nil` part? Hence we can reduce one conditional `result.err != nil` from the if condition.

ring/replication_set.go
charleskorn: I'll clarify this - "If non-nil" at the start refers to `IsTerminalError` being non-nil.

ring/replication_set.go
charleskorn: Maybe this is just me, but as a consumer of `DoUntilQuorum`, I wouldn't expect `IsTerminalError` to be called with a `nil` error, as `nil` means "no error" and therefore it can't be a terminal error.

charleskorn: This also means that implementations of `IsTerminalError` don't need to check for `nil` errors, which simplifies their implementation slightly.

ring/replication_set.go
charleskorn: Resolved in fc2a12f

lamida (approved):

ring/replication_set.go
```diff
@@ -279,6 +292,16 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
 		}
 	}
 
+	terminate := func(err error) ([]T, error) {
+		if cfg.Logger != nil {
+			_ = cfg.Logger.Error(err)
+		}
```
pracucci, line 298: [nit] I think you can just use `logger`.

ring/replication_set.go
```diff
@@ -300,6 +323,13 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
 			resultTracker.startAdditionalRequests()
 		case result := <-resultsChan:
 			resultsRemaining--
+
+			if result.err != nil && cfg.IsTerminalError != nil && cfg.IsTerminalError(result.err) {
+				level.Error(logger).Log("msg", "cancelling all outstanding requests because a terminal error occurred", "err", result.err)
```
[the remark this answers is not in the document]
pracucci: Also, it's getting logged by `terminate()`. Is a double log what we want?

ring/replication_set.go
charleskorn: `logger` only supports logging, whereas we want to set the error flag on the span here.

PR #406, opened 2023-10-16
Allow configuration of GRPC client middleware.

jeschkies:
**What this PR does**:
Similar to the server [config](https://github.com/grafana/dskit/blob/dcd7c1e06a38c25d44664a91cc8794b0d557b77d/server/server.go#L105-L106) one should be able to configure GRPC interceptors for the clients. Loki's ingester client [config](https://github.com/grafana/loki/blob/1ac7410b78416e25179b4fe589e9035793d67025/pkg/ingester/client/client.go#L45-L46) allows this but it should be available for any client.

**Checklist**
- [ ] Tests updated


grpcclient/grpcclient.go
```diff
@@ -37,6 +37,9 @@ type Config struct {
 	// https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md
 	ConnectBackoffBaseDelay time.Duration `yaml:"connect_backoff_base_delay" category:"advanced"`
 	ConnectBackoffMaxDelay  time.Duration `yaml:"connect_backoff_max_delay" category:"advanced"`
+
+	GRPCMiddleware       []grpc.UnaryClientInterceptor  `yaml:"-"`
+	GRPCStreamMiddleware []grpc.StreamClientInterceptor `yaml:"-"`
```
jeschkies, line 42: We could also call it `Middleware` to avoid repeating `GRPCClientConfig.GRPCMiddleware`.

grpcclient/grpcclient.go
```diff
@@ -100,6 +103,9 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 	}
 	opts = append(opts, tlsOpts...)
 
+	unaryClientInterceptors = append(unaryClientInterceptors, cfg.GRPCMiddleware...)
```
56quarters, line 106: Could you document how the configured interceptors are ordered with respect to the ones passed to this method?

grpcclient/grpcclient.go
56quarters: I think calling it `Middleware` makes sense since this is already a `grpc*` package but I don't feel strongly about it.

56quarters (approved):

PR #408, opened 2023-10-17
Use slices.BinarySearch in util.searchToken

pr00se:
**What this PR does**:

The `slices` package introduces a generic `slices.BinarySearch` which looks to be faster than `sort.Search`, especially as the number of items to search increases.

Benchmark:
```
go test -run='TestSearch' -bench='BenchmarkSearch' -timeout='3h' -benchtime='10000x' -cpu=1 -count=10

name                                   old time/op  new time/op  delta
SearchToken/searchToken_3_instances    89.2ns ± 3%  78.2ns ± 3%  -12.36%  (p=0.000 n=9+8)
SearchToken/searchToken_9_instances     100ns ± 3%    87ns ± 3%  -13.03%  (p=0.000 n=10+10)
SearchToken/searchToken_27_instances    107ns ± 4%    92ns ± 4%  -13.28%  (p=0.000 n=10+9)
SearchToken/searchToken_81_instances    138ns ± 3%   121ns ± 3%  -12.25%  (p=0.000 n=9+10)
SearchToken/searchToken_243_instances   187ns ± 6%   149ns ± 3%  -20.47%  (p=0.000 n=10+9)
SearchToken/searchToken_729_instances   313ns ± 4%   215ns ± 3%  -31.35%  (p=0.000 n=10+9)
```

(none)

**Checklist**
- [ ] Tests updated

### DIFF
--- a/ring/util_test.go
+++ b/ring/util_test.go
@@ -13,7 +13,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/exp/slices"
 )
 
 type RingMock struct {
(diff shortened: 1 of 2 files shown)


ring/util_test.go
```diff
@@ -424,3 +425,32 @@ func TestGetTokenDistance(t *testing.T) {
 		require.Equal(t, testData.expected, distance)
 	}
 }
+
+func TestSearchToken(t *testing.T) {
```
56quarters, line 429: Have you verified that this behavior matches the old implementation?

56quarters (approved):
LGTM, just one question about the test.

ring/util_test.go
pr00se: I did, but just to triple-check myself since this is a core part of the code, https://github.com/grafana/dskit/pull/408/commits/9cec3d0c18161df552091809b7e31ad1cea7c96e rolls back to the old function, and adds a test that compares the results of the new and old functions over 1M random keys. All tests passed [here](https://drone.grafana.net/grafana/dskit/3289/1/4). https://github.com/grafana/dskit/pull/408/commits/1275009a299573b0bfc3a88f97d5aeb016e2fd78 reverts those changes.

pracucci (approved):
LGTM! I checked the actual `sort.Search()` vs `slices.SearchBinary()` implementation and I've also written this simple comparison test:

```
func TestSearchToken_EnsureBinarySearchLogicIsEqualToSortSearch(t *testing.T) {
	tokens := []uint32{1, 5, 6, 8, 10}

	searchOld := func(target uint32) int {
		return sort.Search(len(tokens), func(x int) bool {
			return tokens[x] > target
		})
	}

	searchNew := func(target uint32) int {
		i, found := slices.BinarySearch(tokens, target)
		if found {
			// we want the first token > key, not >= key
			i = i + 1
		}
		return i
	}

	for i := uint32(0); i < tokens[len(tokens)-1]+2; i++ {
		assert.Equal(t, searchOld(i), searchNew(i))
	}
}
```

PR #410, opened 2023-10-18
Server: Raise default max-concurrent-streams following change in gRPC

bboreham:
**What this PR does**:

gRPC release 1.58.3 changed the meaning of this setting from a per-client limit to a server-wide limit, thus 100 is very low as a default.

Depending on what your server does, any default will be wrong, but I propose 10,000 as a level which should not constrict most people, which was the intention of the previous default of 100 per client.

https://github.com/grpc/grpc-go/releases/tag/v1.58.3

**Checklist**
- NA Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -173,7 +173,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.HTTPServerIdleTimeout, "server.http-idle-timeout", 120*time.Second, "Idle timeout for HTTP server")
 	f.IntVar(&cfg.GPRCServerMaxRecvMsgSize, "server.grpc-max-recv-msg-size-bytes", 4*1024*1024, "Limit on the size of a gRPC message this server can receive (bytes).")
 	f.IntVar(&cfg.GRPCServerMaxSendMsgSize, "server.grpc-max-send-msg-size-bytes", 4*1024*1024, "Limit on the size of a gRPC message this server can send (bytes).")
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -173,7 +173,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.HTTPServerIdleTimeout, "server.http-idle-timeout", 120*time.Second, "Idle timeout for HTTP server")
 	f.IntVar(&cfg.GPRCServerMaxRecvMsgSize, "server.grpc-max-recv-msg-size-bytes", 4*1024*1024, "Limit on the size of a gRPC message this server can receive (bytes).")
 	f.IntVar(&cfg.GRPCServerMaxSendMsgSize, "server.grpc-max-send-msg-size-bytes", 4*1024*1024, "Limit on the size of a gRPC message this server can send (bytes).")
-	f.UintVar(&cfg.GPRCServerMaxConcurrentStreams, "server.grpc-max-concurrent-streams", 100, "Limit on the number of concurrent streams for gRPC calls per client connection (0 = unlimited)")
+	f.UintVar(&cfg.GPRCServerMaxConcurrentStreams, "server.grpc-max-concurrent-streams", 10000, "Limit on the number of concurrent streams for gRPC calls (0 = unlimited)")
```
colega, line 176: Since the change is subtle and surprising, I'd make it more explicit saying:

```suggestion
	f.UintVar(&cfg.GPRCServerMaxConcurrentStreams, "server.grpc-max-concurrent-streams", 10000, "Limit on the number of concurrent streams for gRPC calls from all clients (0 = unlimited).")
```

bboreham:
It seems I misunderstood the change - it is still a limit per-connection.

PR #411, opened 2023-10-18
memberlist: reresolve members with every 100 nodes on Join

dimitarvdimitrov:
### Description 

This change solves three problems:
1. Context cancelations aren't respected in Join. Sometimes join might take as long as 25 minutes and there is no way to cancel it. Now we check the context on every 100 joined nodes
2. `Join` will attempt to join member IPs that become obsolete between the time discoverMembers runs and memberlist initiates a push-pull with the node. Reresolve `JoinMembers` addresses after joining every 100 nodes. This helps to clean up obsolete IPs that will be tried at the end of the Join procedure
3. (minor) fast joining on startup doesn't respect context cancelations

The side effect of this change is that there will be more DNS resolutions for the JoinMembers.

**Note to reviewers:** `TestJoinMembersWithRetryBackoff` seems to sometimes fail locally, but it should be an unrelated failure. I believe it should be fixed by this https://github.com/grafana/dskit/pull/412.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -591,7 +591,7 @@ func (m *KV) joinMembersWithRetries(ctx context.Context, numAttempts int, logger
 			MaxRetries: numAttempts,
 		}
 		boff               = backoff.New(ctx, cfg)
-		err                = error(nil)
+		err                error
 		successfullyJoined = 0
 	)
 
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -582,7 +582,7 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
 
 // joinMembersWithRetries joins m.cfg.JoinMembers 100 at a time. After each batch of 100 it rediscoveres the members.
 // This helps when the list of members is big and by the time we reach the end the originally resolved addresses may be obsolete.
-// joinMembersWithRetries returns an error iff it couldn't successfully join any node.
+// joinMembersWithRetries returns an error iff it couldn't successfully join any node OR the context was cancelled.
 func (m *KV) joinMembersWithRetries(ctx context.Context, numAttempts int, logger log.Logger) (int, error) {
 	var (
 		cfg = backoff.Config{
@@ -658,12 +658,12 @@ func (m *KV) joinMembersInBatches(ctx context.Context) (int, error) {
 		successfullyJoined += joinedInBatch
 	}
 	if successfullyJoined > 0 {
-		lastErr = nil
+		return successfullyJoined, nil
 	}
 	if successfullyJoined == 0 && lastErr == nil {
 		return 0, errors.New("found no nodes to join")
 	}
-	return successfullyJoined, lastErr
+	return 0, lastErr
 }
 
 // joinMembersBatch returns an error only if it couldn't successfully join any nodes or if ctx is cancelled.
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -647,6 +647,7 @@ func (m *KV) joinMembersInBatches(ctx context.Context) (int, error) {
 				break
 			}
 			batch = append(batch, n)
+			attemptedNodes[n] = true
 		}
 
 		// Join batch
@@ -654,12 +655,7 @@ func (m *KV) joinMembersInBatches(ctx context.Context) (int, error) {
 		if err != nil {
 			lastErr = err
 		}
-
-		// Record joined nodes
 		successfullyJoined += joinedInBatch
-		for _, n := range batch {
-			attemptedNodes[n] = true
-		}
 	}
 	if successfullyJoined > 0 {
 		lastErr = nil
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -597,12 +597,8 @@ func (m *KV) joinMembersInBatches(ctx context.Context, numAttempts int, logger l
 	)
 
 	for ; boff.Ongoing(); boff.Wait() {
-		var (
-			attemptedNodes = make(map[string]bool) // On each attempt we want to retry all nodes.
-			joinedInBatch  int
-			err            error
-		)
-		for batchHasMore := true; boff.Ongoing() && batchHasMore; successfullyJoined += joinedInBatch {
+		attemptedNodes := make(map[string]bool) // On each attempt we want to retry all nodes.
+		for batchHasMore := true; boff.Ongoing() && batchHasMore; {
 			// Rediscover nodes and try to join a subset of them with each batch.
 			// When the list of nodes is large by the time we reach the end of the list some of the
 			// IPs can be unreachable.
@@ -613,10 +609,15 @@ func (m *KV) joinMembersInBatches(ctx context.Context, numAttempts int, logger l
 				break
 			}
 
+			var (
+				joinedInBatch int
+				err           error
+			)
 			batchHasMore, joinedInBatch, err = m.joinMembersBatch(nodes, joinNodesBatch, attemptedNodes)
 			if err != nil {
 				lastErr = err
 			}
+			successfullyJoined += joinedInBatch
 		}
 		if successfullyJoined > 0 {
 			// If there are _some_ successful joins, then we can consider the join done.
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -619,14 +619,17 @@ func (m *KV) joinMembersInBatches(ctx context.Context) (int, error) {
 		successfullyJoined = 0
 		lastErr            error
 		batch              = make([]string, batchSize)
+		nodes              []string
 	)
 	for moreAvailableNodes := true; ctx.Err() == nil && moreAvailableNodes; {
 		// Rediscover nodes and try to join a subset of them with each batch.
 		// When the list of nodes is large by the time we reach the end of the list some of the
 		// IPs can be unreachable.
-		nodes := m.discoverMembers(ctx, m.cfg.JoinMembers)
-		if len(nodes) == 0 {
-			return 0, errors.New("found no nodes to join")
+		newlyResolved := m.discoverMembers(ctx, m.cfg.JoinMembers)
+		if len(newlyResolved) > 0 {
+			// If the resolution fails we keep using the nodes list from the last resolution.
+			// If that failed too, then we fail the join attempt.
+			nodes = newlyResolved
 		}
 
 		// Prepare batch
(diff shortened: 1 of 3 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,88 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
+			MinBackoff: m.cfg.MinJoinBackoff,
+			MaxBackoff: m.cfg.MaxJoinBackoff,
+			MaxRetries: numAttempts,
+		}
+		boff               = backoff.New(ctx, cfg)
+		lastErr            = error(nil)
+		successfullyJoined = 0
+	)
+
+	for ; boff.Ongoing(); boff.Wait() {
```
pracucci, line 599: [nit] To keep the code easier to read, I would move the whole content of the `for` loop to the function `joinMembersInBatches()` and rename this function to `joinMembersInBatchesWithRetries()`. In `joinMembersInBatchers()` instead of checking `boff.Ongoing()` you could just check the `ctx.Err()`.

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,88 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
+		var (
+			attemptedNodes = make(map[string]bool) // On each attempt we want to retry all nodes.
+			joinedInBatch  int
+			err            error
+		)
+		for batchHasMore := true; boff.Ongoing() && batchHasMore; successfullyJoined += joinedInBatch {
```
pracucci, line 605: [nit] To make code easier to read, I move move `successfullyJoined += joinedInBatch` to the end of the for loop and define `joinedInBatch` inside the for loop itself. Keep it outside looks a bit an anti-pattern to me.

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,88 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
-	boff := backoff.New(ctx, cfg)
-	var lastErr error
+	return successfullyJoined, lastErr
+}
 
-	for boff.Ongoing() {
-		// We rejoin all nodes, including those that were joined during "fast-join".
-		// This is harmless and simpler.
-		nodes := m.discoverMembers(ctx, m.cfg.JoinMembers)
+func (m *KV) joinMembersBatch(nodes []string, batchSize int, attemptedNodes map[string]bool) (moreLeft bool, successfullyJoined int, lastErr error) {
```
pracucci, line 632: [nit] From a design perspective, passing all `nodes` and `attemptedNodes` looks a bit an anti-pattern to me. I don't think it's responsibility of the function joining a batch of nodes. The caller could build the list of nodes to join, and `joinMembersBatch()` just does it. Also I would take `ctx` in input here, so that you can break the loop as soon as context is cancelled, without waiting for all 100 nodes in the batch to be tried.

pracucci (approved):
I'm good with the changes, but I think the implementation is a bit more complicated (see it as "harder to follow") than required. A slightly better design may simplify it.

dimitarvdimitrov:
thank you for the review @pracucci. I think I addressed your comments. I decided to go with 3 functions: 1) manage retries, 2) manage batches, 3) join a single batch. The code become a bit longer, but I think it's more clear

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,125 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
-				level.Info(m.logger).Log("msg", "joining memberlist cluster succeeded", "reached_nodes", reached, "elapsed_time", time.Since(startTime))
-				return true
+		// Prepare batch
+		batch = batch[:0]
+		moreAvailableNodes = false
+		for _, n := range nodes {
+			if attemptedNodes[n] {
+				continue
+			}
+			if len(batch) >= batchSize {
```
pstibrany, line 645: nit: should we move this `if` after `batch = append(batch, n)`?

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,125 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
 			lastErr = err
-		} else {
-			level.Warn(m.logger).Log("msg", "joining memberlist cluster: found no nodes to join", "retries", boff.NumRetries())
 		}
 
-		boff.Wait()
+		// Record joined nodes
+		successfullyJoined += joinedInBatch
+		for _, n := range batch {
+			attemptedNodes[n] = true
```
pstibrany, line 661: nit: we can move this to the previous `for` loop, when we add node to batch.

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,125 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
 
-	level.Error(m.logger).Log("msg", "joining memberlist cluster failed", "last_error", lastErr, "elapsed_time", time.Since(startTime))
-	return false
+// joinMembersBatch returns an error only if it couldn't successfully join any nodes or if ctx is cancelled.
+func (m *KV) joinMembersBatch(ctx context.Context, nodes []string) (successfullyJoined int, lastErr error) {
+	for nodeIdx := range nodes {
+		if ctx.Err() != nil {
+			return successfullyJoined, fmt.Errorf("joining batch: %w", context.Cause(ctx))
+		}
+		// Attempt to join a single node. Complexity shouldn't be different from passing all the node IPs to Join.
```
pstibrany, line 679: > Complexity shouldn't be different from passing all the node IPs to Join.

What is this comment trying to say?

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,125 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
+		if successfullyJoined > 0 {
+			// If there are _some_ successful joins, then we can consider the join done.
+			// Mimicking the Join semantics we return an error only when we couldn't join any node at all
+			err = nil
+			break
+		}
+		level.Warn(logger).Log("msg", "joining memberlist cluster", "attempts", boff.NumRetries()+1, "max_attempts", numAttempts, "err", err)
+	}
+	if err == nil && ctx.Err() != nil {
+		err = fmt.Errorf("joining memberlist: %w", context.Cause(ctx))
```
pstibrany, line 609: Does setting `err` here possibly violate the comment saying "joinMembersWithRetries returns an error iff it couldn't successfully join any node."? (What if context is cancelled after we already joined some nodes?)

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,125 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
+// joinMembersWithRetries returns an error iff it couldn't successfully join any node.
+func (m *KV) joinMembersWithRetries(ctx context.Context, numAttempts int, logger log.Logger) (int, error) {
+	var (
+		cfg = backoff.Config{
+			MinBackoff: m.cfg.MinJoinBackoff,
+			MaxBackoff: m.cfg.MaxJoinBackoff,
+			MaxRetries: numAttempts,
+		}
+		boff               = backoff.New(ctx, cfg)
+		err                = error(nil)
```
pstibrany, line 594: nit:
```suggestion
		err                error
```

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,125 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
+		successfullyJoined, err = m.joinMembersInBatches(ctx)
+		if successfullyJoined > 0 {
+			// If there are _some_ successful joins, then we can consider the join done.
+			// Mimicking the Join semantics we return an error only when we couldn't join any node at all
+			err = nil
+			break
+		}
+		level.Warn(logger).Log("msg", "joining memberlist cluster", "attempts", boff.NumRetries()+1, "max_attempts", numAttempts, "err", err)
+	}
+	if err == nil && ctx.Err() != nil {
```
pstibrany, line 608: instead of `ctx.Err`, I would expect that we use `boff.Err()`.

kv/memberlist/memberlist_client.go
```diff
@@ -568,41 +568,125 @@ func (m *KV) joinMembersOnStartup(ctx context.Context) bool {
...
+	if successfullyJoined == 0 && lastErr == nil {
+		return 0, errors.New("found no nodes to join")
 	}
+	return successfullyJoined, lastErr
+}
 
-	level.Error(m.logger).Log("msg", "joining memberlist cluster failed", "last_error", lastErr, "elapsed_time", time.Since(startTime))
-	return false
+// joinMembersBatch returns an error only if it couldn't successfully join any nodes or if ctx is cancelled.
+func (m *KV) joinMembersBatch(ctx context.Context, nodes []string) (successfullyJoined int, lastErr error) {
```
pstibrany, line 674: suggestion: joinMembersBatch -> joinAllMembers

pstibrany (approved):
lgtm overall, comments are not blocking.

kv/memberlist/memberlist_client.go
dimitarvdimitrov: then we don't know whether there are `moreAvailableNodes`. If we have filled the batch size and there is another node which wasn't attempted already, then we know. This is covering the edge case where the batch fills with the last item in `nodes`

kv/memberlist/memberlist_client.go
dimitarvdimitrov: ah, good point

kv/memberlist/memberlist_client.go
dimitarvdimitrov: that it doesn't matter whether we call Join with all nodes once or call it with one node N times. I reworded the comment

kv/memberlist/memberlist_client.go
dimitarvdimitrov: that's right. I updated the comment

kv/memberlist/memberlist_client.go
dimitarvdimitrov: i prefer joinMembersBatch because it complements joinMembersInBatches and it's more clear that InBatches uses Batch

dimitarvdimitrov:
thanks for the reviews @pracucci, @pstibrany. I will merge this after CI passes

dimitarvdimitrov:
there is a race condition with some printing of atomics in the test. I don't expect this was introduced by this PR. I will investigate it and push a fix.

dimitarvdimitrov:
i pushed a commit which gets rid of the data race. The race was there when a service was being described. The race detector treats printing a mutex field and using that mutex as a race condition. This can be circumvented by giving the memberlist kv a name, so it doesn't have to be printed. The change is in https://github.com/grafana/dskit/pull/411/commits/ab6e2d1664431cf2144f7d67da3700f565efc7b3

PR #412, opened 2023-10-18
Fix flaky `TestJoinMembersWithRetryBackoff`

dimitarvdimitrov:
```
    memberlist_client_test.go:898:
        	Error Trace:	/Users/dimitar/grafana/dskit/kv/memberlist/memberlist_client_test.go:898
        	            				/Users/dimitar/grafana/dskit/kv/memberlist/asm_arm64.s:1197
        	            	Member-2 failed to join the cluster: &{[%!f(*errors.errorString=&{Failed to join 127.0.0.1:57493: dial tcp 127.0.0.1:57493: connect: connection refused})] %!f(multierror.ErrorFormatFunc=<nil>)}
        	Test:       	TestJoinMembersWithRetryBackoff/Test_late_start_of_DNS_service
```

The failure was caused by the test clients starting and expecting to connect to a KV which still wasn't started. This is unrelated with the backoff that the test asserts on (the backoff happens effectively in the call `err = services.StartAndAwaitRunning(context.Background(), mkv)`).

### DIFF
--- a/kv/memberlist/memberlist_client_test.go
+++ b/kv/memberlist/memberlist_client_test.go
@@ -891,11 +891,9 @@ func TestJoinMembersWithRetryBackoff(t *testing.T) {
 				clients = append(clients, kv)
 
 				startKVAndRunClient := func(kv *Client, id string, port int) {
-					err = services.StartAndAwaitRunning(context.Background(), mkv)
-					if err != nil {
-						t.Errorf("failed to start KV: %v", err)
-					}
+					err := services.StartAndAwaitRunning(context.Background(), mkv)
 					kvsStarted.Done()
+					require.NoError(t, err, "failed to start KV: %v")
 					err = runClient(kv, id, key, port, time.Second, start, stop)
 					require.NoError(t, err)
 				}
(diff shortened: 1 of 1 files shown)


56quarters (approved):
Nice find!

kv/memberlist/memberlist_client_test.go
```diff
@@ -894,10 +895,12 @@ func TestJoinMembersWithRetryBackoff(t *testing.T) {
 					if err != nil {
 						t.Errorf("failed to start KV: %v", err)
 					}
+					kvsStarted.Done()
```
pstibrany, line 898: We should probably use `defer kvsStarted.Done()` in this function, to unblock the main test goroutine even if starting of KVs fails.

dimitarvdimitrov: we can to run `kvsStarted.Done()` before running `runClient`. And `t.Errorf` continues the execution of the test

Don't we want to abort the test if a kv cannot start? I changed it so that we abort if the kv store cannot start _and_ call `kvsStarted.Done()` regardless if the kv store started successfully or failed. See https://github.com/grafana/dskit/pull/412/commits/da2d51311ff6bb1e31bce42a08815e569a92162c

pstibrany: Thanks, I think that should do it.

pstibrany (approved):

PR #414, opened 2023-10-23
Memcached: allow to configure write and read buffer size

pracucci:
**What this PR does**:
I noticed the memcached client library uses a default 4KB buffer for the read and write. However, in Mimir we transfer way bigger items and I would like to test whether increasing the buffer size will result into better performance. For this reason, I've made it configurable in https://github.com/grafana/gomemcache/pull/15 and I'm proposing to make the configuration option available through YAML / CLI flags (experimental config option).

N/A

**Checklist**
- [x] Tests updated


56quarters (approved):
LGTM, excited to see how this works.

PR #415, opened 2023-10-24
cache: distinguish between connect and request error reason

ortuman:
This PR adds an additional cache error reason (`connect-timeout`) in order to distinguish between timeout errors originated during a cache operation and those resulting from a connection attempt.

### DIFF
--- a/cache/client.go
+++ b/cache/client.go
@@ -22,7 +22,7 @@ const (
 	reasonAsyncBufferFull = "async-buffer-full"
 	reasonMalformedKey    = "malformed-key"
 	reasonConnectTimeout  = "connect-timeout"
-	reasonTimeout         = "timeout"
+	reasonTimeout         = "request-timeout"
 	reasonServerError     = "server-error"
 	reasonNetworkError    = "network-error"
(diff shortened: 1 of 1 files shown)


cache/client.go
```diff
@@ -21,6 +21,7 @@ const (
 	reasonMaxItemSize     = "max-item-size"
 	reasonAsyncBufferFull = "async-buffer-full"
 	reasonMalformedKey    = "malformed-key"
+	reasonConnectTimeout  = "connect-timeout"
 	reasonTimeout         = "timeout"
```
pracucci, line 25: WDYT about changing this to `requestRequestTimeout = "request-timeout"` to clarify it? We don't offer no breaking changes on metrics, so I wouldn't be much concerned changing the label value.

pracucci (approved):
LGTM modulo a nit

cache/client.go
ortuman: The clearer, the better. 👌 

_changed @ 17f669b4474fb14db13d1458684169d0ae7179de_

PR #416, opened 2023-10-24
httpgrpc: Add headers in http grpc error

ying-jeanne:
**What this PR does**:

In the context of incorporating a retry-after mechanism for all errors that can be recovered, we must compute the retry-after duration within the distributor and include it in the httpgrpc error. This PR focuses on encoding the header within the httpgrpc error, enabling it to be subsequently transmitted to the http response.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/httpgrpc/httpgrpc.go
+++ b/httpgrpc/httpgrpc.go
@@ -28,6 +28,14 @@ func Errorf(code int, tmpl string, args ...interface{}) error {
 	})
 }
 
+func ErrorfWithHeaders(code int, headers []*Header, tmpl string, args ...interface{}) error {
+	return ErrorFromHTTPResponse(&HTTPResponse{
+		Code:    int32(code),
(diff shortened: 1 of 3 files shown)


httpgrpc/httpgrpc.go
```diff
@@ -28,6 +28,14 @@ func Errorf(code int, tmpl string, args ...interface{}) error {
 	})
 }
 
+func ErrorfWithHeaders(code int, headers []*Header, tmpl string, args ...interface{}) error {
```
duricanikolic, line 31: Where would this method be called from?
I am currently getting rid of all `httpgrpc.Errorf()` calls in ingester and distributor. Would like to understand where this new method would be needed.

ying-jeanne: after our conversation, I think this PR is no longer needed. Thanks for the headsup

PR #417, opened 2023-10-26
moduleService: Fix return of `nil` from starting function when starting failed

pstibrany:
**What this PR does**:

This is alternative fix to https://github.com/grafana/dskit/pull/409, as discussed in thread https://github.com/grafana/dskit/pull/409#discussion_r1368788306.

**Checklist**
- [x] Tests updated

### DIFF
--- a/modules/module_service.go
+++ b/modules/module_service.go
@@ -86,11 +86,10 @@ func (w *moduleService) start(serviceContext context.Context) error {
 	}
 
 	err = w.service.AwaitRunning(serviceContext)
-	if errors.Is(serviceContext.Err(), context.Canceled) && errors.Is(err, context.Canceled) {
-		// We were asked to cancel, but the underlying service may continue starting and complete its startup after we return from moduleService.start.
-		// If we return an error, then we would be considered Failed and won't have moduleService.stop invoked.
-		// We need to invoke moduleService.stop to stop the underlying service in case it finishes starting successfully later.
-		return nil
+	if err != nil {
+		// Make sure that underlying service is stopped before returning
+		// (eg. in case of context cancellation, AwaitRunning returns early, but service may still be starting).
+		_ = services.StopAndAwaitTerminated(context.Background(), w.service)
 	}
 	return err
 }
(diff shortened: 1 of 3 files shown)


modules/modules_test.go
```diff
@@ -377,16 +377,17 @@ func TestModuleService_InterruptedFastStartup(t *testing.T) {
...
 		<-subserviceStarted
 		cancel()
+		close(finishStarting)
 	}()
 
-	require.ErrorIs(t, services.StartAndAwaitRunning(ctx, moduleSvc), context.Canceled)
+	// Start service using context that's going to be cancelled
+	require.NoError(t, moduleSvc.StartAsync(ctx))
+	// We get context cancelled error from service context cancellation.
+	require.ErrorIs(t, moduleSvc.AwaitRunning(context.Background()), context.Canceled)
```
pstibrany, line 386: Note that original test used `ctx` for waiting, but that wait got canceled in the test.

modules/module_service.go
```diff
@@ -104,19 +103,14 @@ func (w *moduleService) run(serviceContext context.Context) error {
 
 func (w *moduleService) stop(_ error) error {
 	var err error
-	switch w.service.State() {
-	case services.Running:
+	if w.service.State() == services.Running {
```
pstibrany, line 106: This reverts to previous logic, as underlying service can't be in `Starting` state when we get here.

aknuds1 (approved):
LGTM, but I guess @dimitarvdimitrov should review since he has the context (no pun intended).

modules/modules_test.go
```diff
@@ -414,9 +415,12 @@ func TestModuleService_InterruptedSlowStartup(t *testing.T) {
 
 	<-subserviceStarted
 
-	// If moduleService can handle StartingStates, then it should call Stop on the subservice.
-	require.NoError(t, services.StopAndAwaitTerminated(context.Background(), moduleSvc))
+	// We get context.Canceled from moduleService.start.
+	require.ErrorIs(t, services.StopAndAwaitTerminated(context.Background(), moduleSvc), context.Canceled)
```
dimitarvdimitrov, line 419: This is a bit confusing IMO that we get a context cancelation from an unrelated context. I think wrapping the context cancellation will be clearer. I made another suggestion

modules/module_service.go
```diff
@@ -86,11 +86,10 @@ func (w *moduleService) start(serviceContext context.Context) error {
...
-		// We were asked to cancel, but the underlying service may continue starting and complete its startup after we return from moduleService.start.
-		// If we return an error, then we would be considered Failed and won't have moduleService.stop invoked.
-		// We need to invoke moduleService.stop to stop the underlying service in case it finishes starting successfully later.
-		return nil
+	if err != nil {
+		// Make sure that underlying service is stopped before returning
+		// (eg. in case of context cancellation, AwaitRunning returns early, but service may still be starting).
+		_ = services.StopAndAwaitTerminated(context.Background(), w.service)
 	}
 	return err
```
dimitarvdimitrov, line 94:
```suggestion
	return errors.Wrapf(err, "waiting for service %s to start", w.name)
```

dimitarvdimitrov (approved):
LGTM, I don't think deadlock is possible in this case. Thanks for proposing this. You can also edit the changelog entry to mention this PR

modules/module_service.go
pstibrany: 👍 Makes sense. I was thinking the same when writing the test.

PR #418, opened 2023-10-26
hedgedhttp: bump cristalhq/hedgedhttp

cristaloleg:
**What this PR does**:
Updates cristalhq/hedgedhttp library to use new API. Old is still present but probably will be removed in the future.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] hedgedhttp: bump cristalhq/hedgedhttp #418
 * [CHANGE] Updated the minimum required Go version to 1.19 to ensure compatibility with OpenTelemetry version 1.14.0. #350
 * [CHANGE] Change `WaitRingStability` and `WaitInstanceState` methods signature to rely on `ReadRing` instead. #251
 * [CHANGE] Added new `-consul.cas-retry-delay` flag. It has a default value of `1s`, while previously there was no delay between retries. #178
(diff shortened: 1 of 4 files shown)


aknuds1 (commented):
Is it documented anywhere that `hedgedhttp.NewRoundTripper` is deprecated ("old API")? I can't see it in the [docs](https://pkg.go.dev/github.com/cristalhq/hedgedhttp).

cristaloleg:
Hey, 

> Is it documented anywhere that hedgedhttp.NewRoundTripper

Currently not but as the `cristalhq/hedgedhttp` author I can confirm that deprecation will happen in the next release. Not because it's wrong or bad, just because `hedgedhttp.New` (and `hedgedhttp.Client` as a result) implements `http.RoundTripper`.

> Also, can this be moved out of draft mode?

I haven't updated CHANGELOG file and haven't verified are there other things to do to get accepted. If it looks good for you - you/me can change the status.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] hedgedhttp: bump cristalhq/hedgedhttp #418
```
aknuds1, line 3: Would you mind just dropping this change? Shouldn't be necessary, as there shouldn't be any functional difference.

```suggestion
```

aknuds1 (commented):
AFAICT, the bump is straightforward, but I don't think a changelog entry is warranted as it shouldn't make any functional difference.

aknuds1 (approved):
Thanks!

PR #421, opened 2023-10-27
httpgrpc: correct handling of non-loggable errors

duricanikolic:
**What this PR does**:
This PR allows correct handling of non-loggable errors. 

First of all, it introduces `middleware.DoNotLogError` type, representing the errors that should not be logged. Additionally, it introduces `httpgrpc.DoNotLogErrorHeaderKey`, an HTTP header key with value `X-DoNotLogError`. In case of an erroneous `HTTPResponse`, the presence of this header means that the corresponding error should not be logged. 

When `httpgrpc`'s `Server.Handle()` receives a response from the HTTP server, an `HTTPResponse` object is created. `HTTPResponse`s with a status `5xx` are translated into gRPC errors, by using the `httpgrpc` package. Before this PR, all these errors were loggable. This PR allows `Server.Handle()` to return errors that should not be logged. More precisely, the presence of a header with key `httpgrpc.DoNotLogHeader` and any value in an erroneous `HTTPResponse` indicates to `Servier.Handle()` to wrap the gRPC error (created by the `httpgrpc` package) with a `httpgrpc.DoNotLogError`.

One more change introduced by this PR concerns `httpgrpc.HTTPResponseFromError()`. Namely, before this PR, errors built by `httpgrpc` package and wrapped with another error, were not recognized by `httpgrpc.HTTPResponseFromError()`. This PR fixes that behavior.

Part of https://github.com/grafana/mimir/issues/6008.

**Checklist**
- [x] Tests updated

### DIFF
--- a/httpgrpc/httpgrpc.go
+++ b/httpgrpc/httpgrpc.go
@@ -6,10 +6,12 @@ package httpgrpc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/go-kit/log/level"
 	"google.golang.org/grpc/metadata"
+	grpcstatus "google.golang.org/grpc/status"
 
 	spb "github.com/gogo/googleapis/google/rpc"
 	"github.com/gogo/protobuf/types"
@@ -44,7 +46,7 @@ func ErrorFromHTTPResponse(resp *HTTPResponse) error {
 
 // HTTPResponseFromError converts a grpc error into an HTTP response
 func HTTPResponseFromError(err error) (*HTTPResponse, bool) {
-	s, ok := status.FromError(err)
+	s, ok := statusFromError(err)
 	if !ok {
 		return nil, false
 	}
@@ -61,6 +63,29 @@ func HTTPResponseFromError(err error) (*HTTPResponse, bool) {
 	}
 
 	return &resp, true
+}
+
+// statusFromError tries to cast the given error into status.Status.
+// If the given error, or any error from its tree are a status.Status,
+// that status.Status and the outcome true are returned.
+// Otherwise, nil and the outcome false are returned.
+// This implementation differs from status.FromError() because the
+// latter checks only if the given error can be cast to status.Status,
+// and doesn't check other errors in the given error's tree.
+func statusFromError(err error) (*status.Status, bool) {
+	if err == nil {
+		return nil, false
+	}
+	type grpcStatus interface{ GRPCStatus() *grpcstatus.Status }
+	var gs grpcStatus
+	if errors.As(err, &gs) {
+		st := gs.GRPCStatus()
+		if st == nil {
+			return nil, false
+		}
+		return status.FromGRPCStatus(st), true
+	}
+	return nil, false
 }
 
 const (
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -27,6 +27,10 @@ import (
 	"github.com/grafana/dskit/middleware"
 )
 
+var (
+	DoNotLogErrorHeaderKey = http.CanonicalHeaderKey("X-DoNotLogError")
+)
+
 // Server implements HTTPServer.  HTTPServer is a generated interface that gRPC
 // servers must implement.
 type Server struct {
@@ -70,8 +74,8 @@ func (s Server) Handle(ctx context.Context, r *httpgrpc.HTTPRequest) (*httpgrpc.
 	}
 	if recorder.Code/100 == 5 {
 		err := httpgrpc.ErrorFromHTTPResponse(resp)
-		if containsDoNotLogErrorKey(header) {
-			return nil, middleware.DoNotLogError{Err: err}
+		if _, ok := header[DoNotLogErrorHeaderKey]; ok {
+			err = middleware.DoNotLogError{Err: err}
 		}
 		return nil, err
 	}
@@ -238,9 +242,4 @@ func fromHeader(hs http.Header) []*httpgrpc.Header {
 		})
 	}
 	return result
-}
-
-func containsDoNotLogErrorKey(hs http.Header) bool {
-	_, ok := hs[middleware.DoNotLogErrorHeader]
-	return ok
 }
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -62,13 +62,18 @@ func (s Server) Handle(ctx context.Context, r *httpgrpc.HTTPRequest) (*httpgrpc.
 
 	recorder := httptest.NewRecorder()
 	s.handler.ServeHTTP(recorder, req)
+	header := recorder.Header()
 	resp := &httpgrpc.HTTPResponse{
 		Code:    int32(recorder.Code),
-		Headers: fromHeader(recorder.Header()),
+		Headers: fromHeader(header),
 		Body:    recorder.Body.Bytes(),
 	}
 	if recorder.Code/100 == 5 {
-		return nil, httpgrpc.ErrorFromHTTPResponse(resp)
+		err := httpgrpc.ErrorFromHTTPResponse(resp)
+		if containsDoNotLogErrorKey(header) {
+			return nil, middleware.DoNotLogError{Err: err}
+		}
+		return nil, err
 	}
 	return resp, nil
 }
@@ -233,4 +238,9 @@ func fromHeader(hs http.Header) []*httpgrpc.Header {
 		})
 	}
 	return result
+}
+
+func containsDoNotLogErrorKey(hs http.Header) bool {
+	_, ok := hs[middleware.DoNotLogErrorHeader]
+	return ok
 }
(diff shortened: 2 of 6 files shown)


httpgrpc/server/server.go
```diff
@@ -234,3 +239,8 @@ func fromHeader(hs http.Header) []*httpgrpc.Header {
 	}
 	return result
 }
+
+func containsDoNotLogErrorKey(hs http.Header) bool {
```
pstibrany, line 243: This is too shallow. I understand it's that way for testing, but this functionality can be tested together with the rest of server handler too.

httpgrpc/server/server_test.go
```diff
@@ -151,3 +151,37 @@ func TestTracePropagation(t *testing.T) {
...
+			},
+			expectedOutcome: true,
+		},
+	}
+	for testName, testData := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			require.Equal(t, testData.expectedOutcome, containsDoNotLogErrorKey(testData.header))
+		})
+	}
+}
```
pstibrany, line 187: Honestly, this test seems bit too much for testing what's essentially a simple map key-check. I'd suggest removing this test, and including checks for DoNotLog error into already existing tests that handle 500 errors.

httpgrpc/server/server.go
```diff
@@ -62,13 +62,18 @@ func (s Server) Handle(ctx context.Context, r *httpgrpc.HTTPRequest) (*httpgrpc.
...
 		Code:    int32(recorder.Code),
-		Headers: fromHeader(recorder.Header()),
+		Headers: fromHeader(header),
 		Body:    recorder.Body.Bytes(),
 	}
 	if recorder.Code/100 == 5 {
-		return nil, httpgrpc.ErrorFromHTTPResponse(resp)
+		err := httpgrpc.ErrorFromHTTPResponse(resp)
+		if containsDoNotLogErrorKey(header) {
+			return nil, middleware.DoNotLogError{Err: err}
```
pstibrany, line 74:
```suggestion
			err = middleware.DoNotLogError{Err: err}
```

middleware/grpc_logging.go
```diff
@@ -24,11 +25,21 @@ const (
 	gRPC = "gRPC"
 )
 
+var (
+	DoNotLogErrorHeader = http.CanonicalHeaderKey("X-DoNotLogError")
+)
```
pstibrany, line 30: I think this belongs to httpgrpc package, where server is using it. `grpc_logging.go` has nothing to do with HTTP.

pstibrany (commented):
looks good overall, left some minor comments.

httpgrpc/httpgrpc.go
```diff
@@ -63,6 +65,29 @@ func HTTPResponseFromError(err error) (*HTTPResponse, bool) {
 	return &resp, true
 }
 
+// statusFromError tries to cast the given error into status.Status.
+// If the given error, or any error from its tree are a status.Status,
+// that status.Status and the outcome true are returned.
+// Otherwise, nil and the outcome false are returned.
+// This implementation differs from status.FromError() because the
```
56quarters, line 72: It does check that. The implementation is nearly the same as this method:

https://github.com/grafana/mimir/blob/main/vendor/google.golang.org/grpc/status/status.go#L113

duricanikolic: @56quarters you linked the grpc’s status, and not gogo’s status.

56quarters: OK, but that's the one we should be using? That's the one that the circuit breakers use to extract gRPC error codes.

duricanikolic: Hi @56quarters, 
Curently gogo is used both in `dskit` and in `mimir`. It is used even for the generation of Go structs out of `.proto` sources (e.g., [here](https://github.com/grafana/dskit/blob/main/Makefile#L63-L67)).
I agree that we should replace usages of gogo's `status` package with grpc's `status` package, but this should be done in a separate PR, because that change would be very big.

For the moment I changed the implementation of the `statusFromError` and made it use `grpcstatus.FromError()`, since these 2 implementations are, indeed, similar.

WDYT?

middleware/grpc_logging.go
duricanikolic: I have moved this header to `httpgrpc/server.go`.

httpgrpc/server/server.go
duricanikolic: I have removed this method, as well as the corresponding test.

httpgrpc/server/server_test.go
duricanikolic: I have applied the required fix.

httpgrpc/httpgrpc.go
duricanikolic: Update: apparently we already have an [issue](https://github.com/weaveworks/common/issues/268) for removing the deprecated `gogo` dependency. But this would be a breaking change, plus we don't know which package to use instead. The `grpc` package has some big performance issues, that's why the `gogo` project was created.

> That's the one that the circuit breakers in Mimir use to extract gRPC error codes
This is not a problem, since gRPC errors created by both `grpc` and `gogo` can be recognized by each other.

56quarters (approved):

httpgrpc/server/server.go
```diff
@@ -62,13 +66,18 @@ func (s Server) Handle(ctx context.Context, r *httpgrpc.HTTPRequest) (*httpgrpc.
...
 	resp := &httpgrpc.HTTPResponse{
 		Code:    int32(recorder.Code),
-		Headers: fromHeader(recorder.Header()),
+		Headers: fromHeader(header),
 		Body:    recorder.Body.Bytes(),
 	}
 	if recorder.Code/100 == 5 {
-		return nil, httpgrpc.ErrorFromHTTPResponse(resp)
+		err := httpgrpc.ErrorFromHTTPResponse(resp)
+		if _, ok := header[DoNotLogErrorHeaderKey]; ok {
```
pstibrany, line 77: Shall we remove that header from the response? It's internal only. (This would only work for HTTP requests handled via httpgrpc.Server, but that's fine)

httpgrpc/server/server.go
```diff
@@ -27,6 +27,10 @@ import (
 	"github.com/grafana/dskit/middleware"
 )
 
+var (
+	DoNotLogErrorHeaderKey = http.CanonicalHeaderKey("X-DoNotLogError")
```
pstibrany, line 31: Can we add short doc saying how to use this header? If it's present in the response (with any value), and response has status code 5xx, then generated error will be marked to not be logged.

httpgrpc/server/server_test.go
```diff
@@ -73,25 +73,39 @@ func TestBasic(t *testing.T) {
...
+			recorder := httptest.NewRecorder()
+			client.ServeHTTP(recorder, req)
 
-	assert.Equal(t, "foo\n", recorder.Body.String())
-	assert.Equal(t, 500, recorder.Code)
+			assert.Equal(t, "foo\n", recorder.Body.String())
+			assert.Equal(t, 500, recorder.Code)
+			if doNotLog {
+				assert.Equal(t, recorder.Header().Get(DoNotLogErrorHeaderKey), "true")
+			}
```
pstibrany, line 106: The test should not assert on header, but on whether error produced by Server.Handle is marked for logging or not.

something like:
```golang
func TestServerHandleOptionalError(t *testing.T) {
	for _, doNotLog := range []bool{false, true} {
		h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			if doNotLog {
				w.Header().Set(DoNotLogErrorHeaderKey, "true")
			}
			http.Error(w, "foo", http.StatusInternalServerError)
		})

		s := NewServer(h)
		resp, err := s.Handle(...)
		// check err here
	}
}
```

pstibrany (approved):
lgtm overall, but would like to see a better test for httpgrpc.Server.

httpgrpc/server/server.go
duricanikolic: I have removed the `DoNotLogErrorHeader` from the response.

httpgrpc/server/server.go
duricanikolic: I have added a doc, and I have updated CHANGELOG.md.

httpgrpc/server/server_test.go
duricanikolic: A new test has been added.

httpgrpc/server/server.go
```diff
@@ -62,13 +69,19 @@ func (s Server) Handle(ctx context.Context, r *httpgrpc.HTTPRequest) (*httpgrpc.
...
 		Code:    int32(recorder.Code),
-		Headers: fromHeader(recorder.Header()),
+		Headers: fromHeader(header),
 		Body:    recorder.Body.Bytes(),
 	}
 	if recorder.Code/100 == 5 {
-		return nil, httpgrpc.ErrorFromHTTPResponse(resp)
+		err := httpgrpc.ErrorFromHTTPResponse(resp)
+		if _, ok := header[DoNotLogErrorHeaderKey]; ok {
+			recorder.Header().Del(DoNotLogErrorHeaderKey)
```
pstibrany, line 81: We don't need to remove header here, only headers in `resp` are important, and `fromHeader` is covering that already.

pstibrany (approved):
Thank you.

PR #423, opened 2023-10-30
Use ReadHeaderTimeout

pstibrany:
**What this PR does**: Set read header timeout explicitly. This leaves another `ReadTimeout` for body only.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -457,10 +457,11 @@ func newServer(cfg Config, metrics *Metrics) (*Server, error) {
 	}
 
 	httpServer := &http.Server{
-		ReadTimeout:  cfg.HTTPServerReadTimeout,
-		WriteTimeout: cfg.HTTPServerWriteTimeout,
-		IdleTimeout:  cfg.HTTPServerIdleTimeout,
-		Handler:      middleware.Merge(httpMiddleware...).Wrap(router),
+		ReadTimeout:       cfg.HTTPServerReadTimeout,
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -457,10 +459,11 @@ func newServer(cfg Config, metrics *Metrics) (*Server, error) {
 	}
 
 	httpServer := &http.Server{
-		ReadTimeout:  cfg.HTTPServerReadTimeout,
-		WriteTimeout: cfg.HTTPServerWriteTimeout,
-		IdleTimeout:  cfg.HTTPServerIdleTimeout,
-		Handler:      middleware.Merge(httpMiddleware...).Wrap(router),
+		ReadTimeout:       cfg.HTTPServerReadTimeout,
+		ReadHeaderTimeout: cfg.HTTPServerReadHeaderTimeout,
```
pracucci, line 463: Add a comment to explain **why** we want to set the read header timeout.

pracucci (approved):

pstibrany:
Converted to draft again. As it is now, PR is wrong.

pstibrany:
> Converted to draft again. As it is now, PR is wrong. (Mostly documentation).

Updated the description of fields and changelog entry.

server/server.go
pstibrany: PTAL to see if new descriptions make sense.

PR #424, opened 2023-11-01
Allow reporting gRPC codes in "status_code" label of request duration metrics

duricanikolic:
**What this PR does**:
Before this PR `grpc_instrumentation.go` could recognize only the errors built by the `httpgrpc` package and use their error codes in the `status_code` label of the request duration metrics. All other gRPC errors were treated as unknown, and used to be labeled as `error`. This PR fixes that erroneous behaviour, and allows all gRPC status codes to be used as the `status_code` label.

Types `middleware.InstrumentationLabel` and `middleware.InstrumentationOption`, as well as a special value of the latter, `middleware.ReportGRPCStatusOption` have been introduced. This special value can be used to enable reporting of gRPC status codes in `status_code` label (instead of simplified "error", "cancel" or "success" values). `middleware.ReportGRPCStatusOption` can be passed as an optional argument to the following methods:
- `middleware.UnaryServerInstrumentInterceptor`
- `middleware.StreamServerInstrumentInterceptor`
- `middleware.UnaryClientInstrumentInterceptor`
- `middleware.StreamClientInstrumentInterceptor`
- `middleware.Instrument`.

In order to guarantee backwards compatibility, reporting of gRPC codes as labels is disabled by default and could be enabled as follows: 
- on server side by setting the new CLI experimental flag `-server.report-grpc-codes-in-instrumentation-label` to `true`, or by setting `server.Config.ReportGRPCCodesInInstrumentationLabel` to `true`.
- on client side by passing `middleware.ReportGRPCStatusOption` to `middleware.Instrument`.

Part of https://github.com/grafana/mimir/issues/6008.

**Checklist**
- [x] Tests updated


middleware/grpc_instrumentation.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/gogo/status"
```
pstibrany, line 13: Can we not use gogo/status in the new code?

duricanikolic: This has been discussed via Slack. There is already a private method handling this problem in `httpgrpc.statusFromError()`. That method has been moved to the `grpcutil` package, exported and renamed to `grpcutil.ErrorToStatus()`.

Our proto files are converted to structs by using `gogo/protobuf`, so as far as this is the case, `dskit` should prefer `gogo/status` to `grpc/status` for consistency.

middleware/grpc_instrumentation_test.go
```diff
@@ -29,31 +30,81 @@ import (
...
+
+	a = errorCode(nil, "success", true)
+	assert.Equal(t, "success", a)
 }
 
 func TestErrorCode_Any5xx(t *testing.T) {
-	err := httpgrpc.Errorf(http.StatusNotImplemented, "Fail")
-	a := errorCode(err)
-	assert.Equal(t, "5xx", a)
+	for _, maskError := range []bool{true, false} {
```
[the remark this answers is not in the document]
duricanikolic: These tests were already different tests, that's why I kept their original names. But I will change this and merge them into a single test.

grpcutil/status.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// ErrorToStatus tries to cast the given error into gogo's status.Status.
+// If the given error, or any error from its tree are a status.Status,
+// that status.Status and the outcome true are returned.
+// Otherwise, nil and the outcome false are returned.
+// This implementation differs from status.FromError() because the
+// latter checks only if the given error can be cast to status.Status,
+// and doesn't check other errors in the given error's tree.
+func ErrorToStatus(err error) (*status.Status, bool) {
```
duricanikolic, line 17: This current `httpgrpc.statusFromError()` function has been renamed to `ErrorToStatus()`, exported and moved here.

grpcutil/status_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/gogo/status"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+)
+
+func TestErrorToStatus(t *testing.T) {
```
duricanikolic, line 13: The current `httpgrpc.TestStatusFromError()` has been moved here.

grpcutil/metadata_test.go
```diff
@@ -7,24 +7,25 @@ import (
...
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/metadata"
-
-	"github.com/grafana/dskit/httpgrpc"
 )
 
 func TestAppendMessageSizeToOutgoingContext(t *testing.T) {
 	ctx := context.Background()
 
-	req := &httpgrpc.HTTPRequest{
```
duricanikolic, line 17: This has been replaced with `fakeSizer` to avoid a circular dependency. The parameter of `AppendMessageSizeToOutgoingContext` should be any implementation of the `Sizer` interface, and `httpgrpc.HTTPRequest` was just one of them.

middleware/grpc_instrumentation.go
```diff
@@ -174,17 +165,26 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+		if maskHTTPStatuses {
+			statusFamily := int(errResp.Code / 100)
+			return strconv.Itoa(statusFamily) + "xx"
+		}
+		return strconv.Itoa(int(errResp.Code))
+	}
+
+	if stat, ok := grpcutil.ErrorToStatus(err); ok {
+		return stat.Code().String()
+	}
```
pstibrany, line 187: This implementation seems inefficient. First we call `httpgrpc.HTTPResponseFromError` which calls `grpcutil.ErrorToStatus` and checks for "details". If it finds that error is not "HTTPResponse", result is thrown away, only to call `grpcutil.ErrorToStatus` again here.

I'd suggest to intruduce new `httpgrpc.HTTPResponseAndStatusFromError` that returns `*HTTPResponse` (possibly nil) and also `*status.Status` (possibly nil) in one go. Then we don't need another call to `grpcutil.ErrorToStatus(err)` here.

grpcutil/cancel.go
```diff
@@ -9,15 +9,14 @@ import (
...
-	"google.golang.org/grpc/status"
 )
 
 // IsCanceled checks whether an error comes from an operation being canceled
 func IsCanceled(err error) bool {
 	if errors.Is(err, context.Canceled) {
 		return true
 	}
-	s, ok := status.FromError(err)
+	s, ok := ErrorToStatus(err)
```
pstibrany, line 19: Doing full conversion to gogo's `status.Status` seems unnecessary here, if all we care about is code from grpc.Status. Perhaps we can have `ErrorToStatusCode` function too?

middleware/grpc_instrumentation_test.go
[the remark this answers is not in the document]
pstibrany: Merging same test code into one big function was not really what I had in mind.

By single test I was about thinking table-driven test, with inputs and expected outputs, with single "for" loop, covering all cases. I'm sorry for not being clearer in my comment :(

grpcutil/cancel.go
duricanikolic: It is true that we need only the status code here, but in order to get it, we anyway need to call a `gogostatus.FromError()` or `grpcstatus.FromError()` (the current implementation).
My idea here was to unify the way we extract statuses. Because in most of the places it is done by `gogostatus`, and in a couple of places it is done by `grpcstatus`.

The idea is, anyway, to get rid of the `gogo`-related things, but once we find a valid alternative. If we now start mixing `gogo` and `grpc` usages, it will be more difficult to do the change one day.

pstibrany: > If we now start mixing gogo and grpc usages, it will be more difficult to do the change one day.

I don't see how my proposal would do that?

middleware/grpc_instrumentation.go
duricanikolic: WDYT about this:
```
// errorCode converts an error into an error code string.
func errorCode(err error, successCode string, maskHTTPStatuses bool) string {
	if err == nil {
		return successCode
	}

	if grpcutil.IsCanceled(err) {
		return "cancel"
	}

	if stat, ok := grpcutil.ErrorToStatus(err); ok {
		statusFamily := int(stat.Code() / 100)
		if statusFamily > 0 && maskHTTPStatuses {
			return strconv.Itoa(statusFamily) + "xx"
		}
		return stat.Code().String()
	}

	return "error"
}
```
At the end, we don't care whether it is a `httpgrpc` or a pure `gRPC` error, we just care whether it can be converted to the `status` interface or not, and get the corresponding code. This way we reduce the number of calls to `FromError()`

grpcutil/cancel.go
duricanikolic: > I don't see how my proposal would do that?
I was referring to the current implementation: it added a `grpcstatus` usage to `dskit`. And my motivation with the proposed change was to get rid of it. I meant that.

middleware/grpc_instrumentation.go
pstibrany: This seems like a hack, since we're not checking for HTTPResponse at all.

We also need to do this:

```golang
		if statusFamily > 0 {
			// codes bigger than 100 are expected to be HTTP status codes.
			if maskHTTPStatuses {
				return strconv.Itoa(statusFamily) + "xx"
			} 
			return strconv.Itoa(stat.Code())
		}
```

Your example shows why it makes sense to have `ErrorToStatusCode` only -- again, there's no need to do grpc -> gogo status conversion.

pstibrany: I just realized that if we inline `grpcutil.IsCanceled` here, we get rid of yet-another conversion.

duricanikolic: > I just realized that if we inline grpcutil.IsCanceled here, we get rid of yet-another conversion.

I can do that.

grpcutil/cancel.go
pstibrany: > I was referring to the current implementation: it added a `grpcstatus` usage to `dskit`. And my motivation with the proposed change was to get rid of it. I meant that.

We need to check for `GRPCStatus() *grpcstatus.Status` method, is this what you're referring to? I don't see a way around it.

middleware/grpc_instrumentation.go
duricanikolic: > Your example shows why it makes sense to have ErrorToStatusCode only -- again, there's no need to do grpc -> gogo status conversion.

I am afraid I dont' understand your remark about grpc -> gogo status conversion.
My concerns about `ErrorToStatusCode` are:
1) status codes of `HTTPResponse` are of type `int32`
2) both `gogo` and `grpc` statuses have codes of type `codes.Code`
3) we can return the string representation of them, but we need to distinguish between HTTP statuses and gRPC statuses, because the former need to be masked in some cases (e.g., `4xx`)
4) the conversion is needed if we receive, for example, a `DoNotLogError{Err: gogostatus.New(code.Internal, "error")}`. In this cases `gogostatus.FromError()` will not recognize the error

middleware/grpc_instrumentation.go
```diff
@@ -174,17 +166,51 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+// If the given error is nil, errorCode returns "success" or "2xx", depending on whether maskHTTPStatuses
+// is set to false or true respectively.
+// If the given error corresponds to context.Canceled, errorCode returns "cancel", independently of maskHTTPStatuses.
+// If the given error is a gRPC error with an HTTP status code, errorCode returns the string representation of the
+// HTTP status code or the first digit of the HTTP status code followed by "xx", depending on whether maskHTTPStatuses
+// is set to false or true respectively.
+// If the given error is a gRPC error with a gRPC status code different from codes.Unknown, errorCode returns the string
+// representation of the status code, independently of maskHTTPStatuses.
+// If the given error is a gRPC error with gRPC status code codes.Unknown, or if it is a non-gRPC error, errorCode
+// returns "error".
```
pstibrany, line 178: This only repeats what's written below, in the code. I'd remove this duplicity.

middleware/grpc_instrumentation.go
```diff
@@ -174,17 +166,51 @@ func (s *instrumentedClientStream) CloseSend() error {
...
 
-	if errResp, ok := httpgrpc.HTTPResponseFromError(err); ok {
-		statusFamily := int(errResp.Code / 100)
-		return strconv.Itoa(statusFamily) + "xx"
-	} else if grpcutil.IsCanceled(err) {
+	// In order to understand whether this error corresponds to context.Canceled, we don't call
+	// grpcutil.IsCanceled to avoid an additional call to grpcutil.ErrorToStatusCode. Instead,
+	// we check whether statusCodes is codes.Canceled, or whether it is codes.Unknown with the
+	// error being context.Canceled. In that case we return "cancel".
+	// If statusCode is codes.Unknown but the error is not context.Canceled, we return "error".
```
pstibrany, line 192: Same comment, let's remove this.

grpcutil/status.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/gogo/status"
+	"google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+)
+
+// ErrorToStatus tries to cast the given error into gogo's status.Status.
```
pstibrany, line 12: This description seems wrong: there's no casting to gogo's status.Status. Code only checks for `GRPCStatus()` method with specific return type.

I'd suggest to remove this comment, or to update it:

```golang
// ErrorToStatus returns a *github.com/gogo/status.Status representation of err.
//
//   - If err implements the method `GRPCStatus() *google.golang.org/grpc/status.Status` and
//     `GRPCStatus()` does not return nil, or if err wraps a type satisfying this, Status from `GRPCStatus()` is
//     converted to gogo Status, and returned. In that case, ok is true.
//
//   - If err is or GRPCStatus() returns nil, a nil Status is returned and ok is false.
//
//   - Otherwise, err is an error not compatible with this function. In this
//     case, a nil Status is returned and ok is false.
```

grpcutil/status.go
```diff
@@ -0,0 +1,60 @@
...
+			return nil, false
+		}
+		return status.FromGRPCStatus(st), true
+	}
+	return nil, false
+}
+
+// ErrorToStatusCode returns the Code of the error if it is a gogo's status.Status
+// error or if it wraps a status.Status error. If that is not the case, it returns
+// codes.OK if err is nil, or codes.Unknown otherwise.
```
pstibrany, line 37: This description is also bit confusing. Implementation doesn't work with gogo's Status.

```suggestion
// ErrorToStatusCode extracts gRPC status code from error and returns it.
//
//   - If err is nil, codes.OK is returned.
//
//    - If err implements (or wraps error that implements) `GRPCStatus() *google.golang.org/grpc/status.Status`, and this method returns non-nil status, code from the status is returned.
//
//    - Otherwise code.Unknown is returned.
```

grpcutil/status_test.go
```diff
@@ -0,0 +1,157 @@
...
+		},
+		"a wrapped error of a gRPC error built by gogo/status returns the gRPC error's code": {
+			err:                fmt.Errorf("wrapped: %w", status.Error(codes.Unavailable, msgErr)),
+			expectedStatusCode: codes.Unavailable,
+		},
+		"a gRPC error built by grpc/status returns its code": {
+			err:                grpcstatus.Error(codes.FailedPrecondition, msgErr),
+			expectedStatusCode: codes.FailedPrecondition,
+		},
+		"a wrapped error of a gRPC error built by grpc/status creturns the gRPC error's code": {
```
pstibrany, line 96:
```suggestion
		"a wrapped error of a gRPC error built by grpc/status returns the gRPC error's code": {
```

middleware/grpc_instrumentation.go
duricanikolic: I have removed the comment in the code.

server/server.go
```diff
@@ -348,17 +350,21 @@ func newServer(cfg Config, metrics *Metrics) (*Server, error) {
 		WithRequest:              !cfg.ExcludeRequestInLog,
 		DisableRequestSuccessLog: cfg.DisableRequestSuccessLog,
 	}
+	reportGRPCStatusesOption := middleware.DoNotAcceptGRPCStatusesOption
+	if cfg.ReportGRPCCodesInInstrumentationLabel {
+		reportGRPCStatusesOption = middleware.AcceptGRPCStatusesOption
+	}
 	grpcMiddleware := []grpc.UnaryServerInterceptor{
```
pstibrany, line 357: We should pass the option only to ENABLE the new reporting. If old behaviour is wanted, let's not pass any option. That way we only need one, while preserving compatibility with existing code.

middleware/grpc_instrumentation.go
```diff
@@ -6,79 +6,76 @@ package middleware
...
-			respStatus = strconv.Itoa(int(errResp.Code))
-		} else if grpcutil.IsCanceled(err) {
-			respStatus = "cancel"
-		} else {
-			respStatus = "error"
-		}
-	}
-	instrument.ObserveWithExemplar(ctx, hist.WithLabelValues(gRPC, method, respStatus, "false"), duration.Seconds())
+func observe(ctx context.Context, hist *prometheus.HistogramVec, method string, err error, duration time.Duration, instrumentationLabelOptions ...InstrumentationLabelOption) {
+	instrumentationLabel := ApplyInstrumentationLabelOptions(false, instrumentationLabelOptions...)
```
pstibrany, line 25: Let's not do this on every call to `observe`. We can compute whether grpc codes are wanted or not just once per each interceptor, and pass that value to `observe`.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
 	return err
 }
 
-// errorCode converts an error into an error code string.
-func errorCode(err error) string {
-	if err == nil {
-		return "2xx"
+type InstrumentationLabelOption func(*InstrumentationLabel)
```
pstibrany, line 174: suggestion: `InstrumentationOption`.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+func ApplyInstrumentationLabelOptions(maskHTTPStatuses bool, options ...InstrumentationLabelOption) InstrumentationLabel {
+	instrumentationLabel := InstrumentationLabel{maskHTTPStatuses: maskHTTPStatuses}
+	for _, opt := range options {
+		opt(&instrumentationLabel)
+	}
+	return instrumentationLabel
+}
+
+type InstrumentationLabel struct {
+	acceptGRPCStatuses bool
```
pstibrany, line 195: suggestion: `reportGRPCStatus`

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+	maskHTTPStatuses   bool
```
pstibrany, line 196: suggestion: `maskHTTPStatus`

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+var (
+	AcceptGRPCStatusesOption InstrumentationLabelOption = func(instrumentationLabel *InstrumentationLabel) {
+		instrumentationLabel.AcceptGRPCStatuses(true)
 	}
 
-	if errResp, ok := httpgrpc.HTTPResponseFromError(err); ok {
-		statusFamily := int(errResp.Code / 100)
-		return strconv.Itoa(statusFamily) + "xx"
-	} else if grpcutil.IsCanceled(err) {
+	DoNotAcceptGRPCStatusesOption InstrumentationLabelOption = func(instrumentationLabel *InstrumentationLabel) {
```
pstibrany, line 181: Let's remove this one, and only keep one option to control this.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
 }
 
-// errorCode converts an error into an error code string.
-func errorCode(err error) string {
-	if err == nil {
-		return "2xx"
+type InstrumentationLabelOption func(*InstrumentationLabel)
+
+var (
+	AcceptGRPCStatusesOption InstrumentationLabelOption = func(instrumentationLabel *InstrumentationLabel) {
```
pstibrany, line 177: `AcceptGRPCStatusesOption` -> `ReportGRPCStatusOption`.

Also please add short documentation to it, since it's a public var.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+}
+
+type InstrumentationLabel struct {
+	acceptGRPCStatuses bool
+	maskHTTPStatuses   bool
+}
+
+func (i *InstrumentationLabel) AcceptGRPCStatuses(acceptance bool) {
+	i.acceptGRPCStatuses = acceptance
+}
```
pstibrany, line 201: We don't need function for this.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+
+func ApplyInstrumentationLabelOptions(maskHTTPStatuses bool, options ...InstrumentationLabelOption) InstrumentationLabel {
+	instrumentationLabel := InstrumentationLabel{maskHTTPStatuses: maskHTTPStatuses}
+	for _, opt := range options {
+		opt(&instrumentationLabel)
+	}
+	return instrumentationLabel
+}
+
+type InstrumentationLabel struct {
```
pstibrany, line 194: This doesn't need to be public.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+//     depending on whether maskHTTPStatuses is set to false or true respectively.
+//   - If the given status code is codes.Canceled, statusCodeToString returns "cancel"
+//     independently of maskHTTPStatuses.
+//   - If the given status code ia a valid HTTP status code, statusCodeToString returns the
+//     string representation of the HTTP status code or the first digit of the HTTP status
+//     code followed by "xx", depending on whether maskHTTPStatuses is set to false or true
+//     respectively.
+//   - If the given status code is a gRPC status code different from codes.Unknown, statusCodeToString
+//     returns the string representation of the status code, independently of maskHTTPStatuses.
+//   - If the given status code is codes.Unknown, statusCodeToString returns "error".
```
pstibrany, line 222: This once again just duplicates implementation, but in plain english. Let's get rid of it.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+	}
+
+	if statusCode == codes.Canceled {
 		return "cancel"
-	} else {
+	}
+
+	if statusCode == codes.Unknown {
 		return "error"
 	}
```
pstibrany, line 237: I think we should be consistent, and report `OK`, `Canceled` and `Unknown`, when reporting of gRPC statuses is enabled. (or perhaps lowercase version)

When it's disabled, we can go back to `success`, `cancel` or `error`.

WDYT?

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +171,115 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+		return statusCode
+	}
+
+	if isHTTPStatusCode(statusCode) {
+		return statusCode
+	}
+	if i.acceptGRPCStatuses {
+		return statusCode
+	}
+	return codes.Unknown
```
pstibrany, line 277: I think this function is doing too much -- it should just return `grpcutil.ErrorToStatusCode(err)` for non-nil and not-canceled errors. (It also doesn't need to be a method of `InstrumentationLabel`.)

pstibrany (commented):
This PR is getting bigger and bigger :/

middleware/grpc_instrumentation.go
duricanikolic: I have fixed that in all the places where `ApplyInstrumentationLabelOptions` was called.

middleware/grpc_instrumentation.go
duricanikolic: This function is the only place where we actually take into account whether `i.acceptGRPCStatues` is set to true or not. If we move it out from here, we need to put it in `statusCodeToString()` function. I can try to see how it would look like, but it will for sure complicate the test.

middleware/grpc_instrumentation.go
duricanikolic: I have applied your recommendation. I haven't opted for lowercase version of gRPC status codes, because for the majority of them the original form is more readable: e.g., `FailedPrecondition` vs. `failedprecondition`, `ResourceExhausted` vs. `resourceexhausted`, etc.

middleware/grpc_instrumentation_test.go
```diff
@@ -17,44 +17,265 @@ import (
...
-func TestErrorCode_Unknown(t *testing.T) {
-	err := status.Errorf(codes.Unknown, "Fail")
-	a := errorCode(err)
-	assert.Equal(t, "error", a)
+func TestApplyInstrumentationLabelOptions(t *testing.T) {
+	testCases := map[string]struct {
+		instrumentationOptions     []InstrumentationOption
+		expectedAcceptGRPCStatuses bool
+	}{
+		"Applying no InstrimentationOption sets acceptGRPCStatus to false": {
```
pstibrany, line 259: nit: typo in InstrimentationOption

middleware/grpc_instrumentation_test.go
```diff
@@ -17,44 +17,265 @@ import (
...
 }
 
-func TestErrorCode_Unknown(t *testing.T) {
-	err := status.Errorf(codes.Unknown, "Fail")
-	a := errorCode(err)
-	assert.Equal(t, "error", a)
+func TestApplyInstrumentationLabelOptions(t *testing.T) {
+	testCases := map[string]struct {
+		instrumentationOptions     []InstrumentationOption
+		expectedAcceptGRPCStatuses bool
```
pstibrany, line 257: it's now called "reportGRPCStatus".

pstibrany (approved):
Great work :clap:

CHANGELOG.md
```diff
@@ -155,6 +155,13 @@
 * [ENHANCEMENT] Memcached: allow to configure write and read buffer size (in bytes). #414
 * [ENHANCEMENT] Server: Add `-server.http-read-header-timeout` option to specify timeout for reading HTTP request header. It defaults to 0, in which case reading of headers can take up to `-server.http-read-timeout`, leaving no time for reading body, if there's any. #423
 * [ENHANCEMENT] Make httpgrpc.Server produce non-loggable errors when a header with key `httpgrpc.DoNotLogErrorHeaderKey` and any value is present in the HTTP response. #421
+* [ENHANCEMENT] Server: Add `-server.report-grpc-codes-in-instrumentation-label` CLI flag to specify whether gRPC status codes should be used in instrumentation labels. It defaults to false, meaning that gRPC status codes are represented with `error` value. #424
+* [ENHANCEMENT] Instrumentation: `middleware.InstrumentationOption` struct, and a special value `middleware.ReportGRPCStatusOption` have been added to allow both server and clients to configure gRPC status code usages in instrumentation labels. This can be optionally used in the following functions: #424
```
pstibrany, line 159:
```suggestion
* [ENHANCEMENT] Instrumentation: Added `middleware.ReportGRPCStatusOption` that can be passed to the following functions to enable reporting of gRPC status codes in "status_code" label (instead of simplified "error", "cancel" or "success" values): #424
```

CHANGELOG.md
```diff
@@ -155,6 +155,13 @@
 * [ENHANCEMENT] Memcached: allow to configure write and read buffer size (in bytes). #414
 * [ENHANCEMENT] Server: Add `-server.http-read-header-timeout` option to specify timeout for reading HTTP request header. It defaults to 0, in which case reading of headers can take up to `-server.http-read-timeout`, leaving no time for reading body, if there's any. #423
 * [ENHANCEMENT] Make httpgrpc.Server produce non-loggable errors when a header with key `httpgrpc.DoNotLogErrorHeaderKey` and any value is present in the HTTP response. #421
+* [ENHANCEMENT] Server: Add `-server.report-grpc-codes-in-instrumentation-label` CLI flag to specify whether gRPC status codes should be used in instrumentation labels. It defaults to false, meaning that gRPC status codes are represented with `error` value. #424
```
pstibrany, line 158:
```suggestion
* [ENHANCEMENT] Server: Add `-server.report-grpc-codes-in-instrumentation-label` CLI flag to specify whether gRPC status codes should be used in `status_code` label of request duration metric. It defaults to false, meaning that gRPC status codes are represented with `error` value. #424
```

middleware/grpc_instrumentation.go
pstibrany: Looks good, thanks.

middleware/grpc_instrumentation.go
pstibrany: > I can try to see how it would look like, but it will for sure complicate the test.

I'm not sure if test is too complicated or not, but I think this function looks good now. Thank you for giving it a try.

middleware/grpc_instrumentation.go
```diff
@@ -173,18 +172,83 @@ func (s *instrumentedClientStream) CloseSend() error {
...
+
+	return "error"
+}
+
+// errorToStatusCode extracts a status code from the given error, and does the following:
+//
+//   - If the error is nil, codes.OK is returned.
+//   - If the error corresponds to context.Canceled, codes.Canceled is returned.
+//   - Otherwise, the actual status code of the error is returned.
+func errorToStatusCode(err error) codes.Code {
```
pstibrany, line 237: nit: This description of implementation seems unnecessary. Same for next function. (I'd only comment public functions, of functions with tricky behaviour. But this is pretty straightforward)

PR #426, opened 2023-11-06
Log closed connections without any response

pstibrany:
**What this PR does**:

HTTP Server closes connections early (**with no response, or running any handler**) if client fails to send complete HTTP **request headers** within timeout. To help troubleshooting such attempts, this PR adds option `-server.http-log-closed-connections-without-response-enabled`, which enables logging of closed connections to HTTP server that received no response. First 512 bytes read from connection is also logged (while removing credentials from Authorization header).

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,6 +162,7 @@
  * `middleware.UnaryClientInstrumentInterceptor`
  * `middleware.StreamClientInstrumentInterceptor`
  * `middleware.Instrument`
+* [ENHANCEMENT] Server: Added new `-server.http-log-closed-connections-without-response-enabled` option to log details about closed connections that received no response. #426
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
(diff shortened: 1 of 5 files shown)


treid314 (approved):
This makes sense, but I'm not sure how it would be used in a production environment. 

Is the zero response listener something we would always enable or just turn on when we see that there's a lot of slow connections getting closed? For the second use case this makes a lot of sense since you typically don't get the user/tenant info at the gateway level and have to do backtracking via IP addresses at the LB.

pstibrany:
> This makes sense, but I'm not sure how it would be used in a production environment.
> 
> Is the zero response listener something we would always enable or just turn on when we see that there's a lot of slow connections getting closed? For the second use case this makes a lot of sense since you typically don't get the user/tenant info at the gateway level and have to do backtracking via IP addresses at the LB.

I would propose to always enable it in environments where HTTP server is directly exposed to the internet (eg. when using AWS NLB), especially when also configuring shorter HTTP header read timeout. It could help us to make some type of currently-ignored connections more visible, and troubleshoot possible customer problems.

PR #427, opened 2023-11-10
Extend ring.DoBatch's error class tracking to all errors

duricanikolic:
**What this PR does**:

Error class tracking has been added to `ring.DoBatch()` in #201. That PR introduced 2 classes of errors: 
- client errors (tracked by `itemTracker.failed4xx`), and
- server errors (tracked by `itemTracker.failed5xx`).

Unfortunately, this change took into account only errors with an HTTP status. In case of a gRPC error, or any other error not implementing `GRPCStatus() *Status` from `google.golang.org/grpc/status`, the error was implicitly treated as a server error. For example, if ring instances from a replication set return gRPC errors with gRPC codes `FailedPrecondition` (due to a client problem), `FailedPrecondition` (do to a client problem) and `Unavailable` (do to a server being currently unavailable), the current approach would return the last error that arrived, i.e., the server `Unavailable` error, although the majority of the instances would have given the client `FailedPrecondition` error.

#### Main change ####
This PR fixes that problem by introducing a new function, 
```
ring.DoBatchWithClientError()
```
which is a generalization of `ring.DoBatch()` that has an additional parameter `isClientError`, a function specified by the caller that differentiates between client and server errors.

#### Minor changes ####
`itemTracker`'s attributes `failed4xx` and `failed5xx` have been renamed to `failedClient` and `failedServer`.

Part of https://github.com/grafana/mimir/issues/6008

**Checklist**
- [x] Tests updated


ring/batch.go
```diff
@@ -134,49 +167,54 @@ func (b *batchTracker) record(itemTrackers []*itemTracker, err error) {
...
+			errCount := itemTracker.recordError(err)
 			// We should return an error if we reach the maxFailure (quorum) on a given error family OR
-			// we don't have any remaining instances to try.
+			// we don't have any remaining instances to try. In the following we use ClientError and ServerError
+			// to annotate instances of ring.Error whose IsClientError() returns true and false respectively.
 			//
-			// Ex: 2xx, 4xx, 5xx -> return 5xx
-			// Ex: 4xx, 4xx, _ -> return 4xx
-			// Ex: 5xx, _, 5xx -> return 5xx
+			// Ex: _ ClientError, SeriverError -> return ServerError
```
dimitarvdimitrov, line 184: SeriverError has a typo

ring/batch.go
```diff
@@ -4,13 +4,44 @@ package ring
...
+}
+
+type batchError struct {
+	isClientError bool
+	err           error
+}
+
+func newError(err error) Error {
+	var e Error
+	if errors.As(err, &e) {
```
dimitarvdimitrov, line 27: i'm not sure about this. This will unwrap the error and also forces `e` to escape to the heap. Can we instead do a simple type assertion `if e, ok := err.(Error); ok {`? That will retain the error and it will be easier to explain how the contract works

ring/batch.go
```diff
@@ -134,49 +167,54 @@ func (b *batchTracker) record(itemTrackers []*itemTracker, err error) {
...
 	// The use of atomic increments here is needed as:
 	// * rpcsPending and rpcsFailed guarantee only a single goroutine will write to either channel
-	// * succeeded, failed4xx, failed5xx and remaining guarantee that the "return decision" is made atomically
+	// * succeeded, failedClient, failedServer and remaining guarantee that the "return decision" is made atomically
 	// avoiding race condition
-	for i := range itemTrackers {
+	for _, itemTracker := range itemTrackers {
+		if itemTracker.done.Load() {
+			continue
+		}
```
dimitarvdimitrov, line 175: what's the reason for this change?

ring/batch.go
```diff
@@ -25,23 +56,25 @@ type instance struct {
...
+	done         atomic.Bool
 }
 
 func (i *itemTracker) recordError(err error) int32 {
 	i.err.Store(err)
 
-	if s, ok := status.FromError(err); ok && s.Code()/100 == 4 {
-		return i.failed4xx.Inc()
+	e := newError(err)
+	if e.IsClientError() {
```
dimitarvdimitrov, line 73: if we throw `e` away immediately, can we just have one function which returns a boolean? would reduce garbage

ring/batch.go
```diff
@@ -4,13 +4,44 @@ package ring
...
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 
 	"go.uber.org/atomic"
 	"google.golang.org/grpc/status"
 )
 
+type Error interface {
```
dimitarvdimitrov, line 15: I'd rename this to something like ClientServerError. Also why do we need the embedded `error` interface at all?

dimitarvdimitrov (commented):
Thanks for working on this. I left a few comments, but the change looks mostly good

pstibrany:
I would suggest simpler design: passing a function to `DoBatch` that categorises errors to server and client-side, or perhaps does the merging of errors. Default function provided in dskit can work with http status codes, while function passed from Mimir can do its own error categorization. WDYT?

ring/ring_test.go
```diff
@@ -174,7 +172,7 @@ func TestDoBatchZeroInstances(t *testing.T) {
 }
 
 func TestDoBatch_QuorumError(t *testing.T) {
-	// we should run several write request to make sure we dont have any race condition on the batchTracker.record code
+	// we should run several write request to make sure we don't have any race condition on the batchTracker code
 	const numberOfOperations = 10000
 	instanceReturnErrors := [3]error{nil, nil, nil}
```
dimitarvdimitrov, line 177: this is now strangely used. We can have a constant with the number of ingesters and use that to populate the ring. I think it will be more clear if `runDoBatch` takes an array of errors instead of the tests having to share `instanceReturnErrors`

ring/ring_test.go
```diff
@@ -203,71 +201,96 @@ func TestDoBatch_QuorumError(t *testing.T) {
...
+					require.NoError(t, err)
+				} else {
+					found := false
+					for i := 0; i < len(testData.acceptableOutcomes) && !found; i++ {
+						found = err == testData.acceptableOutcomes[i]
+					}
+					if !found {
+						fmt.Println(testName, err)
+					}
+					require.True(t, found)
```
dimitarvdimitrov, line 288: you can use `assert.Contains(t, err, testData.acceptableOutcomes)`

ring/ring_test.go
```diff
@@ -203,71 +201,96 @@ func TestDoBatch_QuorumError(t *testing.T) {
...
+		},
+		"1 client, 1 server ring.Errors and 1 success return one of the two ring.Errors": {
+			errors:             [3]error{ringClientError, ringServerError, nil},
+			acceptableOutcomes: []error{ringClientError, ringServerError},
+		},
+		"1 ring.Error and 1 unhealthy instance should return that error": {
+			errors:               [3]error{ringClientError, nil, nil},
+			unavailableInstances: []string{"2"},
+			acceptableOutcomes:   []error{ringClientError},
+		},
```
dimitarvdimitrov, line 269: should we have a test case with mixed httpgrpc and server/client errors?

dimitarvdimitrov (commented):
the tests look better as table tests 👌

dimitarvdimitrov:
> I would suggest simpler design: passing a function to `DoBatch` that categorises errors to server and client-side, or perhaps does the merging of errors. Default function provided in dskit can work with http status codes, while function passed from Mimir can do its own error categorization. WDYT?

Passing a function will probably create fewer allocations for the client of the library.

But I feel like the interface of DoBatch will become more complex. It already takes one function argument which returns an error. Now it will have another func arg which will receive the same error and return a boolean. I feel like an optional interface on the returned error is less invasive and more straight forward.

pstibrany:
> But I feel like the interface of DoBatch will become more complex. It already takes one function argument which returns an error. Now it will have another func arg which will receive the same error and return a boolean. I feel like an optional interface on the returned error is less invasive and more straight forward.

`DoBatch` signature may not be nice, although we could simplify that. Using function would centralize all this logic into single place, with optional interface there will be mix of errors used all over the code base, some implementing the interface and some not.

duricanikolic:
> > But I feel like the interface of DoBatch will become more complex. It already takes one function argument which returns an error. Now it will have another func arg which will receive the same error and return a boolean. I feel like an optional interface on the returned error is less invasive and more straight forward.
> 
> `DoBatch` signature may not be nice, although we could simplify that. Using function would centralize all this logic into single place, with optional interface there will be mix of errors used all over the code base, some implementing the interface and some not.

I have opted for this solution.

ring/batch.go
duricanikolic: My idea was to set the new `done` attribute of an `itemTracker` to true as soon as that `itemTracker` reaches either its `minSuccess` or its `maxFailure`. Once it is achieved, we don't need to consider that `itemTracker` anymore.

ring/ring_test.go
duricanikolic: I have added a couple of tests for this

ring/batch.go
duricanikolic: As discussed via slack, this will be done in a separate PR, and maybe in a different form. I have reverted this part of the PR.

ring/batch.go
```diff
@@ -55,8 +80,11 @@ func (i *itemTracker) recordError(err error) int32 {
 //
 // cleanup() is always called, either on an error before starting the batches or after they all finish.
 //
+// optionalErrorFilters represent a set of conditions that an error returned
+// by the callback function should satisfy in order to be treated as a client error.
+//
 // Not implemented as a method on Ring so we can test separately.
-func DoBatch(ctx context.Context, op Operation, r ReadRing, keys []uint32, callback func(InstanceDesc, []int) error, cleanup func()) error {
+func DoBatch(ctx context.Context, op Operation, r ReadRing, keys []uint32, callback func(InstanceDesc, []int) error, cleanup func(), optionalErrorFilters ...OptionalErrorFilter) error {
```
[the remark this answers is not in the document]
duricanikolic: I have applied the changes you proposed. Thank you

dimitarvdimitrov (approved):
thanks for addressing my comments

ring/batch.go
```diff
@@ -55,8 +79,11 @@ func (i *itemTracker) recordError(err error) int32 {
 //
 // cleanup() is always called, either on an error before starting the batches or after they all finish.
 //
+// optionalErrorFilters represent a set of conditions that an error returned
+// by the callback function should satisfy in order to be treated as a client error.
```
pstibrany, line 83: This sounds like error needs to satisfy all of them. I'd suggest to support single function only, and call it `isClientError`.

I'd suggest to create new `DoBatchWithClientError` function, and implement `DoBatch` by calling `DoBatchWithClientError` with `isHTTPStatus4xx` function.

duricanikolic: I have applied the required changes.

ring/batch.go
```diff
@@ -25,38 +25,54 @@ type instance struct {
...
 // to send to that instance.
 //
-// cleanup() is always called, either on an error before starting the batches or after they all finish.
+// cleanup() is always called, either on an error before starting the batches
+// or after they all finish.
 //
-// Not implemented as a method on Ring so we can test separately.
-func DoBatch(ctx context.Context, op Operation, r ReadRing, keys []uint32, callback func(InstanceDesc, []int) error, cleanup func()) error {
+// isClientError() represents a condition that an error returned the callback
+// function should satisfy in order to be treated as a client error.
```
pstibrany, line 72:
```suggestion
// isClientError() classifies errors returned by `callback()` into client or server errors.
// See `record` function for details how errors are combined into final error returned
// by DoBatchWithClientError.
```

ring/batch.go
```diff
@@ -128,55 +144,55 @@ func DoBatch(ctx context.Context, op Operation, r ReadRing, keys []uint32, callb
...
 	// If we reach the required number of successful puts on this item, then decrement the
 	// number of pending items by one.
 	//
 	// The use of atomic increments here is needed as:
 	// * rpcsPending and rpcsFailed guarantee only a single goroutine will write to either channel
-	// * succeeded, failed4xx, failed5xx and remaining guarantee that the "return decision" is made atomically
+	// * succeeded, failedClient, failedServer and remaining guarantee that the "return decision" is made atomically
 	// avoiding race condition
-	for i := range itemTrackers {
+	for _, itemTracker := range itemTrackers {
```
pstibrany, line 155: nit: we could shorten `itemTracker` to `it`, to avoid shadowing name of the type.

ring/batch.go
```diff
@@ -128,55 +144,55 @@ func DoBatch(ctx context.Context, op Operation, r ReadRing, keys []uint32, callb
...
-			// we don't have any remaining instances to try.
+			// we don't have any remaining instances to try. In the following we use ClientError and ServerError
+			// to annotate instances of ring.Error whose IsClientError() returns true and false respectively.
 			//
-			// Ex: 2xx, 4xx, 5xx -> return 5xx
-			// Ex: 4xx, 4xx, _ -> return 4xx
-			// Ex: 5xx, _, 5xx -> return 5xx
+			// Ex: _ ClientError, ServerError -> return ServerError
+			// Ex: ClientError, ClientError, _ -> return ClientError
+			// Ex: ServerError, _, ServerError -> return ServerError
```
pstibrany, line 166:
```suggestion
			// Ex: success, ClientError, ServerError -> return ServerError
			// Ex: ClientError, ClientError, success -> return ClientError
			// Ex: ServerError, success, ServerError -> return ServerError
```

ring/batch.go
```diff
@@ -128,55 +144,55 @@ func DoBatch(ctx context.Context, op Operation, r ReadRing, keys []uint32, callb
...
+	for _, itemTracker := range itemTrackers {
 		if err != nil {
 			// Track the number of errors by error family, and if it exceeds maxFailures
 			// shortcut the waiting rpc.
-			errCount := itemTrackers[i].recordError(err)
+			errCount := itemTracker.recordError(err, isClientError)
 			// We should return an error if we reach the maxFailure (quorum) on a given error family OR
-			// we don't have any remaining instances to try.
+			// we don't have any remaining instances to try. In the following we use ClientError and ServerError
+			// to annotate instances of ring.Error whose IsClientError() returns true and false respectively.
```
pstibrany, line 162:
```suggestion
			// to denote errors, for which isClientError() returns true and false respectively.
```

pstibrany (approved):
Thank you. Left some small suggestions for consideration.

PR #429, opened 2023-11-13
Fix benchmarks and run them in CI

colega:
There were two benchmarks that were outdated, and didn't pass.
This fixes the benchmarks, adds `make test-benchmarks` target and runs that in CI with the latest golang version.


duricanikolic (approved):
LGTM

PR #431, opened 2023-11-13
ring.DoBatch with goroutine pool

colega:
This adds an option to `ring.DoBatch` to run the callbacks through a function provided that could make use of a goroutine pool (with concurrency.ReusableGoroutinesPool as an implementation).

This should help in Mimir's distributors to avoid extra stack allocation on every new goroutine created to perform a push to an ingester.

Since there was a recent change to `DoBatch` that added an `isClientError` param, I made a more generic version `DoBatchWithOptions`. I also made the cleanup optional, since I see it's
not so common to pass it down.

With this change, I also marked as deprecated the default `ring.DoBatch` version.

I did two benchmark versions, showing some imrovement in CPU time on the pooled version:

```
$ benchstat -col /go bench.txt
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-11700K @ 3.60GHz
                 │   default    │  concurrency.NonBlockingWorkerPool   │
                 │    sec/op    │    sec/op     vs base                │
Batch10x100-16     69.04µ ±  0%   64.53µ ±  0%   -6.53% (p=0.000 n=20)
Batch100x100-16    131.6µ ±  1%   123.8µ ±  1%   -5.94% (p=0.000 n=20)
Batch100x1000-16   2.088m ± 30%   1.146m ± 39%  -45.12% (p=0.001 n=20)
geomean            266.7µ         209.2µ        -21.57%
```

### DIFF
--- a/concurrency/worker.go
+++ b/concurrency/worker.go
@@ -1,0 +1,38 @@
+package concurrency
+
+// NewReusableGoroutinesPool creates a new worker pool with the given size.
+// These workers will run the workloads passed through Go() calls.
+// If all workers are busy, Go() will spawn a new goroutine to run the workload.
+func NewReusableGoroutinesPool(size int) *ReusableGoroutinesPool {
+	p := &ReusableGoroutinesPool{
+		jobs: make(chan func()),
+	}
+	for i := 0; i < size; i++ {
+		go func() {
+			for f := range p.jobs {
+				f()
+			}
+		}()
+	}
+	return p
+}
+
+type ReusableGoroutinesPool struct {
+	jobs chan func()
+}
+
+// Go will run the given function in a worker of the pool.
+// If all workers are busy, Go() will spawn a new goroutine to run the workload.
+func (p *ReusableGoroutinesPool) Go(f func()) {
+	select {
+	case p.jobs <- f:
+	default:
+		go f()
+	}
+}
+
+// Close stops the workers of the pool.
+// No new Do() calls should be performed after calling Close().
+// Close does NOT wait for all jobs to finish, it is the caller's responsibility to ensure that in the provided workloads.
+// Close is intended to be used in tests to ensure that no goroutines are leaked.
+func (p *ReusableGoroutinesPool) Close() { close(p.jobs) }
--- a/concurrency/worker_test.go
+++ b/concurrency/worker_test.go
@@ -1,0 +1,48 @@
+package concurrency
+
+import (
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReusableGoroutinesPool(t *testing.T) {
+	baseGoroutines := runtime.NumGoroutine()
+	const workers = 2
+
+	w := NewReusableGoroutinesPool(workers)
+	require.Equal(t, baseGoroutines+workers, runtime.NumGoroutine())
+
+	// Wait a little bit so both goroutines would be waiting on the jobs chan.
+	time.Sleep(10 * time.Millisecond)
+
+	ch := make(chan struct{})
+	w.Go(func() { <-ch })
+	require.Equal(t, baseGoroutines+workers, runtime.NumGoroutine())
+	w.Go(func() { <-ch })
+	require.Equal(t, baseGoroutines+workers, runtime.NumGoroutine())
+	w.Go(func() { <-ch })
+	require.Equal(t, baseGoroutines+workers+1, runtime.NumGoroutine())
+
+	// end workloads, we should have only the workers again.
+	close(ch)
+	for i := 0; i < 1000; i++ {
+		if runtime.NumGoroutine() == baseGoroutines+workers {
+			break
+		}
+		time.Sleep(time.Millisecond)
+	}
+	require.Equal(t, baseGoroutines+workers, runtime.NumGoroutine())
+
+	// close the workers, eventually they should be gone.
+	w.Close()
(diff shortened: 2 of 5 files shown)


ring/batch.go
```diff
@@ -50,32 +50,56 @@ func isHTTPStatus4xx(err error) bool {
...
+	// IsClientError classifies errors returned by `callback()` into client or server errors.
+	// See `batchTracker.record()` function for details about how errors are combined into final error returned by DoBatchWithClientError.
+	// If nil, a default implementation is used that classifies grpc errors containing status codes 4xx as client errors.
+	IsClientError func(error) bool
+
+	// Go will be used to spawn the callback goroutines, and can be used to use a worker pool like concurrency.ReusableGoroutinesPool.
+	Go func(func())
+}
+
+func (o *DoBatchOptions) setDefaults() {
```
duricanikolic, line 78: Could we rename this method in something like `setDefaultsIfNeeded()`.
IMO seeing `o.setDefaults()` could be interpreted as "we get `DoBatchOptions` o, and we set default values on all its fields`. WDYT?

ring/batch.go
```diff
@@ -50,32 +50,56 @@ func isHTTPStatus4xx(err error) bool {
...
+// we track the relationship between batch RPCs and the items within them.
 //
-// isClientError() classifies errors returned by `callback()` into client or
-// server errors. See `batchTracker.record()` function for details about how
-// errors are combined into final error returned by DoBatchWithClientError.
+// See comments on DoBatchOptions for available options for this call.
 //
 // Not implemented as a method on Ring, so we can test separately.
-func DoBatchWithClientError(ctx context.Context, op Operation, r ReadRing, keys []uint32, callback func(InstanceDesc, []int) error, cleanup func(), isClientError func(error) bool) error {
+func DoBatchWithOptions(ctx context.Context, op Operation, r ReadRing, keys []uint32, callback func(InstanceDesc, []int) error, o DoBatchOptions) error {
```
duricanikolic, line 98: Did you remove `DoBatchWithClientError()`? 
I would put it back and possibly mark it as deprecated, since it will now be part of `r264`.

duricanikolic:
Could you please update `CHANGELOG.md` and write the improtant changes this PR brings:
- `ring.DoBatchWithOptions()` has been added.
- `ring.DoBatch()` and `ring.DoBatchWithClientError()` have been deprecated.

ring/batch.go
colega: We don't do server in dskit. We're already very generous at leaving the `DoBatch` with the old signature: I didn't change it just  to make the update path easier for codebases with tangled dependencies.

I don't see any need to leave `DoBatchWithClientError` here.

concurrency/worker.go
```diff
@@ -0,0 +1,38 @@
+package concurrency
+
+// NewReusableGoroutinesPool creates a new worker pool with the given size.
+// These workers will run the workloads passed through Go() calls.
+// If all workers are busy, Go() will spawn a new goroutine to run the workload.
+func NewReusableGoroutinesPool(size int) *ReusableGoroutinesPool {
```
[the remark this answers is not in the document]
colega: Yes, there will be a CLI flag just like there's one in the dskit/server for grpc workers.

ring/batch.go
colega: I can rename it but given it's an unexported method with a single usage that fits on the same screen I don't think that any naming can lead to confusion.

colega: WDYT about `replaceZeroValuesWithDefaults()` then?

duricanikolic: > WDYT about replaceZeroValuesWithDefaults() then?

That looks good.

ring/batch.go
duricanikolic: Then we should replace `DoBatchWithClientError` with `DoBatchWithOptions` in `CHANGELOG` too, and mention both PRs there. 
The usage of `DoBatchWithClientError` should be removed from Mimir once the vendored `dskit` is updated.

ring/batch.go
colega: Renamed.

colega:
I have updated the `CHANGELOG.md`, however since we don't do releases, and changelog is sorted by section type (CHANGE/FEATURE/ENHANCEMENT/BUGFIX) instead of the date, I think it's quite useless.

duricanikolic (approved):
Thank you for addressing my comments.
LGTM

krajorama (changes requested):

concurrency/worker_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReusableGoroutinesPool(t *testing.T) {
+	baseGoroutines := runtime.NumGoroutine()
+	const workers = 2
+
+	w := NewReusableGoroutinesPool(workers)
+	require.Equal(t, baseGoroutines+workers, runtime.NumGoroutine())
```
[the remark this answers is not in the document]
colega: Here's a test from golang/go doing the same thing, for reference: https://github.com/golang/go/blob/dd88f23a2006307f835f42063b5168ec56c2c428/src/net/http/httputil/dump_test.go#L254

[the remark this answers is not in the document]
krajorama: Ack

krajorama (approved):
Not an expert on goroutines, but LGTM

PR #432, opened 2023-11-17
Remove an un-necessary go routine in a test

duricanikolic:
**What this PR does**:
`TestBatchTracker_Record` introduced in https://github.com/grafana/dskit/pull/430 contains an un-necessary go routine in which we just wait for a `WaitGroup`, as remarked [here](https://github.com/grafana/dskit/pull/430#discussion_r1394985293). This PR removes that go routine.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/batch_test.go
+++ b/ring/batch_test.go
@@ -69,9 +69,7 @@ func executeScenario(bt batchTracker, scenarios []scenario) error {
 		}
 	}()
 
-	go func() {
-		wg.Wait()
-	}()
+	wg.Wait()
 
 	select {
 	case err := <-bt.err:


aknuds1 (approved):

ring/batch_test.go
```diff
@@ -69,9 +69,7 @@ func executeScenario(bt batchTracker, scenarios []scenario) error {
 		}
 	}()
 
-	go func() {
-		wg.Wait()
-	}()
+	wg.Wait()
```
charleskorn, line 72: Why do we need to wait on the `WaitGroup` at all? Won't the `select` below block?

If you want to ensure the goroutine terminates, perhaps moving this below the `select` would be best. Or this could even be simplified to have the goroutine close a channel once the loop is complete, and then the test can wait for that channel to be closed.

PR #433, opened 2023-11-17
Ring: add method to compute token ranges owned by given instance.

pstibrany:
**What this PR does**:
This PR adds `GetTokenRangesForInstance` method to `*Ring`. This method returns token ranges owned by given instance. Implementation is simplified to only support case when zones are enabled, and number of zones is equal to replication factor.

This PR also adds `numberOfKeysOwnedByInstance` method to the `*Ring`, which is more universal way of computing key ownership (it supports zones, instance states, and `zoneCount!=RF` case... but is also 14x slower). This is built on existing code for finding instances for given key, and is used to verify implementation of `(*Ring).GetTokenRangesForInstance` + `(TokenRanges).IncludesKey` methods.

We would like to use this code to implement better series limiting in Mimir by considering series ownership.

**Checklist**
- [x] Tests updated


ring/token_range.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"math"
+
+	"github.com/pkg/errors"
+	"golang.org/x/exp/slices" // using exp/slices until moving to go 1.21.
+)
+
+// TokenRanges describes token ranges owned by an instance.
+// It consists of [start, end] pairs, where both start and end are inclusive.
+type TokenRanges []uint32
```
duricanikolic, line 12: I would be more expressive here and I would say something about the length of the slice, and also about the special cases you handle in the implementation.
E.g., this slice contains an even number of elements, elements at even and odd positions represent range starts and ends respectively, etc.

ring/token_range.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (tr TokenRanges) IncludesKey(key uint32) bool {
```
duricanikolic, line 14: I don't know whether it might be of interest for this implementation, but there is a struct called `ringToken` declared in `ownership_priority_queue.go` as:
```
type ringToken struct {
	token     uint32
	prevToken uint32
}
```

ring/token_range.go
```diff
@@ -0,0 +1,136 @@
...
+		// hash would be inserted after the start of a range (even index)
+		return true
+	default:
+		return false
+	}
+}
+
+// GetTokenRangesForInstance returns the token ranges owned by an instance in the ring.
+//
+// Current implementation only works with multizone setup, where number of zones is equal to replication factor.
```
duricanikolic, line 41: I don't know whether it might be of interest for this implementation, but there is a struct called `ringToken` declared in `ownership_priority_queue.go` as:
```
type ringToken struct {
	token     uint32
	prevToken uint32
}
```

ring/token_range.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	// at this point zone-aware replication is enabled, and rf == numZones
+	// this means that we will write to one replica in each zone, so we can just consider the zonal ring for our instance
+	subringTokens, ok := r.ringTokensByZone[instance.Zone]
+	if !ok || len(subringTokens) == 0 {
+		return nil, errors.New("no tokens for zone")
+	}
+
+	ranges := make([]uint32, 0, 2*(len(instance.Tokens)+1)) // 1 range (2 values) per token + one additional if we need to split the rollover range
```
duricanikolic, line 72: Is the rollover range a range having the start token higher than the end token? E.g., $[2^{32} - 10, 15]$

duricanikolic (commented):
I left some comments

ring/token_range_test.go
```diff
@@ -0,0 +1,306 @@
...
+				"instance-0-0": {0, math.MaxUint32},
+			},
+		},
+		"simple ranges": {
+			zoneTokens: map[string][]uint32{
+				"instance-0-0": {25, 75},
+				"instance-0-1": {10, 50, 100},
+			},
+			expected: map[string]TokenRanges{
+				"instance-0-0": {10, 24, 50, 74},
```
duricanikolic, line 45: This is something that I personally don’t agree with: if token 10 belongs to an instance, I’d expect the corresponding range to include that token

ring/token_range_test.go
```diff
@@ -0,0 +1,306 @@
...
+		_, _ = ring.GetTokenRangesForInstance("instance-1")
+	}
+}
+
+func TestCheckingOfKeyOwnership(t *testing.T) {
+	const instancesPerZone = 100
+	const numZones = 3
+	const numTokens = 512
+	const replicationFactor = numZones // This is the only config supported by GetTokenRangesForInstance right now.
+
```
pracucci, line 174: [nit] You may want to init and log the rand seed here to be able to eventually reproduce a failure.

ring/token_range_test.go
```diff
@@ -0,0 +1,306 @@
...
+	for uid, tokens := range userTokens {
+		shardSize := shardSizes[uid]
+
+		subRing := ring.ShuffleShard(uid, shardSize)
+		sr := subRing.(*Ring)
+
+		// find some instance in subring
+		var instanceID string
+		for id := range sr.ringDesc.Ingesters {
```
pracucci, line 212: [nit] I would run this test for every instance in the ring. It's an easy change to do, but would increase our confidence.

pracucci (approved):
Good job! The logic makes sense to me. I left nits, but I haven't found any issue.

ring/token_range.go
pstibrany: Documenting implementation details would commit to this implementation. Given the early nature of the development, we may still want or need to change it. I prefer to hide implementation details in supporting functions/methods, and only document what's necessary for clients to use.

ring/token_range.go
pstibrany: I wouldn't reuse the type only because it has same two uint32 fields. I was considering to create custom type with similar fields (start/end), but decided to keep implementation as-is for now, as the code is pretty straightforward to follow (imho).

ring/token_range.go
pstibrany: > Is the rollover range a range having the start token higher than the end token? E.g., [232−10,15]

Yes, but it's stored as two ranges in the result.

ring/token_range_test.go
pstibrany: I agree with you that it's confusing, and wouldn't mind if we fixed that (in some other PR).

ring/token_range.go
```diff
@@ -0,0 +1,136 @@
...
+
+	rf := r.cfg.ReplicationFactor
+	numZones := len(r.ringTokensByZone)
+
+	// To simplify computation of token ranges, we currently only support case where zone-awareness is enabled,
+	// and replicaction factor is equal to number of zones.
+	if !r.cfg.ZoneAwarenessEnabled || rf != numZones {
+		// if zoneAwareness is disabled we need to treat the whole ring as one big zone, and we would
+		// need to walk the ring backwards looking for RF-1 tokens from other instances to determine the range.
+		return nil, errors.New("can't use ring configuration for computing token ranges")
```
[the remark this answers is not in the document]
pstibrany: >  I would rather explain "why" telling that this function only support the zone-aware ring with RF = num zones.

True "why" is because we don't need it for our purposes, and it wasn't worth the effort to add support for all cases. I think we should eventually do it, so that Mimir feature using this has full support of various Mimir setups.

I don't think we need to explain this reasoning beyond what's mentioned in the comment:
```
// Current implementation only works with multizone setup, where number of zones is equal to replication factor.
```

ring/token_range.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, ErrInconsistentTokensInfo
+	}
+
+	if firstTokeninfo.InstanceID == instanceID {
+		// we'll start by looking for the beginning of the range that ends with math.MaxUint32
+		rangeEnd = math.MaxUint32
+	}
+
+	// walk the ring backwards, alternating looking for ends and starts of ranges
+	for i := len(subringTokens) - 1; i > 0; i-- {
```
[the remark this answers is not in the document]
pstibrany: Thank you. I agree there's room for improvement (also see previous comment about not supporting other cases). @duricanikolic also had some ideas (mentioned privately) how to simplify this.

I will keep the current implementation to focus on the main feature we're working on, but will keep this in mind as possible optimization.

ring/token_range_test.go
pstibrany: I've added `initTokenGenerator` and passing it to `generateRingInstances` now, modified many tests in the progress. (But not all, there's still `GenerateTokens` function that generates random tokens without logging the seed)

ring/token_range_test.go
pstibrany: Great idea.

pstibrany: Sure enough, it found bugs!

pstibrany: It turned out that our ring tests can be flaky, because our tests could produce conflicting tokens. After fixing token generation in tests, test is now reliable.

ring/token_range.go
```diff
@@ -0,0 +1,136 @@
...
+	} else {
+		if firstTokeninfo.InstanceID == instanceID {
+			ranges = append(ranges, rangeEnd, 0)
+		} else {
+			ranges = append(ranges, rangeEnd, firstToken)
+		}
+	}
+
+	// Ensure returned ranges are sorted.
+	slices.Sort(ranges)
```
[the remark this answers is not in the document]
pstibrany: You're right that reversing should work, if we preserve order of adding of tokens to the slice.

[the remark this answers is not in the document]
pstibrany: I gave it a quick try, and tests started to fail, so I'm keeping current code to move on. May look at this later.

ring/token_range_test.go
pstibrany: Fixed that in other tests too: https://github.com/grafana/dskit/pull/437

ring/token_range.go
duricanikolic: I agree that we shouldn’t document implementationsl details, but giving an example saying something like “_e.g., [5, 10, 20, 30] means that the corresponding instance covers tokens 5-10 and 20-30_” wouldn’t damage.

ring/token_range.go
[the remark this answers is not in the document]
pr00se: My plan was to use `slices.Reverse()` (which is why the `append`s above are slightly awkward), but I ran into (likely) similar issues to Peter, so I opted to just move on at the time.

ring/token_range_test.go
pr00se: I also agree that it's confusing, but it's a function of [this code](https://github.com/grafana/dskit/blob/main/ring/util.go#L129-L140), which causes items that map directly to a token to actually belong to the range owned by the NEXT token.

We could have used half-open ranges in the implementation (`[start, end)`), but that would have made the `IncludesKey` logic more complicated, since we'd need to do an additional check if the `binarySearch` found the key in the ranges slice.

In my opinion, having the token ranges be closed is also easier to reason about, since you don't need to remember the implementation detail of ring instances not actually "owning" any items that map directly to the tokens they have claimed.

pstibrany: Note, by "fixing" this, I didn't mean to change how token ranges work. I agree that closed ranges make sense.

What I meant by fixing is modifying logic in `searchToken` method, so that if ingester owns the token, it actually "owns" it :)

PR #434, opened 2023-11-17
encapsulate httpgrpc to http request unwrapping as public fn

francoposa:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -58,7 +58,7 @@ func (n nopCloser) BytesBuffer() *bytes.Buffer { return n.Buffer }
 
 // Handle implements HTTPServer.
 func (s Server) Handle(ctx context.Context, r *httpgrpc.HTTPRequest) (*httpgrpc.HTTPResponse, error) {
-	req, err := http.NewRequest(r.Method, r.Url, nopCloser{Buffer: bytes.NewBuffer(r.Body)})
+	req, err := UnwrapHTTPRequest(ctx, r)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


httpgrpc/server/server.go
```diff
@@ -165,8 +165,8 @@ func NewClient(address string) (*Client, error) {
 	}, nil
 }
 
-// HTTPRequest wraps an ordinary HTTPRequest with a gRPC one
-func HTTPRequest(r *http.Request) (*httpgrpc.HTTPRequest, error) {
+// WrapHTTPRequest wraps an ordinary http.Request up into an httpgrpc.HTTPRequest
+func WrapHTTPRequest(r *http.Request) (*httpgrpc.HTTPRequest, error) {
```
francoposa, line 169: there is one usage of this in Mimir/GEM that would need updated with the new name, but if we're super adverse to changing that, it could remain as it was.

httpgrpc/server/server.go
```diff
@@ -179,6 +179,19 @@ func HTTPRequest(r *http.Request) (*httpgrpc.HTTPRequest, error) {
...
+// UnwrapHTTPRequest unwraps an ordinary http.Request from an httpgrpc.HTTPRequest
+func UnwrapHTTPRequest(ctx context.Context, r *httpgrpc.HTTPRequest) (*http.Request, error) {
+	req, err := http.NewRequest(r.Method, r.Url, nopCloser{Buffer: bytes.NewBuffer(r.Body)})
+	if err != nil {
+		return nil, err
+	}
+	toHeader(r.Headers, req.Header)
+	req = req.WithContext(ctx)
+	req.RequestURI = r.Url
+	req.ContentLength = int64(len(r.Body))
```
leizor, line 191: We don't need to repeat these steps in `Handle` above, do we?

francoposa: yep, just fixed

leizor (approved):
LGTM!

httpgrpc/server/server.go
pstibrany: It's not "wrapping" (new message doesn't keep original request), but full conversion, so the new naming is wrong.

PR #437, opened 2023-11-22
Use seeded random generator in tests.

pstibrany:
**What this PR does**:

This PR replaces use of `GenerateTokens` in tests with generator initialized by `initTokenGenerator(b)`, which also logs used seed.

Extracted from https://github.com/grafana/dskit/pull/433, to make it smaller.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/token_generator.go
+++ b/ring/token_generator.go
@@ -3,6 +3,7 @@ package ring
 import (
 	"math/rand"
 	"sort"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


ring/token_generator.go
```diff
@@ -35,16 +43,17 @@ func (t *RandomTokenGenerator) GenerateTokens(requestedTokensCount int, allTaken
...
-
 	used := make(map[uint32]bool, len(allTakenTokens))
 	for _, v := range allTakenTokens {
 		used[v] = true
 	}
 
 	tokens := make([]uint32, 0, requestedTokensCount)
 	for i := 0; i < requestedTokensCount; {
-		candidate := r.Uint32()
+		t.m.Lock()
```
duricanikolic, line 53: Why do we need this?

ring/token_generator_test.go
```diff
@@ -34,8 +34,3 @@ func TestRandomTokenGenerator_IgnoresOldTokens(t *testing.T) {
 		}
 	}
 }
-
-// GenerateTokens generates numTokens unique, random and sorted tokens for testing purposes.
-func GenerateTokens(tokensCount int, takenTokens []uint32) Tokens {
```
duricanikolic, line 39: [nit] WDYT if we keep this function, and initialize the token generator inside it?
This way the diff in `ring_test.go` would be minimizined.

ring/token_generator.go
pstibrany: Because access to `*math.Rand` from different goroutines needs to be serialized.

ring/token_generator_test.go
pstibrany: Many tests call that function multiple times, which would then initialize multiple generators. I tried to avoid having too many generators per test, although there may be some tests where there is more than one used.

ring/token_generator.go
duricanikolic: I see, previously we used to create a new instance of `Rand` inside the method itself, and it was used by only 1 go routine. But it is not the case anymore.

duricanikolic (approved):
LGTM
Thank you

PR #439, opened 2023-11-23
ClientPool: expose method to remove client by value

dimitarvdimitrov:
This PR adds a method to remove a client from the pool only if it already hasn't been removed.

This helps in cases where concurrent users of the client want to remove the same client instance and then immediately after that get a new instance. Clients can remove each other's new instances since the access to the pool is serialized via the mutex. If the clients retry removing a client on failed connections this can lead to live-lock.


pstibrany (approved):
Thank you.

PR #440, opened 2023-11-24
`DoUntilQuorum`: add the ability to prioritise zones when request minimisation is enabled

charleskorn:
**What this PR does**:

This PR adds support for prioritising zones when using `DoUntilQuorum` with request minimisation and zone awareness.

This is useful when the caller has additional knowledge about the zones and wants to prioritise using some over others. For example, the caller might have observed that some zones are faster or under less load and want to preference these over other, slower or more heavily loaded zones.

Related: https://github.com/grafana/mimir/pull/6726

**Checklist**
- [x] Tests updated


pstibrany (approved):
Looks reasonable.

PR #441, opened 2023-11-24
Token ranges: test different instance states.

pstibrany:
**What this PR does**:

This PR modifies `TestCheckingOfKeyOwnership` test to randomize instance states.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/token_range_test.go
+++ b/ring/token_range_test.go
@@ -3,6 +3,7 @@ package ring
 import (
 	"fmt"
 	"math"
+	"math/rand"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bboreham (commented):
I worry that we are now not testing the steady-state of all ingesters ready.
If you agree, maybe running the test once with all-ready and once with randomized states would be more thorough?

ring/token_range_test.go
```diff
@@ -195,6 +200,12 @@ func TestCheckingOfKeyOwnership(t *testing.T) {
 
 	// Generate ring
 	ringDesc := &Desc{Ingesters: generateRingInstances(gen, instancesPerZone*numZones, numZones, numTokens)}
+	// Switch states of some ingesters
```
pracucci, line 203: I have a feedback similar to Bryan's one. The comment says "some ingesters" but I read the code as "all ingesters".

pstibrany:
> I worry that we are now not testing the steady-state of all ingesters ready. If you agree, maybe running the test once with all-ready and once with randomized states would be more thorough?

Good point. I've added test without states randomization too.

pracucci (approved):

PR #444, opened 2023-11-28
Add ability to log all source hosts from http header instead of only the first one.

bubu11e:
This PR add the ability to extract all hosts from a custom header or the standard one Forwarded, X-Real-IP, X-Forwraded-For.
    
This list of hosts can then be used for informational purposes.
    
Previous behavior remains the default one and new one can be triggered by setting a configuration flag `server.log-source-ips-full to true.`

This PR follow [this conversation](https://grafana.slack.com/archives/C039863E8P7/p1700840526610699).

### DIFF
--- a/middleware/source_ips.go
+++ b/middleware/source_ips.go
@@ -18,6 +18,9 @@ var (
 	// De-facto standard header keys.
 	xForwardedFor = http.CanonicalHeaderKey("X-Forwarded-For")
 	xRealIP       = http.CanonicalHeaderKey("X-Real-IP")
+	// Allows to extract the host from the X-Forwarded-For header.
+	// Will strip out any spaces or double quote surrounding host.
+	xForwardedForRegex = regexp.MustCompile(`(?: *"?([^,]+)"? *)`)
 )
 
 var (
@@ -25,9 +28,9 @@ var (
 	// existing use of X-Forwarded-* headers.
 	// e.g. Forwarded: for=192.0.2.60;proto=https;by=203.0.113.43
 	forwarded = http.CanonicalHeaderKey("Forwarded")
-	// Allows for a sub-match of the first value after 'for=' to the next
-	// comma, semi-colon or space. The match is case-insensitive.
-	forRegex = regexp.MustCompile(`(?i)(?:for=)([^(;|,| )]+)`)
+	// Allows to extract the host from the for clause of the Forwarded header.
+	// Will strip out any spaces or double quote surrounding host.
+	forwardedRegex = regexp.MustCompile(`(?i)(?:for=)(?: *"?([^;,]+)"? *)`)
 )
 
 // SourceIPExtractor extracts the source IPs from a HTTP request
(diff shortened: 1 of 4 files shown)


middleware/source_ips.go
```diff
@@ -72,7 +78,15 @@ func extractHost(address string) string {
...
 func (sips SourceIPExtractor) Get(req *http.Request) string {
-	fwd := extractHost(sips.getIP(req))
+	hosts := []string{}
+
+	// Remove port informations from extracted address
+	for _, addr := range sips.getIP(req) {
+		hosts = append(hosts, extractHost(addr))
+	}
+
+	fwd := strings.Join(hosts, ", ")
```
56quarters, line 88: nit: Should we prefer `strings.Join(hosts, ",")` so that consumers don't have to trim whitespace?

middleware/source_ips.go
```diff
@@ -94,52 +108,45 @@ func (sips SourceIPExtractor) Get(req *http.Request) string {
...
-		s := strings.Index(fwd, ",")
-		if s == -1 {
-			s = len(fwd)
+		addrs = sips.extractHeader(fwd, xForwardedForRegex)
+	}
+
+	return addrs
+}
+
+// Just a toolbox function that will parse a header content with a regex and return a list
```
56quarters, line 135:
```suggestion
// extractHeader is a toolbox function that will parse a header content with a regex and return a list
```

server/server.go
```diff
@@ -194,6 +195,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.LogFormat, "log.format", log.LogfmtFormat, "Output log messages in the given format. Valid formats: [logfmt, json]")
 	cfg.LogLevel.RegisterFlags(f)
 	f.BoolVar(&cfg.LogSourceIPs, "server.log-source-ips-enabled", false, "Optionally log the source IPs.")
+	f.BoolVar(&cfg.LogSourceIPsFull, "server.log-source-ips-full", false, "Log all source hosts instead of only the first one. Only used if server.log-source-ips-enabled is true")
```
56quarters, line 198: Can you clarify this message a bit? e.g.
```suggestion
	f.BoolVar(&cfg.LogSourceIPsFull, "server.log-source-ips-full", false, "Log all source IPs instead of only the originating one. Only used if server.log-source-ips-enabled is true")
```

middleware/source_ips_test.go
```diff
@@ -192,7 +192,198 @@ func TestGetSourceIPs(t *testing.T) {
...
+		},
+		{
+			name: "Forwarded returns hostname",
+			req: &http.Request{
+				RemoteAddr: "192.168.1.100:3454",
+				Header: map[string][]string{
+					http.CanonicalHeaderKey(forwarded): {"for=workstation.local"},
+				},
+			},
+			want: "workstation.local, 192.168.1.100",
```
56quarters, line 381: Can you add a test for `Forwarded` that verifies the matching is case insensitive? e.g. `For=192.168.1.1`

56quarters (approved):
LGTM with a few small changes

middleware/source_ips.go
bubu11e: Change made

middleware/source_ips.go
bubu11e: Change made

server/server.go
bubu11e: Change made

middleware/source_ips.go
bubu11e: That would change the current behavior as in current implementation this space is present. However if it's ok with you we can change it ?

middleware/source_ips_test.go
bubu11e: Good idea, implemented.

middleware/source_ips.go
56quarters: Nah, good point. Let's leave it.

PR #445, opened 2023-11-29
Remove global state from tenant ID parsing package

### LINKED ISSUE #443: Tenant package relies on mutable global state
The `tenant` package is used for (among other things) extracting a tenant ID from a `context.Context`. 

There are two separate implementations of this logic:
* Logic that understands tenant IDs separated by a `|` and so can return multiple tenant IDs from a context.
* Logic that does not understand tenant IDs separate by a `|` and so returns an error in this case.

There is a `WithDefaultResolver` that changes which logic is used, [via mutating a global](https://github.com/grafana/dskit/blob/4cfd2dc0d14c7e79e6482f231c8c58ae6ceea6d7/tenant/resolver.go#L14-L17). [Mimir](https://github.com/grafana/mimir/blob/e974d3d816e42b8c182b7acbc0e205ce5251da44/pkg/mimir/mimir.go#L720-L727) and [Loki](https://github.com/grafana/loki/blob/bd505f8e2d37172ff35a89f4ac42efec9566a263/pkg/loki/modules.go#L332-L337) do this in response to some form of multi-tenant queries being enabled or disabled by configuration.

This is a problem because the behavior of huge portions of each database can change by changes made far away from the code in question. It also makes it more difficult to unit test the code in question since adjusting the global state for a single test affects the behavior of every other test.

I propose:

* Remove the `WithDefaultResolver` to prevent mutation of global state
* Remove all global state since it's unnecessary. The logic of `MultiResolver` would move to the global `tenant.TenantID()` and `tenant.TenantIDs()` methods. 
* This would also make `SingleResolver`, `MultiResolver`, and `Resolver` unnecessary

This would require changes in Mimir and Loki to reject multi-tenant queries when they aren't enabled. However, this would result in a better design anyway since the databases are checking for and enforcing exactly what they intend to enforce.

56quarters:
**What this PR does**:

Remove the ability to set a global "default" parser for tenant IDs since this makes uses of this package fragile and bug prone. This also removes the `SingleResolver` struct which does not support multiple tenant IDs and was previously the default logic.

All code for parsing tenant IDs is now aware of multiple tenant IDs separated by a `|` character. Consumers that don't want to support multiple tenant IDs at all should use the `TenantID()` method which returns an error if there are multiple tenant IDs. Consumers that wish to optionally support multiple tenant IDs should validate incoming input to ensure only a single ID is present.

This changes the default behavior of `TenantID()` and `TenantIDs()` in two
ways:
    
* `SingleResolver` did not previously enforce a limit on the length of a tenant
   ID. A limit of 150 characters is now enforced. This has always been the
   documented behavior as far back as Cortex, where this code originated. Not
   enforcing it was an oversight.
* `SingleResolver` previously allowed tenant IDs to contain the `|` character.
   This is no longer allowed as part of a tenant ID and instead will be treated
   as a divider between multiple tenant IDs. This has always been the documented
   behavior as far back as Cortex, where this code originated. Not enforcing it
   was an oversight.

Fixes #443

**Checklist**
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,7 +60,7 @@
 * [CHANGE] server: fix incorrect spelling of "gRPC" in `server.Config` fields. #422
 * [CHANGE] memberlist: re-resolve `JoinMembers` during full joins to the cluster (either at startup or periodic). The re-resolution happens on every 100 attempted nodes. This helps speed up joins and respect context cancelation #411
 * [CHANGE] ring: `ring.DoBatch()` was deprecated in favor of `DoBatchWithOptions()`. #431
-* [CHANGE] tenant: Remove `tenant.WithDefaultResolver()` and `SingleResolver` in favor of global functions `tenant.TenantID()`, `tenant.TenantIDs()`,  or `MultiResolver`. #TBD
+* [CHANGE] tenant: Remove `tenant.WithDefaultResolver()` and `SingleResolver` in favor of global functions `tenant.TenantID()`, `tenant.TenantIDs()`,  or `MultiResolver`. #445
 * [FEATURE] Cache: Add support for configuring a Redis cache backend. #268 #271 #276
 * [FEATURE] Add support for waiting on the rate limiter using the new `WaitN` method. #279
 * [FEATURE] Add `log.BufferedLogger` type. #338
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@
 * [CHANGE] server: fix incorrect spelling of "gRPC" in `server.Config` fields. #422
 * [CHANGE] memberlist: re-resolve `JoinMembers` during full joins to the cluster (either at startup or periodic). The re-resolution happens on every 100 attempted nodes. This helps speed up joins and respect context cancelation #411
 * [CHANGE] ring: `ring.DoBatch()` was deprecated in favor of `DoBatchWithOptions()`. #431
+* [CHANGE] tenant: Remove `tenant.WithDefaultResolver()` and `SingleResolver` in favor of global functions `tenant.TenantID()`, `tenant.TenantIDs()`,  or `MultiResolver`. #TBD
 * [FEATURE] Cache: Add support for configuring a Redis cache backend. #268 #271 #276
 * [FEATURE] Add support for waiting on the rate limiter using the new `WaitN` method. #279
 * [FEATURE] Add `log.BufferedLogger` type. #338
--- a/tenant/resolver.go
+++ b/tenant/resolver.go
@@ -2,14 +2,9 @@ package tenant
 
 import (
 	"context"
-	"errors"
 	"strings"
 
 	"github.com/grafana/dskit/user"
-)
-
-var (
-	errInvalidTenantID = errors.New("invalid tenant ID")
 )
 
 // TenantID returns exactly a single tenant ID from the context. It should be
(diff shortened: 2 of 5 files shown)


56quarters:
An example of how multi-tenant queries could be rejected when tenant-federation is not enabled in Mimir: https://github.com/grafana/mimir/pull/6959/files#diff-2a3eea939d04482548f6edf1e004ecc720e1bb218963865cf9e9ad706f5fd5b8

leizor (approved):
Makes sense to me 👍

tenant/resolver.go
```diff
@@ -3,18 +3,14 @@ package tenant
...
 )
 
-var defaultResolver Resolver = NewSingleResolver()
-
-// WithDefaultResolver updates the resolver used for the package methods.
-func WithDefaultResolver(r Resolver) {
-	defaultResolver = r
-}
+var (
+	errInvalidTenantID = errors.New("invalid tenant ID")
```
dimitarvdimitrov, line 12: can we add some details about what's invalid (`"tenant ID is ., .. or contains / or \"`)

56quarters:
> SGTM, not suuper sure of the things that might break with this change.

* Anyone using `|` in a tenant ID and not intending to treat it as the separator between multiple IDs. This doesn't affect Mimir or Loki in Grafana Cloud. It doesn't affect on-prem GEM or GEL since we don't allow `|` in tenant names.  It may affect on-prem installations or Mimir or Loki _but_ `|` has been documented as not being allowed since Mimir has existed.

* Anyone with a tenant ID longer than 150 bytes. This doesn't affect Mimir or Loki in Grafana Cloud. It doesn't affect on-prem GEM or GEL since we limit tenant names to 63 characters there. It may affect on-prem installations of Mimir or Loki but again this has been documented as not being allowed since Mimir has existed.

dimitarvdimitrov:
> SGTM, not suuper sure of the things that might break with this change.

I was concerned about places that under the SingleResolver used to call `TenantID` and assume it will be concatenated list tenant IDs. Now effectively we're only using the MultiResolver.

But you said this would have only happened when multi-tenant is disabled and the tenant ID contains `|`, which as you pointed out is not supported (at least in Mimir and Loki's context). So I'm less concerned now. Even if other clients of dskit hit some problems I think it's better to hit them with a change like this than as bug later on. LGTM

tenant/resolver.go
56quarters: I'm realizing now that it doesn't make sense to have `containsUnsafePathSegments` separate from the logic in `ValidTenantID`. WDYT about moving the logic from `containsUnsafePathSegments` to `ValidTenantID`?

56quarters: Fixed in cb0052572fc643943975681cad008d9610787447 and moved that logic into `ValidTenantID()`

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@
 * [CHANGE] server: fix incorrect spelling of "gRPC" in `server.Config` fields. #422
 * [CHANGE] memberlist: re-resolve `JoinMembers` during full joins to the cluster (either at startup or periodic). The re-resolution happens on every 100 attempted nodes. This helps speed up joins and respect context cancelation #411
 * [CHANGE] ring: `ring.DoBatch()` was deprecated in favor of `DoBatchWithOptions()`. #431
+* [CHANGE] tenant: Remove `tenant.WithDefaultResolver()` and `SingleResolver` in favor of global functions `tenant.TenantID()`, `tenant.TenantIDs()`,  or `MultiResolver`. #TBD
```
56quarters, line 63:
```suggestion
* [CHANGE] tenant: Remove `tenant.WithDefaultResolver()` and `SingleResolver` in favor of global functions `tenant.TenantID()`, `tenant.TenantIDs()`,  or `MultiResolver`. #445
```

PR #447, opened 2023-12-06
Add URL clustering package

eskirk:
**What this PR does**:
This PR aims to add a new package to dskit called `clusterurl` which will process a string URL and return a generic "page route".

For example, the path "/foo/42/baz" would be replaced with "/foo/*/baz".

This solves issue https://github.com/grafana/app-o11y-kwl-endpoint/issues/283. Additionally, since both the Frontend Observability and Beyla/Instrumentations teams plan on using this functionality, it is best to put it in a place where it is easily reusable and discoverable. 

**Checklist**
- [ ] Tests updated

**Notes**
Thanks to @grcevski and the Beyla team for creating this code - I am just packaging it up for them :)


56quarters (changes requested):
Blocking because dskit is isn't meant to be the only way to share dependencies between projects, another repo would make more sense for this particular functionality.

eskirk:
@56quarters got it - that makes sense to me. I was told this would be a good place to put it so I am not super invested in it necessarily being in this repo. going to close this PR in that case 👍🏼

PR #448, opened 2023-12-08
Add SpanProfiler

kolesnikovae:
The PR introduces `spanprofiler`. Its purpose is to facilitate "request-scoped" profiles with minimal effort and impact on application performance. The use of OpenTracing and [OpenTelemetry](https://github.com/grafana/otel-profiling-go) SDK interfaces appears to be the simplest solution to this problem. For a detailed description, please refer to the [README](https://github.com/grafana/dskit/blob/feat/add-spanprofiler/spanprofiler/README.md) note.

We've successfully integrated and tested this feature in our internal deployments over the past two months with no reported issues.

**Checklist**
- [x] Tests updated

### DIFF
--- a/spanprofiler/spanprofiler.go
+++ b/spanprofiler/spanprofiler.go
@@ -1,0 +1,107 @@
+package spanprofiler
+
+import (
+	"context"
+	"runtime/pprof"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/uber/jaeger-client-go"
+)
+
+// StartSpanFromContext starts and returns a Span with `operationName`, using
+// any Span found within `ctx` as a ChildOfRef. If no such parent could be
+// found, StartSpanFromContext creates a root (parentless) Span.
+//
+// The call sets `operationName` as `span_name` pprof label, and the new span
+// identifier as `span_id` pprof label, if the trace is sampled.
+//
+// The second return value is a context.Context object built around the
+// returned Span.
+//
+// Example usage:
+//
+//	SomeFunction(ctx context.Context, ...) {
+//	    sp, ctx := opentracing.StartSpanFromContext(ctx, "SomeFunction")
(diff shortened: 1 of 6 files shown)


spanprofiler/tracer.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// samplingState is a pointer to a struct that has "localRootSpan" member,
+	// which we could probably use: that would allow omitting quite expensive
+	// parentSpanContextFromRef call. However, interpreting the pointer and
+	// the complex struct memory layout is more complicated and dangerous.
+	samplingState uintptr
+
+	// remote indicates that span context represents a remote parent
+	remote bool
+}
```
kolesnikovae, line 109: This is, perhaps, the most controversial part. If someone knows a way to avoid such unsafe access, I'll be happy to change it.

spanprofiler/spanprofiler.go
```diff
@@ -0,0 +1,107 @@
...
+	// labels when span finishes. The need of this wrapper is questioned:
+	// as we do not have the original context, we could leave the goroutine
+	// labels – normally, span is finished at the very end of the goroutine's
+	// lifetime, so no significant side effects should take place.
+	w := spanWrapper{
+		parentPprofCtx:  parentCtx,
+		currentPprofCtx: ctx,
+	}
+	w.Span = span.SetTag(profileIDTagKey, spanID)
+	return &w
```
kolesnikovae, line 85: It's worth noting that we probably could avoid wrapping of `jaeger.Span`, and preserve the original type

spanprofiler/tracer.go
cyriltovena: To me this seems fine, we could consider running our fork of open tracing while we transition off.

cyriltovena (approved):
LGTM

PR #449, opened 2023-12-11
`DoUntilQuorum`: provide causes when cancelling contexts, and don't log cancellation as an error

charleskorn:
**What this PR does**:

This PR improves the handling of context cancellations in `DoUntilQuorum`:

* if a cause is provided when cancelling the parent context passed to `DoUntilQuorum`, that cause will be returned by `DoUntilQuorum` if it terminates due to cancellation (if no cause is available, `context.Cause()` will fall back to `context.Canceled`, preserving the existing behaviour)
* `DoUntilQuorum` will provide a cause when cancelling contexts it controls
* `DoUntilQuorum` will no longer mark the trace span as failed if the call fails due to cancellation

Note that this introduces a breaking change to the signature of `DoUntilQuorumWithoutSuccessfulContextCancellation`: the function provided must now accept a `context.CancelCauseFunc` instead of a `context.CancelFunc`.

(none)

**Checklist**
- [x] Tests updated


dimitarvdimitrov (approved):
LGTM

PR #450, opened 2023-12-12
server: import godeltaprof/http/pprof

korniltsev:
**What this PR does**:
This PR imports `github.com/grafana/pyroscope-go/godeltaprof/http/pprof` in addition to `net/http/pprof`

When pyroscope is scraping cumulative profiles (`alloc_{space,objects}`, `mutex`, `block`) it needs to compute delta (difference / substraction from the previous profile). This is currently happening at [grafana-agent  ](https://github.com/grafana/agent/tree/main/component/pyroscope/scrape/internal/fastdelta)after scraping. This requires keeping state of previous profiles requiring much memory and also spends a bunch of cpu time (decompress,deserialize,substract,serialize,compress). 

[godeltaprof](https://github.com/grafana/pyroscope-go/tree/main/godeltaprof)  is an efficient delta profiler for memory, mutex, and block.  

It does compute this delta on the target side instead of the agent side before any compression/serialization, so a lot of work is avoided. Using godeltaprof removes the need to compute the delta on the grafna-agent side - this saves cpu and memory. Also the scraping godeltaprof itself is more efficient compared to scraping net/http/pprof because after delta is computed a lot of samples are rejected and the output profile is much smaller. So it is a win-win for grafana-agent <-> targets. Some [ benchmarks ](https://github.com/grafana/pyroscope-go/tree/main/godeltaprof#benchmarks).

This PR **does not** change `net/http/pprof` endpoints, just makes the server **ready** to scrape godeltaprof profiles. Additional changes to scraping configuration is required.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Changelog
 
-* [CHANGE] server: import godeltaprof/http/pprof
+* [CHANGE] server: import godeltaprof/http/pprof adding /debug/pprof/delta_{heap,mutex,block} endpoints to DefaultServeMux #450
 * [CHANGE] Move httpgrpc request & response utils up from httpgrpc/server to httpgrpc package #434, #435
 * [CHANGE] Updated the minimum required Go version to 1.20 and update Drone config #420
 * [CHANGE] Change `WaitRingStability` and `WaitInstanceState` methods signature to rely on `ReadRing` instead. #251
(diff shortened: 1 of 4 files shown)


cyriltovena (approved):
Lgtm

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] server: import godeltaprof/http/pprof
```
pstibrany, line 3: Can you please specify which endpoints this adds?

korniltsev: Done

pstibrany (approved):
Thanks.

PR #451, opened 2023-12-13
Fix issue where applying httpgrpc tracing middleware results in all HTTP requests having duplicate trace spans

charleskorn:
**What this PR does**:

This PR fixes the issue where applications that use the httpgrpc tracing middleware (through `middleware.InitHTTPGRPCMiddleware`) would have traces for all incoming HTTP calls duplicated.

For example, note the duplicate `HTTP POST - api_v1_push` spans in this trace from a Mimir distributor:

Screenshot 2023-12-13 at 12 50 20 pm

The fix changes the HTTP middleware to a gRPC interceptor that is then able to accurately filter incoming requests and apply the additional information only to httpgrpc requests. 

One downside of this is that we'll convert the request to a `http.Request` twice now: once in the interceptor, and again in the httpgrpc handler, but this seems like the neatest solution. 

Related: https://github.com/grafana/dskit/pull/353

(none)

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -200,3 +200,4 @@
 * [BUGFIX] Ring: use zone-aware logging when all zones are required for quorum. #403
 * [BUGFIX] Services: moduleService could drop stop signals to its underlying service. #409 #417
 * [BUGFIX] ring: don't mark trace spans as failed if `DoUntilQuorum` terminates due to cancellation. #449 
+* [BUGFIX] middleware: fix issue where applications that used the httpgrpc tracing middleware would generate duplicate spans for incoming HTTP requests. #451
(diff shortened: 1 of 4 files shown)


francoposa (approved):
LGTM!
If you were feeling particularly enterprising, you could figure out how to use the an OTEL tracer in the tests instead of this jaeger one.
The jaeger tracer is the one dskit used to use for opentracing before it was converted to OTEL. It stuck around in this test and maybe others but isn't actually used in the application anymore

charleskorn:
> If you were feeling particularly enterprising, you could figure out how to use the an OTEL tracer in the tests instead of this jaeger one.
> The jaeger tracer is the one dskit used to use for opentracing before it was converted to OTEL. It stuck around in this test and maybe others but isn't actually used in the application anymore

Sounds good, I'll take a look at this in a separate PR.

Issue #452, opened 2023-12-18
Refactor cache package and interfaces to something intentional


56quarters:
The current `cache` package is a mix of things that came from multiple packages with varying levels of abstraction in Thanos. We didn't really clean them up or decide what an appropriate cache interface would look like when removing the dependency on Thanos and copying code into dskit. Let's do that now.

I propose, in roughly the following order:
* `memcachedClient` and `redisClient` are renamed to `MemcachedClient` and `RedisClient` respectively and the constructors for these structs return `*MemcachedClient` and `*RedisClient` instead of the interface `RemoteCacheClient`. This will mean they are exported for consumers of dskit to use.
   * https://github.com/grafana/dskit/pull/453
* Remove legacy metrics emitted by Memcached clients
  * https://github.com/grafana/dskit/pull/461
* We standardize on a single "cache" interface and gradually migrate the various implementations to it. Example:
   ```go
   type Cache interface {
       Get(ctx context.Context, key string, opts ...Option) []byte
       GetMulti(ctx context.Context, keys []string, opts ...Option) map[string][]byte
       SetAsync(key string, value []byte, ttl time.Duration) error
       SetMultiAsync(data map[string][]byte, ttl time.Duration) error
       Delete(ctx context.Context, key string) error
       Stop()
       Name() string
    }
    ```
    * This involves collapsing the distinction between `MemcachedClient` and `MemcachedCache` and the corresponding Redis implementations. The `remoteCache` struct isn't doing much except for emitting a few metrics. We'll have to make sure any metrics emitted by these various things are preserved through the migration.
      * https://github.com/grafana/dskit/pull/465
      * https://github.com/grafana/dskit/pull/466
      * https://github.com/grafana/dskit/pull/471
      * https://github.com/grafana/dskit/pull/520

PR #453, opened 2023-12-19
Export Memcached and Redis client types

56quarters:
**What this PR does**:

Instead of returning the `RemoteCacheClient` interface from their constructors, return the actual type of the client. This lets callers decide how they want to use the clients and not be restricted to only the methods supported by the interface.

Part of #452

**Checklist**
- [NA] Tests updated


rlankfo (approved):
LGTM

PR #454, opened 2023-12-20
ring.DoBatch: check context cancellation

colega:
**What this PR does**:

This changes ring.DoBatch & ring.DoBatchWithOptions to check the context cancellation, as it might take a while to calculate the instances for all keys and sometimes it doesn't make sense to proceed with the callbacks.

None.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/ring_test.go
+++ b/ring/ring_test.go
@@ -199,7 +199,6 @@ func TestDoBatchZeroInstances(t *testing.T) {
 }
 
 func TestDoBatchWithOptionsContextCancellation(t *testing.T) {
-	// It takes ~4s to do 5M keys on my M3 laptop. So we're going to cancel the context after 400ms.
 	const (
 		numKeys      = 5e6
 		numInstances = 100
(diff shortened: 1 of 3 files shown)


ring/ring_test.go
```diff
@@ -198,6 +198,59 @@ func TestDoBatchZeroInstances(t *testing.T) {
 	require.Error(t, DoBatch(ctx, Write, &r, keys, callback, cleanup))
 }
 
+func TestDoBatchWithOptionsContextCancellation(t *testing.T) {
+	// It takes ~4s to do 5M keys on my M3 laptop. So we're going to cancel the context after 400ms.
```
duricanikolic, line 202: I would remove this comment

duricanikolic (approved):
LGTM
with a small nit

ring/ring_test.go
colega, line 202:
```suggestion
```
Removing the comment.

colega: Removed.

dimitarvdimitrov (approved):
LGTM, makes sense

PR #455, opened 2023-12-20
Make httpgrpc.Server return 4xx errors

duricanikolic:
**What this PR does**:
Currently, `httpgrpc`'s `Server.Handle()` returns an error only if a response received from the HTTP server contains a `5xx` HTTP status code. Responses with any other HTTP status code (e.g., `4xx`) are converted to non-erroneous `HTTPResponse` objects. As a consequence, the correlated `status_code` labels in the request duration metrics is `success`.

This PR allows `Server.Handle()` to return errors even in case of a `4xx` HTTP status code. Namely, the presence of a header with key `httpgrpc.Return4XXErrorsKey` and any value in a response received from HTTP server indicates to `Server.Handle()` to convert that response to an error. This way the correlated `status_code` labels in the request duration metrics will be `4xx`.

Part of https://github.com/grafana/mimir/issues/1830

**Checklist**
- [x] Tests updated


pracucci:
I discourage using an header to enable the behaviour. The feature flag approach is better. I would close this PR in favour of https://github.com/grafana/dskit/pull/457.

duricanikolic:
I am closing this PR in favour of #457

PR #457, opened 2023-12-21
Allow reporting HTTP 4xx codes in status_code label of request duration metrics

duricanikolic:
**What this PR does**:
Currently, `httpgrpc`'s `Server.Handle()` returns an error only if a response received from the HTTP server contains a `5xx` HTTP status code. Responses with any other HTTP status code (e.g., `4xx`) are converted to non-erroneous `HTTPResponse` objects. As a consequence, the correlated requests are reported with `status_code` labels `success` in the request duration metrics.

This PR allows `Server.Handle()` to return errors even in case of a `4xx` HTTP status code. Namely, when constructor `httpgrpc.NewSerer()` is invoked with the newly introduced `httpgrpc.WithReturn4XXErrors` server option, the created `httpgrpc.Server` will return errors with `4xx` HTTP status codes. This way the correlated requests will be denoted with `status_code` label `4xx` in the request duration metrics.

In order to guarantee backwards compatibility, reporting of HTTP `4xx` status codes as labels is disabled by default and could be enabled by setting the new `server.Config.ReportHTTP4XXCodesInInstrumentationLabel` to `true`.

Part of https://github.com/grafana/mimir/issues/1830

**Checklist**
- [x] Tests updated

### DIFF
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -32,15 +32,13 @@ var (
 	DoNotLogErrorHeaderKey = http.CanonicalHeaderKey("X-DoNotLogError")
 )
 
-type Options func(*Server)
+type Option func(*Server)
 
-var (
-	Return4XXErrorsOption Options = func(s *Server) {
-		s.return4XXErrors = true
-	}
-)
+func WithReturn4XXErrors(s *Server) {
+	s.return4XXErrors = true
+}
 
-func applyServerOptions(s *Server, opts ...Options) *Server {
+func applyServerOptions(s *Server, opts ...Option) *Server {
 	for _, opt := range opts {
 		opt(s)
 	}
@@ -55,7 +53,7 @@ type Server struct {
 }
 
 // NewServer makes a new Server.
-func NewServer(handler http.Handler, opts ...Options) *Server {
+func NewServer(handler http.Handler, opts ...Option) *Server {
 	return applyServerOptions(&Server{handler: handler}, opts...)
 }
 
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -32,17 +32,31 @@ var (
 	DoNotLogErrorHeaderKey = http.CanonicalHeaderKey("X-DoNotLogError")
 )
 
+type Options func(*Server)
+
+var (
+	Return4XXErrorsOption Options = func(s *Server) {
+		s.return4XXErrors = true
+	}
+)
+
+func applyServerOptions(s *Server, opts ...Options) *Server {
+	for _, opt := range opts {
+		opt(s)
+	}
+	return s
+}
+
 // Server implements HTTPServer.  HTTPServer is a generated interface that gRPC
 // servers must implement.
 type Server struct {
-	handler http.Handler
+	handler         http.Handler
+	return4XXErrors bool
 }
 
 // NewServer makes a new Server.
-func NewServer(handler http.Handler) *Server {
-	return &Server{
-		handler: handler,
-	}
+func NewServer(handler http.Handler, opts ...Options) *Server {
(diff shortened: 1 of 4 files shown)


httpgrpc/server/server.go
```diff
@@ -32,17 +32,31 @@ var (
 	DoNotLogErrorHeaderKey = http.CanonicalHeaderKey("X-DoNotLogError")
 )
 
+type Options func(*Server)
```
pracucci, line 35: This define a single option function:

```suggestion
type Option func(*Server)
```

httpgrpc/server/server.go
```diff
@@ -32,17 +32,31 @@ var (
...
+
+var (
+	Return4XXErrorsOption Options = func(s *Server) {
```
pracucci, line 38: [nit] A couple of nits:
- Looks a cumbersome way of defining a function
- It's commong to start such functions with `With...` and not having the `Option` suffix

So my suggestion is:

```suggestion
func WithReturnErrorOn4XX(s *Server) {
```

server/server.go
```diff
@@ -94,10 +94,11 @@ type Config struct {
 	HTTPTLSConfig TLSConfig `yaml:"http_tls_config"`
 	GRPCTLSConfig TLSConfig `yaml:"grpc_tls_config"`
 
-	RegisterInstrumentation               bool `yaml:"register_instrumentation"`
-	ReportGRPCCodesInInstrumentationLabel bool `yaml:"report_grpc_codes_in_instrumentation_label_enabled"`
-	ExcludeRequestInLog                   bool `yaml:"-"`
-	DisableRequestSuccessLog              bool `yaml:"-"`
+	RegisterInstrumentation                  bool `yaml:"register_instrumentation"`
+	ReportGRPCCodesInInstrumentationLabel    bool `yaml:"report_grpc_codes_in_instrumentation_label_enabled"`
+	ReportHTTP4XXCodesInInstrumentationLabel bool `yaml:"report_http_4xx_codes_in_instrumentation_label_enabled"`
```
pracucci, line 99: I don't think should be configurable via YAML / flag. This option should be configurable only in code (we want it enabled in Mimir everywhere).

pracucci (commented):
Thanks Yuri for working on this. I agree on the proposed change, but I think the feature flag shouldn't be exposed via CLI / YAML, but only configurable in code.

Once we successfully moved to it in Mimir, please inform other teams using it. Ideally we should get all teams doing this change (return error on 4xx) and then we'll remove the feature flag.

httpgrpc/server/server.go
duricanikolic: I will implement this recommendation. Thank you

server/server.go
```diff
@@ -550,9 +551,12 @@ func (s *Server) Run() error {
 		}
 	}()
 
-	// Setup gRPC server
-	// for HTTP over gRPC, ensure we don't double-count the middleware
-	httpgrpc.RegisterHTTPServer(s.GRPC, httpgrpc_server.NewServer(s.HTTP))
+	var serverOptions []httpgrpc_server.Option
+	if s.cfg.ReportHTTP4XXCodesInInstrumentationLabel {
+		serverOptions = []httpgrpc_server.Option{httpgrpc_server.WithReturn4XXErrors}
```
pracucci, line 556: [nit] Doesn't change the current logic, but makes the intention more clear (you don't want to replace the options, but add an option).

```suggestion
		serverOptions = append(serverOptions, httpgrpc_server.WithReturn4XXErrors)
```

pracucci (approved):
Thanks Yuri. I'm good with these changes. I left a final minor comment.

Note: as discussed offline, I suggest to merge this PR only once we've confirmed this solves the problem we have in Mimir.

dimitarvdimitrov (approved):
SGTM

PR #458, opened 2024-01-05
Change default memberlist.stream-timeout from `10s` to `2s`

56quarters:
**What this PR does**:

We have our own `packet_dial_timeout` and `packet_write_timeout` settings that we use for regular memberlist messages. However, special messages such as push-pull are managed by the memberlist library. A `10s` timeout is too large and causes joins to nonexistent IPs to take a really long time.

We've been using a `2s` timeout on Mimir in production for several weeks at this point with no adverse affects.

N/A

**Checklist**
- [ ] Tests updated


dimitarvdimitrov (approved):
LGTM. Thanks for carrying this change out

PR #464, opened 2024-01-10
Ring: Add `HasReplicationSetChangedWithoutStateOrAddr`

pr00se:
**What this PR does**:

The existing `ring.HasReplicationSetChangedWithoutState` function will detect a replication state change if a member was rebooted and rejoins the ring with a different IP address. In Mimir we want to ignore these changes as, in some cases, we only care about the distribution of the tokens in the ring changing.

This PR:
1. changes the sorting of the replication sets in `ring.hasReplicationSetChangedExcluding` from `ByAddr` to `ByID`, since we want to ignore the `Addr` field in some situations.
2. adds `ring.HasReplicationSetChangedWithoutStateOrAddr` to detect replication set changes ignoring the State and Addr fields.

Fixes N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/model.go
+++ b/ring/model.go
@@ -14,12 +14,12 @@ import (
 	"github.com/grafana/dskit/loser"
 )
 
-// ByAddr is a sortable list of InstanceDesc.
-type ByAddr []InstanceDesc
+// ByID is a sortable list of InstanceDesc.
+type ByID []InstanceDesc
 
-func (ts ByAddr) Len() int           { return len(ts) }
-func (ts ByAddr) Swap(i, j int)      { ts[i], ts[j] = ts[j], ts[i] }
-func (ts ByAddr) Less(i, j int) bool { return ts[i].Addr < ts[j].Addr }
+func (ts ByID) Len() int           { return len(ts) }
+func (ts ByID) Swap(i, j int)      { ts[i], ts[j] = ts[j], ts[i] }
+func (ts ByID) Less(i, j int) bool { return ts[i].Id < ts[j].Id }
 
 // ProtoDescFactory makes new Descs
 func ProtoDescFactory() proto.Message {
--- a/ring/model.go
+++ b/ring/model.go
@@ -14,6 +14,13 @@ import (
 	"github.com/grafana/dskit/loser"
 )
 
+// ByAddr is a sortable list of InstanceDesc.
+type ByAddr []InstanceDesc
+
+func (ts ByAddr) Len() int           { return len(ts) }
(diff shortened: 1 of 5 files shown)


ring/model.go
```diff
@@ -21,6 +21,13 @@ func (ts ByAddr) Len() int           { return len(ts) }
 func (ts ByAddr) Swap(i, j int)      { ts[i], ts[j] = ts[j], ts[i] }
 func (ts ByAddr) Less(i, j int) bool { return ts[i].Addr < ts[j].Addr }
 
+// ByID is a sortable list of InstanceDesc.
```
pr00se, line 24: I've left `ByAddr` because it is used [outside of the library](https://github.com/search?q=org%3Agrafana++ring.ByAddr&type=code)

ring/replication_set.go
```diff
@@ -468,6 +468,17 @@ func HasReplicationSetChangedWithoutState(before, after ReplicationSet) bool {
 	})
 }
 
+// Has HasReplicationSetChangedWithoutStateOrAddr returns false if two replications sets
```
56quarters, line 471: You don't have to do it as part of this PR but we seem to be accumulating quiet a few different methods that all do roughly the same thing. Maybe we could apply the option pattern here.

e.g.
```go
func HasReplicationSetChanged(before, after ReplicationSet, opt... RingOption) bool { /* ... */ }
```
Called like
```go
if ring.HasReplicationSetChanged(a, b, ring.ExcludeAddress(), ring.ExcludeState()) {
    /* ... */
}
```

56quarters (approved):
LGTM, one non-blocking comment

ring/replication_set.go
pr00se: I had the same thought, but it would mean breaking changes [not only for Mimir](https://github.com/search?q=org%3Agrafana%20HasReplicationSetChanged&type=code). Since dskit doesn't have a formal versioning process we'd need to coordinate changing over (or have two implementations for some time). We also don't have an easy way to signal that breaking change to any external users of dskit (are there any?).

We could add something like

```
func HasReplicationSetChangedFn(opt... ExcludeOption) func (before, after ReplicationSet) bool { /* ... */ }
```

and replace the [usage of `HasReplicationSetChangedWithoutXXXXX`](https://github.com/search?q=org%3Agrafana+HasReplicationSetChangedWithout&type=code), but I don't foresee us needing another flavor of this kind of function in the future (there aren't many other fields left to ignore). If we do end up adding a third kind, though, we should revisit this.

PR #465, opened 2024-01-10
Move requests/hits metrics to cache clients

56quarters:
**What this PR does**:

Move tracking of keys requested and hits from the `remoteCache` to each respective cache client (Memcached and Redis). This is part of a refactor that will eventually remove `remoteCache` since it doesn't provide enough value to justify itself.

Related #452

**Checklist**
- [ ] Tests updated

### DIFF
--- a/cache/client.go
+++ b/cache/client.go
@@ -38,6 +38,10 @@ type clientMetrics struct {
 	dataSize   *prometheus.HistogramVec
 }
 
+// newClientMetrics creates a new bundle of metrics about an instance of a cache client. Note
+// that there may be multiple cache clients at any given time so the prometheus.Registerer passed
+// to this method should include labels unique to this particular client (e.g. a name for each
+// different cache being used).
 func newClientMetrics(reg prometheus.Registerer) *clientMetrics {
 	cm := &clientMetrics{}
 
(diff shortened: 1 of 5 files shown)


cache/client.go
```diff
@@ -39,6 +41,14 @@ type clientMetrics struct {
 func newClientMetrics(reg prometheus.Registerer) *clientMetrics {
 	cm := &clientMetrics{}
 
+	cm.requests = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "requests_total",
+		Help: "Total number of items requests to cache.",
```
56quarters, line 46: Note that we don't set a constant label with the name of the cache anymore: Both `MemcachedClient` and `RedisClient` already wrap the registries they use in their respective constructors with the cache name label.

francoposa: maybe just a short comment or docstring about this - something like "registries passed in should be wrapped with an implementation-specific value for the `labelCacheBackend` label"

francoposa (approved):
LGTM - prefer docstring as mentioned but just a nit

cache/client.go
56quarters: Fixed in 739f646a4ebf59eff78a659aae90440c31732a59

PR #466, opened 2024-01-11
Remove errors from set methods on RemoteCacheClient interface

56quarters:
**What this PR does**:

This change does a few things:

* Removes the error from the set method on the `RemoteCacheClient` interface. The only error that can be returned was an error when the async queue used for `set` operations was full. All other errors were logged and a counter incremented. Even for this error, it was special-cased and `nil` was returned instead. Thus, the interface included an `error` return value that could never actually happen.
* Adds another method, `SetMultiAsync` to the interface for setting multiple key-value pairs at once. This was the only functionality that the `remoteCache` struct was providing beyond the cache clients it was wrapping. With this new method, there's no longer a reason to keep it around (will address in a follow-up PR).
* Uses relative TTLs for Memcached items. Memcached supports using a relative TTL (e.g. 300 for an item that expires after 5 minutes) or absolute TTL (a particular UNIX timestamp). We were converting `time.Duration` objects to an absolute TTL for no reason.

Part of #452

**Checklist**
- [ ] Tests updated


cache/memcached_client.go
```diff
@@ -305,12 +305,22 @@ func (c *MemcachedClient) Stop() {
 	c.client.Close()
 }
 
-func (c *MemcachedClient) SetAsync(key string, value []byte, ttl time.Duration) error {
-	return c.setAsync(key, value, ttl, func(key string, buf []byte, ttl time.Duration) error {
+func (c *MemcachedClient) SetMultiAsync(data map[string][]byte, ttl time.Duration) {
+	c.setMultiAsync(data, ttl, func(key string, buf []byte, ttl time.Duration) error {
 		return c.client.Set(&memcache.Item{
 			Key:        key,
-			Value:      value,
```
56quarters, line 312: The old method was actually using the wrong var here, setting `value` from the argument to `SetAsync` instead of `buf` being passed to the closure.

cache/cache.go
```diff
@@ -15,17 +15,22 @@ import (
...
-	// GetMulti fetches multiple keys at once from remoteCache. In case of error,
+	// GetMulti fetches multiple keys at once from a cache. In case of error,
 	// an empty map is returned and the error tracked/logged. One or more Option
 	// instances may be passed to modify the behavior of this GetMulti call.
 	GetMulti(ctx context.Context, keys []string, opts ...Option) map[string][]byte
 
-	// SetAsync enqueues an asynchronous operation to store a key into memcached.
+	// SetAsync enqueues an asynchronous operation to store a key into a cache.
 	// Returns an error in case it fails to enqueue the operation. In case the
 	// underlying async operation will fail, the error will be tracked/logged.
```
dimitarvdimitrov, line 25: this mentions the returned error still

cache/cache.go
```diff
@@ -15,17 +15,22 @@ import (
...
-	SetAsync(key string, value []byte, ttl time.Duration) error
+	SetAsync(key string, value []byte, ttl time.Duration)
 
-	// Delete deletes a key from memcached.
+	// SetMultiAsync enqueues asynchronous operations to store a keys and values
+	// into a cache. Returns an error in case it fails to enqueue the operation. In
```
dimitarvdimitrov, line 29: the comment still mentions the error

dimitarvdimitrov (approved):
SGTM. I don't mind the error returned from SetAsync, especially because adding it back later might break linters and it's only an `error`. But maybe we'll cross that bridge when we come to it. 

Found two outdated comments

56quarters:
> SGTM. I don't mind the error returned from SetAsync, especially because adding it back later might break linters and it's only an `error`. But maybe we'll cross that bridge when we come to it.

I'm OK removing the errors from `RemoteCacheClient` because we don't actually use that anywhere in Mimir except `RemoteIndexCache` in the store-gateway. Even there, the errors are logged instead of being returned (but again, they couldn't actually ever happen with the current implementations).  
 
> Found two outdated comments

Fixed, thank you.

dimitarvdimitrov:
> I'm OK removing the errors from `RemoteCacheClient` because we don't actually use that anywhere in Mimir except `RemoteIndexCache` in the store-gateway

hopefully one day dskit becomes a real library, not just a sub-package of mimir :D

PR #470, opened 2024-01-24
grpcclient: rename backoff retry middleware

narqo:
**What this PR does**:

The changes rename the `grpcclient.NewBackoffRetry` to `grpcclient.NewRateLimitRetrier`. The later is more obvious in term of signaliing to the (potencial) user, and a reader of the code, that the middleware is not a common-case retrier. That is, the middleware specifically retries RPC that were rate limited (see the discussion below https://github.com/grafana/dskit/pull/470#issuecomment-1910571457).

**Checklist**
- [ ] Tests updated

### DIFF
--- a/grpcclient/backoff_retry.go
+++ b/grpcclient/backoff_retry.go
@@ -2,6 +2,7 @@ package grpcclient
 
 import (
 	"context"
+	"errors"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov:
this middleware seems to be turned on when `back_off_on_rate_limits: true`

https://github.com/grafana/mimir/blob/1997cf598be8d94a2d2761915a78c940c82dbc93/vendor/github.com/grafana/dskit/grpcclient/grpcclient.go#L110-L112


Doesn't ResourceExhausted refer to cases where the a rate limit has been exceeded? That's what some of the comments/strings say in [mimir code](https://github.com/search?q=org%3Agrafana+ResourceExhausted+repo%3Agrafana%2Fmimir&type=code) and [tempo code](https://github.com/search?q=org%3Agrafana+ResourceExhausted+repo%3Agrafana%2Ftempo&type=code)

narqo:
> this middleware seems to be turned on when `back_off_on_rate_limits: true`

That's what I missed, indeed. How about we rename the middleware to something more obvious, maybe `NewRateLimitRetrier()`?

For the context, of why I'm looking at that, from https://github.com/grafana/mimir/issues/6609, gRPC server returns the `codes.ResourceExhausted` code if the payload exceeds the configured limit. I looked around for examples of where the code used across the code, and bumped into this one.

dimitarvdimitrov:
> How about we rename the middleware to something more obvious, maybe `NewRateLimitRetrier()`?

makes sense to me. At a [glance](https://github.com/search?q=NewBackoffRetry&type=code&p=4) I couldn't find it being widely used outside of the linked dskit usage, so we might be ok to do this breaking change.

dimitarvdimitrov (approved):
LGTM, thank you

PR #471, opened 2024-01-29
Cache: Make the purpose of remoteCache adapter explicit

56quarters:
**What this PR does**:

Strip down and export the `remoteCache` struct to only act as an adapter between the two cache interfaces in dskit: `Cache` and `RemoteCacheClient`.

This change also removes `MemcachedCache` and `RedisCache` since they were only wrappers around the `remoteCache` struct and not referenced outside of dskit itself.

Part of #452

**Checklist**
- [ ] Tests updated


bboreham (approved):
lgtm

PR #474, opened 2024-01-31
Add partition ring basic support

pracucci:
**What this PR does**:

In Mimir we're working on a experimental next-generation architecture where sharding is based on the concept of "partitions". Data is sharded between partitions using an hash ring. However, the `ring.Ring` is not suited for the partitions, so we've created a new ring called `PartitionRing`.

In this PR I'm introducing `PartitionRing` data structure and some basic utilities (e.g. `PartitionRingWatcher` to watch for changes). I've done my best to comment the structs and code, to make it as much as possible self-explanatory.

This is the just first PR of a series, where more features will be added to the `PartitionRing` (e.g. shuffle sharding support, ring state web page, ...).

N/A

**Checklist**
- [x] Tests updated


ring/partition_ring_model.go
```diff
@@ -0,0 +1,336 @@
...
+				change.Partitions[pid] = thisPart
+			}
+		}
+	}
+
+	// Now let's handle owners. Owners don't have tokens, which simplifies things compared to "normal" ring.
+	for id, otherOwner := range other.Owners {
+		thisOwner := m.Owners[id]
+
+		if otherOwner.UpdatedTimestamp > thisOwner.UpdatedTimestamp || (otherOwner.UpdatedTimestamp == thisOwner.UpdatedTimestamp && otherOwner.State == OwnerDeleted && thisOwner.State != OwnerDeleted) {
```
pstibrany, line 244: For owners we accept "Deleted" state even if timestamp hasn't changed, but for partitions we don't. 

Two comments:
* I'd suggest to be consistent for both partitions and owners -- either we keep the check or remove it. To be honest, I don't remember why this case exists in the original Ring code. I think it would be fine to remove this special check.
* If we decide to keep this check, let's add test for it.

ring/partition_ring_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+	w.ringMx.Unlock()
+
+	// Update metrics.
+	for state, count := range desc.countPartitionsByState() {
+		w.numPartitionsGaugeVec.WithLabelValues(state.CleanName()).Set(float64(count))
+	}
+}
+
+// GetRing returns the most updated snapshot of the PartitionRing. The returned instance
+// is immutable and will not updated if new changes will be done to the ring.
```
pstibrany, line 95:
```suggestion
// is immutable and will not be updated if new changes are done to the ring.
```

pstibrany (approved):
Nice work!

ring/partition_ring_desc.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+
+enum PartitionState {
+  PartitionUnknown = 0;
+
+  // Active partition in read-write mode.
+  PartitionActive = 1;
+
+  // Inactive partition in read-only mode. This partition will be deleted after a grace period.
+  PartitionInactive = 2;
```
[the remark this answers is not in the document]
pracucci: Addressed in https://github.com/grafana/dskit/pull/474/commits/5c0a105b387c6bd1a06f27cbd4b59ea2da7c7e21.

ring/partition_ring_model.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (m *PartitionRingDesc) mergeWithTime(mergeable memberlist.Mergeable, localCAS bool, now time.Time) (memberlist.Mergeable, error) {
+	if mergeable == nil {
+		return nil, nil
+	}
+
+	other, ok := mergeable.(*PartitionRingDesc)
+	if !ok {
+		// This method only deals with non-nil rings.
```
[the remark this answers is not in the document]
pracucci: Removed both here and `Ring` because we actually handle `nil` in both cases: https://github.com/grafana/dskit/pull/474/commits/ca4eaff6260631a99a0ca06a09200c27faa94072

ring/partition_ring_model.go
pracucci: > To be honest, I don't remember why this case exists in the original Ring code.

Maybe to cover the case it was removed in less than 1s from the last update? The timestamp has second precision.

pstibrany: I think the reason is that original ring code did not update timestamp when setting entry state to LEFT (tombstone). See `Merge` method in https://github.com/cortexproject/cortex/pull/1727/files#diff-e773119be1c2446745f5b5eeb2d7b436e99cb21c02e1fdfc40e42d041d541ad8 and how `localCAS` is handled.

Your point about 1s precision is good reason why we may want to keep it.

pstibrany (approved):
Thank you for addressing my feedback!

PR #476, opened 2024-02-01
Add shuffle sharding support to partitions ring

pracucci:
**What this PR does**:

In #474 I've introduced the experimental `PartitionRing`. In this I'm adding support for shuffle sharding. The shuffle sharding implementation follows what we've already done in `Ring`. The main differences are:

- Partitions ring has no concept of zones (cause owners will be deployed cross zones, but that's another story)
- Partitions can either be in ACTIVE or INACTIVE state. This change the logic around which partitions are eligible to be picked by the shuffle sharding (see comments to `ShardSize()` and `ShardSizeWithLookback()` functions).
- Implementing cache for `PartitionRing` subrings is easier. `PartitionRing` is immutable by design, so we can just cache subrings inside it.

Note to reviewers:
- I haven't unit tested `partitionRingShuffleShardCache` because it's tested at an upper level by `TestPartitionRing_ShuffleShardWithLookback_Caching` and `TestPartitionRing_ShuffleShardWithLookback_CachingConcurrency`. I'm open to feedback here. I can also unit tests for `partitionRingShuffleShardCache` if desired.
- Most of the new tests added in `partition_ring_test.go` are an adaptation of existing shuffle sharding tests from `ring_test.go`. These tests have been very useful, in particular `TestPartitionRing_ShuffleShardWithLookback_CorrectnessWithFuzzy` which allowed me to spot a bug in the first `shuffleShard()` implementation, which was detected by other (simpler) unit tests.

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_ring.go
+++ b/ring/partition_ring.go
@@ -111,7 +111,8 @@ func (r *PartitionRing) ShuffleShard(identifier string, size int) (*PartitionRin
 		return cached, nil
 	}
 
-	subring, err := r.shuffleShard(identifier, size, 0, time.Now())
+	// No need to pass the time if there's no lookback.
+	subring, err := r.shuffleShard(identifier, size, 0, time.Time{})
 	if err != nil {
 		return nil, err
 	}
@@ -149,7 +150,10 @@ func (r *PartitionRing) shuffleShard(identifier string, size int, lookbackPeriod
 		size = len(r.desc.Partitions)
 	}
 
-	lookbackUntil := now.Add(-lookbackPeriod).Unix()
+	var lookbackUntil int64
+	if lookbackPeriod > 0 {
+		lookbackUntil = now.Add(-lookbackPeriod).Unix()
+	}
 
 	// Initialise the random generator used to select instances in the ring.
 	// There are no zones
--- a/ring/partition_ring.go
+++ b/ring/partition_ring.go
@@ -199,28 +199,10 @@ func (r *PartitionRing) shuffleShard(identifier string, size int, lookbackPeriod
 			}
 
 			var (
-				shouldInclude bool
-				shouldExtend  bool
+				withinLookbackPeriod = lookbackPeriod > 0 && p.GetStateTimestamp() >= lookbackUntil
+				shouldExtend         = withinLookbackPeriod
+				shouldInclude        = p.IsActive() || withinLookbackPeriod
 			)
-
-			// Check whether the partition should be included in the result and/or the result set extended.
-			if lookbackPeriod > 0 {
-				if p.IsActive() {
-					// The partition is active. We should include it. The result set is extended only
-					// if the partition became active within the lookback window.
-					shouldInclude = true
-					shouldExtend = p.GetStateTimestamp() >= lookbackUntil
-				} else {
-					// The partition is inactive. We should include it (and then extend the result set)
-					// only if became inactive within the lookback window.
-					shouldInclude = p.GetStateTimestamp() >= lookbackUntil
-					shouldExtend = shouldInclude
-				}
-			} else {
-				// No lookback was provided. We only include active partitions and don't lookback.
-				shouldInclude = p.IsActive()
-				shouldExtend = false
-			}
 
 			// Either include or exclude the found partition.
 			if shouldInclude {
--- a/ring/partition_ring.go
+++ b/ring/partition_ring.go
@@ -195,8 +195,8 @@ func (r *PartitionRing) shuffleShard(identifier string, size int, lookbackPeriod
 			}
 
 			var (
-				shouldInclude = false
-				shouldExtend  = false
+				shouldInclude bool
+				shouldExtend  bool
 			)
 
 			// Check whether the partition should be included in the result and/or the result set extended.
--- a/ring/partition_ring.go
+++ b/ring/partition_ring.go
@@ -122,7 +122,7 @@ func (r *PartitionRing) ShuffleShard(identifier string, size int) (*PartitionRin
 }
 
 // ShuffleShardWithLookback is like ShuffleShard() but the returned subring includes all instances
-// that have been part of the identifier's shard since "now - lookbackPeriod".
+// that have been part of the identifier's shard in [now - lookbackPeriod, now] time window.
 //
 // This function can return a mix of ACTIVE and INACTIVE partitions. INACTIVE partitions are only
 // included if they were part of the identifier's shard within the lookbackPeriod.
--- a/ring/partition_ring.go
+++ b/ring/partition_ring.go
@@ -3,6 +3,12 @@ package ring
 import (
 	"bytes"
 	"fmt"
+	"math/rand"
+	"time"
+
+	"golang.org/x/exp/slices"
+
+	shardUtil "github.com/grafana/dskit/ring/shard"
 )
 
 var ErrNoActivePartitionFound = fmt.Errorf("no active partition found")
@@ -27,6 +33,9 @@ type PartitionRing struct {
 
 	// ownersByPartition is a map where the key is the partition ID and the value is a list of owner IDs.
 	ownersByPartition map[int32][]string
+
+	// shuffleShardCache is used to cache subrings generated with shuffle sharding.
+	shuffleShardCache *partitionRingShuffleShardCache
 }
 
 func NewPartitionRing(desc PartitionRingDesc) *PartitionRing {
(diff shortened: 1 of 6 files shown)


ring/partition_ring.go
```diff
@@ -75,11 +85,227 @@ func (r *PartitionRing) ActivePartitionForKey(key uint32) (int32, error) {
...
+//     subring with no more then 1 difference.
+//
+//   - Shuffling: probabilistically, for a large enough cluster each identifier gets a different
+//     set of instances, with a reduced number of overlapping instances between two identifiers.
+func (r *PartitionRing) ShuffleShard(identifier string, size int) (*PartitionRing, error) {
+	if cached := r.shuffleShardCache.getSubring(identifier, size); cached != nil {
+		return cached, nil
+	}
+
+	subring, err := r.shuffleShard(identifier, size, 0, time.Now())
```
pstibrany, line 114: nit: do we need "time.Now()" when not using lookback?

ring/partition_ring.go
```diff
@@ -75,11 +85,227 @@ func (r *PartitionRing) ActivePartitionForKey(key uint32) (int32, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	r.shuffleShardCache.setSubring(identifier, size, subring)
+	return subring, nil
+}
+
+// ShuffleShardWithLookback is like ShuffleShard() but the returned subring includes all instances
+// that have been part of the identifier's shard since "now - lookbackPeriod".
```
pstibrany, line 124:
```suggestion
// that have been part of the identifier's shard in [now - lookbackPeriod, now] time window.
```

ring/partition_ring.go
```diff
@@ -75,11 +85,227 @@ func (r *PartitionRing) ActivePartitionForKey(key uint32) (int32, error) {
...
+			}
+
+			// We can stop searching for other partitions only if this partition was included
+			// and no extension was requested, which means it's the "stop partition" for this cycle.
+			if shouldInclude && !shouldExtend {
+				found = true
+			}
+		}
+
+		// If we iterated over all tokens, and no new partition has been found, we can stop looking for more partitions.
```
pstibrany, line 240: This means that shard is smaller than requested size. Should we return error in that case?

ring/partition_ring.go
```diff
@@ -75,11 +85,227 @@ func (r *PartitionRing) ActivePartitionForKey(key uint32) (int32, error) {
...
+					// The partition is inactive. We should include it (and then extend the result set)
+					// only if became inactive within the lookback window.
+					shouldInclude = p.GetStateTimestamp() >= lookbackUntil
+					shouldExtend = shouldInclude
+				}
+			} else {
+				// No lookback was provided. We only include active partitions and don't lookback.
+				shouldInclude = p.IsActive()
+				shouldExtend = false
+			}
```
pstibrany, line 219: I find this slightly easier to read (don't think comments are needed):

```suggestion
			shouldExtend := lookbackPeriod > 0 && p.GetStateTimestamp() >= lookbackUntil
			shouldInclude := p.IsActive() || shouldExtend
```

pstibrany (commented):
Early comments, going to check tests.

ring/partition_ring_test.go
```diff
@@ -143,6 +149,775 @@ func BenchmarkPartitionRing_ActivePartitionForKey(b *testing.B) {
...
+				2: {userToken(userID, "", 0) + 1},
+			},
+			timeline: []event{
+				{what: add, partitionID: 0, partitionDesc: generatePartitionWithInfo(PartitionActive, outsideLookback)},
+				{what: add, partitionID: 1, partitionDesc: generatePartitionWithInfo(PartitionActive, outsideLookback)},
+				{what: add, partitionID: 2, partitionDesc: generatePartitionWithInfo(PartitionActive, outsideLookback)},
+				{what: test, shardSize: 2, expected: []int32{2, 0}},
+				{what: update, partitionID: 2, partitionDesc: generatePartitionWithInfo(PartitionInactive, withinLookback)},
+				{what: test, shardSize: 2, expected: []int32{2, 1, 0}},
+				// Partition 2 still inactive, but now falls outside the lookback period, there's no need to look back to partition 1
```
pstibrany, line 506:
```suggestion
				// Partition 2 still inactive, but now falls outside the lookback period, there's no need to include to partition 1
```

ring/partition_ring_test.go
```diff
@@ -143,6 +149,775 @@ func BenchmarkPartitionRing_ActivePartitionForKey(b *testing.B) {
...
+				{what: update, partitionID: 2, partitionDesc: generatePartitionWithInfo(PartitionInactive, outsideLookback)},
+				{what: test, shardSize: 2, expected: []int32{1, 0}},
+				// Partition 2 becomes active again
+				{what: update, partitionID: 2, partitionDesc: generatePartitionWithInfo(PartitionActive, withinLookback)},
+				{what: test, shardSize: 2, expected: []int32{2, 1, 0}},
+				// Partition 2 still active, but now falls outside the lookback period, there's no need to look back to partition 1
```
pstibrany, line 512:
```suggestion
				// Partition 2 still active, but now falls outside the lookback period, there's no need to include to partition 1
```

(comment applies in few more places in subsequent tests too)

ring/partition_ring_test.go
```diff
@@ -143,6 +149,775 @@ func BenchmarkPartitionRing_ActivePartitionForKey(b *testing.B) {
...
+			currTime := time.Now()
+			subring, err := ring.shuffleShard(userID, shardSize, 0, currTime)
+			require.NoError(t, err)
+
+			history := map[time.Time][]int32{
+				currTime: subring.PartitionIDs(),
+			}
+
+			// Advance the time. The simulation assumes the initial ring contains partitions added
+			// since more than the lookback period.
```
pstibrany, line 627:
```suggestion
			// outside of the lookback period.
```

ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (r *partitionRingShuffleShardCache) setSubringWithLookback(identifier string, size int, lookbackPeriod time.Duration, now time.Time, subring *PartitionRing) {
+	if subring == nil {
+		return
+	}
+
+	var (
+		lookbackWindowStart                   = now.Add(-lookbackPeriod).Unix()
+		validForLookbackWindowsStartingBefore = int64(math.MaxInt64)
+	)
```
pstibrany, line 53: nit: Don't you find this nicer and easier to read?

```suggestion
	lookbackWindowStart := now.Add(-lookbackPeriod).Unix()
	validForLookbackWindowsStartingBefore := int64(math.MaxInt64)
```

pstibrany (approved):
Amazing work, great suite of tests!

ring/partition_ring.go
pracucci: Agree. Done in https://github.com/grafana/dskit/pull/476/commits/19af036f63181298290f54ce0d30d1544acedc63

ring/partition_ring.go
pracucci: I don't think so. The shard can be smaller than the requested size (e.g. size could be > than number of partitions in the ring). Am I missing anything?

ring/partition_ring_test.go
pracucci: Done in https://github.com/grafana/dskit/pull/476/commits/42f25898ee5aee2bdf4c0a5bd72b2b4eda900131

ring/partition_ring_test.go
pracucci: Done in https://github.com/grafana/dskit/pull/476/commits/42f25898ee5aee2bdf4c0a5bd72b2b4eda900131

ring/partition_ring_test.go
pracucci: Done https://github.com/grafana/dskit/pull/476/commits/055b60599dcc71d8d3f7b7464ae0f87f4828806c

ring/partition_ring.go
pracucci: Do you feel strong about this? I find your edit quite cryptic. This logic is complex and I don't think the goal should be compress it as much as possible, but make each case intention as clearest as possible.

ring/partitions_ring_shuffle_shard_cache.go
pracucci: I honestly do find it easier to read when `=` are aligned :)

ring/partition_ring.go
pstibrany: No, you're right.

ring/partition_ring.go
pracucci: Discussed offline. Agreed on introducing a 3rd variable to make it easier to understand the rationale: https://github.com/grafana/dskit/pull/476/commits/7552bd65f97657ce2b96e65c37490c0a53484eaa

pstibrany (approved):
Thanks for addressing my feedback.

PR #477, opened 2024-02-02
Add more methods to memcachedClient

ananasovich:
Add Increment, Touch, CompareAndSwap, and FlushAll methods to the underlying Memcache client.

Include associated wrappers for the exported MemcachedClient object.

**What this PR does**:
Increase the number of MemcachedClient methods accessible to consumers.

This pull request addresses an issue in a private repository, so it may not be directly linkable. 

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@
 * [FEATURE] Server: Add support for `GrpcInflightMethodLimiter` -- limiting gRPC requests before reading full request into the memory. This can be used to implement global or method-specific inflight limits for gRPC methods. #377 #392
 * [FEATURE] Server: Add `-grpc.server.num-workers` flag that configures the `grpc.NumStreamWorkers()` option. This can be used to start a fixed base amount of workers to process gRPC requests and avoid stack allocation for each call. #400
 * [FEATURE] Add `PartitionRing`. The partitions ring is hash ring to shard data between partitions. #474 #476 #478
+* [FEATURE] Add methods `Increment`, `FlushAll`, `CompareAndSwap`, `Touch` to `cache.MemcachedClient` #477
 * [ENHANCEMENT] Add ability to log all source hosts from http header instead of only the first one. #444
 * [ENHANCEMENT] Add configuration to customize backoff for the gRPC clients.
 * [ENHANCEMENT] Use `SecretReader` interface to fetch secrets when configuring TLS. #274
--- a/cache/memcached_client.go
+++ b/cache/memcached_client.go
@@ -44,7 +44,11 @@ type memcachedClientBackend interface {
 	GetMulti(keys []string, opts ...memcache.Option) (map[string]*memcache.Item, error)
 	Set(item *memcache.Item) error
 	Delete(key string) error
+	Increment(key string, delta uint64) (newValue uint64, err error)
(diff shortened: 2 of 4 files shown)


rlankfo:
@56quarters these add the additional methods we need to the memcached client. This should allow us to get off the fork we're using that ejects the underlying `*memcache.Client` client so we can use the wrapper directly.

56quarters (commented):
I appreciate the effort to make the new methods look like the existing code, but:

* I don't think it's necessary to split these methods between `baseClient` and `MemcachedClient` since they're pretty Memcached specific. It's fine to just put them right in `MemcachedClient`
* Using `c.asyncQueue.submit()` isn't really necessary since in all cases it looks like we're waiting for the operation to complete before returning. Fine to just call `c.client.Method()` directly.  If you were adding a method for bulk `Add()` or `Replace()` it'd make sense but for the methods added, just calling them and returning an error is fine.

Thank you for adding corresponding metrics for all the new operations.

ananasovich:
@56quarters
> * I don't think it's necessary to split these methods between `baseClient` and `MemcachedClient` since they're pretty Memcached specific. It's fine to just put them right in `MemcachedClient`
I am not sure what it means in practice. Does it mean to change the signatures of the methods to be the same? Wouldn't it be beneficial in this case to maintain some level of dependency isolation?

cache/memcached_client.go
```diff
@@ -564,3 +597,114 @@ func (c *MemcachedClient) resolveAddrs() error {
...
+		)
+		c.trackError(opTouch, err)
+	} else {
+		c.metrics.duration.WithLabelValues(opTouch).Observe(time.Since(start).Seconds())
+	}
+
+	return err
+}
+
+func (c *MemcachedClient) increment(ctx context.Context, key string, delta uint64, f func(ctx context.Context, key string, delta uint64) (uint64, error)) (uint64, error) {
```
56quarters, line 651: Thank you but now there's not really any reason to keep these methods around. Instead, they could be combined with their corresponding public methods.  WDYT?

ananasovich: I prefer atomized functions, but in this case, the function stays small, so this is okay.

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@
 * [FEATURE] Server: Add support for `GrpcInflightMethodLimiter` -- limiting gRPC requests before reading full request into the memory. This can be used to implement global or method-specific inflight limits for gRPC methods. #377 #392
 * [FEATURE] Server: Add `-grpc.server.num-workers` flag that configures the `grpc.NumStreamWorkers()` option. This can be used to start a fixed base amount of workers to process gRPC requests and avoid stack allocation for each call. #400
 * [FEATURE] Add `PartitionRing`. The partitions ring is hash ring to shard data between partitions. #474
+* [FEATURE] Add methods `Increment`, `FlushAll`, `CompareAndSwap`, `Touch` to `cache.MemcachedClient`
```
rlankfo, line 79:
```suggestion
* [FEATURE] Add methods `Increment`, `FlushAll`, `CompareAndSwap`, `Touch` to `cache.MemcachedClient` #477
```

rlankfo (approved):
LGTM, thanks @ananasovich!

PR #479, opened 2024-02-06
Add PartitionInstanceRing

pracucci:
**What this PR does**:

This PR is another piece of the work we're doing on the experimental partitions ring support. In this PR I'm introducing `PartitionInstanceRing`, which is a component holding a partitions ring and instances ring, and then provides functions to lookup the intersection of the two.

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_instance_ring.go
+++ b/ring/partition_instance_ring.go
@@ -50,11 +50,11 @@ func (r *PartitionInstanceRing) GetReplicationSetsForOperation(op Operation) ([]
 	result := make([]ReplicationSet, 0, len(partitionsRingDesc.Partitions))
 
 	for partitionID := range partitionsRingDesc.Partitions {
-		ownerIDs := partitionsRing.UnsafePartitionOwnerIDs(partitionID)
+		ownerIDs := partitionsRing.PartitionOwnerIDs(partitionID)
 		instances := make([]InstanceDesc, 0, len(ownerIDs))
 
 		for _, instanceID := range ownerIDs {
-			instance, err := r.instancesRing.UnsafeGetInstance(instanceID)
+			instance, err := r.instancesRing.GetInstance(instanceID)
 			if err != nil {
 				// If an instance doesn't exist in the instances ring we don't return an error
 				// but lookup for other instances of the partition.
--- a/ring/partition_instance_ring.go
+++ b/ring/partition_instance_ring.go
@@ -86,8 +86,9 @@ func (r *PartitionInstanceRing) GetReplicationSetsForOperation(op Operation) ([]
 		}
 
 		result = append(result, ReplicationSet{
-			Instances: instances,
-			MaxErrors: len(instances) - 1, // We need response from at least 1 owner.
+			Instances:            instances,
+			MaxUnavailableZones:  len(instances) - 1, // We need response from at least 1 owner.
+			ZoneAwarenessEnabled: true,               // Partitions has no concept of zone, but we enable it in order to support ring's requests minimization feature.
 		})
 	}
 	return result, nil
--- a/ring/partition_instance_ring.go
+++ b/ring/partition_instance_ring.go
@@ -12,12 +12,9 @@ type PartitionRingReader interface {
 // PartitionInstanceRing holds a partitions ring and a instances ring, and provide functions
 // to look up the intersection of the two (e.g. healthy instances by partition).
 type PartitionInstanceRing struct {
-	// partitionsRingReader and partitionsRing are mutually exclusive.
 	partitionsRingReader PartitionRingReader
-	partitionsRing       *PartitionRing
-
-	instancesRing    *Ring
-	heartbeatTimeout time.Duration
+	instancesRing        *Ring
+	heartbeatTimeout     time.Duration
 }
 
 func NewPartitionInstanceRing(partitionsRingWatcher PartitionRingReader, instancesRing *Ring, heartbeatTimeout time.Duration) *PartitionInstanceRing {
@@ -28,20 +25,8 @@ func NewPartitionInstanceRing(partitionsRingWatcher PartitionRingReader, instanc
 	}
 }
 
-func newStaticPartitionInstanceRing(partitionsRing *PartitionRing, instancesRing *Ring, heartbeatTimeout time.Duration) *PartitionInstanceRing {
-	return &PartitionInstanceRing{
-		partitionsRing:   partitionsRing,
-		instancesRing:    instancesRing,
-		heartbeatTimeout: heartbeatTimeout,
-	}
-}
-
 func (r *PartitionInstanceRing) PartitionRing() *PartitionRing {
-	if r.partitionsRingReader != nil {
-		return r.partitionsRingReader.PartitionRing()
-	}
-
-	return r.partitionsRing
+	return r.partitionsRingReader.PartitionRing()
 }
 
 func (r *PartitionInstanceRing) InstanceRing() *Ring {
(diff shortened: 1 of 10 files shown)


ring/partition_ring_watcher.go
```diff
@@ -91,9 +91,9 @@ func (w *PartitionRingWatcher) updatePartitionRing(desc *PartitionRingDesc) {
 	}
 }
 
-// GetRing returns the most updated snapshot of the PartitionRing. The returned instance
+// PartitionRing returns the most updated snapshot of the PartitionRing. The returned instance
 // is immutable and will not be updated if new changes are done to the ring.
-func (w *PartitionRingWatcher) GetRing() *PartitionRing {
+func (w *PartitionRingWatcher) PartitionRing() *PartitionRing {
```
pracucci, line 96: Note to reviewers: renamed to adhere to the `PartitionRingReader` interface.

ring/partition_instance_ring.go
```diff
@@ -0,0 +1,123 @@
...
+type PartitionRingReader interface {
+	PartitionRing() *PartitionRing
+}
+
+// PartitionInstanceRing holds a partitions ring and a instances ring, and provide functions
+// to look up the intersection of the two (e.g. healthy instances by partition).
+type PartitionInstanceRing struct {
+	// partitionsRingReader and partitionsRing are mutually exclusive.
+	partitionsRingReader PartitionRingReader
+	partitionsRing       *PartitionRing
```
pstibrany, line 17: Sinnce `PartitionRingReader` is an interface, instead of having `partitionsRing` field, we could provide another implementation of `PartitionRingReader` that returns given partition ring. Then there's no need for this mutual-exclusivity.

ring/partition_instance_ring.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+func (r *PartitionInstanceRing) PartitionRing() *PartitionRing {
+	if r.partitionsRingReader != nil {
+		return r.partitionsRingReader.PartitionRing()
+	}
+
+	return r.partitionsRing
+}
```
pstibrany, line 45: If you accept my above suggestion about "static" implementation of `PartitionRingReader`, we can get simplify this.

ring/partition_instance_ring.go
```diff
@@ -0,0 +1,123 @@
+package ring
+
+import (
+	"fmt"
+	"time"
+)
+
+type PartitionRingReader interface {
+	PartitionRing() *PartitionRing
```
pstibrany, line 9: Let's mention that we never expect to get `nil` from this method.

ring/partition_instance_ring.go
```diff
@@ -0,0 +1,123 @@
...
+
+	now := time.Now()
+	result := make([]ReplicationSet, 0, len(partitionsRingDesc.Partitions))
+
+	for partitionID := range partitionsRingDesc.Partitions {
+		ownerIDs := partitionsRing.UnsafePartitionOwnerIDs(partitionID)
+		instances := make([]InstanceDesc, 0, len(ownerIDs))
+
+		for _, instanceID := range ownerIDs {
+			instance, err := r.instancesRing.UnsafeGetInstance(instanceID)
```
pstibrany, line 70: "Unsafe" prefix seems very weird. There's nothing "unsafe" about this. Even the instance can be modified, only the tokens slice is shared. I think comment on `GetInstance` is enough, I'd suggest dropping `Unsafe` prefix.

ring/partition_instance_ring.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if len(partitionsRingDesc.Partitions) == 0 {
+		return nil, ErrEmptyRing
+	}
+
+	now := time.Now()
+	result := make([]ReplicationSet, 0, len(partitionsRingDesc.Partitions))
+
+	for partitionID := range partitionsRingDesc.Partitions {
+		ownerIDs := partitionsRing.UnsafePartitionOwnerIDs(partitionID)
```
pstibrany, line 66: Instead of `Unsafe` prefix, perhaps should we call the *other* method to indicate that it returns `Copy` (ie. we would have `PartitionOwnerIDs` and `PartitionOwnerIDsCopy`)? I get what you're trying to say with it, but it seems weird that we're using "unsafe" methods.

ring/partition_ring_watcher.go
pstibrany: That's fine, but it can return `nil` value if watcher is not running yet. We should avoid that.

ring/ring.go
```diff
@@ -866,16 +866,32 @@ func mergeTokenGroups(groupsByName map[string][]uint32) []uint32 {
 	return merged
 }
 
-// GetInstanceState returns the current state of an instance or an error if the
-// instance does not exist in the ring.
-func (r *Ring) GetInstanceState(instanceID string) (InstanceState, error) {
+// UnsafeGetInstance return the InstanceDesc for the given instanceID or an error
+// if the instance doesn't exist in the ring. The returned InstanceDesc is NOT a
+// deep copy, so the caller should never modify it.
+func (r *Ring) UnsafeGetInstance(instanceID string) (InstanceDesc, error) {
```
pstibrany, line 872: There's nothing unsafe about *calling* this method. Returned InstanceDesc is a shallow-copy, and all of its attributes except tokens can also be freely changed. I disagree with using "Unsafe" prefix here (we don't use this anywhere in dskit), and think that enhanced comment is enough.

Perhaps naming return value as `(doNotModify InstanceDesc, _ error)` would also indicate what you want to say here.

pstibrany (commented):
code lgtm, my biggest objection is about introducing "unsafe" prefix on methods (not used anywhere in dskit). Methods themselves are **not** unsafe, what you're trying to communicate is that returned values should not be modified. I think comment and perhaps naming of return values is enough.

ring/partition_instance_ring.go
pracucci: Good idea. Done in https://github.com/grafana/dskit/pull/479/commits/b457572c7c8cbf10b4e7cce98ae3180920821c91

ring/partition_instance_ring.go
pracucci: Sure. Done in https://github.com/grafana/dskit/pull/479/commits/b457572c7c8cbf10b4e7cce98ae3180920821c91.

ring/partition_instance_ring.go
pracucci: Agree. Done in https://github.com/grafana/dskit/pull/479/commits/48ffab9b07a12f8f692da0fc5d47404c510712ff

ring/ring.go
pracucci: > and all of its attributes except tokens can also be freely changed

Reason why I called it unsafe was exactly because of the tokens.

ring/partition_instance_ring.go
pracucci: Addressed in https://github.com/grafana/dskit/pull/479/commits/e33bb87dc4c8eb13e91aeff9d22a8f35a4246950

ring/partition_instance_ring.go
pracucci: Addressed in https://github.com/grafana/dskit/pull/479/commits/e33bb87dc4c8eb13e91aeff9d22a8f35a4246950

ring/ring.go
pracucci: Addressed in https://github.com/grafana/dskit/pull/479/commits/e33bb87dc4c8eb13e91aeff9d22a8f35a4246950

ring/partition_ring_watcher.go
pracucci: Good job spotting it! I haven't realised it, and was unintentional. Fixed in https://github.com/grafana/dskit/pull/479/commits/6fcc58211fbb3e843f990eab2de285586e13bb43

pstibrany (approved):
Thanks for addressing my feedback.

PR #481, opened 2024-02-07
Add pending state to partitions (experimental)

pracucci:
**What this PR does**:

We're working on a experimental support to partitions hash ring. We re-iterated on the design and we're introducing a new partition state: `PENDING`. The idea is that a PENDING partition is not actively used in the write and read path yet, but allows to instances to pre-register as owners of that partition, that will be soon become ACTIVE.

While working on this PR I've found out that `TestPartitionRing_ShuffleShardWithLookback_CorrectnessWithFuzzy` had a bug: the ID of the partition to change when processing random events was not the right now. I've fixed it in this PR, given that test has also been touched by this work.

N/A

**Checklist**
- [x] Tests updated


ring/partition_ring_desc.proto
```diff
@@ -32,16 +32,22 @@ message PartitionDesc {
 enum PartitionState {
   PartitionUnknown = 0;
 
+  // Pending partition is a partition that is about to be switched to ACTIVE. This state is used
+  // to let owners to attach to the partition and get ready to handle the partition.
+  //
+  // When a partition is in this state, it must not be used for writing or reading.
+  PartitionPending = 1;
```
pracucci, line 39: Note to reviewers: this will break (again) existing rings, but given the feature it's under development I would prefer to sort states by their expected sequence of events.

pstibrany: :+1: Makes sense.

pstibrany (approved):

PR #484, opened 2024-02-14
Add DoBatch() support for partitions ring

pracucci:
**What this PR does**:

As part of the experimental partitions ring support, in this PR I'm adding `DoBatch()` and `DoBatchWithOptions()` support for the partitions ring.

`DoBatch()` and `DoBatchWithOptions()` currently takes in input `ReadRing` interface, which is quite a big interface with several methods that don't make sense for partitions ring. However, `DoBatch()` and `DoBatchWithOptions()` only require 3 methods which I've defined into a new interface `DoBatchRing`.

Then I've created `ActivePartitionBatchRing`, which is a `PartitionRing` wrapper implementing `DoBatchRing`. Why didn't I added `DoBatchRing` methods directly to `PartitionRing`? The reason is that `DoBatchRing` methods behaviour needs to be contextualised based on what you really want to "batch" (in Mimir will be used on the write path where we need to write to ACTIVE partitions only). To keep a cleaner design, I've created `ActivePartitionBatchRing` which is the specific `DoBatchRing` implementation which lookup keys only among ACTIVE partitions.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/ring/batch.go
+++ b/ring/batch.go
@@ -49,9 +49,22 @@ func isHTTPStatus4xx(err error) bool {
 	return code/100 == 4
 }
 
+// DoBatchRing defines the interface required by a ring implementation to use DoBatch() and DoBatchWithOptions().
+type DoBatchRing interface {
+	// Get returns a ReplicationSet containing the instances to which the input key should be sharded to
+	// for the input Operation.
+	Get(key uint32, op Operation, bufInstances []InstanceDesc, bufStrings1, bufStrings2 []string) (ReplicationSet, error)
+
+	// ReplicationFactor returns the number of instances each key is expected to be sharded to.
+	ReplicationFactor() int
+
+	// InstancesCount returns the number of instances in the ring eligible to get any key sharded to.
+	InstancesCount() int
+}
+
 // DoBatch is a deprecated version of DoBatchWithOptions where grpc errors containing status codes 4xx are treated as client errors.
 // Deprecated. Use DoBatchWithOptions instead.
-func DoBatch(ctx context.Context, op Operation, r ReadRing, keys []uint32, callback func(InstanceDesc, []int) error, cleanup func()) error {
+func DoBatch(ctx context.Context, op Operation, r DoBatchRing, keys []uint32, callback func(InstanceDesc, []int) error, cleanup func()) error {
 	return DoBatchWithOptions(ctx, op, r, keys, callback, DoBatchOptions{
 		Cleanup:       cleanup,
 		IsClientError: isHTTPStatus4xx,
@@ -94,14 +107,14 @@ func (o *DoBatchOptions) replaceZeroValuesWithDefaults() {
 // See comments on DoBatchOptions for available options for this call.
 //
 // Not implemented as a method on Ring, so we can test separately.
-func DoBatchWithOptions(ctx context.Context, op Operation, r ReadRing, keys []uint32, callback func(InstanceDesc, []int) error, o DoBatchOptions) error {
+func DoBatchWithOptions(ctx context.Context, op Operation, r DoBatchRing, keys []uint32, callback func(InstanceDesc, []int) error, o DoBatchOptions) error {
 	o.replaceZeroValuesWithDefaults()
 
 	if r.InstancesCount() <= 0 {
 		o.Cleanup()
 		return fmt.Errorf("DoBatch: InstancesCount <= 0")
 	}
-	expectedTrackers := len(keys) * (r.ReplicationFactor() + 1) / r.InstancesCount()
+	expectedTrackersPerInstance := len(keys) * (r.ReplicationFactor() + 1) / r.InstancesCount()
 	itemTrackers := make([]itemTracker, len(keys))
 	instances := make(map[string]instance, r.InstancesCount())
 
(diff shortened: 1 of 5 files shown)


ring/partition_ring.go
```diff
@@ -340,3 +352,60 @@ func (r *PartitionRing) String() string {
...
+
+	bufInstances[0] = InstanceDesc{
+		Addr:      partitionIDString,
+		Timestamp: 0,
+		State:     ACTIVE,
+		Id:        partitionIDString,
+	}
+
+	return ReplicationSet{
+		Instances:            bufInstances,
```
pstibrany, line 406: Since we return instances in shared buffer, `ReplicaSet` is only valid until next call to `Get` with the same buffer. This is very tricky behaviour, and should be documented. (`DoBatch` is doing the right thing -- it uses `ReplicaSet.Instances` and then doesn't use same `ReplicaSet`.)

If we moved buffer into `ActivePartitionBatchRing` as suggested above, this would still be tricky, but at least it would be clearer which buffer is "the same".

ring/batch.go
```diff
@@ -49,9 +49,22 @@ func isHTTPStatus4xx(err error) bool {
 	return code/100 == 4
 }
 
+// DoBatchRing defines the interface required by a ring implementation to use DoBatch() and DoBatchWithOptions().
+type DoBatchRing interface {
```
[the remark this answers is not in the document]
pracucci: I've considered it but I personally don't like the adapter approach very much. I personally don't see any issue passing generic buffers even if they're not used. I also think it's safer to pass buffers from outside, instead of hardcoding them internally in a adapter because the caller has control whether buffers are safe to be reused or not.

ring/partition_ring.go
pracucci: I think we should document that buffer is used, but for the rest of the comment I see it differently:

- Given buffer is explicitly passed from outside, it's in caller control whether to reuse it or not (but I agree should be documented)
- The adapter approach, instead, mask it and it's outside of caller control whether a buffer should be reused or not

pstibrany (approved):
Thanks for considering my feedback.

ring/batch.go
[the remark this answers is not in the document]
dimitarvdimitrov: I agree that reusing the buffers without coordinating with the caller looks dangerous. Letting the caller control the buffers (DoBatch or DoBatchWithOptions) seems slightly more clear

ring/partition_ring.go
```diff
@@ -340,3 +352,60 @@ func (r *PartitionRing) String() string {
 
 	return fmt.Sprintf("PartitionRing{ownersCount: %d, partitionsCount: %d, partitions: {%s}}", len(r.desc.Owners), len(r.desc.Partitions), buf.String())
 }
+
+// ActivePartitionBatchRing wraps PartitionRing and implements DoBatchRing to lookup ACTIVE partitions.
+type ActivePartitionBatchRing struct {
+	ring *PartitionRing
+}
+
+func NewActivePartitionBatchRing(ring *PartitionRing) *ActivePartitionBatchRing {
```
dimitarvdimitrov, line 361: shouldn't the `Operation` parameter to `Get` determine which instances (partitions in our case) get included in the replication set? Having a ring implementation for a specific partition state seems different from how the classic ring and DoBatch work.

We do have PartitionState and not InstanceStates, but maybe we can implement a mapping for the purposes of an Operation.

dimitarvdimitrov (commented):
i'm wondering if we can do away without `ActivePartitionBatchRing`

ring/batch.go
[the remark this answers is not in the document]
pstibrany: > I agree that reusing the buffers without coordinating with the caller looks dangerous. Letting the caller control the buffers (DoBatch or DoBatchWithOptions) seems slightly more clear

We could document that repeated calls to `Get` invalidate previous results, and if that's not desired, new instance of "adapter" is required. I think that still looks nicer than passing 3 buffers via method in the interface. But it's mostly a matter of preference.

ring/partition_ring.go
dimitarvdimitrov: another option is to extend `Operation` to also support these methods, which are similar to the existing methods for `InstanceState`


```go
func (op Operation) IsPartitionInStateHealthy(s PartitionState) bool {}
func (op Operation) ShouldExtendReplicaSetOnPartitionState(s PartitionState) bool {}
```

ring/batch.go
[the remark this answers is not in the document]
dimitarvdimitrov: or ask clients to copy the buffers.


I also prefer the clearer interface and am less concerned about the breaking change because it's trivial to fix and there aren't a lot of clients in reality (2-3 projects?)

Another option is to put the 3 buffers in a struct so that it's less verbose and you only pass one thing, not 3.

```go
type DoBatchBuffers struct {
  Instances []InstanceDesc
  Addresses []string
  Zones []string
}
```

i'll leave the consensus to you two because I don't have a strong preference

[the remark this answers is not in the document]
pracucci: > I also prefer the clearer interface and am less concerned about the breaking change

The adapter solution requires to wrap the `Ring` with the adapter in every place we call `DoBatch...()`. If it's about clean design I don't think that's clean too.

Anyway, I think we reached the "a matter of tastes" point and I don't think this discussion is foundamentally changing the partitions ring implementation. I'm going to keep it as is, which is mimicking `Ring.Get()` without requiring any wrapper or interface change.

ring/partition_ring.go
pracucci: Explained buffers in the interface doc: https://github.com/grafana/dskit/pull/484/commits/1f95b9219b6c4e59da02c45c7420233b65c85595

ring/partition_ring.go
pracucci: > shouldn't the Operation parameter to Get determine which instances (partitions in our case) get included in the replication set?

Yes. I think we can solve the problem with generics (I've a local copy based on generics that looks working). Let me merge this PR as is and then open a follow up PR, so that I'm not blocked on other PRs based on this.

dimitarvdimitrov: i'm not even sure we need generics - why not just reuse `Operation`? it's only a `uint32`

PR #485, opened 2024-02-14
Display corrupted partitions in ring UI page

pracucci:
**What this PR does**:

The partitions ring model could end up in a "corrupted" state if there are owners owning a partition which doesn't exist. This is a state which is not easy to detect and I would like to display it in the ring UI page.

Note: we rely on the ring UI page quite a lot and its correctness is quite important. For this reason I've added some assertions on the displayed content. However, asserting on HTML is quite difficult using the golang stdlib, so I've introduced `goquery` (which is a library I'm used to because of a personal project). If there are strong opinions about not importing such library I can revert the tests.

N/A

**Checklist**
- [ ] Tests updated


pstibrany (commented):
Wouldn't it be easier to check the model passed to page template?

For rendered page, we could look for basic strings like "Active", "Inactive", "Corrupted" and `bgcolor="FFDEDE".

ring/partition_ring_http_test.go
```diff
@@ -47,4 +55,30 @@ func TestPartitionRingPageHandler(t *testing.T) {
...
+
+	rows := doc.Find("table tbody tr")
+	require.Equal(t, 3, rows.Length())
+
+	row := doc.Find("table tbody tr:nth-child(1)")
+	require.Equal(t, 1, row.Length())
+	assert.Equal(t, "1", strings.TrimSpace(row.Find("td:nth-child(1)").Text()))
+	assert.Equal(t, "Active", strings.TrimSpace(row.Find("td:nth-child(2)").Text()))
+	assert.Contains(t, strings.TrimSpace(row.Find("td:nth-child(4)").Text()), "ingester-zone-a-0")
+	assert.Contains(t, strings.TrimSpace(row.Find("td:nth-child(4)").Text()), "ingester-zone-b-0")
```
dimitarvdimitrov, line 70: do we need `TrimSpace` if we're already using `Contains`? Is it possible to use an exact match - in that case it's less likely we mess up the query and match `Contains` on a big chunk of the HTML page

dimitarvdimitrov (approved):
LGTM

ring/partition_ring_http.go
```diff
@@ -38,26 +38,54 @@ func (h *PartitionRingPageHandler) ServeHTTP(w http.ResponseWriter, req *http.Re
...
-		return partitionsData[i].ID < partitionsData[j].ID
+	// Look for owners of non-existing partitions. We want to provide visibility for such case
+	// and we report the partition in corrupted state.
+	for ownerID, owner := range ringDesc.Owners {
+		partition, exists := partitionsByID[owner.OwnedPartition]
+
+		if !exists {
+			partition = partitionPageData{
+				ID:             owner.OwnedPartition,
+				State:          "Corrupted",
```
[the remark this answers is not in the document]
pracucci: It doesn't make any difference to me. Renamed in https://github.com/grafana/dskit/pull/485/commits/bfb382839f3646bce7b1a8de2d858f9cdbe23405

ring/partition_ring_http_test.go
pracucci: I've removed goquery dependency at all and used regexps so we don't introduce a big dependecy just because of this test.

pracucci:
> Wouldn't it be easier to check the model passed to page template?
> For rendered page, we could look for basic strings like "Active", "Inactive", "Corrupted" and `bgcolor="FFDEDE".
> Alternatively, we could compare final rendered HTML with expected string.

I've done the assertion using regexps. All in all, it's not that difficult to read. WDYT?

dimitarvdimitrov (approved):
LGTM, I don't think it's now harder to read than with the goquery syntax (maybe it's even easier)

PR #486, opened 2024-02-14
Add concurrency.ForEachJobMergeResults() utility

pracucci:
**What this PR does**:

In this PR I propose to add the general purpose utility function `concurrency.ForEachJobMergeResults()` that concurrently runs jobs like `ForEachJob()` but also collect and merge results.

N/A

**Checklist**
- [x] Tests updated


ring/replication_set.go
```diff
@@ -394,6 +395,80 @@ type instanceResult[T any] struct {
...
+		fnCtx, cancelFnCtx = context.WithCancel(ctx)
+	)
+
+	defer func() {
+		close(done)
+
+		// Ensure the fnCtx is canceled if it wasn't already.
+		if !errors.Is(fnCtx.Err(), context.Canceled) {
+			cancelFnCtx()
+		}
```
dimitarvdimitrov, line 421: cancel funcs they're usually idempotent, not sure we need the check here


```go
// A CancelFunc tells an operation to abandon its work.
// A CancelFunc does not wait for the work to stop.
// A CancelFunc may be called by multiple goroutines simultaneously.
// After the first call, subsequent calls to a CancelFunc do nothing.
type CancelFunc func()
```

ring/replication_set.go
```diff
@@ -394,6 +395,80 @@ type instanceResult[T any] struct {
...
+		close(resultsC)
+	}()
+
+	result := make([]T, 0, len(sets)) // expect at least one result from each replication set.
+	for {
+		select {
+		case err := <-errorC:
+			// An error occurred in one of the callback functions. We can cancel fnCtx to early
+			// stop the remaining running callback functions.
+			cancelFnCtx()
```
dimitarvdimitrov, line 459: is this also done in the `defer` virtually immediately after this call?

ring/replication_set.go
```diff
@@ -394,6 +395,80 @@ type instanceResult[T any] struct {
 	instance *InstanceDesc
 }
 
+// DoForReplicationSets runs fn for each ring.ReplicationSet in parallel, and returns a list with all merged results.
+// If any invocation returns an error, DoForReplicationSets returns without waiting for the other invocations to finish.
+func DoForReplicationSets[T any](ctx context.Context, sets []ReplicationSet, fn func(context.Context, ReplicationSet) ([]T, error)) ([]T, error) {
```
dimitarvdimitrov, line 400: i realized this does very similar job to `concurrency.ForEachJob`. Should we move `DoForReplicationSets` there and make the interface more similar to `ForEachJob(context.Context, int)`? At which point we can even reuse `ForEachJob` and remove most of the code of this function

dimitarvdimitrov (commented):
we may be able to do the same with less code - see my comment on DoForReplicationSets

ring/replication_set.go
pracucci: Closing because outdated.

ring/replication_set.go
pracucci: Closing because outdated.

ring/replication_set.go
pracucci: Very true. I've introduced a new (simpler) `ForEachJobMergeResults()` in https://github.com/grafana/dskit/pull/486/commits/7ebe017a47919e60e73fda0c5d832f1a3f58732e. WDYT?

concurrency/runner.go
```diff
@@ -113,3 +117,33 @@ func ForEachJob(ctx context.Context, jobs int, concurrency int, jobFunc func(ctx
 	// Wait until done (or context has canceled).
 	return g.Wait()
 }
+
+// ForEachJobMergeResults is like ForEachJob but expects jobFunc to return a slice of results which are then
+// merged with results from all jobs. This function returns no results if an error occurred running any jobFunc.
+func ForEachJobMergeResults[T any, R any](ctx context.Context, jobs []T, concurrency int, jobFunc func(ctx context.Context, job T) ([]R, error)) ([]R, error) {
```
pstibrany, line 123: nit: shall we rename `T` to `J` (as in Job)?

pstibrany (approved):
looks like useful addition.

concurrency/runner.go
pracucci: Ah yes.

pracucci: Done in https://github.com/grafana/dskit/pull/486/commits/a880baf577deb957457f7711b720a53db60c2090.

dimitarvdimitrov (approved):
LGTM

concurrency/runner.go
```diff
@@ -83,11 +83,15 @@ func CreateJobsFromStrings(values []string) []interface{} {
...
 // ForEachJob runs the provided jobFunc for each job index in [0, jobs) up to concurrency concurrent workers.
+// If the concurrency value is <= 0 all jobs will be executed in parallel.
 // The execution breaks on first error encountered.
 func ForEachJob(ctx context.Context, jobs int, concurrency int, jobFunc func(ctx context.Context, idx int) error) error {
 	if jobs == 0 {
 		return nil
 	}
+	if concurrency <= 0 {
+		concurrency = jobs
+	}
```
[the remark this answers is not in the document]
pracucci: True, done in https://github.com/grafana/dskit/pull/486/commits/2a1741c4587e6850ba90f68d2dd86c8d5f670467

PR #487, opened 2024-02-15
Change exemplar label from "traceID" to "trace_id"

bboreham:
**What this PR does**:

Change exemplar label from "traceID" to "trace_id".

This is consistent with [the OpenTelemetry standard](https://github.com/open-telemetry/opentelemetry-specification/blob/89aa01348139/specification/metrics/data-model.md#exemplars), and [an example in OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/138654493130/specification/OpenMetrics.md#exemplars-1).

~Sorry it is a breaking change.~  I believe it was me that picked the wrong string in the first place.
EDIT: not a breaking change until we remove the old one.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/instrument/instrument.go
+++ b/instrument/instrument.go
@@ -75,7 +75,7 @@ func ObserveWithExemplar(ctx context.Context, histogram prometheus.Observer, sec
 	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
 		histogram.(prometheus.ExemplarObserver).ObserveWithExemplar(
 			seconds,
-			prometheus.Labels{"trace_id": traceID},
+			prometheus.Labels{"trace_id": traceID, "traceID": traceID},
 		)
 		return
 	}
--- a/instrument/instrument.go
+++ b/instrument/instrument.go
@@ -75,7 +75,7 @@ func ObserveWithExemplar(ctx context.Context, histogram prometheus.Observer, sec
 	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
 		histogram.(prometheus.ExemplarObserver).ObserveWithExemplar(
 			seconds,
-			prometheus.Labels{"traceID": traceID},
+			prometheus.Labels{"trace_id": traceID},
 		)
 		return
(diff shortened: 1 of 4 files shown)


pracucci (approved):
LGTM

instrument/instrument.go
```diff
@@ -75,7 +75,7 @@ func ObserveWithExemplar(ctx context.Context, histogram prometheus.Observer, sec
 	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
 		histogram.(prometheus.ExemplarObserver).ObserveWithExemplar(
 			seconds,
-			prometheus.Labels{"traceID": traceID},
+			prometheus.Labels{"trace_id": traceID},
```
56quarters, line 78: I guess this isn't a huge deal since exemplars are short-lived and only in memory but should we send both for a period of time?

bboreham: That does seem like an attractive idea.

middleware/logging.go
```diff
@@ -58,7 +58,7 @@ func (l Log) logWithRequest(r *http.Request) log.Logger {
 	localLog := l.Log
 	traceID, ok := tracing.ExtractTraceID(r.Context())
 	if ok {
-		localLog = log.With(localLog, "traceID", traceID)
+		localLog = log.With(localLog, "trace_id", traceID)
```
bboreham, line 61: This is logging-only; is it better to be consistent with OpenTelemetry here, or consistent with what we used to do?

56quarters: Internally, our Loki datasources are configured to support either. I think making it consistent is better but I don't feel strongly about it.

56quarters (approved):

middleware/logging.go
bboreham: TIL, thanks.  Since this is a public open source project we shouldn't be particularly guided by what happens inside Grafana Labs, but I'll take comfort that anyone can apply the same setting.
image

PR #489, opened 2024-02-19
Fix MaxUnavailableZones computed by PartitionInstanceRing.GetReplicationSetsForOperation()

pracucci:
**What this PR does**:

`PartitionInstanceRing.GetReplicationSetsForOperation()` currently returns `ReplicationSet` with `MaxUnavailableZones` computed as `len(instances) - 1`. This works fine if all instances are in different zones, but we have some (or all) instances in the same zone it will not work as expected because `MaxUnavailableZones` will be >= the actual number of zones among `instances`. For example, when that's the case, the `zoneAwareResultTracker` will not issue any request at all because if `MaxUnavailableZones` is >= the actual number of zones... well the quorum is reached even without sending any request.

In this PR I propose to change `MaxUnavailableZones` to `len(zones) - 1`. Naturally this has an impact on quorum if some (or all instances) run in the same zone, but we expect not to be the case when using the partitions ring. However, if that will happen (e.g. local dev env, a demo, ...) after this change the system will keep working as expected, instead of weird bugs being surfaced (e.g. no requests sent to instances at all).

Benchmark:

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
                                                        │ before.txt  │             after.txt              │
                                                        │   sec/op    │   sec/op     vs base               │
PartitionInstanceRing_GetReplicationSetsForOperation-12   37.99µ ± 1%   41.91µ ± 1%  +10.33% (p=0.002 n=6)

                                                        │  before.txt  │            after.txt            │
                                                        │     B/op     │     B/op      vs base           │
PartitionInstanceRing_GetReplicationSetsForOperation-12   32.88Ki ± 0%   32.88Ki ± 0%  ~ (p=1.000 n=6) ¹
¹ all samples are equal

                                                        │ before.txt │           after.txt           │
                                                        │ allocs/op  │ allocs/op   vs base           │
PartitionInstanceRing_GetReplicationSetsForOperation-12   101.0 ± 0%   101.0 ± 0%  ~ (p=1.000 n=6) ¹
¹ all samples are equal
```

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_instance_ring_test.go
+++ b/ring/partition_instance_ring_test.go
@@ -55,19 +55,19 @@ func TestPartitionInstanceRing_GetReplicationSetsForOperation(t *testing.T) {
 					2: {State: PartitionInactive},
 				},
 				Owners: map[string]OwnerDesc{
-					"instance-1-a": {OwnedPartition: 1},
-					"instance-2-a": {OwnedPartition: 2},
-					"instance-2-b": {OwnedPartition: 2},
+					"instance-zone-a-1": {OwnedPartition: 1},
+					"instance-zone-a-2": {OwnedPartition: 2},
+					"instance-zone-b-2": {OwnedPartition: 2},
 				},
 			},
 			instancesRing: &Desc{Ingesters: map[string]InstanceDesc{
-				"instance-1-a": {Id: "instance-1-a", State: ACTIVE, Zone: "a", Timestamp: now.Unix()},
-				"instance-2-a": {Id: "instance-2-a", State: ACTIVE, Zone: "a", Timestamp: now.Unix()},
-				"instance-2-b": {Id: "instance-2-b", State: ACTIVE, Zone: "b", Timestamp: now.Unix()},
+				"instance-zone-a-1": {Id: "instance-zone-a-1", State: ACTIVE, Zone: "a", Timestamp: now.Unix()},
+				"instance-zone-a-2": {Id: "instance-zone-a-2", State: ACTIVE, Zone: "a", Timestamp: now.Unix()},
+				"instance-zone-b-2": {Id: "instance-zone-b-2", State: ACTIVE, Zone: "b", Timestamp: now.Unix()},
 			}},
 			expectedSets: []comparableReplicationSet{
-				{instances: []string{"instance-1-a"}, maxUnavailableZones: 0},
-				{instances: []string{"instance-2-a", "instance-2-b"}, maxUnavailableZones: 1},
+				{instances: []string{"instance-zone-a-1"}, maxUnavailableZones: 0},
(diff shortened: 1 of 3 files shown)


dimitarvdimitrov:
> we have some (or all) instances in the same zone

With the implementation in this PR would we end up querying both instances in the same zone or only one of them?

dimitarvdimitrov (commented):
LGTM, but `TestDefaultResultTracker_StartMinimumRequests_NoFailingRequests` is failing and it _sounds_ somewhat related. Have you checked it?

pracucci:
> TestDefaultResultTracker_StartMinimumRequests_NoFailingRequests

I don't think it's related because the modified code doesn't touch the code run by that test at all, but I'm not in a hurry on this PR and tomorrow I will look at why `TestDefaultResultTracker_StartMinimumRequests_NoFailingRequests` is flaky.

pracucci:
> > TestDefaultResultTracker_StartMinimumRequests_NoFailingRequests
> 
> I don't think it's related because the modified code doesn't touch the code run by that test at all, but I'm not in a hurry on this PR and tomorrow I will look at why `TestDefaultResultTracker_StartMinimumRequests_NoFailingRequests` is flaky.

Yes, it's unrelated: https://github.com/grafana/dskit/pull/491

ring/partition_instance_ring_test.go
```diff
@@ -55,11 +61,14 @@ func TestPartitionInstanceRing_GetReplicationSetsForOperation(t *testing.T) {
 				},
 			},
 			instancesRing: &Desc{Ingesters: map[string]InstanceDesc{
-				"instance-1-a": {Id: "instance-1-a", State: ACTIVE, Timestamp: now.Unix()},
-				"instance-2-a": {Id: "instance-2-a", State: ACTIVE, Timestamp: now.Unix()},
-				"instance-2-b": {Id: "instance-2-b", State: ACTIVE, Timestamp: now.Unix()},
+				"instance-1-a": {Id: "instance-1-a", State: ACTIVE, Zone: "a", Timestamp: now.Unix()},
+				"instance-2-a": {Id: "instance-2-a", State: ACTIVE, Zone: "a", Timestamp: now.Unix()},
+				"instance-2-b": {Id: "instance-2-b", State: ACTIVE, Zone: "b", Timestamp: now.Unix()},
```
dimitarvdimitrov, line 66: the reverse `partition-zone` naming is confusing

dimitarvdimitrov (approved):
LGTM

ring/partition_instance_ring_test.go
pracucci: Agree. Fixed in https://github.com/grafana/dskit/pull/489/commits/a2609c759a5226d700fb74d8b806bc955f719464

PR #490, opened 2024-02-19
Implement per tenant duration

simonswine:
This will allow a per tenant metric request duration collection, based on an injected callback.

I have an example of this (in the private PR https://github.com/grafana/backend-enterprise/pull/5933, GL emplyoees only)

### DIFF
--- a/middleware/instrument.go
+++ b/middleware/instrument.go
@@ -32,11 +32,11 @@ type RouteMatcher interface {
 // PerTenantCallback is a function that returns a tenant ID for a given request. When the returned tenant ID is not empty, it is used to label the duration histogram.
 type PerTenantCallback func(context.Context) string
 
-func (f PerTenantCallback) shouldInstrument(ctx context.Context) (tenantID string, ok bool) {
+func (f PerTenantCallback) shouldInstrument(ctx context.Context) (string, bool) {
 	if f == nil {
 		return "", false
 	}
-	tenantID = f(ctx)
+	tenantID := f(ctx)
 	if tenantID == "" {
 		return "", false
 	}
--- a/middleware/instrument.go
+++ b/middleware/instrument.go
@@ -5,6 +5,7 @@
 package middleware
 
 import (
+	"context"
 	"io"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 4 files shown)


middleware/instrument.go
```diff
@@ -28,13 +29,29 @@ type RouteMatcher interface {
 	Match(*http.Request, *mux.RouteMatch) bool
 }
 
+// PerTenantCallback is a function that returns a tenant ID for a given request. When the returned tenant ID is not empty, it is used to label the duration histogram.
+type PerTenantCallback func(context.Context) string
```
bryanhuhta, line 33: nit: `PerTenantCallback` doesn't seem accurate as a name since the function is supposed to return a tenant id. Maybe `TenantIDCallback`?

middleware/instrument.go
```diff
@@ -28,13 +29,29 @@ type RouteMatcher interface {
...
+func (f PerTenantCallback) shouldInstrument(ctx context.Context) (tenantID string, ok bool) {
+	if f == nil {
+		return "", false
+	}
+	tenantID = f(ctx)
+	if tenantID == "" {
+		return "", false
+	}
+	return tenantID, true
+}
```
bryanhuhta, line 44: nit: It doesn't seem like named return values are common in dskit and it doesn't appear that the serve to increase readability/ergonomics of this function. Maybe they can be dropped here?

```suggestion
func (f PerTenantCallback) shouldInstrument(ctx context.Context) (string, bool) {
	if f == nil {
		return "", false
	}
	tenantID := f(ctx)
	if tenantID == "" {
		return "", false
	}
	return tenantID, true
}
```

bryanhuhta (approved):
LGTM 👍

middleware/instrument.go
bryanhuhta: Actually coming back to this, I'm indifferent. I think in the broader context this naming is fine.

server/metrics.go
```diff
@@ -46,6 +47,15 @@ func NewServerMetrics(cfg Config) *Metrics {
 			NativeHistogramMaxBucketNumber:  100,
 			NativeHistogramMinResetDuration: time.Hour,
 		}, []string{"method", "route", "status_code", "ws"}),
+		PerTenantRequestDuration: reg.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace:                       cfg.MetricsNamespace,
+			Name:                            "per_tenant_request_duration_seconds",
+			Help:                            "Time (in seconds) spent serving HTTP requests for a particular tenant.",
+			Buckets:                         instrument.DefBuckets,
```
krajorama, line 54: Hi, do you really need the classic histogram series? I know that currently exemplars don't work without the classic buckets, but maybe you don't need exemplars per tenant anyway and it will be solved as well

simonswine: I was planning to subtract a particular tenant from our existing SLO metrics, which rely on classic histograms (AFAIK) with the same buckets. So I do think I rely on them (until the main SLOs) use native histograms.

middleware/instrument.go
simonswine: My thinking was more that in the metrics themselves, we call the label `tenant` and the metric `per_tenant_request_duration_seconds` and adding the "_id" wouldn't seem right in my opinion.

`per_tenant_id_request_duration_seconds`
`per_tenant_id_request_duration_seconds{tenant_id="xx"}`

So I think it is better to stick with that naming theme without ID.

PR #492, opened 2024-02-20
Add id to PartitionDesc

pracucci:
**What this PR does**:

In `InstanceDesc` (instances ring) we recently added the instance `id` as part of the struct too. It makes it handy to work with `InstanceDesc`. In `PartitionDesc` we don’t have it, and we don’t need it right now, but adding it now is easier than doing later. In this PR I'm proposing to add it. Since it's expected to be immutable there's no special handling to do, except setting it when a partition is added to the ring.

N/A

**Checklist**
- [ ] Tests updated


ring/partition_ring_model.go
```diff
@@ -133,6 +133,7 @@ func (m *PartitionRingDesc) AddPartition(id int32, state PartitionState, now tim
 	spreadMinimizing := NewSpreadMinimizingTokenGeneratorForInstanceAndZoneID("", int(id), 0, false)
 
 	m.Partitions[id] = PartitionDesc{
+		Id:             id,
```
pstibrany, line 136: This will only work for newly added partitions. What about partitions that already exist in the ring (like on our dev cluster), how will we make sure that those get correct `Id` as well?

pstibrany: For instance-ring, we covered that by https://github.com/grafana/dskit/blob/5d716e13ec96347b7196e89087b954705a0369b3/ring/ring.go#L313-L316 for readers of the ring. We could do something similar in `PartitionRingWatcher`.

pracucci: > This will only work for newly added partitions. What about partitions that already exist in the ring (like on our dev cluster), how will we make sure that those get correct Id as well?

Do we really need to support it? Given it's under development, my idea was to simply ignore the migration problem.

pracucci: I checked the actual implementation and adding support for migration is not difficult, but looks too weird to me that we have to migrate for a feature which is still in development. I will just re-create it in the dev cell.

PR #493, opened 2024-02-21
Add `PartitionRing.ShuffleShardSize`.

pstibrany:
**What this PR does**:

This PR adds `PartitionRing.ShuffleShardSize` method that returns number of partitions that would be in the shuffle shard with the same size. This is useful when this count is needed, but actual partitions are not. In that case, we can simplify the calculation.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_ring.go
+++ b/ring/partition_ring.go
@@ -97,7 +97,6 @@ func (r *PartitionRing) ShuffleShardSize(size int) int {
 		return r.activePartitionsCount
 	}
 
-	// return min(size, r.activePartitionsCount)
 	if size < r.activePartitionsCount {
 		return size
 	}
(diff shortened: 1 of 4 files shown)


ring/partition_ring.go
```diff
@@ -87,6 +91,19 @@ func (r *PartitionRing) ActivePartitionForKey(key uint32) (int32, error) {
 	return 0, ErrNoActivePartitionFound
 }
 
+// ShuffleShardSize returns number of partitions that would be in the result of ShuffleShard call with the same size.
+func (r *PartitionRing) ShuffleShardSize(size int) int {
+	if size <= 0 || size > r.activePartitionsCount {
+		return r.activePartitionsCount
+	}
+
+	// return min(size, r.activePartitionsCount)
```
pracucci, line 100: [nit] I would remove this line.

pracucci (approved):
LGTM, thanks!

ring/partition_ring.go
pstibrany: Done. I left it there as a version to use after we switch to Go 1.21 :)

PR #494, opened 2024-02-21
Expose InstancesInZoneCount and ZonesCount from ring.ReadRing.

pstibrany:
**What this PR does**:

This PR adds `InstancesInZoneCount` and `ZonesCount` methods to `ring.ReadRing` interface and `*ring.Ring` implementation. These methods are already available via lifecycler, but we would like to switch Mimir's limiter from using lifecycler to using `ring.ReadRing` instead in https://github.com/grafana/mimir/pull/7411 (to avoid a bug caused by starting lifecycler after the `ownedSeriesService`), so we are exposing some necessary methods.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -78,6 +78,12 @@ type ReadRing interface {
 
 	// GetTokenRangesForInstance returns the token ranges owned by an instance in the ring
 	GetTokenRangesForInstance(instanceID string) (TokenRanges, error)
+
+	// InstancesInZoneCount returns the number of instances in the ring that are registered in given zone.
+	InstancesInZoneCount(zone string) int
+
+	// ZonesCount returns the number of zones for which there's at least 1 instance registered in the ring.
+	ZonesCount() int
 }
 
 var (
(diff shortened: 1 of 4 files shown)


pracucci (approved):
Nice job! The typical case where tests took longer than code to write, but I'm glad you added them ;)

ring/util_test.go
```diff
@@ -71,6 +71,14 @@ func (r *RingMock) GetTokenRangesForInstance(_ string) (TokenRanges, error) {
 	return []uint32{0, math.MaxUint32}, nil
 }
 
+func (r *RingMock) InstancesInZoneCount(zone string) int {
```
pracucci, line 74: Just to make the linter happy.

```suggestion
func (r *RingMock) InstancesInZoneCount(_ string) int {
```

pracucci (approved):

ring/ring.go
```diff
@@ -78,6 +78,12 @@ type ReadRing interface {
 
 	// GetTokenRangesForInstance returns the token ranges owned by an instance in the ring
 	GetTokenRangesForInstance(instanceID string) (TokenRanges, error)
+
+	// InstancesInZoneCount returns the number of instances in the ring that are registered in given zone.
+	InstancesInZoneCount(zone string) int
+
+	// ZonesCount returns the number of zones for which there's at least 1 instance registered in the ring.
+	ZonesCount() int
```
pr00se, line 86: Do we also need `InstancesCount` to fulfill the `RingCount` [interface](https://github.com/grafana/mimir/blob/16877ec44bc52d4d2f4459ee544dad0e31c8e1f5/pkg/ingester/limiter.go#L18-L22)?

pr00se: it [exists](https://github.com/grafana/dskit/pull/494/files#diff-b51c6d6dfebcf2757957f711c664d9579e1789505a8e771edbe04ede79ff4c9fR59)

PR #495, opened 2024-02-21
Add ring.DoMultiUntilQuorumWithoutSuccessfulContextCancellation() and fix context cancellation in concurrency.ForEachJob() when there's 1 job

pracucci:
**What this PR does**:

In this PR:

1. Fix context cancellation in `concurrency.ForEachJob()` when there's 1 job
   - In https://github.com/grafana/dskit/pull/486 I've introduced an optimization to `concurrency.ForEachJob()` to shortcut the invocation when there's only 1 job. However, this introduced a behavior change in how context is handled: the function is expected to cancel the context passed to the callback. This PR fixes it.
3. Add `ring.DoMultiUntilQuorumWithoutSuccessfulContextCancellation()`
   - `DoMultiUntilQuorumWithoutSuccessfulContextCancellation()` is a wrapper around `DoUntilQuorumWithoutSuccessfulContextCancellation()` which allows to call it for multiple `ReplicationSet` and collect the results.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/concurrency/runner.go
+++ b/concurrency/runner.go
@@ -84,12 +84,19 @@ func CreateJobsFromStrings(values []string) []interface{} {
 
 // ForEachJob runs the provided jobFunc for each job index in [0, jobs) up to concurrency concurrent workers.
 // If the concurrency value is <= 0 all jobs will be executed in parallel.
+//
 // The execution breaks on first error encountered.
+//
+// ForEachJob cancels the context.Context passed to each invocation of jobFunc before ForEachJob returns.
 func ForEachJob(ctx context.Context, jobs int, concurrency int, jobFunc func(ctx context.Context, idx int) error) error {
 	if jobs == 0 {
 		return nil
 	}
 	if jobs == 1 {
+		// Honor the function contract, cancelling the context passed to the jobFunc once it completed.
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
 		return jobFunc(ctx, 0)
 	}
 	if concurrency <= 0 {
@@ -123,6 +130,8 @@ func ForEachJob(ctx context.Context, jobs int, concurrency int, jobFunc func(ctx
 
 // ForEachJobMergeResults is like ForEachJob but expects jobFunc to return a slice of results which are then
 // merged with results from all jobs. This function returns no results if an error occurred running any jobFunc.
+//
+// ForEachJobMergeResults cancels the context.Context passed to each invocation of jobFunc before ForEachJobMergeResults returns.
 func ForEachJobMergeResults[J any, R any](ctx context.Context, jobs []J, concurrency int, jobFunc func(ctx context.Context, job J) ([]R, error)) ([]R, error) {
 	var (
 		resultsMx sync.Mutex
--- a/ring/replication_set.go
+++ b/ring/replication_set.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"sync"
 	"time"
 
 	kitlog "github.com/go-kit/log"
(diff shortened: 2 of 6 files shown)


ring/replication_set_tracker.go
```diff
@@ -465,3 +466,74 @@ func (t *zoneAwareContextTracker) cancelAllContexts(cause error) {
...
+
+	return &inflightInstanceTracker{
+		inflight:            inflight,
+		expectMoreInstances: true,
+	}
+}
+
+// addInstance adds the instance for replicationSetIdx to the tracker.
+//
+// addInstance is idempotent.
```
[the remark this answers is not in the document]
pracucci: > Is this property important for anything?

It's an important property for `removeInstance()` cause there's no guarantee that the context cancel function is called exactly once (I just realised there's no test on it, going to add it). I commented the property for `addInstance()` to keep consistency between the two.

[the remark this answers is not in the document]
pracucci: > allInstancesAdded is also idempotent, and it's not mentioned

Added in https://github.com/grafana/dskit/pull/495/commits/a59f9741bb5ffbef0072142fd7a3700f72d53e57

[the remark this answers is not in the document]
pracucci: > there's no guarantee that the context cancel function is called exactly once (I just realised there's no test on it, going to add it)

Added in https://github.com/grafana/dskit/pull/495/commits/c49d0c00bf26efc49b6011c35cbeaaba9a38827d

ring/replication_set_tracker.go
```diff
@@ -465,3 +466,74 @@ func (t *zoneAwareContextTracker) cancelAllContexts(cause error) {
...
+func (t *inflightInstanceTracker) addInstance(replicationSetIdx int, instance *InstanceDesc) {
+	t.mx.Lock()
```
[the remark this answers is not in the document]
pracucci: At first I was panicking but then I removed it. Reason is that I looked at `DoUntilQuorumWithoutSuccessfulContextCancellation()`. Code is tricky, but I think there could be some races in case of errors where the callback function may called after `DoUntilQuorumWithoutSuccessfulContextCancellation()` returns with a canceled context, so I've removed the panic.

For our case, that we care is that we'll wait for all instances added before `DoUntilQuorumWithoutSuccessfulContextCancellation()` returns. If a race adds an instance slightly after it will still be handled correctly, but there's no correctness issue if won't be waited for.

ring/replication_set_tracker.go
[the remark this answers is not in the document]
pstibrany: > cancel function is called exactly once

It's OK to call it multiple times. Only first time counts.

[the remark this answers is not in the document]
pracucci: > > cancel function is called exactly once
> 
> It's OK to call it multiple times. Only first time counts.

Exactly. So our wrapped needs to be idempotent too.

concurrency/runner.go
```diff
@@ -123,6 +130,8 @@ func ForEachJob(ctx context.Context, jobs int, concurrency int, jobFunc func(ctx
 
 // ForEachJobMergeResults is like ForEachJob but expects jobFunc to return a slice of results which are then
 // merged with results from all jobs. This function returns no results if an error occurred running any jobFunc.
+//
+// ForEachJobMergeResults cancels the context.Context passed to each invocation of jobFunc before ForEachJobMergeResults returns.
 func ForEachJobMergeResults[J any, R any](ctx context.Context, jobs []J, concurrency int, jobFunc func(ctx context.Context, job J) ([]R, error)) ([]R, error) {
```
dimitarvdimitrov, line 135: should we delete this function? We don't need it anymore and I doubt anyone was using it apart from mimir since the time it was added

ring/replication_set.go
```diff
@@ -388,6 +389,111 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
...
+
+		workersGroup                 = sync.WaitGroup{}
+		workersCtx, cancelWorkersCtx = context.WithCancelCause(ctx)
+
+		inflightTracker = newInflightInstanceTracker(sets)
+	)
+
+	cancelWorkersCtxIfSafe := func() {
+		if inflightTracker.allInstancesCompleted() {
+			cancelWorkersCtx(errors.New("all requests completed"))
```
dimitarvdimitrov, line 432: perhaps worth noting that this error should never be seen ever anywhere. I wouldn't open a follow-up PR just for this but if you do end up with one you can add a comment or change the error string to make this more clear

ring/replication_set_tracker.go
```diff
@@ -465,3 +466,76 @@ func (t *zoneAwareContextTracker) cancelAllContexts(cause error) {
 		delete(t.cancelFuncs, instance)
 	}
 }
+
+type inflightInstanceTracker struct {
+	mx       sync.Mutex
+	inflight []map[*InstanceDesc]struct{}
```
dimitarvdimitrov, line 472: nit: assuming that each RS will have 2-3 instances, maybe a slice would have been more efficient

dimitarvdimitrov (commented):
mostly nitpicks, not sure if we shouldn't remove `ForEachJobMergeResults`. otherwise LGTM

concurrency/runner.go
pracucci: We're still using it [here](https://github.com/grafana/mimir/blob/main/pkg/distributor/distributor.go#L1508).

ring/replication_set.go
pracucci: Agree, but it's handy to assert in tests the context was cancelled for the expected reason ([example](https://github.com/grafana/dskit/blob/main/ring/replication_set_test.go#L1575)).

ring/replication_set_tracker.go
pracucci: Will give it a try.

pracucci: Yes! It's better: https://github.com/grafana/dskit/pull/498

PR #497, opened 2024-02-22
Add option to force partition state change from UI

pracucci:
**What this PR does**:

In this PR I'm proposing to add an option to force partition state change from the UI.

##### UX

- To keep the UI simple, I propose to only allow a single state switch based on the partition state:
  - PENDING -> ACTIVE
  - ACTIVE -> INACTIVE
  - INACTIVE -> ACTIVE
- Asks confirmation before proceeding (it's done in javascript)

https://github.com/grafana/dskit/assets/1701904/f405351e-ac2c-495c-b59c-ea4f42cbf583

##### Implementation

- We don't have a component that allows to easily manipulate the partition ring in the store, so I've introduced `PartitionRingEditor`

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_ring_editor.go
+++ b/ring/partition_ring_editor.go
@@ -1,0 +1,60 @@
+package ring
+
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/grafana/dskit/kv"
+)
+
+// PartitionRingEditor is standalone component that can be used to modify the partitions ring.
+// If you want to implement the partition lifecycle you should use PartitionInstanceLifecycler instead.
+type PartitionRingEditor struct {
+	ringKey string
+	store   kv.Client
+}
+
+func NewPartitionRingEditor(ringKey string, store kv.Client) *PartitionRingEditor {
+	return &PartitionRingEditor{
+		ringKey: ringKey,
+		store:   store,
+	}
+}
+
+// ChangePartitionState changes the partition state to toState.
+// This function returns ErrPartitionDoesNotExist if the partition doesn't exist,
+// and ErrPartitionStateChangeNotAllowed if the state change is not allowed.
+func (l *PartitionRingEditor) ChangePartitionState(ctx context.Context, partitionID int32, toState PartitionState) error {
+	return l.updateRing(ctx, func(ring *PartitionRingDesc) (bool, error) {
+		partition, exists := ring.Partitions[partitionID]
+		if !exists {
+			return false, ErrPartitionDoesNotExist
+		}
+
+		if partition.State == toState {
+			return false, nil
+		}
+
+		if !isPartitionStateChangeAllowed(partition.State, toState) {
+			return false, errors.Wrapf(ErrPartitionStateChangeNotAllowed, "change partition state from %s to %s", partition.State.CleanName(), toState.CleanName())
+		}
+
+		return ring.UpdatePartitionState(partitionID, toState, time.Now()), nil
+	})
+}
+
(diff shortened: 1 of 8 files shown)


ring/partition_ring_status.gohtml
```diff
@@ -34,6 +42,19 @@
                         {{$ownerID}} <br />
                     {{ end }}
                 </td>
+                <td>
+                    <!-- Allow to force a state change -->
+                    {{ if and (not .Corrupted) (ne (index $stateChanges .State) 0) }}
+                        {{ $toState := index $stateChanges .State }}
+                        <form action="" method="POST" onsubmit="return confirm('Do you confirm you want to change the state to {{ $toState.CleanName }}?');">
+                            <input type="hidden" name="csrf_token" value="$__CSRF_TOKEN_PLACEHOLDER__">
```
pracucci, line 50: Note to reviewers: we do the same for other forms (e.g. ingester ring), but I can't find anywhere in dskit or Mimir where the token is injected and checked. There's a code comment mentioning it can be done through a middleware, so I've just added the token here as well, but I'm not sure who/where we use it.

pstibrany: > it can be done

perhaps dskit authors thought that some external users will add this on their own? :)

pstibrany (approved):

ring/partition_ring_status.gohtml
pracucci: I don't know... these dskit authors, bad people :D

ring/partition_ring_editor.go
```diff
@@ -0,0 +1,60 @@
...
+		ringKey: ringKey,
+		store:   store,
+	}
+}
+
+// ChangePartitionState changes the partition state to toState.
+// This function returns ErrPartitionDoesNotExist if the partition doesn't exist,
+// and ErrPartitionStateChangeNotAllowed if the state change is not allowed.
+func (l *PartitionRingEditor) ChangePartitionState(ctx context.Context, partitionID int32, toState PartitionState) error {
+	return l.updateRing(ctx, func(ring *PartitionRingDesc) (bool, error) {
```
dimitarvdimitrov, line 30: this is copied from the `PartitionInstanceLifecycler`. Was there a specific reason for doing this? Can we instead change the `PartitionInstanceLifecycler` to delegate to an PartitionRingEditor and don't duplicate the tests and the code?

pracucci: > Can we instead change the PartitionInstanceLifecycler to delegate to an PartitionRingEditor

Unfortunately not because the lifecycler needs to run all changes through a single goroutine to avoid concurrency issues. Refactoring will probably require more code than such a little copying.

dimitarvdimitrov: can't the lifecycler still invoke the editor through a single goroutine?

```go 
func (l *PartitionInstanceLifecycler) ChangePartitionState(ctx context.Context, toState PartitionState) error {
	return l.runOnLifecyclerLoop(func() error {
		return l.editor.ChangePartitionState(ctx, l.cfg.PartitionID, toState)
	})
}
```

pracucci: I took a similar approach without directly using the editor. I moved the change partition logic to a dedicated function which is shared between the two: https://github.com/grafana/dskit/pull/497/commits/2e611f818a566c41cadda35d53d02f5871bdf6c9

dimitarvdimitrov: that looks better, but updateRing is still duplicated. updateRing is more boilerplate code, so i'm happy to keep it as-is

dimitarvdimitrov (approved):

PR #504, opened 2024-03-07
Add support for http.TimeoutHandler

joe-elliott:
**What this PR does**:
Adds the ability to use a `http.TimeoutHandler` instead of the http server's `WriteTimeout`. The timeout handler both correctly cancels context for a query and ends it after the configured timeout (unlike WriteTimeout).

**Checklist**
- [x] Tests updated


pstibrany (commented):
Have you considered providing this functionality as a new middleware?

```
func TimeoutMiddleware(dt time.Duration, msg string) Func {
	return func(next http.Handler) http.Handler {
		return http.TimeoutHandler(next, dt, msg)
	}
}
```

server/server.go
```diff
@@ -104,6 +104,10 @@ type Config struct {
 	HTTPServerReadHeaderTimeout   time.Duration `yaml:"http_server_read_header_timeout"`
 	HTTPServerWriteTimeout        time.Duration `yaml:"http_server_write_timeout"`
 	HTTPServerIdleTimeout         time.Duration `yaml:"http_server_idle_timeout"`
+	// HTTPServerWriteTimeout has some misleading behavior. Consider using the below fields to
+	// enable a timeout handler that will write a response after the timeout. See tests for details.
+	HTTPServerUseTimeoutHandler     bool   `yaml:"-"`
+	HTTPServerTimeoutHandlerMessage string `yaml:"-"`
```
dimitarvdimitrov, line 110: why not make those configurable via CLI too?

joe-elliott:
@pstibrany 

> Have you considered providing this functionality as a new middleware?

Good call out. I've added and used a timeout middleware which I think reads cleaner. Thanks!

server/server.go
```diff
@@ -104,6 +104,10 @@ type Config struct {
 	HTTPServerReadHeaderTimeout   time.Duration `yaml:"http_server_read_header_timeout"`
 	HTTPServerWriteTimeout        time.Duration `yaml:"http_server_write_timeout"`
 	HTTPServerIdleTimeout         time.Duration `yaml:"http_server_idle_timeout"`
+	// HTTPServerWriteTimeout has some misleading behavior. Consider using the below fields to
+	// enable a timeout handler that will write a response after the timeout. See tests for details.
+	HTTPServerUseTimeoutHandler     bool   `yaml:"-"`
```
dimitarvdimitrov, line 109: can we maybe include that's misleading? tests aren't available when the library is venodred. Perhaps something along the lines of 

```
Use http.TimeoutHandler instead of http.Server.WriteTimeout (See also middleware.TimeoutMiddleware).
Using the WriteTimeout means that the request's context is not cancelled 
immediatelly after hitting the timeout and that the TCP connection is closed.
When using the TimeoutMiddleware, HTTP 503 is instead gracefully returned.
When using the TimeoutMiddleware any later handler will be bound by its timeout,
whereas if you're relying on the server write timeout,
any handler in the chain can modify the write deadline via a ResponseController.
```

server/server.go
dimitarvdimitrov: aha, i see your point. Some more ideas: `HTTPServerGracefulTimeoutsEnabled`, `HTTPServerGracefulTimeoutsMessage`. But your suggestions looks good to me, I'm happy to go with them too. 

I'm not sure about adding a customized message, can't we get away with a hardcoded string `request timeout reached`?

server/server.go
pstibrany: `http.TimeoutHandler` also has one serious drawback -- it buffers entire result in memory until handler finishes, and disallows streaming.

pstibrany:
I'm not sure about suitability of `TimeoutHandler` as a generic replacement for `WriteTimeout`. I think both have their place in the same codebase, and which one is used depends on actual handler. For handlers that take long time to compute results, `TimeoutHandler` may be better. For handlers that can start streaming results fast it's not suitable. I would suggest to apply  middleware for handlers on individual basis, instead of trying to replace WriteTimeout with it.

`TimeoutMiddleware` could disable write deadline on individual requests, while preserving global `WriteTimeout` as-is.

pstibrany:
> `TimeoutMiddleware` could disable write deadline on individual requests, while preserving global `WriteTimeout` as-is.

What I mean is: if someone adds `TimeoutMiddleware` to global HTTP server middlewares (as opposed to per-route middlewares), this can disable write timeout. This way the entire change is localized in the `TimeoutMiddleware`.

middleware/http_timeout.go
```diff
@@ -0,0 +1,48 @@
...
+// NewTimeoutMiddleware returns a new timeout middleware that returns a 503 Service Unavailable
+// using the http.TimeoutHandler. Note also that the middleware disables the http server write timeout
+// to ensure the two timeouts don't conflict. We disable the server write timeout b/c it's behavior may
+// be unintuitive. See below.
+//
+// Server.WriteTimeout:
+//   - does not cancel context and instead allows the request to go until the next write. in practice this
+//     means that an http server with a write timeout of 10s may go for significantly longer
+//   - closes the tcp connection on the next write after the timeout has elapsed instead of sending a
+//     meaningful http response
```
pstibrany, line 20: - allows streaming of response

middleware/http_timeout.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// http.TimeoutHandler
+//   - cancels context allowing downstream code to abandon the request
+//   - returns a 503 Service Unavailable with the provided message
```
pstibrany, line 24: - buffers response in memory until handler returns

server/server.go
joe-elliott: moved all comments to middleware and removed Server changes. also added @pstibrany's concern about response buffering.

pstibrany (approved):
Would it be possible to add simple test to check for the behaviour? For example start HTTP server with 1s server timeout, install handler that uses timeout middleware with 2s timeout and handler waits for context cancellation. Client should observe 2s timeout, and all handlers should finish correctly. WDYT?

joe-elliott:
> Would it be possible to add simple test to check for the behaviour? For example start HTTP server with 1s server timeout, install handler that uses timeout middleware with 2s timeout and handler waits for context cancellation. Client should observe 2s timeout, and all handlers should finish correctly. WDYT?

Apologies! I forgot to `git add` my test. It's there now!

pstibrany:
> Would it be possible to add simple test to check for the behaviour? 

Oh wait, there IS a test. I missed it. 🤦

PR #507, opened 2024-03-19
Add an option to disable gRPC server stats tracking

na--:
**What this PR does**:

Disabling these stats should actually allow gRPC to reuse the memory buffers used for receiving incoming requests. 

Right now, the gRPC library only returns the `[]byte` buffer back to the pool [if the `payInfo` variable is `nil` here](https://github.com/grpc/grpc-go/blob/dadbbfa2863a67ed640aac924b7b7fd18b50a429/rpc_util.go#L784-L793). And that variable (of `*payloadInfo` type) is initialized [here](https://github.com/grpc/grpc-go/blob/dadbbfa2863a67ed640aac924b7b7fd18b50a429/server.go#L1318-L1321) for unary RPC (and in a bunch of places for Streams, e.g. [here](https://github.com/grpc/grpc-go/blob/dadbbfa2863a67ed640aac924b7b7fd18b50a429/stream.go#L1714-L1716)) if `shs` (i.e. the `StatsHandler`) is not `nil` and if we don't have binary loggers. 

I couldn't see anywhere where we use [the `binarylog` package](https://pkg.go.dev/google.golang.org/grpc@v1.62.1/binarylog), so removing the stats handler might allow us to achieve some memory usage optimizations by also enabling the recv buffer pools :crossed_fingers: 

**Checklist**
- [ ] Tests updated


seizethedave:
Interesting. Just a drive-by comment- you probably already learned this but I noticed that `recvBufferPool` is a no-op pool unless explicitly configured. https://github.com/grpc/grpc-go/blob/dadbbfa2863a67ed640aac924b7b7fd18b50a429/server.go#L184

na--:
@seizethedave, I saw that, but thanks for pointing it out! I figured that I should be able to configure it by passing the result of [`experimental.RecvBufferPool()`](https://pkg.go.dev/google.golang.org/grpc/experimental#RecvBufferPool) to dskit's [`server.Config.GRPCOptions`](https://github.com/grafana/dskit/blob/73feada6c0d7d02d419415a63fd5a31ae2a40830/server/server.go#L110) to start with. If it works out as I hope, I can make another PR and expose this as an actual option in the Config? Or do you think I should add it now?

na--:
Added the buffer pools as a separate option in a new PR: https://github.com/grafana/dskit/pull/510

They are still an experimental feature, so I'm not sure we should add them to dskit just yet, when individual users of the library can do it manually on their own (once this PR has been merged).

server/server.go
```diff
@@ -191,6 +192,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.GRPCServerTimeout, "server.grpc.keepalive.timeout", time.Second*20, "After having pinged for keepalive check, the duration after which an idle connection should be closed, Default: 20s")
 	f.DurationVar(&cfg.GRPCServerMinTimeBetweenPings, "server.grpc.keepalive.min-time-between-pings", 5*time.Minute, "Minimum amount of time a client should wait before sending a keepalive ping. If client sends keepalive ping more often, server will send GOAWAY and close the connection.")
 	f.BoolVar(&cfg.GRPCServerPingWithoutStreamAllowed, "server.grpc.keepalive.ping-without-stream-allowed", false, "If true, server allows keepalive pings even when there are no active streams(RPCs). If false, and client sends ping when there are no active streams, server will send GOAWAY and close the connection.")
+	f.BoolVar(&cfg.GRPCServerStatsTrackingEnabled, "server.grpc.stats-tracking-enabled", true, "If true, the request_message_bytes, response_message_bytes, and inflight_requests metrics will be tracked, however the memory used for parsing gRPC request bodies will not be reused.")
```
slim-bean, line 195:
```suggestion
	f.BoolVar(&cfg.GRPCServerStatsTrackingEnabled, "server.grpc.stats-tracking-enabled", true, "If true, the request_message_bytes, response_message_bytes, and inflight_requests metrics will be tracked. Enabling this option prevents the use of memory pools for parsing gRPC request bodies and may lead to more memory allocations.")
```

PR #508, opened 2024-03-19
Add PROXY protocol support to `Server`

drclau:
**What this PR does**:
Adds support for PROXY protocol to `Server`, by using `go-proxyproto` package.

Fixes N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -17,13 +17,13 @@ import (
 	"strings"
 	"time"
 
-	_ "github.com/grafana/pyroscope-go/godeltaprof/http/pprof" // anonymous import to get godelatprof handlers registered
-
 	gokit_log "github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
+	_ "github.com/grafana/pyroscope-go/godeltaprof/http/pprof" // anonymous import to get godelatprof handlers registered
 	otgrpc "github.com/opentracing-contrib/go-grpc"
 	"github.com/opentracing/opentracing-go"
+	"github.com/pires/go-proxyproto"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/config"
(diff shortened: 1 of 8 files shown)


server/server.go
```diff
@@ -592,3 +599,13 @@ func (s *Server) Shutdown() {
 	_ = s.HTTPServer.Shutdown(ctx)
 	s.GRPC.GracefulStop()
 }
+
+func newProxyProtocolListener(httpListener net.Listener, readHeaderTimeout time.Duration) net.Listener {
+	// Wraps the listener with a proxy protocol listener.
+	// NOTE: go-proxyproto supports non-PROXY, PROXY v1 and PROXY v2 protocols via the same listener.
+	// Therefore, enabling this feature does not break existing setups.
```
grafanalf, line 606: I suggest you move this comment into a more visible place. Perhaps into the README.md at the root of the repository?

server/server_test.go
```diff
@@ -723,36 +740,72 @@ func (f *FakeLogger) assertNotContains(t *testing.T, content string) {
...
+		require.NoError(t, err)
+
+		logger.assertContains(t, "sourceIPs=127.0.0.1")
 	})
 
-	go func() {
-		require.NoError(t, server.Run())
-	}()
-	defer server.Shutdown()
+	t.Run("without PROXY protocol", func(t *testing.T) {
```
grafanalf, line 784: The name of the test is wrong: should be `with PROXY protocol` (not without) as per line 787.

server/server.go
drclau: I like your suggestion. Though, the `README.md` at the root does not seem to be the right place, considering how generic it is and how little information it contains. What do you think about creating a new `README.md` in the `server` directory instead?
I noticed that other directories do have their own `README`s.

grafanalf: Sounds great.

aarongodin (approved):
Beyond the suggestions already made, LGTM

server/server.go
drclau: I went for adding `PROXYPROTOCOL.md`, because the information is very specific.

pracucci (approved):
LGTM, thanks!

PR #510, opened 2024-03-27
Add gRPC server recv buffer pool option

na--:
**What this PR does**:

This PR builds on top of https://github.com/grafana/dskit/pull/507 and adds a dedicated option for enabling gRPC [`RecvBufferPool`](https://pkg.go.dev/google.golang.org/grpc/experimental#RecvBufferPool). 

That's an experimental gRPC feature, so I decided to add it in a separate PR, given that individual users of dskit can enable it locally by manually adding it to `Cfg.Server.GRPCOptions`. However, https://github.com/grafana/dskit/pull/507 is necessary even for that, because it prevents anyone from using the feature due to this limitation:
> Note: The shared buffer pool feature will not be active if any of the following options are used: StatsHandler, EnableTracing, or binary logging. In such cases, the shared buffer pool will be ignored. 

**Checklist**
- [x] Tests updated


server/server.go
```diff
@@ -193,6 +195,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.GRPCServerMinTimeBetweenPings, "server.grpc.keepalive.min-time-between-pings", 5*time.Minute, "Minimum amount of time a client should wait before sending a keepalive ping. If client sends keepalive ping more often, server will send GOAWAY and close the connection.")
 	f.BoolVar(&cfg.GRPCServerPingWithoutStreamAllowed, "server.grpc.keepalive.ping-without-stream-allowed", false, "If true, server allows keepalive pings even when there are no active streams(RPCs). If false, and client sends ping when there are no active streams, server will send GOAWAY and close the connection.")
 	f.BoolVar(&cfg.GRPCServerStatsTrackingEnabled, "server.grpc.stats-tracking-enabled", true, "If true, the request_message_bytes, response_message_bytes, and inflight_requests metrics will be tracked, however the memory used for parsing gRPC request bodies will not be reused.")
+	f.BoolVar(&cfg.GRPCServerRecvBufferPoolsEnabled, "server.grpc.recv-buffer-pools-enabled", false, "If true, gGPC's buffer pools will be used to handle incoming requests. This option is incompatible with `server.grpc.stats-tracking-enabled=true`!")
```
slim-bean, line 198:
```suggestion
	f.BoolVar(&cfg.GRPCServerRecvBufferPoolsEnabled, "server.grpc.recv-buffer-pools-enabled", false, "If true, gGPC's buffer pools will be used to handle incoming requests. Enabling this feature can reduce memory allocation, but also requires disabling GRPC server stats tracking by setting `server.grpc.stats-tracking-enabled=false`")
```

server/server.go
```diff
@@ -426,6 +429,13 @@ func newServer(cfg Config, metrics *Metrics) (*Server, error) {
 		)
 	}
 
+	if cfg.GRPCServerRecvBufferPoolsEnabled {
+		if cfg.GRPCServerStatsTrackingEnabled {
+			return nil, fmt.Errorf("options grpc_server_stats_tracking_enabled and grpc_server_recv_buffer_pools_enabled cannot both be enabled")
```
slim-bean, line 434:
```suggestion
			return nil, fmt.Errorf("grpc_server_stats_tracking_enabled must be set to false if grpc_server_recv_buffer_pools_enabled is true")
```

slim-bean (approved):
LGTM!

PR #511, opened 2024-03-28
Implement UNIX socket support for gRPC and HTTP listeners

### LINKED ISSUE #475: Support listening on unix sockets
It seems this can almost work since [you can specify the network](https://github.com/grafana/dskit/blob/9af49a5b011a71a2bbee9df83e61ef71e750b8d0/server/server.go#L82), the only problem is that it [adds the port to the address](https://github.com/grafana/dskit/blob/9af49a5b011a71a2bbee9df83e61ef71e750b8d0/server/server.go#L260). Perhaps it should only do this if the network equals 'tcp' and ignore the port field, then we could configure Grafana with something like:

```
http_listen_network: unix
http_listen_address: /var/run/grafana.socket
```

hbjydev:
**What this PR does**:

This PR resolves an issue when trying to set the `HTTPListenNetwork` or `GRPCListenNetwork` config values to `unix` where a port would be appended to the file path, preventing the socket from being bound.

As an example, take this configuration:

```yaml
server:
  http_listen_network: unix
  http_listen_address: /var/run/http.sock
```

This would lead to a listener being created for `unix:///var/run/http.sock:80`, including `:80`.

This PR adds a brief check to see if the network is `unix` and if so, removes the port from the address, leading to the correct address being bound (`unix:///var/run/http.sock`).

Fixes #475

**Checklist**
- [x] Tests updated

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -48,6 +48,8 @@ const (
 	DefaultNetwork = "tcp"
 	// NetworkTCPV4 for IPV4 only
 	NetworkTCPV4 = "tcp4"
+  // NetworkUnix for UNIX sockets
+	NetworkUnix = "unix"
 )
 
 // SignalHandler used by Server.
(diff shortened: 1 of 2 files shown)


pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

PR #513, opened 2024-04-03
Removed unused time.Duration parameter from ShouldLog() function in middle.OptionalLogging interface

pracucci:
**What this PR does**:

Yesterday I was working on conditional error logging in Mimir and I noticed that the duration parameter of `ShouldLog()` is not used anywhere. I talked to Bryan which mentioned it was originally added in `weaveworks/common` (before we forked it into `dskit`) "just in case". I propose to remove it since we still don't have a practical use case for it. Given `dskit` is under our control (and we don't mind breaking interfaces), we can always add it back if needed.

N/A

**Checklist**
- [ ] Tests updated


narqo (approved):
🔥 Works for me

Not sure, we have an easy way to scan across the internal repos, to check no one relies on that (which I doubt), just to minimize "surprises".

pracucci:
> Not sure, we have an easy way to scan across the internal repos, to check no one relies on that (which I doubt), just to minimize "surprises".

I checked Loki, Tempo and Pyroscope `main` and none of them use `ShoudLog()` at all.

pstibrany (approved):
I'm fine with that, if Bryan agrees.

PR #514, opened 2024-04-03
Return sampling frequency from ShouldLog()

pracucci:
**What this PR does**:

In order to address https://github.com/grafana/mimir/issues/7690, I propose to push down the "sampled" suffix to `GRPCServerLog`. In order to do it, `GRPCServerLog` needs to know the extra information to add to the logged error, so I've modified `ShouldLog()` to also return a textual "reason" (optional).

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,7 @@
 * [CHANGE] Remove `RouteHTTPToGRPC` option and related functionality #460
 * [CHANGE] Cache: Remove `MemcachedCache` and `RedisCache` structs in favor of `RemoteCacheAdapter` or using the underlying clients directly. #471
 * [CHANGE] Removed unused `time.Duration` parameter from `ShouldLog()` function in `middle.OptionalLogging` interface. #513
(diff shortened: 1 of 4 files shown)


pstibrany (approved):
This only works with specific way of sampling, but it avoids confusion when "sampled (1/x)" was added even to non-sampled messages. We can revisit this when different sampling methods are added.

pracucci:
> This only works with specific way of sampling, but it avoids confusion when "sampled (1/x)" was added even to non-sampled messages. We can revisit this when different sampling methods are added.

Alternative, instead of returning the sampling frequency, we could return something like a message (e.g. string) to describe the actual sampling? I'm open to feedback.

pstibrany:
> Alternative, instead of returning the sampling frequency, we could return something like a message (e.g. string) to describe the actual sampling? I'm open to feedback.

Either way works for me. I joked we could return a function to decorate error too. We may end up doing that eventually, but I'd keep it simple for now and use either `int` or `string`.

pracucci:
> > Alternative, instead of returning the sampling frequency, we could return something like a message (e.g. string) to describe the actual sampling? I'm open to feedback.
> 
> Either way works for me. I joked we could return a function to decorate the error too. We may end up doing that eventually, but I'd keep it simple for now and use either `int` or `string`.

I've pushed a commit https://github.com/grafana/dskit/pull/514/commits/d27330463557055a33281ca4345abbd502b1dee4 moving from "frequency" to a more flexible "reason".

pstibrany (approved):

PR #516, opened 2024-04-10
Expose IngestersWithTokensCount and IngestersWithTokensInZone count from ring.ReadRing

pr00se:
**What this PR does**:

This PR adds `InstancesWithTokensCount` and `InstancesWithTokensInZoneCount` methods to the `ring.ReadRing` interface and `*ring.Ring` implementation. These methods provide global and per-zone counts of registered instances that have some tokens in the ring, regardless of instance state. In Mimir we need these to exclude instances with no tokens when calculating limits (since these instances won't actually handle a portion of the global limit).

This PR also moves the `InstancesCount` implementation to `ring.go`, with the other instance counting functions.

Relates to https://github.com/grafana/mimir/issues/7847

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/model.go
+++ b/ring/model.go
@@ -518,11 +518,35 @@ func (d *Desc) getOldestRegisteredTimestamp() int64 {
 	return result
 }
 
+func (d *Desc) instancesWithTokensCount() int {
+	count := 0
+	if d != nil {
+		for _, ingester := range d.Ingesters {
+			if len(ingester.Tokens) > 0 {
+				count++
+			}
+		}
+	}
+	return count
+}
+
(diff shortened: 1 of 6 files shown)


jhesketh (approved):

ring/model.go
```diff
@@ -518,6 +518,18 @@ func (d *Desc) getOldestRegisteredTimestamp() int64 {
 	return result
 }
 
+func (d *Desc) instancesWithTokensCount() int {
+	count := 0
+	if d != nil {
```
pstibrany, line 523: super-nit: `if d == nil` we can return 0, and then reduce indentation in the following code.

pstibrany (approved):

PR #519, opened 2024-05-06
consul mock: respect context cancellations in Get

dimitarvdimitrov:
**What this PR does**:

What this PR does:

This makes tearing down tests faster. The *consul.Client will wait until *consul.mockKV.Get returns. However, *mockKV.Get will wait 1.5s on average (two broadcasts of the shared sync.Cond) before returning because it's waiting for a new update of the key, which will never come because the test is shutting down. However, in tests that is usually close to 2s because the test started recently.

I tried waiting for the sync.Cond and the context in a separate goroutine, but that becomes more complicated with the mutex and the sync.Cond.

After this change the test will finish after the first sync.Cond broadcast, so 500ms on average and 1s in tests instead of the previous 1.5s and 2s.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/kv/consul/mock.go
+++ b/kv/consul/mock.go
@@ -181,6 +181,10 @@ func (m *mockKV) Get(key string, q *consul.QueryOptions) (*consul.KVPair, *consu
 			if value != nil {
 				valueModifyIndex = value.ModifyIndex
 			}
+			if err := q.Context().Err(); err != nil {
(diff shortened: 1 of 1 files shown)


56quarters (approved):

dimitarvdimitrov:
tests are failing and it seems like related to this PR, I'll take a look later

PR #254, opened 2023-01-19
multierror: Add support for errors.Is()

dimitarvdimitrov:
**What this PR does**:

This allows to use `multierror.MultiError` with `errors.Is()`.
`MultiError` will in turn call `errors.Is()` on every error value.

**Which issue(s) this PR fixes**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #522, opened 2024-05-09
multierror: Add support for errors.As()

### LINKED ISSUE #521: Allow casting multierror.MultiError to the types of its composing errors
It is currently possible to verify if a specific error has been added to the list of errors represented by an object of type `multierror.MultiError` by using `errors.Is()` function. 

For example, assuming that there is a type `customError` implementing the `error` interface, both variables `isCustomError` and `isContextCanceled` will be `true` at the end of the following code snippet:
```
cErr              := newCustomError()
mErr              := multierror.New(context.Canceled, cErr).Err()
isCustomError     := errors.Is(mErr, cErr)             // true
isContextCanceled := errors.Is(mErr, context.Canceled) // true
```

Using `multierror.MultError` withing `errors.As()`, though, do not provide correct results: `multierror.MultiError` represents a list of errors, each of a certain error type, but `multierror.MultiError` cannot be cast to any of those error types.

The problem can be illustrated by using the example from above: although `mErr` contains error `cErr` of type `customError`, `errors.As()` returns false.
```
cErr := newCustomError()
mErr := multerror.New(context.Canceled, cErr).Err()
var target customError
isCast := errors.As(mErr, &target) // false
```

The goal of this issue is to fix that problem.

duricanikolic:
**What this PR does**:
This PR allows that both `errors.As()` can be correctly used with `multierror.MultiError`. 
Please consult the issue for more details.

This is achieved by implementing function `Unwrap() []error`. Since `Unwrap()` covers both `errors.Is()` and `errors.As()`, the implementation of `Is(error) bool` added in #254 has been removed.

Fixes #521 

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] Multierror: Implement `Unwrap() []error`. This allows to use `multierror.MultiError` with both `errors.Is()` and `errors.As()`. Previously implemented `Is(error) bool` has been removed. #522
 * [CHANGE] Add a new `grpc_server_recv_buffer_pools_enabled` option that enables recv buffer pools in the gRPC server (assuming `grpc_server_stats_tracking_enabled` is disabled). #510
 * [CHANGE] Add a new `grpc_server_stats_tracking_enabled` option that allows us to disable stats tracking and potentially improve server memory reuse. #507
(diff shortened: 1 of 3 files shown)


ortuman (approved):
looks good. thanks! 👍

PR #523, opened 2024-05-14
Create CI/CD test and build in the Github action

lamida:
**What this PR does**:

Create CI pipeline in Github Action. Drone removal will be implemented in a different PR.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,4 +1,4 @@
-name: ci
+name: CI
 on:
   push:
     branches:
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -15,36 +15,17 @@ concurrency:
 jobs:
   test-build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version ['1.20', '1.21.x']
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Go
-        uses: actions/setup-go@v4
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v5
         with:
-          go-version: '1.21'
-      - name: mod-check
-        run: make mod-check
-      - name: lint
-        run: make lint
-      - name: test
-        run: make test
-      - name: test-benchmarks
-        run: make test-benchmarks
-      - name: check-protos
-        run: |
-          apt-get update && apt-get -y install unzip
-          go mod vendor
-          make check-protos
-
-  # same like above jobs, but use go1.20
-  # todo: find a way to parameterise the go version
-  test-build-go-1-20:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - name: Set up Go
-        uses: actions/setup-go@v4
-        with:
-          go-version: '1.20'
+          go-version: ${{ matrix.go-version }}
+      - name: Display Go Version
+        run: go version
       - name: mod-check
         run: make mod-check
       - name: lint
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,0 +1,52 @@
+name: ci
+on:
+  push:
+  pull_request:
+
+concurrency:
+  # Cancel any running workflow for the same branch when new commits are pushed.
+  # We group both by ref_name (available when CI is triggered by a push to a branch/tag)
+  # and head_ref (available when CI is triggered by a PR).
+  group: "${{ github.ref_name }}-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+jobs:
+  test-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.21'
+      - name: mod-check
+        run: make mod-check
+      - name: lint
+        run: make lint
+      - name: test
+        run: make test
+      - name: test-benchmarks
+        run: make test-benchmarks
+      - name: check-protos
+        run: make check-protos
+
+  # same like above jobs, but use go1.20
+  # todo: find a way to parameterise the go version
+  test-build-go-1.20:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+      - name: mod-check
+        run: make mod-check
+      - name: lint
+        run: make lint
+      - name: test
+        run: make test
+      - name: test-benchmarks
+        run: make test-benchmarks
+      - name: check-protos
+        run: make check-protos
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -27,8 +27,8 @@ jobs:
           go-version: ${{ matrix.go-version }}
       - name: Display Go Version
         run: go version
-      - name: mod-check
-        run: Run Go mod check
+      - name: Run Go mod check
+        run: make mod-check
       - name: Run lint
         run: make lint
       - name: Run test
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version ['1.20', '1.21.x']
+        go-version: ['1.20', '1.21.x']
     steps:
       - uses: actions/checkout@v4
       - name: Set up Go ${{ matrix.go-version }}
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -13,8 +13,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-build:
-    name: Build and test
+  build:
+    name: Build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


server/server_test.go
```diff
@@ -949,6 +949,7 @@ func TestGrpcOverProxyProtocol(t *testing.T) {
 	var cfg Config
 	cfg.RegisterFlags(flag.NewFlagSet("", flag.ExitOnError))
 	cfg.ProxyProtocolEnabled = true
+	cfg.HTTPListenPort = 8080
```
lamida, line 952: Github action runner complained about port permission in `make test` without this change

```
level=info msg="server listening on addresses" http=127.0.0.1:36861 grpc=127.0.0.1:42313
--- FAIL: TestGrpcOverProxyProtocol (0.00s)
    server_test.go:973: 
        	Error Trace:	/home/runner/work/dskit/dskit/server/server_test.go:9[73](https://github.com/grafana/dskit/actions/runs/9073285038/job/24930131832#step:6:74)
        	            	listen tcp :80: bind: permission denied
        	Test:       	TestGrpcOverProxyProtocol
```

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,41 @@
+name: ci
```
charleskorn, line 1: [nit] Could be capitalised, this is meant to be a human-friendly name used in the UI

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,41 @@
...
+  # and head_ref (available when CI is triggered by a PR).
+  group: "${{ github.ref_name }}-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+jobs:
+  test-build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version: ['1.20', '1.21.x']
```
charleskorn, line 20: Do we need to run all of this on both Go versions? The Drone pipeline runs everything with the most recent Go version, with the exception of the unit tests which are run on both.

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,41 @@
...
+concurrency:
+  # Cancel any running workflow for the same branch when new commits are pushed.
+  # We group both by ref_name (available when CI is triggered by a push to a branch/tag)
+  # and head_ref (available when CI is triggered by a PR).
+  group: "${{ github.ref_name }}-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+jobs:
+  test-build:
+    runs-on: ubuntu-latest
```
charleskorn, line 17: [nit] Might be nice to set a human-friendly name for the job with `name`, something like `Build and test on Go ${{ matrix.go-version }}` perhaps?

server/server_test.go
charleskorn: Could we just set this to 0 to have it choose a random port? Seems like the test doesn't care about the value of the HTTP port, and this will prevent conflicts for anyone with something else already listening on port 8080.

.github/workflows/test-build.yml
charleskorn: Also, any reason to use `1.20` and `1.21.x` (rather than say `1.20.x` and `1.21.x`)?

lamida: The original drone pipeline runs `make test` in the two versions too. I guess to make sure dskit run fine in those last two go minor versions. I changed the matrix to use `1.20.x` and `1.21.x` for consistency.

.github/workflows/test-build.yml
lamida: I tried to change this, but it seems the job id strictly just the alphabet and `-` and `_` with no space. Of course no templating too apparently. So I reverted to the original ID `test-build` which clearly describes what the job do. Let me know if you have a better suggestion

lamida: Oh now I just realized you are talking about run-name. Let try to apply that.

lamida: Applied in ff313735d90e6298a72ffa63bc6ef246e493c81b

.github/workflows/test-build.yml
lamida: To be honest, I am not sure what are the differences between `1.20` vs `1.20.x` in this context. I believe both should use the latest patch versions.

server/server_test.go
```diff
@@ -949,7 +949,8 @@ func TestGrpcOverProxyProtocol(t *testing.T) {
 	var cfg Config
 	cfg.RegisterFlags(flag.NewFlagSet("", flag.ExitOnError))
 	cfg.ProxyProtocolEnabled = true
-	cfg.HTTPListenPort = 8080
+	// Set this to 0 to have it chooses a random port
```
charleskorn, line 952: [nit]

```suggestion
	// Set this to 0 to have it chooses a random port
```

.github/workflows/test-build.yml
```diff
@@ -26,15 +27,15 @@ jobs:
...
+      - name: Run lint
         run: make lint
-      - name: test
+      - name: Run test
         run: make test
-      - name: test-benchmarks
+      - name: Run test benchmarks
         run: make test-benchmarks
-      - name: check-protos
+      - name: Run check protos and generate Go code from the proto files
```
charleskorn, line 38: [nit] We could drop the "run" from each of these names, for example:

* Check go.mod
* Lint
* Test
* Test benchmarks
* Check protos

charleskorn, line 38: I'm not sure "...and generate Go code from the proto files" is accurate - doesn't this target just check that the generated files are up-to-date?

.github/workflows/test-build.yml
charleskorn: > The original drone pipeline runs `make test` in the two versions too. I guess to make sure dskit run fine in those last two go minor versions. I changed the matrix to use `1.20.x` and `1.21.x` for consistency.

Yep, but it doesn't run all of the other targets on both versions - only the unit tests are run on both versions. Do we need to run all the targets on both versions? Seems unnecessary.

.github/workflows/test-build.yml
lamida: > doesn't this target just check that the generated files are up-to-date?

Yes, but it will run also the `clean-protos` and `protos` target if they are not up to date. That's why I put the initial name for the this step with that description. However, at the end maybe I will simplify it to Check protos

.github/workflows/test-build.yml
lamida: Do you have any suggestions to run the test step with different golang versions?

In Drone, we can define container image in the step level, hence we can have a Go container with a different Golang version running the step, one for 1.20 and another for 1.21. However, in Github action we can only define runner at the job level and there is no thing that is called "go runner". The canonical way to have Golang CI is by using actions/setup-go and then Github action provide a matrix strategy to run the combination of versions across jobs and will run the jobs in parallel. 

Stepping back from talking about test in multiple Golang versions, do you think we need to do that at all? Or should we just one latest go version in the pipeline?

charleskorn: >Stepping back from talking about test in multiple Golang versions, do you think we need to do that at all? Or should we just one latest go version in the pipeline?

We have projects that consume dskit that use different versions of Go, so I believe the intent of this is to ensure that we don't start relying on something added in (for example) Go 1.21 when some projects still use 1.20.

>Do you have any suggestions to run the test step with different golang versions?

We'll need to define two separate jobs: one that runs the unit tests with different Go versions with a matrix like you have here, and another job that runs the other targets on just one Go version.

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,56 @@
...
+concurrency:
+  # Cancel any running workflow for the same branch when new commits are pushed.
+  # We group both by ref_name (available when CI is triggered by a push to a branch/tag)
+  # and head_ref (available when CI is triggered by a PR).
+  group: "${{ github.ref_name }}-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build
```
charleskorn, line 17: [nit] "Check" might be a more accurate name here - there's no building going on.

```suggestion
  check:
    name: Check
```

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Display Go version
+        run: go version
+      - name: Go mod check
+        run: make mod-check
+      - name: Lint
+        run: make lint
```
charleskorn, line 54: I don't think running `mod-check` and `lint` again are necessary - just the unit tests are enough.

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: |
+          apt-get update && apt-get -y install unzip
+          go mod vendor
+          make check-protos
+
+  test:
+    name: Test on Go ${{ matrix.go-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
charleskorn, line 41: It might be worth adding a comment here to explain why we run the tests on both Go versions.

.github/workflows/test-build.yml
lamida: Updated in 485a1db93

.github/workflows/test-build.yml
lamida: Updated in 485a1db93

charleskorn (approved):

Issue #192, opened 2022-07-18
Proposal: make memberlist TCP Transport non blocking


pracucci:
_This proposal is based on a theory I have, but haven't proved if it can significantly improve the messages propagation yet._

The dskit's memberlist client propagate changes (e.g. hash ring changes) using broadcast messages:
https://github.com/grafana/dskit/blob/ead3f9308bb7b413ce997182dd4d7c6e038bc68f/kv/memberlist/memberlist_client.go#L1042-L1058

Messages are push to a queue, along with messages received from other nodes and enqueue for re-broadcasting, and then a single goroutine periodically pulls messages from the queue and send them to some other (random) nodes:
https://github.com/grafana/memberlist/blob/09ffed8adbbed3ee3ace7de977c5e1cade90da87/state.go#L604-L626

The `gossip()` function, responsible to pick up messages from the queue and send them to some random nodes, is called at a regular interval, configued by `-memberlist.gossip-interval` (`200ms` by default):
https://github.com/grafana/memberlist/blob/09ffed8adbbed3ee3ace7de977c5e1cade90da87/state.go#L131-L135

For each gossip interval, the `gossip()` function picks up messages up to `UDPBufferSize`, which is hardcoded to 10MB in dskit (because dskit uses a custom transport based on TCP):
https://github.com/grafana/dskit/blob/ead3f9308bb7b413ce997182dd4d7c6e038bc68f/kv/memberlist/memberlist_client.go#L420-L422

We assume that we broadcast new messages and re-broadcast received message every `-memberlist.gossip-interval=200ms` but that's not guaranteed: the `gossip()` function make take significantly longer, because the dskit's custom TCP transport blocks on each call to `rawSendMsgPacket()`. In particular, the call to `rawSendMsgPacket()` blocks for the whole duration it takes to the TCP transport to create a new TCP connection, write the data and close it:
https://github.com/grafana/dskit/blob/ead3f9308bb7b413ce997182dd4d7c6e038bc68f/kv/memberlist/tcp_transport.go#L438-L439

For example, if it can't reach a node and the connection times out the gossiping is blocked for 5 seconds (when running with the default config `-memberlist.packet-dial-timeout=5s`).

## Proposal

The `gossip()` in memberlist library assumes an UDP transport, where sending packet is a "fire and forget", and basically doesn't block for a long time. On the other side, the dskit's TCP transport blocks even if it swallows any error:
[...]

PR #525, opened 2024-05-14
Add concurrency to the memberlist transport's `WriteTo` method

aldernero:
**What this PR does**:
This PR adds concurrency to the WriteTo function with a configurable maximum writes parameter to control the number of goroutines.

This allows the memberlist to keep working if one of the addresses it's trying to write to is unreachable

Fixes https://github.com/grafana/dskit/issues/192

**Checklist**
- [x] Tests updated

### DIFF
--- a/kv/memberlist/tcp_transport.go
+++ b/kv/memberlist/tcp_transport.go
@@ -19,7 +19,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"go.uber.org/atomic"
 
 	dstls "github.com/grafana/dskit/crypto/tls"
 	"github.com/grafana/dskit/flagext"
@@ -100,11 +99,11 @@ type TCPTransport struct {
 	tcpListeners []net.Listener
 	tlsConfig    *tls.Config
 
-	writeMu sync.RWMutex
 	writeCh chan writeRequest
 	writeWG sync.WaitGroup
 
-	shutdown atomic.Int32
+	shutdown   bool
+	shutdownMu sync.RWMutex
 
 	advertiseMu   sync.RWMutex
 	advertiseAddr string
@@ -145,6 +144,7 @@ func NewTCPTransport(config TCPTransportConfig, logger log.Logger, registerer pr
 	}
 
 	for i := 0; i < concurrentWrites; i++ {
+		t.writeWG.Add(1)
 		go t.writeWorker()
 	}
 
@@ -227,7 +227,10 @@ func (t *TCPTransport) tcpListen(tcpLn net.Listener) {
 	for {
 		conn, err := tcpLn.Accept()
 		if err != nil {
-			if s := t.shutdown.Load(); s == 1 {
+			t.shutdownMu.RLock()
+			isShuttingDown := t.shutdown
+			t.shutdownMu.RUnlock()
+			if isShuttingDown {
 				break
 			}
 
@@ -446,17 +449,17 @@ func (t *TCPTransport) getAdvertisedAddr() string {
 // WriteTo is a packet-oriented interface that fires off the given
 // payload to the given address.
 func (t *TCPTransport) WriteTo(b []byte, addr string) (time.Time, error) {
-	if t.shutdown.Load() == 1 {
+	t.shutdownMu.RLock()
+	defer t.shutdownMu.RUnlock() // Unlock at the end to protect the chan
+	if t.shutdown {
 		return time.Time{}, errors.New("transport is shutting down")
 	}
-	t.writeMu.RLock()
-	defer t.writeMu.RUnlock()
-	t.writeWG.Add(1)
 	t.writeCh <- writeRequest{b: b, addr: addr}
 	return time.Now(), nil
 }
 
 func (t *TCPTransport) writeWorker() {
+	defer t.writeWG.Done()
 	for req := range t.writeCh {
 		b, addr := req.b, req.addr
 		t.sentPackets.Inc()
@@ -473,7 +476,6 @@ func (t *TCPTransport) writeWorker() {
 			}
 			logLevel.Log("msg", "WriteTo failed", "addr", addr, "err", err)
 		}
-		t.writeWG.Done()
 	}
 }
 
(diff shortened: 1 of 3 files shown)


kv/memberlist/tcp_transport.go
```diff
@@ -426,7 +432,15 @@ func (t *TCPTransport) getAdvertisedAddr() string {
 func (t *TCPTransport) WriteTo(b []byte, addr string) (time.Time, error) {
 	t.sentPackets.Inc()
 	t.sentPacketsBytes.Add(float64(len(b)))
+	t.writeCh <- struct{}{}
+	go func() {
+		defer func() { <-t.writeCh }()
+		t.writeToAsync(b, addr)
+	}()
```
pstibrany, line 439: Instead of starting new goroutine for each write, shall we have a pool of writing goroutines, and pass them messages to write via channel?

kv/memberlist/tcp_transport.go
```diff
@@ -426,7 +432,15 @@ func (t *TCPTransport) getAdvertisedAddr() string {
 func (t *TCPTransport) WriteTo(b []byte, addr string) (time.Time, error) {
 	t.sentPackets.Inc()
 	t.sentPacketsBytes.Add(float64(len(b)))
+	t.writeCh <- struct{}{}
```
pstibrany, line 435: Is `WriteTo` only called from single goroutine? If not, it just became more blocking.

kv/memberlist/tcp_transport.go
julienduchesne: I've updated the PR to have that

kv/memberlist/tcp_transport.go
julienduchesne: I've added a test for this. Before my PR, calling `WriteTo` x times possibly takes `x * timeout delay`, following my PR, the calls are non-blocking

kv/memberlist/tcp_transport.go
```diff
@@ -424,27 +446,35 @@ func (t *TCPTransport) getAdvertisedAddr() string {
 // WriteTo is a packet-oriented interface that fires off the given
 // payload to the given address.
 func (t *TCPTransport) WriteTo(b []byte, addr string) (time.Time, error) {
-	t.sentPackets.Inc()
-	t.sentPacketsBytes.Add(float64(len(b)))
+	if t.shutdown.Load() == 1 {
+		return time.Time{}, errors.New("transport is shutting down")
+	}
```
pstibrany, line 451: I think there is a race between WriteTo and Shutdown. `t.shutdown.Load()` may return 0, but by the time `WriteTo` gets to writing to channel, it may be closed.

We should check `shutdown` flag and write to the channel inside the same lock. (It would mean changing the `shutdown` flag into simple bool and protect it with a mutex).

In `Shutdown` method, we should take a lock, set shutdown to true, close the channel, release the lock, and then wait for all goroutines to stop. (No need to hold lock for that anymore)

kv/memberlist/tcp_transport.go
```diff
@@ -424,27 +446,35 @@ func (t *TCPTransport) getAdvertisedAddr() string {
...
+	t.writeMu.RLock()
+	defer t.writeMu.RUnlock()
+	t.writeWG.Add(1)
```
pstibrany, line 454: Instead of calling `Add` for each message, I'd suggest to call `Add` for each started goroutine. Closing channel will then trigger stop of goroutines (and calling `Done` once, when `writeWorker` stops, instead of after every message).

This makes Add/Wait/Done logic a bit simpler, and doesn't require use of locks for correct usage. (We still need exclusion for shutdown flag though)

pstibrany (commented):
This doesn't make WriteTo non-blocking -- only increases concurrency, but when all write goroutines are used, blocking will still occur. We should clarify that.

julienduchesne:
> This doesn't make WriteTo non-blocking -- only increases concurrency, but when all write goroutines are used, blocking will still occur. We should clarify that

I considered giving it a buffered channel which would push back the blocking part. It would still be blocking to a point

An unlimited amount of goroutines could cause unexpected problems

pstibrany:
> I considered giving it a buffered channel which would push back the blocking part. It would still be blocking to a point
> 
> An unlimited amount of goroutines could cause unexpected problems

Yes, I think the only *safe* way to make it non-blocking would be to drop a packet if workers are busy and buffer is full (when used). Having unbounded number of goroutines or buffer could lead to memory exhaustion.

kv/memberlist/tcp_transport.go
julienduchesne: Done. Hadn't thought of doing it that way 👍

julienduchesne:
I've instead renamed the PR and changelog. While, semantically, it doesn't really make it non-blocking, it does resolve the core issue which https://github.com/grafana/dskit/issues/192 was meant to solve, would you agree?

kv/memberlist/tcp_transport.go
```diff
@@ -91,7 +99,11 @@ type TCPTransport struct {
 	tcpListeners []net.Listener
 	tlsConfig    *tls.Config
 
-	shutdown atomic.Int32
+	writeCh chan writeRequest
+	writeWG sync.WaitGroup
+
+	shutdown   bool
+	shutdownMu sync.RWMutex
```
pstibrany, line 106: nit: mutex is typically on the top of the group of variables it protects. ("mutex hat")

kv/memberlist/tcp_transport.go
```diff
@@ -424,27 +449,34 @@ func (t *TCPTransport) getAdvertisedAddr() string {
...
 // payload to the given address.
 func (t *TCPTransport) WriteTo(b []byte, addr string) (time.Time, error) {
-	t.sentPackets.Inc()
-	t.sentPacketsBytes.Add(float64(len(b)))
+	t.shutdownMu.RLock()
+	defer t.shutdownMu.RUnlock() // Unlock at the end to protect the chan
+	if t.shutdown {
+		return time.Time{}, errors.New("transport is shutting down")
+	}
+	t.writeCh <- writeRequest{b: b, addr: addr}
```
pstibrany, line 457: If this blocks, it will also block shutdown.

I'd suggest to have some very short "handover timeout" (250ms?), and if it's reached first, silently drop the packet (while increasing some metric), to make sure we don't block here for too long.

WDYT?

kv/memberlist/tcp_transport.go
```diff
@@ -91,7 +99,11 @@ type TCPTransport struct {
 	tcpListeners []net.Listener
 	tlsConfig    *tls.Config
 
-	shutdown atomic.Int32
+	writeCh chan writeRequest
+	writeWG sync.WaitGroup
```
pstibrany, line 103: I'd suggest to move this under `shutdownMu`, since sends to channel are done with read-lock and channel closing is done with write-lock (also worth mentioning in the comment).

kv/memberlist/tcp_transport_test.go
```diff
@@ -61,6 +67,58 @@ func TestTCPTransport_WriteTo_ShouldNotLogAsWarningExpectedFailures(t *testing.T
...
+}
+
+func TestTCPTransportWriteToUnreachableAddr(t *testing.T) {
+	writeCt := 50
+
+	// Listen for TCP connections on a random port
+	freePorts, err := getFreePorts(1)
+	require.NoError(t, err)
+	addr := &net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: freePorts[0]}
+	listener, err := net.ListenTCP("tcp", addr)
```
pstibrany, line 84:
```suggestion
	// Listen for TCP connections on a random port
	listener, err := net.Listen("tcp", "127.0.0.1:0")
```

and then later `listener.Addr().String()` instead of `addr.String()`.

kv/memberlist/tcp_transport.go
pstibrany: Sorry, I meant the channel, not `WaitGroup`.

kv/memberlist/tcp_transport.go
```diff
@@ -91,7 +103,11 @@ type TCPTransport struct {
 	tcpListeners []net.Listener
 	tlsConfig    *tls.Config
 
-	shutdown atomic.Int32
+	shutdownMu sync.RWMutex
+	shutdown   bool
+	writeCh    chan writeRequest // this channel is protected by shutdownMu
```
pstibrany, line 108:
```suggestion
	writeCh    chan writeRequest // reading and closing this channel is protected by shutdownMu
```

pstibrany (approved):
Thank you for addressing my feedback.

PR #527, opened 2024-05-27
Extract route name logic to a single middleware and use it everywhere

charleskorn:
**What this PR does**:

This PR introduces a new HTTP middleware, `RouteInjector`, that extracts and stores the active request's route name for use in telemetry.

This route name can then be used in traces, metrics and logs. The existing tracing and metrics middlewares have been updated to use this single source of the route name, and applications can call `middleware.ExtractRouteName` to include the route name in logs when appropriate.

(none)

**Checklist**
- [x] Tests updated


middleware/route_injector.go
```diff
@@ -0,0 +1,95 @@
...
+// This method should generally only be used in tests: in production code, use RouteInjector instead.
+func WithRouteName(r *http.Request, routeName string) *http.Request {
+	ctx := context.WithValue(r.Context(), contextKeyRouteName, routeName)
+	return r.WithContext(ctx)
+}
+
+// ExtractRouteName returns the route name associated with this request.
+// This is the same route name used for trace and metric names, and is already suitable for use as a Prometheus label
+// value.
+func ExtractRouteName(r *http.Request) string {
```
pstibrany, line 41: Nits:
* I'd suggest to take `context.Context` parameter only, since that's there we take the name from
* I'd also suggest to document that for this function to work, one must include `RouteInjector` middleware as one of the first middlewares, since that's what stores the route into the context.

middleware/grpc_stats.go
```diff
@@ -31,6 +31,7 @@ type contextKey int
 
 const (
 	contextKeyMethodName contextKey = 1
+	contextKeyRouteName  contextKey = 2
```
pstibrany, line 34: nit: move to route_injector.go?

pstibrany (approved):

middleware/grpc_stats.go
charleskorn: I wanted to keep all the `contextKey` values in one place, as otherwise there's a risk that we define two `contextKey`s with the same value (eg. two `contextKey`s with value `3`).

middleware/route_injector.go
charleskorn: Addressed in 3996871

PR #528, opened 2024-05-30
spanprofiler: do less work on unsampled traces

bboreham:
**What this PR does**:

Skip the work to examine the span and decorate it, if the whole trace is not sampled hence the data is going nowhere.
This is 4% of all garbage in one production Mimir installation I looked at, maybe 2% of all CPU.

```
BenchmarkTracer/unsampled-28             2866868               405.6 ns/op           979 B/op         10 allocs/op
BenchmarkTracer/sampled-28               1000000              1024 ns/op            2346 B/op         25 allocs/op
```

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -210,6 +210,7 @@
 * [ENHANCEMENT] Add `outcome` label to `gate_duration_seconds` metric. Possible values are `rejected_canceled`, `rejected_deadline_exceeded`, `rejected_other`, and `permitted`. #512
 * [ENHANCEMENT] Expose `InstancesWithTokensCount` and `InstancesWithTokensInZoneCount` in `ring.ReadRing` interface. #516
 * [ENHANCEMENT] Middleware: determine route name in a single place, and add `middleware.ExtractRouteName()` method to allow consuming applications to retrieve the route name. #527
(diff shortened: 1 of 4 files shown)


colega (approved):

kolesnikovae:
I probably should have made this more clear, but there is a reason for collecting sampled traces: `span_name` attribution (aggregated span profiles):

https://github.com/grafana/dskit/blob/514184267d5cdd82edbababb51462d5049507552/spanprofiler/spanprofiler.go#L66-L71

On the other hand, I suppose that probabilistic _trace_ sampling might not impact the results of _profiling_ sampling (structurally and proportionally), so that even a fraction of span profiles forms a trustworthy picture.

I guess it might make sense to add an option that controls the behaviour

bboreham:
Sorry I missed that nuance. 

It will affect the result: suppose a particular span name does 1x work in traces sampled 0.01 and 100x work in traces sampled 0.1. 

But by this reasoning, the aggregate seems low-value. I need to know the cost in context.

kolesnikovae:
I'm going to add an option that controls this behavior. By default, sampled traces will be accounted for in span aggregate profiles. Would you mind?

PR #537, opened 2024-07-02
Use GetClientCertificate to allow client cert to be reloaded

### LINKED ISSUE #549: Supporting reloading client certificates
Currently I see some logic with the use of GetCertificates that allows to reload the certificate when it rotates. But there is no option for the client certificates.

rubenvp8510:
**What this PR does**:
Use `GetClientCertificate` to reload the client certificates from the disk, similar to what has been done with  `GetCertificate` in this library https://github.com/prometheus/exporter-toolkit/blob/master/web/tls_config.go#L204 which is used here:  https://github.com/grafana/dskit/blob/main/server/server.go#L317 by the server.

This is useful because the certificates can rotate, and with this change we will reload the certificate from the disk without need to restart the application for it.

Fixes #549 

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -218,6 +218,7 @@
 * [ENHANCEMENT] memberlist: use separate queue for broadcast messages that are result of local updates, and prioritize locally-generated messages when sending broadcasts. On stopping, only wait for queue with locally-generated messages to be empty. #539
 * [ENHANCEMENT] memberlist: Added `-<prefix>memberlist.broadcast-timeout-for-local-updates-on-shutdown` option to set timeout for sending locally-generated updates on shutdown, instead of previously hardcoded 10s (which is still the default). #539
 * [ENHANCEMENT] tracing: add ExtractTraceSpanID function.
+* [EHNANCEMENT]  Supporting reloading client certificates #549
 * [CHANGE] Backoff: added `Backoff.ErrCause()` which is like `Backoff.Err()` but returns the context cause if backoff is terminated because the context has been canceled. #538
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -218,7 +218,7 @@
 * [ENHANCEMENT] memberlist: use separate queue for broadcast messages that are result of local updates, and prioritize locally-generated messages when sending broadcasts. On stopping, only wait for queue with locally-generated messages to be empty. #539
 * [ENHANCEMENT] memberlist: Added `-<prefix>memberlist.broadcast-timeout-for-local-updates-on-shutdown` option to set timeout for sending locally-generated updates on shutdown, instead of previously hardcoded 10s (which is still the default). #539
 * [ENHANCEMENT] tracing: add ExtractTraceSpanID function.
-* [EHNANCEMENT]  Supporting reloading client certificates #537
+* [EHNANCEMENT] crypto/tls: Support reloading client certificates #537
 * [CHANGE] Backoff: added `Backoff.ErrCause()` which is like `Backoff.Err()` but returns the context cause if backoff is terminated because the context has been canceled. #538
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -218,7 +218,7 @@
 * [ENHANCEMENT] memberlist: use separate queue for broadcast messages that are result of local updates, and prioritize locally-generated messages when sending broadcasts. On stopping, only wait for queue with locally-generated messages to be empty. #539
 * [ENHANCEMENT] memberlist: Added `-<prefix>memberlist.broadcast-timeout-for-local-updates-on-shutdown` option to set timeout for sending locally-generated updates on shutdown, instead of previously hardcoded 10s (which is still the default). #539
 * [ENHANCEMENT] tracing: add ExtractTraceSpanID function.
-* [EHNANCEMENT]  Supporting reloading client certificates #549
+* [EHNANCEMENT]  Supporting reloading client certificates #537
 * [CHANGE] Backoff: added `Backoff.ErrCause()` which is like `Backoff.Err()` but returns the context cause if backoff is terminated because the context has been canceled. #538
 * [BUGFIX] spanlogger: Support multiple tenant IDs. #59
 * [BUGFIX] Memberlist: fixed corrupted packets when sending compound messages with more than 255 messages or messages bigger than 64KB. #85
--- a/crypto/tls/test/tls_integration_go_1_20_test.go
+++ b/crypto/tls/test/tls_integration_go_1_20_test.go
@@ -4,3 +4,4 @@ package test
 
 // The error message changed in Go 1.21.
 const badCertificateErrorMessage = "remote error: tls: bad certificate"
+const mismatchCAAndCerts = "remote error: tls: unknown certificate authority"
--- a/crypto/tls/test/tls_integration_go_1_20_test.go
+++ b/crypto/tls/test/tls_integration_go_1_20_test.go
@@ -4,4 +4,3 @@ package test
 
 // The error message changed in Go 1.21.
 const badCertificateErrorMessage = "remote error: tls: bad certificate"
-const mismatchCAAndCerts = "remote error: tls: unknown certificate authority"
--- a/crypto/tls/test/tls_integration_go_1_21_test.go
+++ b/crypto/tls/test/tls_integration_go_1_21_test.go
@@ -4,3 +4,4 @@ package test
 
 // The error message changed in Go 1.21.
 const badCertificateErrorMessage = "remote error: tls: certificate required"
+const mismatchCAAndCerts = "remote error: tls: unknown certificate authority"
--- a/crypto/tls/test/tls_integration_go_1_21_test.go
+++ b/crypto/tls/test/tls_integration_go_1_21_test.go
@@ -4,4 +4,3 @@ package test
 
 // The error message changed in Go 1.21.
 const badCertificateErrorMessage = "remote error: tls: certificate required"
-const mismatchCAAndCerts = "remote error: tls: unknown certificate authority"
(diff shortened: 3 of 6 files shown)


zalegrala (commented):
This looks reasonable to me.  Adhere to the lint check.

Is the expectation here that the serer certificate will be reloaded by some other means?

rubenvp8510:
> This looks reasonable to me. Adhere to the lint check.
> 
> Is the expectation here that the serer certificate will be reloaded by some other means?

The expectation is that at some point the certificate expired, and usually in some platforms there is a logic for cert renewal. But if the certificate is loaded only one time in memory, even if the file changed this won't take effect. Unless you reload it as this PR does. may be an additional improvement will be to not reload each time

crypto/tls/tls.go
```diff
@@ -109,29 +122,42 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
...
+				return nil, errors.Wrapf(err, "error loading client key: %s", cfg.KeyPath)
+			}
+
+			clientCert, err := tls.X509KeyPair(cert, key)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to load TLS certificate %s,%s", cfg.CertPath, cfg.KeyPath)
+			}
+			return &clientCert, nil
 		}
+		return nil, err
```
aknuds1, line 144: I suggest to drop the `cfg.validateCertificatePaths` call, since the assumption is already that a TLS client certificate is configured. If, between calls to `loadCert`, both `cfg.CertPath` and `cfg.KeyPath` were to be set to the empty string, the code would break.

```suggestion
		cert, err := reader.ReadSecret(cfg.CertPath)
		if err != nil {
			return nil, errors.Wrapf(err, "error loading client cert: %s", cfg.CertPath)
		}
		key, err := reader.ReadSecret(cfg.KeyPath)
		if err != nil {
			return nil, errors.Wrapf(err, "error loading client key: %s", cfg.KeyPath)
		}

		clientCert, err := tls.X509KeyPair(cert, key)
		if err != nil {
			return nil, errors.Wrapf(err, "failed to load TLS certificate %s,%s", cfg.CertPath, cfg.KeyPath)
		}
		return &clientCert, nil

```

crypto/tls/tls.go
```diff
@@ -109,29 +122,42 @@ func (cfg *ClientConfig) GetTLSConfig() (*tls.Config, error) {
...
-		}
-		key, err := reader.ReadSecret(cfg.KeyPath)
-		if err != nil {
-			return nil, errors.Wrapf(err, "error loading client key: %s", cfg.KeyPath)
+	useClientCerts, err := cfg.validateCertificatePaths()
+	if err != nil {
+		return nil, err
+	}
+
+	if useClientCerts {
```
aknuds1, line 152: It makes more sense to me to only validate the configuration parameters before setting `config.GetClientCertificate`, unless you want to fully handle configuration parameters potentially changing:

```suggestion
	// Read Client Certificate
	if cfg.CertPath != "" || cfg.KeyPath != "" {
		if cfg.CertPath == "" {
			return nil, errCertMissing
		}
		if cfg.KeyPath == "" {
			return nil, errKeyMissing
		}
```

aknuds1 (commented):
The idea makes sense to me, but the current implementation doesn't really.

crypto/tls/test/tls_integration_go_1_20_test.go
```diff
@@ -4,3 +4,4 @@ package test
 
 // The error message changed in Go 1.21.
 const badCertificateErrorMessage = "remote error: tls: bad certificate"
+const mismatchCAAndCerts = "remote error: tls: unknown certificate authority"
```
aknuds1, line 7: Please move this to crypto/tls/test/tls_integration_test.go, since it doesn't differ between Go versions.

crypto/tls/test/tls_integration_go_1_21_test.go
```diff
@@ -4,3 +4,4 @@ package test
 
 // The error message changed in Go 1.21.
 const badCertificateErrorMessage = "remote error: tls: certificate required"
+const mismatchCAAndCerts = "remote error: tls: unknown certificate authority"
```
aknuds1, line 7: Please move this to crypto/tls/test/tls_integration_test.go, since it doesn't differ between Go versions.

crypto/tls/tls_test.go
```diff
@@ -109,7 +109,7 @@ func TestGetTLSConfig_ClientCerts(t *testing.T) {
 	tlsConfig, err := c.GetTLSConfig()
 	assert.NoError(t, err)
 	assert.Equal(t, false, tlsConfig.InsecureSkipVerify, "make sure we default to not skip verification")
-	assert.Equal(t, 1, len(tlsConfig.Certificates), "ensure a certificate is returned")
+	assert.NotNil(t, tlsConfig.GetClientCertificate, "ensure a get certificate function is returned")
```
aknuds1, line 112: I suggest also testing what GetClientCertificate returns.
```suggestion
	require.NotNil(t, tlsConfig.GetClientCertificate, "ensure GetClientCertificate is set")
	cert, err := tlsConfig.GetClientCertificate(nil)
	require.NoError(t, err)
	assert.NotNil(t, cert, "ensure GetClientCertificate returns a certificate")
```

crypto/tls/tls.go
```diff
@@ -83,6 +83,19 @@ func (cfg *ClientConfig) GetTLSCipherSuitesLongDescription() string {
...
+		if cfg.CertPath == "" {
+			return true, errCertMissing
+		}
+		if cfg.KeyPath == "" {
+			return true, errKeyMissing
+		}
+		return true, nil
+	}
+	return false, nil
+}
```
aknuds1, line 97: This is no longer in use.
```suggestion
```

CHANGELOG.md
```diff
@@ -218,6 +218,7 @@
 * [ENHANCEMENT] memberlist: use separate queue for broadcast messages that are result of local updates, and prioritize locally-generated messages when sending broadcasts. On stopping, only wait for queue with locally-generated messages to be empty. #539
 * [ENHANCEMENT] memberlist: Added `-<prefix>memberlist.broadcast-timeout-for-local-updates-on-shutdown` option to set timeout for sending locally-generated updates on shutdown, instead of previously hardcoded 10s (which is still the default). #539
 * [ENHANCEMENT] tracing: add ExtractTraceSpanID function.
+* [EHNANCEMENT]  Supporting reloading client certificates #537
```
aknuds1, line 221:
```suggestion
* [EHNANCEMENT] crypto/tls: Support reloading client certificates #537
```

aknuds1 (changes requested):
Looks pretty good now, but `validateCertificatePaths` is now dead and should be removed. Plus, some other simplifications should be made.

aknuds1 (approved):
LGTM, thanks for addressing my feedback!

charleskorn:
This change has broken the `TestSingleBinaryWithMemberlist/tls` integration test in Mimir ([sample logs](https://github.com/grafana/mimir/actions/runs/10137802610/job/28029505013#step:15:206)), it is failing with:

```
failed to create transport: failed to start TLS TCP listener on "0.0.0.0" port 8000: tls: neither Certificates, GetCertificate, nor GetConfigForClient set in Config
```

Could you please take a look @rubenvp8510? You should be able to pull the branch from https://github.com/grafana/mimir/pull/8833 to reproduce this.

rubenvp8510:
Hello @charleskorn I'll take a look today, it seems like the test is expecting one of these to be defined: `Certificates`, `GetCertificate`, nor `GetConfigForClient` . Why is not valid to expect a `GetClientCertificate` ? My understanding if that if you set that you don't need to set `Certificates`. Also this is for clients., so `GetCertificate` won't work on this case. 

Other thing I can do is set `GetConfigForClient`, but wondering why this is needed as I only want to make the part of the certificate to be reloaded (which is why I'm using `GetClientCertificate)

zalegrala:
How goes the testing?  I'm seeing the above error in Tempo also after the dskit update.

aknuds1:
I'm trying to reproduce the test error on my side.

aknuds1:
Strange, on my side the memberlist-kv module doesn't fail for the reason Charles linked to in CI. For me it fails with a non-informative error: 

```
16:43:20 mimir-1: ts=2024-08-01T14:43:20.847969237Z caller=mimir.go:907 level=error msg="module failed" module=memberlist-kv err="starting module memberlist-kv: invalid service state: Stopping, expected: Running"
```

zalegrala:
I'm seeing this in Tempo.
```
level=error ts=2024-08-01T15:06:58.78706799Z caller=app.go:223 msg="module failed" module=memberlist-kv err="service memberlist_kv failed: failed to create transport: failed to start TLS TCP listener on \"::\" port 7946: tls: neither Certificates, GetCertificate, nor GetConfigForClient set in Config"
```

rubenvp8510:
It seems like in order to not break anything we will need to provide a `GetConfigForClient` instead, not sure why `GetClientCertificate` is not enough  I'll do the change and send the PR , I expected to have something between today and tomorrow.

aknuds1:
@charleskorn @rubenvp8510 @zalegrala I think the problem is that memberlist/kv uses a client certificate also for server purposes, that's the reason for the error. As a fix, shall we also set the `Certificates` field?

rubenvp8510:
@aknuds1 that make sense

According to go documentation

```
Server configurations must set one of Certificates, GetCertificate or
GetConfigForClient. Clients doing client-authentication may set either
Certificates or GetClientCertificate.
```
This is confusing because the struture is named `ClientConfig`. 

About setting certificates, What I don't know is , if I set certificates then `GetClientCertificate` will be called? or if this is also used for server, we can set algo GetCertificates method.

aknuds1:
@rubenvp8510 I agree it's confusing, I don't know yet why memberlist/kv appears to use `ClientConfig` for a server. Maybe it shouldn't?

> About setting certificates, What I don't know is , if I set certificates then GetClientCertificate will be called? or if this is also used for server, we can set algo GetCertificates method.

The `GetClientCertificate` documentation makes it clear that it masks `Certificates` for clients.

```golang
// GetClientCertificate, if not nil, is called when a server requests a
// certificate from a client. If set, the contents of Certificates will
// be ignored.
```

aknuds1:
I've tested also setting `Certificates` for backwards compat purposes, that makes the integration test pass.

rubenvp8510:
> I've tested also setting `Certificates` for backwards compat purposes, that makes the integration test pass. Making a PR.

Great! then we can set it with confidence :)

aknuds1:
I've made a [PR](https://github.com/grafana/dskit/pull/552) to fix the issue.

PR #538, opened 2024-07-03
Add Backoff.ErrCause()

pracucci:
**What this PR does**:

In this PR I propose a add `Backoff.ErrCause()` which is like `Backoff.Err()` but returns the context cause if backoff is terminated because the context has been canceled.

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -214,6 +214,7 @@
 * [ENHANCEMENT] SpanProfiler: do less work on unsampled traces. #528
 * [ENHANCEMENT] Log Middleware: if the trace is not sampled, log its ID as `trace_id_unsampled` instead of `trace_id`. #529
 * [EHNANCEMENT] httpgrpc: httpgrpc Server can now use error message from special HTTP header when converting HTTP response to an error. This is useful when HTTP response body contains binary data that doesn't form valid utf-8 string, otherwise grpc would fail to marshal returned error. #531
+* [CHANGE] Backoff: added `Backoff.ErrCause()` which is like `Backoff.Err()` but returns the context cause if backoff is terminated because the context has been canceled. #538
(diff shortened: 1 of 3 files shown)


pracucci:
I just realised we require golang 1.20 in dskit. Damn this repo lives in the past!

aknuds1 (approved):
LGTM - except I guess we need to increase the minimum Go version to 1.21. Does anything hinder us from doing so?

pstibrany:
> If there are objections to this change, then an alternative may be adding another function `Backoff.ErrCause()` which is like `Err()` but returns the context cause instead.

This seems like safer option to me, to avoid breaking changes to clients using `Backoff` today.

simonswine:
> > If there are objections to this change, then an alternative may be adding another function `Backoff.ErrCause()` which is like `Err()` but returns the context cause instead.
> 
> This seems like safer option to me, to avoid breaking changes to clients using `Backoff` today.

I would agree with @pstibrany: A subtle breaking change like this will not come up when randomly upgrading the dskit package, and not too sure that everyone fully check the diff between go mod updates.

I could also think of something that would at least make the `errors.Is` still work by wrapping the cause: https://go.dev/play/p/tT6qg_rnDxd, but still prefer the simplicity of a separate method.

pracucci:
Thanks for the feedback. Switching to draft cause it requires golang 1.21 anyway, so we would need to upgrade first.

pracucci:
Thanks for the feedback people. I've added `Backoff.ErrCause()` instead, so that this PR is not a breaking change. Still blocked on https://github.com/grafana/dskit/pull/540 tho.

pracucci:
> Still blocked on #540 tho.

I merged #540

simonswine (approved):
LGTM

PR #539, opened 2024-07-03
memberlist: Use separate queue for CAS messages and only wait for CAS messages queue to be empty when stopping

pstibrany:
**What this PR does**:

This PR changes memberlist client to send out CAS updates faster, and to reduce chance of dropping CAS updates when memberlist is stopping.

* memberlist client now keeps two separate queues for messages that should be broadcast. One queue for messages that were created as a result of local CAS operation, other queue for remaining messages, created as a result of gossip updates.
* CAS messages are prioritized when gossiping messages to other nodes
* When memberlist client is stopping:
  * as before, no more CAS messages can be enqueued. However other messages can be still enqueued and are forwarded.
  * stopping method will wait only for CAS queue to be empty.
* Wait timeout on shutdown is now configurable, and can be disabled completely. Previously it was hardcoded to 10 seconds.

Using 2 separate queues avoids optimization performed by the queue, when subsequent updates affecting the same "change" (eg. instances in case of instance ring) can be merged together. However CAS updates typically modify different parts of the key (ie. different instances in the instance ring) than incoming gossip messages, so this doesn't look like a problem.

**Checklist**
- [x] Tests updated


kv/memberlist/memberlist_client.go
```diff
@@ -1163,32 +1193,25 @@ func (m *KV) processValueUpdate(workerCh <-chan valueUpdate, key string) {
 	}
 }
 
-func (m *KV) queueBroadcast(key string, content []string, version uint, message []byte) {
```
pstibrany, line 1166: inlined into `broadcastNewValue`

kv/memberlist/memberlist_client_test.go
```diff
@@ -718,7 +718,7 @@ func testMultipleClientsWithConfigGenerator(t *testing.T, members int, configGen
 
 	startTime := time.Now()
 	firstKv := clients[0]
-	ctx, cancel := context.WithTimeout(context.Background(), casInterval*3/2) // Watch for 1.5 cas intervals.
+	ctx, cancel := context.WithTimeout(context.Background(), casInterval*3) // Watch for 3x cas intervals.
```
pstibrany, line 721: This change is unrelated, only done to avoid test flakiness.

kv/memberlist/memberlist_client.go
```diff
@@ -231,10 +233,11 @@ type KV struct {
 	// dns discovery provider
 	provider DNSProvider
 
-	// Protects access to memberlist and broadcasts fields.
-	delegateReady atomic.Bool
-	memberlist    *memberlist.Memberlist
-	broadcasts    *memberlist.TransmitLimitedQueue
+	// Protects access to memberlist and gossipBroadcasts fields.
```
seizethedave, line 236:
```suggestion
	// Protects access to memberlist and broadcast queues.
```

kv/memberlist/memberlist_client.go
```diff
@@ -1335,7 +1358,7 @@ func (m *KV) MergeRemoteState(data []byte, _ bool) {
 			level.Error(m.logger).Log("msg", "failed to store received value", "key", kvPair.Key, "err", err)
 		} else if newver > 0 {
 			m.notifyWatchers(kvPair.Key)
-			m.broadcastNewValue(kvPair.Key, change, newver, codec)
+			m.broadcastNewValue(kvPair.Key, change, newver, codec, false)
```
seizethedave, line 1361: These messages we're persisting to the local KV that come from external broadcasts are non-CAS by definition? (Only the local mutations are called CAS?)

seizethedave (approved):
LGTM. A nice improvement.

kv/memberlist/memberlist_client.go
pstibrany: Correct. Currently only CAS operation can modify KV store. There is also Delete operation in [KV client interface](https://github.com/grafana/dskit/blob/47b410a939dfc243124b650db6214e87b0f5a540/kv/client.go#L91-L125), but memberlist implementation [doesn't support it yet](https://github.com/grafana/dskit/blob/4beef469a18ec34fdc12d1fd678b1f0c018de4f4/kv/memberlist/memberlist_client.go#L74-L77). But we should implement it eventually, so perhaps it would be better to call it "local updates", instead of "cas updates". WDYT? I think I'll rename it.

seizethedave: Sounds good! CAS is an implementation detail and doesn't really explain that it's a local mod.

pstibrany: I've done this change in https://github.com/grafana/dskit/pull/539/commits/979c0311f20e86684374328c29dd386be86e4b30. I've also updated `memberlist_client_messages_in_broadcast_queue` metric to report values for both queues individually.

seizethedave (approved):
Nice, LGTM.

PR #543, opened 2024-07-05
Cache jump-hash: use xxhash v2

bboreham:
This is consistent with what we use elsewhere


paul1r (approved):

cache/memcached_server_selector.go
```diff
@@ -7,7 +7,7 @@ import (
 	"net"
 	"sync"
 
-	"github.com/cespare/xxhash"
+	"github.com/cespare/xxhash/v2"
```
colega, line 10: I am not familiar with the differences: do they yield the same hash?

colega: @pstibrany responded on Slack, it should be the same: https://xxhash.com/

PR #544, opened 2024-07-05
Neater string-to-byteslice conversions

bboreham:
`unsafe.Slice` and `unsafe.StringData` were added in Go 1.20

`yoloBuf` previously did not fill in the `cap` member, which didn't matter in the way it was used but was potentially risky in future.


cyriltovena (approved):
LGTM

PR #550, opened 2024-07-17
tracing: add ExtractTraceSpanID

krajorama:
**What this PR does**:

Add ExtractTraceSpanID helper function for enhanced logging.

Related to https://github.com/grafana/mimir/pull/8599

**Checklist**
- [ ] Tests updated


flxbk (approved):
LGTM

tracing/tracing.go
```diff
@@ -55,16 +55,30 @@ func NewFromEnv(serviceName string, options ...jaegercfg.Option) (io.Closer, err
...
+		return nil
 	}
 	sctx, ok := sp.Context().(jaeger.SpanContext)
 	if !ok {
-		return "", false
+		return nil
 	}
-
-	return sctx.TraceID().String(), true
+	return &sctx
```
pstibrany, line 81: By returning `&sctx`, `jaeger.SpanContext` needs to be allocated on a heap first, and then pointer to it returned, right?

Perhaps we could return just traceID and spanID directly? Or `jaeger.SpanContext` without the pointer.

pstibrany (approved):

tracing/tracing.go
krajorama: I thought about that , but that would be duplicated code. I can definitely do a copy of the jeager.SpanContext thought, if that's preferred?

krajorama: By the way , I don't know if the compiler is smart enough to avoid allocation on the heap

ying-jeanne: Can we keep ExtractTraceID, and call ExtractTraceSpanID inside, just discard spanID when return result. In that case no need to refectory this extractJaegerContext but leave all logics in ExtractTraceSpanID

krajorama: facepalm, of course

ying-jeanne (approved):
LGTM

tracing/tracing.go
ying-jeanne: I think it is a fair comment, current code LGTM, approved.

krajorama: it did inspire me to go ahead and look at the assembly :) Got the final result by deferring the formatting to the public function , but only returning the two relevant fields from the helper function. It's the least extra instructions.

PR #551, opened 2024-07-18
memberlist: Remove noisy debug log message

pstibrany:
**What this PR does**:
This PR removes noisy log message: Invalidating previous messages that are still in the queue by new messages is a common case, we don't need to log a message each time it happens.

Found while watching https://www.youtube.com/watch?v=iH0Ufv2bD1U, where it shows up as most frequent log message in the demo.

**Checklist**
- [na] Tests updated

### DIFF
--- a/kv/memberlist/broadcast.go
+++ b/kv/memberlist/broadcast.go
@@ -1,10 +1,7 @@
 package memberlist
 
 import (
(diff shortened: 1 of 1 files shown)


cyriltovena (approved):
LGTM

PR #552, opened 2024-08-01
crypto/tls.ClientConfig: Set Config.Certificates for backwards compat

aknuds1:
**What this PR does**:
A [recent change](https://github.com/grafana/dskit/pull) in `crypto/tls.ClientConfig.GetTLSConfig` that allows for reloading of client certs, turned out to [break](https://github.com/grafana/dskit/pull/537#issuecomment-2254953303) `memberlist/kv.NewTCPTransport`. Apparently the latter uses `ClientConfig` for a server, which then fails since there's no longer any certificate config for servers.

I propose working around the problem by again setting `crypto/tls.Config.Certificates`, as it used to be prior to the aforementioned change. However, since we now also set `GetClientCertificate`, _clients will prefer the latter_:

```golang
// GetClientCertificate, if not nil, is called when a server requests a
// certificate from a client. If set, the contents of Certificates will
// be ignored.
```

I've tested that this change fixes the broken integration test in Mimir (`TestSingleBinaryWithMemberlist/tls`).

**Checklist**
- [x] Tests updated


rubenvp8510 (approved):

zalegrala (approved):

zalegrala:
Looks good to me.

PR #553, opened 2024-08-02
Add read only ingester support to lifecycler and ring

jhalterman:
**What this PR does**:

This PR is 1 of 2 that will add read-only ingester support to the ring and lifecycler. It's based on proposal 1 in [this design doc](https://docs.google.com/document/d/1V002gl5THZY8_pZyfj02gimwv85aSdReTRjMm3GenaY/edit#heading=h.5sybau7waq2q). This PR:

- Updates ring.proto to support two new fields in `InstanceDesc`:
  - `read_only`
  - `read_only_updated_timestamp`
- Updates the lifecycler to support reading/writing the new fields
- Updates ring_http to expose the new fields

To keep the scope of this PR reasonable, a followup PR will add support for read only ingesters when computing shuffle shards.

Fixes #<issue number>

**Checklist**
- [x] Tests updated


ring/ring_http.go
```diff
@@ -36,15 +36,17 @@ type httpResponse struct {
...
-	Tokens              []uint32  `json:"tokens"`
-	NumTokens           int       `json:"-"`
-	Ownership           float64   `json:"-"`
+	ID                       string    `json:"id"`
+	State                    string    `json:"state"`
+	Address                  string    `json:"address"`
+	HeartbeatTimestamp       time.Time `json:"timestamp"`
+	RegisteredTimestamp      time.Time `json:"registered_timestamp"`
+	ReadOnly                 bool      `json:"readonly"`
+	ReadOnlyUpdatedTimestamp time.Time `json:"readonly_updated_timestamp"`
```
pracucci, line 45: [nit, non blocking] Have you considered `read_only*` (separated by `_`) to keep it consistent with the protobuf message definition?

ring/ring.proto
```diff
@@ -44,6 +44,16 @@ message InstanceDesc {
 
 	// ID of the instance. This value is the same as the key in the ingesters map in Desc.
 	string id = 9;
+
+	// Timestamp when the read_only flag was updated. This is used to find other instances that could have possibly
+	// owned a specific token in the past on the write path, due to *this* instance being read-only.
+	// This value should only increase.
+	int64 read_only_updated_timestamp = 10;
```
pracucci, line 51: [nit] Can you clarify the unit (e.g. unix timestamp with seconds precision) in the field doc, please?

test/poll.go
```diff
@@ -6,21 +6,36 @@ import (
 	"time"
 )
 
-// Poll repeatedly calls a function until the function returns the correct response or until poll timeout.
-func Poll(t testing.TB, d time.Duration, want interface{}, have func() interface{}) {
+// Eventually repeatedly calls the fn until it returns true or until the duration is exceeded.
+// The fn will be called up to 100 times.
+func Eventually(t testing.TB, duration time.Duration, fn func() bool) {
```
pracucci, line 11: [nit] I don't fully understand this. Don't you get the same with `require.Eventually()`?

ring/model.go
```diff
@@ -45,26 +45,35 @@ func NewDesc() *Desc {
...
 
-	ingester := InstanceDesc{
-		Id:                  id,
-		Addr:                addr,
-		Timestamp:           time.Now().Unix(),
-		RegisteredTimestamp: registeredTimestamp,
-		State:               state,
-		Tokens:              tokens,
-		Zone:                zone,
+func (d *Desc) AddInstance(id string, ingester InstanceDesc) InstanceDesc {
```
pracucci, line 74: [nit]
```suggestion
func (d *Desc) AddInstance(id string, instance InstanceDesc) InstanceDesc {
```

ring/model.go
```diff
@@ -142,6 +151,28 @@ func (i *InstanceDesc) GetRegisteredAt() time.Time {
 	return time.Unix(i.RegisteredTimestamp, 0)
 }
 
+func (i *InstanceDesc) SetRegisteredAt(t time.Time) {
+	if t.IsZero() {
```
pracucci, line 155: [nit] Can you use `timeToUnixSecons` instead?

ring/model.go
```diff
@@ -600,6 +631,14 @@ func (d *Desc) RingCompare(o *Desc) CompareResult {
 			return Different
 		}
 
+		if ing.ReadOnly != oing.ReadOnly {
```
pracucci, line 634: [nit] Please add test cases to `TestDesc_RingsCompare`.

ring/model.go
```diff
@@ -142,6 +151,28 @@ func (i *InstanceDesc) GetRegisteredAt() time.Time {
 	return time.Unix(i.RegisteredTimestamp, 0)
 }
 
+func (i *InstanceDesc) SetRegisteredAt(t time.Time) {
```
pracucci, line 154: Why is this change required? The registered timestamp should never change after an instance is added to the ring. Having an exported function to change the setting looks risky to me.

ring/basic_lifecycler.go
```diff
@@ -316,7 +316,8 @@ func (l *BasicLifecycler) registerInstance(ctx context.Context) error {
 		// Always overwrite the instance in the ring (even if already exists) because some properties
 		// may have changed (stated, tokens, zone, address) and even if they didn't the heartbeat at
 		// least did.
-		instanceDesc = ringDesc.AddIngester(l.cfg.ID, l.cfg.Addr, l.cfg.Zone, tokens, state, registeredAt)
+		// TODO: add support for read-only state and timestamp.
```
pracucci, line 319: Do you plan to address these TODOs soon? If no, I would rather update the `BasicLifecycler struct` go doc to mention that this lifecycler doesn't support read-only state for now.

ring/lifecycler.go
```diff
@@ -631,30 +677,37 @@ func (i *Lifecycler) initRing(ctx context.Context) error {
 		if !ok {
 			// The instance doesn't exist in the ring, so it's safe to set the registered timestamp
```
pracucci, line 678: [nit] Can you move this comment 1 line below, please?

ring/model.go
pracucci, line 74: Why is this function required? The benefit of having only `AddIngester()` (despite the "ingester" naming) is that it makes it very explicit all fields to set, but this function hides it (it allows you to set a pre-populated `InstanceDesc`).

ring/basic_lifecycler_test.go
```diff
@@ -102,6 +102,19 @@ func TestBasicLifecycler_RegisterOnStart(t *testing.T) {
 			registerState:  ACTIVE,
 			registerTokens: Tokens{1, 2, 3, 4, 5},
 		},
+		"initial ring contains read only instance": {
+			initialInstanceID: testInstanceID,
+			initialInstanceDesc: &InstanceDesc{
+				Addr:                     "1.1.1.1",
+				State:                    ACTIVE,
+				Tokens:                   Tokens{6, 7, 8, 9, 10},
```
pracucci, line 110: I find a bit misleading the different tokens here. I don't think it matters for this specific test. We just want to make sure the instance will still be in RO state, isn't it?

pracucci (approved):
Nice work, LGTM! I left few minor comments I would be glad if you could take a look before merging.

ring/model.go
jhalterman: I don't think this is needed anymore. Removing.

ring/basic_lifecycler_test.go
jhalterman: Yes that's right. Updated.

ring/model.go
jhalterman: I agree, it's not needed.

test/poll.go
jhalterman: The behavior is the same, the only difference is `require.Eventually` requires a `tick` time argument whereas `test.Poll` just uses `duration / 100`. I thought it would be nice to use something more like `Eventually`, but keeping the tick time. Not a big deal either way.

ring/basic_lifecycler.go
jhalterman: Converted these to a doc comment as suggested.

PR #554, opened 2024-08-02
Add read only support for computing shuffle shards

jhalterman:
**What this PR does**:

This is PR 2 of 2 (following #553) to add read only support for ingesters. This PR integrates read only ingester support with shuffle shard computation. 

On the write path, read only ingesters are excluded from shuffle shards. On the read path, ingesters with a read only timestamp within the lookback period are included in addition to another ingesters, similar to how recently registered ingesters are handled.

This PR also updates shuffle shard caching with lookback to consider the read only timestamp.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -706,6 +706,11 @@ func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
 // This function supports caching, but the cache will only be effective if successive calls for the
 // same identifier are with the same lookbackPeriod and increasing values of now.
 func (r *Ring) ShuffleShardWithLookback(identifier string, size int, lookbackPeriod time.Duration, now time.Time) ReadRing {
+	// Nothing to do if the shard size is not smaller than the actual ring.
+	if lookbackPeriod > 0 && (size <= 0 || r.InstancesCount() <= size) {
+		return r
+	}
+
 	if cached := r.getCachedShuffledSubringWithLookback(identifier, size, lookbackPeriod, now); cached != nil {
 		return cached
 	}
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -35,6 +35,7 @@ const (
 )
 
 // ReadRing represents the read interface to the ring.
+// Support for read-only instances requires use of ShuffleShard or ShuffleShardWithLookback prior to getting a ReplicationSet.
 type ReadRing interface {
 	// Get returns n (or more) instances which form the replicas for the given key.
 	// bufDescs, bufHosts and bufZones are slices to be overwritten for the return value
@@ -166,6 +167,7 @@ type instanceInfo struct {
 }
 
 // Ring is a Service that maintains an in-memory copy of a ring and watches for changes.
+// Support for read-only instances requires use of ShuffleShard or ShuffleShardWithLookback prior to getting a ReplicationSet.
 type Ring struct {
 	services.Service
 
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -819,7 +819,7 @@ func (r *Ring) shuffleShard(identifier string, size int, lookbackPeriod time.Dur
 				instanceID := info.InstanceID
 				instance := r.ringDesc.Ingesters[instanceID]
 
-				// On write path (lookbackPeriod == 0), read only instances are excluded.
+				// The lookbackPeriod is 0 when this function is called by ShuffleShard(). In this case, we want read only instances excluded.
 				if lookbackPeriod == 0 && instance.ReadOnly {
 					continue
 				}
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -678,6 +678,8 @@ func (r *Ring) updateRingMetrics(compareResult CompareResult) {
 //
 // - Shuffling: probabilistically, for a large enough cluster each identifier gets a different
 // set of instances, with a reduced number of overlapping instances between two identifiers.
+//
+// Subring returned by this method does not contain instances that have read-only field set.
 func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
 	// Use all instances if shuffle sharding is disabled, or it covers all instances anyway.
 	// Reason for not returning entire ring directly is that we need to filter out read-only instances.
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -680,7 +680,7 @@ func (r *Ring) updateRingMetrics(compareResult CompareResult) {
 // set of instances, with a reduced number of overlapping instances between two identifiers.
 func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
 	// Use all instances if shuffle sharding is disabled, or it covers all instances anyway.
-	// Reason is that we need to filter out read-only instances.
+	// Reason for not returning entire ring directly is that we need to filter out read-only instances.
 	instances := r.InstancesCount()
 	if size <= 0 || instances <= size {
 		size = instances
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -722,11 +722,6 @@ func (r *Ring) ShuffleShardWithLookback(identifier string, size int, lookbackPer
 func (r *Ring) shuffleShard(identifier string, size int, lookbackPeriod time.Duration, now time.Time) *Ring {
 	lookbackUntil := now.Add(-lookbackPeriod).Unix()
 
-	includeReadOnly := false
-	if lookbackPeriod > 0 {
-		includeReadOnly = true
-	}
-
 	r.mtx.RLock()
 	defer r.mtx.RUnlock()
 
@@ -789,8 +784,8 @@ func (r *Ring) shuffleShard(identifier string, size int, lookbackPeriod time.Dur
 				instanceID := info.InstanceID
 				instance := r.ringDesc.Ingesters[instanceID]
 
-				// If the instance is read only, do not include it in the shard.
-				if instance.ReadOnly && !includeReadOnly {
+				// If the instance is read only without a lookback, do not include it in the shard.
+				if lookbackPeriod == 0 && instance.ReadOnly {
 					continue
 				}
 
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -812,12 +812,12 @@ func (r *Ring) shuffleShard(identifier string, size int, lookbackPeriod time.Dur
 					continue
 				}
 
-				// If the lookback is enabled, and this instance has switched its read-only state within the lookback period,
-				// then we should include it in the subring, but continue selecting more instances.
+				// If the lookback is enabled, and this instance is read only or has switched its read-only state
+				// within the lookback period, then we should include it in the subring, but continue selecting more instances.
 				//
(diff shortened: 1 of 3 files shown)


ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -48,15 +48,15 @@ func (r *partitionRingShuffleShardCache) setSubringWithLookback(identifier strin
...
+		lookbackWindowStart       = now.Add(-lookbackPeriod).Unix()
+		validForLookbackWindowEnd = int64(math.MaxInt64)
 	)
 
 	for _, partition := range subring.desc.Partitions {
 		stateChangedDuringLookbackWindow := partition.StateTimestamp >= lookbackWindowStart
 
-		if stateChangedDuringLookbackWindow && partition.StateTimestamp < validForLookbackWindowsStartingBefore {
-			validForLookbackWindowsStartingBefore = partition.StateTimestamp
+		if stateChangedDuringLookbackWindow && partition.StateTimestamp < validForLookbackWindowEnd {
```
jhalterman, line 58: Do we want to support read only timestamps here as well? /cc @pracucci

ring/ring.go
```diff
@@ -705,11 +707,6 @@ func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
 // This function supports caching, but the cache will only be effective if successive calls for the
 // same identifier are with the same lookbackPeriod and increasing values of now.
 func (r *Ring) ShuffleShardWithLookback(identifier string, size int, lookbackPeriod time.Duration, now time.Time) ReadRing {
-	// Nothing to do if the shard size is not smaller then the actual ring.
-	if size <= 0 || r.InstancesCount() <= size {
```
pracucci, line 709: It's pity we lose this optimization. What can we do here?

pracucci: But why was removed in the first place? If shuffle sharding is disabled, we should always returns the full ring, isn't it?

ring/ring.go
```diff
@@ -224,9 +224,9 @@ type subringCacheKey struct {
 }
 
 type cachedSubringWithLookback[R any] struct {
-	subring                               R
-	validForLookbackWindowsStartingAfter  int64 // if the lookback window is from T to S, validForLookbackWindowsStartingAfter is the earliest value of T this cache entry is valid for
-	validForLookbackWindowsStartingBefore int64 // if the lookback window is from T to S, validForLookbackWindowsStartingBefore is the latest value of T this cache entry is valid for
+	subring                  R
+	validLookbackWindowStart int64 // if the lookback window is from T to S, validLookbackWindowStart is the earliest value of T this cache entry is valid for
+	validLookbackWindowEnd   int64 // if the lookback window is from T to S, validLookbackWindowEnd is the latest value of T this cache entry is valid for
```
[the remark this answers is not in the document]
jhalterman: It's probably better to address renaming in a separate PR. Reverting this.

ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -48,15 +48,15 @@ func (r *partitionRingShuffleShardCache) setSubringWithLookback(identifier strin
 	}
 
 	var (
-		lookbackWindowStart                   = now.Add(-lookbackPeriod).Unix()
-		validForLookbackWindowsStartingBefore = int64(math.MaxInt64)
+		lookbackWindowStart       = now.Add(-lookbackPeriod).Unix()
+		validForLookbackWindowEnd = int64(math.MaxInt64)
```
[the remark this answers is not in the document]
jhalterman: This was just to be consistent with the instances ring renaming, but I'll revert both.

ring/ring.go
```diff
@@ -726,19 +723,14 @@ func (r *Ring) ShuffleShardWithLookback(identifier string, size int, lookbackPer
 func (r *Ring) shuffleShard(identifier string, size int, lookbackPeriod time.Duration, now time.Time) *Ring {
 	lookbackUntil := now.Add(-lookbackPeriod).Unix()
 
+	includeReadOnly := false
+	if lookbackPeriod > 0 {
+		includeReadOnly = true
+	}
```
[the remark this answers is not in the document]
jhalterman: The determines whether we include an instance in the shard on the write path. I think the code is more consistent below without this variable though, so I removed it. Let me know what you think.

ring/ring.go
```diff
@@ -726,19 +723,14 @@ func (r *Ring) ShuffleShardWithLookback(identifier string, size int, lookbackPer
...
 	r.mtx.RLock()
 	defer r.mtx.RUnlock()
 
-	// If all instances have RegisteredTimestamp within the lookback period,
-	// then all instances would be included in the resulting ring, so we can
-	// simply return this ring.
-	//
-	// If any instance had RegisteredTimestamp equal to 0 (it would not cause additional lookup of next instance),
-	// then r.oldestRegisteredTimestamp is zero too, and we skip this optimization.
-	if lookbackPeriod > 0 && r.oldestRegisteredTimestamp > 0 && r.oldestRegisteredTimestamp >= lookbackUntil {
```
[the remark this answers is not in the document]
jhalterman: This optimization was removed early on, but it looks like it can be retained. I restored it.

ring/ring.go
jhalterman: I think this optimization can be restored as well. Thanks for raising this.

jhalterman:
I added the mentioned test cases and also ran the benchmarks. The resulting ns/op were nearly the same (< 1% difference) before and after the change, for all the `BenchmarkRing_ShuffleShard*` tests.

ring/ring.go
pstibrany: `ShuffleShardWithLookback` now filters out read-only instances with `ReadOnlyUpdatedTimestamp` outside of lookback period. If we return entire ring, read-only instances will remain in the ring.

Filtering out read-only instances outside of lookback window is an optimization and not strictly required. Returning entire ring is a valid choice.

I think we should make code consistent:
* either remove this shortcut, and keep filtering of read-only instances with state change outside of lookback window
* or remove the filtering (which I added yesterday)

I don't have strong opinion, and I'm fine if we remove the filtering, and always return read-only instances (but still extend the shard!) when lookback period is > 0.

ring/ring_test.go
```diff
@@ -1632,6 +1660,19 @@ func TestRing_ShuffleShard_Consistency(t *testing.T) {
...
+			case enableReadOnly:
+				for id, desc := range ringDesc.Ingesters {
+					desc.ReadOnly = true
+					desc.ReadOnlyUpdatedTimestamp = time.Now().Unix()
+					ringDesc.Ingesters[id] = desc
+					break
+				}
+			case disableReadOnly:
+				for id, desc := range ringDesc.Ingesters {
+					desc.ReadOnly = false
```
pstibrany, line 1672: Update of `ReadOnly` field should be done together with update of `ReadOnlyUpdatedTimestamp` field.

ring/ring.go
pstibrany: I'll go ahead and remove the filtering mentioned in my comment.

ring/ring_test.go
pstibrany: done in https://github.com/grafana/dskit/pull/554/commits/5d2ced6ec53289458d86768113f93fa9194b83b9.

ring/ring.go
pstibrany: Done in https://github.com/grafana/dskit/pull/554/commits/a29a1ea7035791d19113d2638b5fb4c55090fcc4.

ring/ring.go
```diff
@@ -677,9 +679,11 @@ func (r *Ring) updateRingMetrics(compareResult CompareResult) {
 // - Shuffling: probabilistically, for a large enough cluster each identifier gets a different
 // set of instances, with a reduced number of overlapping instances between two identifiers.
 func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
```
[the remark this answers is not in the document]
pstibrany: Done in https://github.com/grafana/dskit/pull/554/commits/e2d1a8d639d7dc38b13d4dd627a53dca2bf9ed3e.

ring/ring_test.go
```diff
@@ -2234,82 +2342,99 @@ func TestRing_ShuffleShardWithLookback_CorrectnessWithFuzzy(t *testing.T) {
...
+										break
 									}
-
-									ringState.Instances = append(ringState.Instances[:idx], ringState.Instances[idx+1:]...)
-									history[ringTime] = ringState
-									break
 								}
 							}
+
+							// Removed instance can't be read-only.
```
pracucci, line 2381: Uhm... I don't understand this comment. Is it correct?

pstibrany: It is. "Removed instance" was already removed from the ring, but we also need to remove it from the `readOnlyInstances` map. Previous version of the test forgot to do that, and then test could try to switch such instance back to "read write" mode, which re-created entry in the ring with no tokens.

ring/ring.go
pstibrany: > @pracucci voiced concern about NOT filtering out such instances:
> 
> > Isn't there the risk of failing quorum, due to race conditions with shutdown? Like shutdown a read-only instance that was just returned by `ShuffleShardWithLookback()`
> 
> And also pointed out that we _do_ filter out read-only partitions on shuffle shard with lookback.

Let's address this in the next PR.

ring/ring.go
```diff
@@ -720,8 +726,8 @@ func (r *Ring) ShuffleShard(identifier string, size int) ReadRing {
 // This function supports caching, but the cache will only be effective if successive calls for the
 // same identifier are with the same lookbackPeriod and increasing values of now.
 func (r *Ring) ShuffleShardWithLookback(identifier string, size int, lookbackPeriod time.Duration, now time.Time) ReadRing {
-	// Nothing to do if the shard size is not smaller then the actual ring.
-	if size <= 0 || r.InstancesCount() <= size {
+	// Nothing to do if the shard size is not smaller than the actual ring.
+	if lookbackPeriod > 0 && (size <= 0 || r.InstancesCount() <= size) {
```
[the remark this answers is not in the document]
jhalterman: Yea, this was a safety check just in case someone calls ShuffleShardWithLookback with a lookbackPeriod == 0, but I'll remove.

ring/ring_test.go
```diff
@@ -1714,6 +1742,20 @@ func TestRing_ShuffleShard_Consistency(t *testing.T) {
...
 					break
 				}
+			case enableReadOnly:
+				for id, desc := range ringDesc.Ingesters {
+					desc.ReadOnly = true
+					desc.ReadOnlyUpdatedTimestamp = time.Now().Unix()
+					ringDesc.Ingesters[id] = desc
+					break
+				}
+			case disableReadOnly:
```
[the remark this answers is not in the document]
jhalterman: Updated the test to mark some instances as read only initially.

PR #570, opened 2024-08-19
Upgrade to golangci-lint v1.60.1

aknuds1:
**What this PR does**:
Upgrade to golangci-lint v1.60.1, in which `staticcheck` forbids calling `printf` style functions without a formatting string/arguments.

**Checklist**
- [x] Tests updated
- [NA] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/ring/replication_set.go
+++ b/ring/replication_set.go
@@ -316,7 +316,7 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
 			ext.Error.Set(cfg.Logger.Span, true)
 		}
 
-		contextTracker.cancelAllContexts(cancellation.NewErrorf(cause))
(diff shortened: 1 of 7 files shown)


narqo (approved):
🔥

ring/replication_set.go
```diff
@@ -316,7 +316,7 @@ func DoUntilQuorumWithoutSuccessfulContextCancellation[T any](ctx context.Contex
 			ext.Error.Set(cfg.Logger.Span, true)
 		}
 
-		contextTracker.cancelAllContexts(cancellation.NewErrorf(cause))
```
colega, line 319: I can't see why the new version is better.

aknuds1: It's generally because the linter thinks the non-formatting version should be used, if there's no string formatting.

PR #571, opened 2024-08-21
Support gzipped runtime configurations

colega:
Some runtime configs can be too large to fit in one kubernetes configmap. However, Kubernetes configmaps accept binary data, so we can gzip the yaml, which offers up to 10x compression.

This would decompress using gzip runtime config files that have the `.gz` suffix.


pstibrany:
I'd suggest to remove the config option, and simply check for filename suffix (`.gz`).

pstibrany (approved):
lgtm, thanks.

PR #573, opened 2024-08-23
Add lifecycler_read_only metric to observe read-only state of the lifecycler.

pstibrany:
**What this PR does**:

This PR adds `lifecycler_read_only` metric to observe read-only state of the Lifecycler.

This PR also changes `ReadOnlyUpdatedTimestamp` from current time to zero when initializing ring entry by lifecycler in case that lifecycler's entry doesn't exist in the ring yet. This is cosmetic change for testing and making Ring HTML page simpler by not including too many timestamps.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -415,6 +415,11 @@ func (i *Lifecycler) setReadOnlyState(readOnly bool, readOnlyLastUpdated time.Ti
 	defer i.stateMtx.Unlock()
 	i.readOnly = readOnly
 	i.readOnlyLastUpdated = readOnlyLastUpdated
+	if readOnly {
+		i.lifecyclerMetrics.readonly.Set(1)
+	} else {
+		i.lifecyclerMetrics.readonly.Set(0)
+	}
 }
 
 // ClaimTokensFor takes all the tokens for the supplied ingester and assigns them to this ingester.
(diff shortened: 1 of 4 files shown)


jhalterman (approved):

ring/lifecycler.go
```diff
@@ -415,6 +415,11 @@ func (i *Lifecycler) setReadOnlyState(readOnly bool, readOnlyLastUpdated time.Ti
 	defer i.stateMtx.Unlock()
 	i.readOnly = readOnly
 	i.readOnlyLastUpdated = readOnlyLastUpdated
+	if readOnly {
+		i.lifecyclerMetrics.readonly.Set(1)
+	} else {
+		i.lifecyclerMetrics.readonly.Set(0)
```
pr00se, line 421: What if rather than exporting a `0` value (which will be the case the majority of the time, and always for lifecyclers not using read-only state) we just stop exporting the metric entirely when not in read-only state?

pstibrany: You are right that this will be mostly 0, but I don't think removing the metric when it's 0 is a good practice. It makes it difficult to see the difference between metric being 0 vs metric not working correctly for some reason. It's also easier to work with metric that exists all the time vs metric that can be absent.

PR #565, opened 2024-08-15
Filter out read-only instances, but skip filtering when it's not required.

pstibrany:
**What this PR does**:

This PR updates shuffle-sharding to 1) filter out read-only instances on shuffle shard with lookback, if instance changed its state before lookback window, 2) avoid computing shuffle shard if no filtering is required.

Benchmark compared to `main` (updated after https://github.com/grafana/dskit/pull/565/commits/95e95648d2c85553387565584c8e394ec79eaf0f):

```
benchstat main-new-bench.txt final2.txt
goos: darwin
goarch: arm64
pkg: github.com/grafana/dskit/ring
                                                                                  │ main-new-bench.txt │              final2.txt              │
                                                                                  │       sec/op       │    sec/op     vs base                │
Ring_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_0-10                657443.00n ± ∞ ¹   59.71n ± ∞ ¹   -99.99% (p=0.008 n=5)
Ring_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_3-10                    19.83µ ± ∞ ¹   19.34µ ± ∞ ¹         ~ (p=0.151 n=5)
Ring_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_10-10                   53.21µ ± ∞ ¹   53.59µ ± ∞ ¹         ~ (p=0.841 n=5)
Ring_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_30-10                   163.3µ ± ∞ ¹   163.5µ ± ∞ ¹         ~ (p=0.310 n=5)
Ring_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_100-10                  651.8µ ± ∞ ¹   272.4µ ± ∞ ¹   -58.21% (p=0.008 n=5)
Ring_ShuffleShard/num_instances_=_50,_num_zones_=_1,_shard_size_=_1000-10                 652.2µ ± ∞ ¹   271.9µ ± ∞ ¹   -58.31% (p=0.008 n=5)
Ring_ShuffleShard/num_instances_=_50,_num_zones_=_3,_shard_size_=_0-10                680198.00n ± ∞ ¹   58.80n ± ∞ ¹   -99.99% (p=0.008 n=5)
[...]

PR #575, opened 2024-08-23
update-ring

pstibrany:
**What this PR does**:

This PR extracts code for setting internal `*Ring` state from `*Desc`, and reuses the method from unit tests. There is no visible change to ring code api.

This was raised during recent reviews, unfortunately I can't find the comment anymore (I thought it came from https://github.com/grafana/dskit/pull/565, but it's not there).

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -321,7 +321,7 @@ func (r *Ring) starting(ctx context.Context) error {
 func (r *Ring) loop(ctx context.Context) error {
 	// Update the ring metrics at start of the main loop.
 	r.mtx.Lock()
-	r.updateRingMetrics(Different)
+	r.updateRingMetrics()
 	r.mtx.Unlock()
 
 	r.KVClient.WatchKey(ctx, r.key, func(value interface{}) bool {
@@ -362,7 +362,9 @@ func (r *Ring) updateRingState(ringDesc *Desc) {
 		// when watching the ring for updates).
 		r.mtx.Lock()
 		r.ringDesc = ringDesc
-		r.updateRingMetrics(rc)
+		if rc != Equal {
+			r.updateRingMetrics()
+		}
 		r.mtx.Unlock()
 		return
 	}
@@ -405,7 +407,7 @@ func (r *Ring) updateRingState(ringDesc *Desc) {
 		r.shuffledSubringWithLookbackCache = make(map[subringCacheKey]cachedSubringWithLookback[*Ring])
 	}
 
-	r.updateRingMetrics(rc)
+	r.updateRingMetrics()
 }
 
 // Get returns n (or more) instances which form the replicas for the given key.
@@ -645,11 +647,7 @@ func (r *Desc) CountTokens() map[string]int64 {
 }
 
 // updateRingMetrics updates ring metrics. Caller must be holding the Write lock!
-func (r *Ring) updateRingMetrics(compareResult CompareResult) {
-	if compareResult == Equal {
-		return
-	}
-
+func (r *Ring) updateRingMetrics() {
 	numByState := map[string]int{}
 	oldestTimestampByState := map[string]int64{}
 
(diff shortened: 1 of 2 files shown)


ring/ring.go
```diff
@@ -677,10 +685,6 @@ func (r *Ring) updateRingMetrics(compareResult CompareResult) {
 		r.oldestTimestampGaugeVec.WithLabelValues(state).Set(float64(timestamp))
 	}
 
-	if compareResult == EqualButStatesAndTimestamps {
```
pstibrany, line 680: This check used to be useful when we did more updates in `updateRingMetrics` after this point, but now we only update single gauge, so it seems useless.

aknuds1:
I don't understand why linting fails in Drone. Doesn't happen on my machine. Any idea @pstibrany?

ring/ring.go
```diff
@@ -362,11 +362,17 @@ func (r *Ring) updateRingState(ringDesc *Desc) {
...
 		r.ringDesc = ringDesc
-		r.updateRingMetrics(rc)
+		if rc != Equal {
+			r.updateRingMetrics()
+		}
 		r.mtx.Unlock()
 		return
 	}
 
+	r.setRingStateFromDesc(ringDesc, true, false, true)
```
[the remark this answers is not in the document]
pstibrany: Oh, great catch. I introduced that by mistake by using refactor tool in my IDE, which automatically added default value.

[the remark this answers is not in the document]
aknuds1: Thanks for fixing!

ring/ring_test.go
```diff
@@ -189,12 +201,9 @@ func TestDoBatchZeroInstances(t *testing.T) {
...
 	}
 	desc := NewDesc()
-	r := Ring{
-		cfg:      Config{},
-		ringDesc: desc,
-		strategy: NewDefaultReplicationStrategy(),
-	}
-	require.Error(t, DoBatch(ctx, Write, &r, keys, callback, cleanup))
+	r := newRingForTesting(Config{}, false)
+	r.setRingStateFromDesc(desc, false, false, true)
```
aknuds1, line 205: Why true for `updateReadOnlyInstances`?

ring/ring_test.go
```diff
@@ -208,26 +217,16 @@ func TestDoBatchWithOptionsContextCancellation(t *testing.T) {
...
-		strategy:              NewDefaultReplicationStrategy(),
-		lastTopologyChange:    time.Now(),
-	}
+	r := newRingForTesting(Config{
+		HeartbeatTimeout:     time.Hour,
+		ZoneAwarenessEnabled: true,
+		SubringCacheDisabled: true,
+		ReplicationFactor:    numZones,
+	}, true)
+	r.setRingStateFromDesc(desc, false, false, true)
```
aknuds1, line 226: Why true for `updateReadOnlyInstances`?

aknuds1 (commented):
Generally LGTM, but there are some cases where I don't understand why you pass `true` for `updateReadOnlyInstances`.

ring/ring_test.go
pstibrany: That seems unnecessary indeed, and not what previous code was doing. I'll remove it.

ring/ring_test.go
pstibrany: Also here the code was not doing that before. Let me recheck all places again. I suspect I changed my mind about it mid-work (don't remember), so it's not consistent now.

aknuds1: @pstibrany yes, there were some more call sites, but I wanted to hear from you first :D

aknuds1 (approved):
Nice cleanup!

pstibrany:
> Nice cleanup!

Thank you for your review!

ring/ring.go
```diff
@@ -362,11 +362,17 @@ func (r *Ring) updateRingState(ringDesc *Desc) {
...
+			r.updateRingMetrics()
+		}
 		r.mtx.Unlock()
 		return
 	}
 
+	r.setRingStateFromDesc(ringDesc, true, true, true)
+}
+
+func (r *Ring) setRingStateFromDesc(ringDesc *Desc, updateMetrics, updateRegisteredTimestampCache, updateReadOnlyInstances bool) {
```
pracucci, line 375: [nit] Would be nice to comment that this function should be called with lock held.

pstibrany: That would be incorrect. This function takes the lock on its own, after computing new values, but before setting them.

PR #577, opened 2024-08-26
Report context cancel cause in DoBatchWithOptions instead of ctx.Err().

### LINKED ISSUE #576: Report context cancel cause in `DoBatchWithOptions` instead of `ctx.Err()`.
In https://github.com/grafana/tempo/issues/3957 it was found that analyzing the root cause of `context cancelled` logs can be quite hard to analyze. In the [discussion](https://github.com/grafana/tempo/issues/3957#issuecomment-2286941935) on how to improve the situation, it was found that reporting the cancel cause, introduced by [`WithCancelCause`](https://pkg.go.dev/context#WithCancelCause) in go 1.20, would be more helpful in these situations. 

Note that this won't change the behavior of error reporting when `WithCancel` (without cause) is used.

dastrobu:
**What this PR does**:

Report context cancel cause in `DoBatchWithOptions` instead of `ctx.Err()`.

Closes #576

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated - the order ox entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 
 * [CHANGE] Update the gRPC dependency to v1.66.0 and deprecate the `grpc_server_recv_buffer_pools_enabled` option that is no longer supported by it. #580
+* [CHANGE] `ring.DoBatchWithOptions` (and `ring.DoBatch`) reports the cancelation cause when the context is canceled instead of `context.Canceled`.
 * [CHANGE] Multierror: Implement `Unwrap() []error`. This allows to use `multierror.MultiError` with both `errors.Is()` and `errors.As()`. Previously implemented `Is(error) bool` has been removed. #522
 * [CHANGE] Add a new `grpc_server_recv_buffer_pools_enabled` option that enables recv buffer pools in the gRPC server (assuming `grpc_server_stats_tracking_enabled` is disabled). #510
 * [CHANGE] Add a new `grpc_server_stats_tracking_enabled` option that allows us to disable stats tracking and potentially improve server memory reuse. #507
--- a/ring/batch.go
+++ b/ring/batch.go
@@ -131,7 +131,7 @@ func DoBatchWithOptions(ctx context.Context, op Operation, r DoBatchRing, keys [
 		// Get call below takes ~1 microsecond for ~500 instances.
 		// Checking every 10K calls would be every 10ms.
 		if i%10e3 == 0 {
-			if err := ctx.Err(); err != nil {
+			if err := context.Cause(ctx); err != nil {
 				o.Cleanup()
 				return err
 			}
@@ -161,7 +161,7 @@ func DoBatchWithOptions(ctx context.Context, op Operation, r DoBatchRing, keys [
 	}
 
 	// One last check before calling the callbacks: it doesn't make sense if context is canceled.
-	if err := ctx.Err(); err != nil {
+	if err := context.Cause(ctx); err != nil {
 		o.Cleanup()
 		return err
 	}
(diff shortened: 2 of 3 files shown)


joe-elliott:
I'm good on this change. It seems straightforward and propagates more info (if available) back to the caller. The ring is very specialized and highly critical code. Let me try to summon an expert to confirm this won't do anything unexpected.

joe-elliott (dismissed):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Changelog
 
 * [CHANGE] Update the gRPC dependency to v1.66.0 and deprecate the `grpc_server_recv_buffer_pools_enabled` option that is no longer supported by it. #580
+* [CHANGE] `ring.DoBatchWithOptions` (and `ring.DoBatch`) reports the cancellation cause when the context is cancelled instead of `context.Cancelled`.
```
pstibrany, line 4: (To follow spelling used in Go)

```suggestion
* [CHANGE] `ring.DoBatchWithOptions` (and `ring.DoBatch`) reports the cancelation cause when the context is canceled instead of `context.Canceled`. #577 
```

pstibrany (commented):
Cannot this introduce a subtle bugs in code that already uses cancelation with cause, but calls to `DoBatchWithOptions` do expect standard context errors to be returned?

WDYT about adding an option to `DoBatchOptions` struct to enable return of `context.Cause(ctx)` errors instead?

dastrobu:
> Cannot this introduce a subtle bugs in code that already uses cancelation with cause, but calls to `DoBatchWithOptions` do expect standard context errors to be returned?

I think it is very unlikely, but definitely possible. I don't have a good idea right now why one would do that.

> WDYT about adding an option to `DoBatchOptions` struct to enable return of `context.Cause(ctx)` errors instead?

It is a viable option. It adds some complexity and makes it less likely that callers of `DoBatchOptions` will adopt the new (preferred) error policy, as adding an extra flag is required.

If you prefer the safer but more complex option, I can adapt the PR accordingly.

pstibrany:
> > Cannot this introduce a subtle bugs in code that already uses cancelation with cause, but calls to `DoBatchWithOptions` do expect standard context errors to be returned?
> 
> I think it is very unlikely, but definitely possible. I don't have a good idea right now why one would do that.

I can imagine this scenario very easily, with different layers having different expectations. I don't see a good way to find this in practice, other than just making the change and testing it.

> > WDYT about adding an option to `DoBatchOptions` struct to enable return of `context.Cause(ctx)` errors instead?
> 
> It is a viable option. It adds some complexity and makes it less likely that callers of `DoBatchOptions` will adopt the new (preferred) error policy, as adding an extra flag is required.

We could also make the flag to be opt-out -- enable the new behaviour by default, but provide way to disable it in case it proves troublesome. (It is a possibly breaking change after all).

I would ask @colega for his opinion too, since he introduced context checks into this code.

dastrobu:
> We could also make the flag to be opt-out -- enable the new behaviour by default, but provide way to disable it in case it proves troublesome. 

I like the idea. It would be a good compromise between moving forward and reducing risk. 
If this is consensus, I could add something like `ReportContextErrInsteadOfCause` to the `DoBatchOptions`.

colega:
As far as I understand context cancellations with cause, this LGTM.

I think adding the option would only make it harder to reason about in the future (as we all know nobody would remove that option).

colega (approved):

charleskorn:
The `cancellation.NewError` / `cancellation.NewErrorf` functions might be helpful here: they're intended to be used as the cancellation cause and satisfy `errors.Is(cause, context.Canceled)`.

pstibrany (approved):
Thank you!

PR #580, opened 2024-08-29
Update the gRPC dependency to v1.66.0 and deprecate `RecvBufferPools`

na--:
**What this PR does**:

It updates the `google.golang.org/grpc` dependency to the latest stable v1.66.0 version. It also removes an experimental `RecvBufferPool` feature that I previously added in https://github.com/grafana/dskit/pull/510 because that was deprecated in the same version and now results in a compilation error. The good news is that it got replaced (https://github.com/grpc/grpc-go/pull/7356) with something even better that's turned on by default, so we no longer need that config at all.

**Checklist**
- [x] Tests updated


.github/workflows/test-build.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       matrix:
         # We want to make sure dskit can support multiple golang versions
         # by ensuring the test would pass using all these supported versions.
-        go-version: ['1.21.x', '1.22.x']
+        go-version: ['1.21.x', '1.22.x', '1.23.x']
```
pstibrany, line 44: Is this required? If not, let's do that in separate PR.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/alicebob/miniredis v2.5.0+incompatible
 	github.com/armon/go-metrics v0.3.10
 	github.com/aws/aws-sdk-go v1.44.321
-	github.com/cespare/xxhash/v2 v2.2.0
+	github.com/cespare/xxhash/v2 v2.3.0
```
pstibrany, line 10: All the module updates seem unrelated?

pstibrany: They come from https://github.com/grpc/grpc-go/blob/v1.66.0/go.mod.

pstibrany (approved):
Thank you.

.github/workflows/test-build.yml
na--: Somewhat required, Loki already uses Go 1.23. A separate PR would have been better, but I don't want to wait another 25 minutes for the CI to run :sweat_smile:

PR #584, opened 2024-09-20
Alerting: adapt sum of gauges to use options

titolins:
To be able to filter 0 value gauges per tenant, we need to adapt the helper as it was done already before for other metrics.

Trying to follow the name convention, but not sure if we should introduce a new function instead. Despite being simple to adapt, it would require changes

### DIFF
--- a/metrics/tenant_registries.go
+++ b/metrics/tenant_registries.go
@@ -219,20 +219,16 @@ func (d MetricFamiliesPerTenant) SendSumOfGaugesWithLabels(out chan<- prometheus
 
 // SendSumOfGaugesPerTenant provides metrics on a per-tenant basis.
 // This function assumes that `tenant` is the first label on the provided metric Desc.
-func (d MetricFamiliesPerTenant) SendSumOfGaugesPerTenant(out chan<- prometheus.Metric, desc *prometheus.Desc, gauge string) {
-	d.SendSumOfGaugesPerTenantWithLabels(out, desc, gauge)
-}
+func (d MetricFamiliesPerTenant) SendSumOfGaugesPerTenant(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, options ...MetricOption) {
+	opts := applyMetricOptions(options...)
 
-// SendSumOfGaugesPerTenantWithLabels provides metrics with the provided label names on a per-tenant basis. This function assumes that `tenant` is the
-// first label on the provided metric Desc
-func (d MetricFamiliesPerTenant) SendSumOfGaugesPerTenantWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
 	for _, tenantEntry := range d {
 		if tenantEntry.tenant == "" {
 			continue
 		}
 
 		result := singleValueWithLabelsMap{}
-		tenantEntry.metrics.sumOfSingleValuesWithLabels(metric, labelNames, gaugeValue, result.aggregateFn, false)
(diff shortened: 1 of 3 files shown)


alexweav (approved):

metrics/tenant_registries.go
```diff
@@ -219,20 +219,16 @@ func (d MetricFamiliesPerTenant) SendSumOfGaugesWithLabels(out chan<- prometheus
...
-// first label on the provided metric Desc
-func (d MetricFamiliesPerTenant) SendSumOfGaugesPerTenantWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
 	for _, tenantEntry := range d {
 		if tenantEntry.tenant == "" {
 			continue
 		}
 
 		result := singleValueWithLabelsMap{}
-		tenantEntry.metrics.sumOfSingleValuesWithLabels(metric, labelNames, gaugeValue, result.aggregateFn, false)
+		tenantEntry.metrics.sumOfSingleValuesWithLabels(metric, opts.labelNames, gaugeValue, result.aggregateFn, opts.skipZeroValueMetrics)
```
pstibrany, line 231: Could you please add unit test showing that skipping of zero values works?

metrics/tenant_registries.go
```diff
@@ -219,20 +219,16 @@ func (d MetricFamiliesPerTenant) SendSumOfGaugesWithLabels(out chan<- prometheus
...
 // This function assumes that `tenant` is the first label on the provided metric Desc.
-func (d MetricFamiliesPerTenant) SendSumOfGaugesPerTenant(out chan<- prometheus.Metric, desc *prometheus.Desc, gauge string) {
-	d.SendSumOfGaugesPerTenantWithLabels(out, desc, gauge)
-}
+func (d MetricFamiliesPerTenant) SendSumOfGaugesPerTenant(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, options ...MetricOption) {
+	opts := applyMetricOptions(options...)
 
-// SendSumOfGaugesPerTenantWithLabels provides metrics with the provided label names on a per-tenant basis. This function assumes that `tenant` is the
-// first label on the provided metric Desc
-func (d MetricFamiliesPerTenant) SendSumOfGaugesPerTenantWithLabels(out chan<- prometheus.Metric, desc *prometheus.Desc, metric string, labelNames ...string) {
```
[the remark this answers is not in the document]
titolins: Good point, no breaking changes then. Added it now and reverted unit tests also 👍

metrics/tenant_registries.go
```diff
@@ -219,25 +219,29 @@ func (d MetricFamiliesPerTenant) SendSumOfGaugesWithLabels(out chan<- prometheus
...
+		tenantEntry.metrics.sumOfSingleValuesWithLabels(metric, opts.labelNames, gaugeValue, result.aggregateFn, opts.skipZeroValueMetrics)
 		result.prependTenantLabelValue(tenantEntry.tenant)
 		result.WriteToMetricChannel(out, desc, prometheus.GaugeValue)
 	}
 }
 
+// SendSumOfGaugesPerTenantWithLabels provides metrics with the provided label names on a per-tenant basis. This function assumes that `tenant` is the
+// first label on the provided metric Desc
+//
+// Deprecated: use SendSumOfGaugesPerTenant instead.
```
pstibrany, line 240:
```suggestion
// Deprecated: use SendSumOfGaugesPerTenant with WithLabels option instead.
```

pstibrany (approved):
Thank you. Can you add an enhancement changelog entry?

titolins:
> Thank you. Can you add an enhancement changelog entry?

Added now. Thanks for the review 🙌

pstibrany:
Thank you.

PR #587, opened 2024-09-25
Change how read-only fields disaply in ring HTML page

pr00se:
**What this PR does**:

When dealing with a ring status page with hundreds of instances, it can be difficult to tell which ones are currently read-only. This PR makes two changes:
- for instances that are not read-only, leave the `Read-Only` field in the status page empty, rather than showing `false`
- for instances that are read-only, also show the duration that they have been read-only in the `Read-Only Updated` field

Before:
Screenshot 2024-09-25 at 2 44 52 AM

After:
Screenshot 2024-09-25 at 2 56 16 AM

Fixes #N/A

**Checklist**
- [ ] Tests updated


pstibrany (approved):
Thank you!

PR #591, opened 2024-10-03
cache: Add `.Set` and `.Add` methods to cache clients

56quarters:
**What this PR does**:

This change adds a synchronous version of `.Set` to Memcached and Redis clients as well as the various `Cache` wrapper implementations. This allows callers to set a key and be sure it exists in the cache. This change also adds an `.Add` method which conditionally adds an item to the cache only if it does not already exist.

This change is a prerequisite for grafana/mimir#9386

**Checklist**
- [x] Tests updated

### DIFF
--- a/cache/client.go
+++ b/cache/client.go
@@ -17,6 +17,7 @@ import (
 // Common functionality shared between the Memcached and Redis Cache implementations
 
 const (
+	opAdd            = "add"
 	opSet            = "set"
 	opGetMulti       = "getmulti"
 	opDelete         = "delete"
@@ -29,6 +30,8 @@ const (
 	reasonMaxItemSize     = "max-item-size"
 	reasonAsyncBufferFull = "async-buffer-full"
 	reasonMalformedKey    = "malformed-key"
+	reasonInvalidTTL      = "invalid-ttl"
+	reasonNotStored       = "not-stored"
 	reasonConnectTimeout  = "connect-timeout"
 	reasonTimeout         = "request-timeout"
 	reasonServerError     = "server-error"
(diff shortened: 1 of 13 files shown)


cache/client.go
```diff
@@ -17,6 +17,7 @@ import (
 // Common functionality shared between the Memcached and Redis Cache implementations
 
 const (
+	opAdd            = "add"
```
dimitarvdimitrov, line 20: this isn't used to initialize `cm.operations`

cache/lru.go
```diff
@@ -103,6 +103,41 @@ func (l *LRUCache) SetMultiAsync(data map[string][]byte, ttl time.Duration) {
...
+	return err
+}
+
+func (l *LRUCache) Add(ctx context.Context, key string, value []byte, ttl time.Duration) error {
+	err := l.c.Add(ctx, key, value, ttl)
+
+	// When a caller uses the Add method, the presence of absence of an entry in the cache
+	// has significance. In order to maintain the semantics of that, we only add an entry to
+	// the LRU when it was able to be successfully added to the shared cache.
+	if err == nil {
```
dimitarvdimitrov, line 127: i didn't think of this. Can you add a tiny test for it?

cache/memcached_client.go
```diff
@@ -28,6 +28,7 @@ import (
 
 const (
 	dnsProviderUpdateInterval = 30 * time.Second
+	maxTTL                    = 30 * 24 * time.Hour
```
dimitarvdimitrov, line 31: where does this come from? number of seconds in an in32? hmm, no, that's way more.

cache/memcached_client.go
```diff
@@ -339,6 +374,20 @@ func (c *MemcachedClient) setSingleItem(key string, value []byte, ttl time.Durat
 	})
 }
 
+// TODO: Docs
```
dimitarvdimitrov, line 377: 👀

56quarters:
> nice one, I'm not sure about the 30d TTL, otherwise LGTM

When you give Memcached a TTL longer than 30d, it treats the number of seconds as a Unix timestamp. E.g. if you give it a 31d TTL your item will expire on February 1st 1970 -- immediately.

cache/memcached_client.go
56quarters: :facepalm: I'll fix

cache/client.go
56quarters: Fixed

cache/lru.go
56quarters: Good idea, will do :+1:

PR #592, opened 2024-10-04
Memberlist: support for debouncing notifications

seizethedave:
**What this PR does**:
- Adds a layer of buffering to Memberlist's notification handling so that notifications are fired at most once per `NotifyInterval`, at which point it will deliver notifications using the most recently-observed data.
- Adds a new config flag to control this interval: `-memberlist.notify-interval` which defaults to 0 (off).

Motivation for this change:

In clusters where the memberlist KVStore watched by Ring has many replicas, redeploying those replicas can cause `WatchKey` and `updateRingState` to be called hundreds of times per second. When there are many concurrent goroutines calling `ring.ShuffleShard`, the high rate of `updateRingState` calls (which take locks and clear caches) can create heavy lock contention and latency as `ShuffleShard` attempts to take locks in order to repopulate those caches.

**Checklist**
- [x] Tests updated


CHANGELOG.md
```diff
@@ -80,9 +80,11 @@
 * [CHANGE] Changed `ShouldLog()` function signature in `middleware.OptionalLogging` interface to `ShouldLog(context.Context) (bool, string)`: the returned `string` contains an optional reason. When reason is valued, `GRPCServerLog` adds `(<reason>)` suffix to the error. #514
 * [CHANGE] Cache: Remove superfluous `cache.RemoteCacheClient` interface and unify all caches using the `cache.Cache` interface. #520
 * [CHANGE] Updated the minimum required Go version to 1.21. #540
+* [CHANGE] Backoff: added `Backoff.ErrCause()` which is like `Backoff.Err()` but returns the context cause if backoff is terminated because the context has been canceled. #538
 * [CHANGE] memberlist: Metric `memberlist_client_messages_in_broadcast_queue` is now split into `queue="local"` and `queue="gossip"` values. #539
 * [CHANGE] memberlist: Failure to fast-join a cluster via contacting a node is now logged at `info` instead of `debug`. #585
 * [CHANGE] `Service.AddListener` and `Manager.AddListener` now return function for stopping the listener. #564
+* [CHANGE] ring: KVStrore updates now processed at most once per `-ring.update-interval`. #592
```
pstibrany, line 87:
```suggestion
* [CHANGE] ring: ring updates now processed at most once per `-ring.update-interval`. #592
```

ring/ring.go
```diff
@@ -324,18 +330,46 @@ func (r *Ring) loop(ctx context.Context) error {
 	r.updateRingMetrics()
 	r.mtx.Unlock()
 
+	// Debounce WatchKey updates, as they can be frequent enough to cause lock
+	// contention. The most recent update is the one we'll use when we
+	// periodically update the ring.
+
+	go func() {
+		t := time.NewTicker(r.cfg.UpdateInterval)
```
pstibrany, line 338: We should probably validate that `r.cfg.UpdateInterval` is > 0 in `New` function, or this will panic. Or use some sane positive value instead.

pstibrany (approved):
Thank you, PR looks fine to me. Tests that do not initialize `UpdateInterval` are now failing.

I think the change is safe to enable it by default, without the need to go back to previous behaviour.

Alternative to this would be implementing rate-limiting in memberlist KV client's WatchKey/WatchPrefix implementations, which would automatically cover all rings, but only benefit memberlist users. (However we would like to get rid of supporting other KV stores eventually, so that's not a big deal)

ring/ring.go
seizethedave: I think if we don't turn a 0 interval into a sane default, this will be an annoying breaking change for other repos (incl. Mimir) when this PR is merged. As they construct ring.Config{} in their own tests.

pstibrany: Makes sense, that should also minimize changes required to make tests pass.

ring/ring.go
```diff
@@ -162,6 +163,7 @@ func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	cfg.KVStore.RegisterFlagsWithPrefix(prefix, "collectors/", f)
 
 	f.DurationVar(&cfg.HeartbeatTimeout, prefix+"ring.heartbeat-timeout", time.Minute, "The heartbeat timeout after which ingesters are skipped for reads/writes. 0 = never (timeout disabled).")
+	f.DurationVar(&cfg.UpdateInterval, prefix+"ring.update-interval", 0, "How often to recompute ring state when a change is detected from the KVStore. 0 = no delay.")
```
pstibrany, line 166: I'd suggest mentioning that nonzero interval means rate-limiting the updates, which can have positive effect on cpu usage.

pstibrany:
> I vendored this change locally and it actually does break tests in Mimir, as ruler tests attempt to wait 100ms for ring changes to propagate. I may just bite the bullet and make this change a no-op by default (synchronous propagation) and make the added delay configurable. @pstibrany let me know if you have thoughts.

I'm fine with having it disabled by default.

One previously mentioned alternative would be to move this to memberlist KV store -- not only would it apply to all rings, but clients of memberlist KV store already expect updates to be delayed. WDYT?

pstibrany (approved):
Thank you, I'm fine with keeping this disabled by default. Once we

seizethedave:
> > I vendored this change locally and it actually does break tests in Mimir, as ruler tests attempt to wait 100ms for ring changes to propagate. I may just bite the bullet and make this change a no-op by default (synchronous propagation) and make the added delay configurable. @pstibrany let me know if you have thoughts.
> 
> I'm fine with having it disabled by default.
> 
> One previously mentioned alternative would be to move this to memberlist KV store -- not only would it apply to all rings, but clients of memberlist KV store already expect updates to be delayed. WDYT?

I had taken a cursory look at this earlier and it looked like I've have to pepper a bunch of ad-hoc timer code into the bodies of WatchKey/Prefix. But today I gave it another look and realized it could cleanly go in the memberlist notification handling before the notifs make it to the Watch methods. I like this better! PTAL in your limited time. :)

pstibrany (approved):
Very nice! Thank you!

Issue #593, opened 2024-10-04
Test flake TestReusableGoroutinesPool


seizethedave:
```
--- FAIL: TestReusableGoroutinesPool (0.01s)
    worker_test.go:23: 
        	Error Trace:	/home/runner/work/dskit/dskit/concurrency/worker_test.go:23
        	Error:      	Not equal: 
        	            	expected: 8
        	            	actual  : 7
        	Test:       	TestReusableGoroutinesPool
```

Seems to happen more often (or exclusively?) in the new Go 1.23.x CI job. Can't replicate it on my local Mac when I run the test thousands of times.

[Log](https://github.com/grafana/dskit/actions/runs/11183125949/job/31092499090?pr=592)

PR #596, opened 2024-10-04
Fix flaky `TestReusableGoroutinesPool` test

julienduchesne:
Fixes https://github.com/grafana/dskit/issues/593
Count goroutines from the current test rather than for the whole process


seizethedave (approved):
Very nice!

PR #597, opened 2024-10-07
test: Use interface for passing instance ring to partition instance ring

benclive:
**What this PR does**:

Define a InstanceRingReader interface for creating a new PartitionInstanceRing.
* This is to enable mocking the Partition Instance Ring during tests external to this package.
* Specifically, Loki already uses an ingester ring mock to test certain functionality, such as looking up ingesters for querying, without defining a full, in-memory ring. I would like to extend this test to use a partition instance ring without rewriting the whole test suite, but this isn't possible right now.
* I cannot use approaches such as `&Ring{ringDesc: &Desc{}}` outside of the `ring` package because `ringDesc` is private, and the existing `ReadRing` interface does not provide the `GetInstance(string)` method.

n/a

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@
 * [CHANGE] memberlist: Metric `memberlist_client_messages_in_broadcast_queue` is now split into `queue="local"` and `queue="gossip"` values. #539
 * [CHANGE] memberlist: Failure to fast-join a cluster via contacting a node is now logged at `info` instead of `debug`. #585
 * [CHANGE] `Service.AddListener` and `Manager.AddListener` now return function for stopping the listener. #564
+* [CHANGE] Add `InstanceRingReader` interface to `ring` package.
 * [FEATURE] Cache: Add support for configuring a Redis cache backend. #268 #271 #276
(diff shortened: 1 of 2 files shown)


pstibrany:
This change seems unnecessary to me. You can as well just keep reference to the ring provided to `NewPartitionInstanceRing` function, without introducing new interface.

benclive:
> keep reference to the ring provided to NewPartitionInstanceRing function

The problem I want to solve, which isn't present in this repo, is to be able to provide a mocked ring implementation to NewPartitionInstanceRing so I can test external behaviour that depends on the it without initialising a full Ring, KVStore & running service.

Specifically, in Loki I'm looking to provide a static list of ingesters & partitions in tests so I can assert that the query engine generates the correct behaviour in various scenarios. I can't easily do that right now because I need to provide a real Ring implementation to the constructor. 

I'm aware it is possible to set up the real implementation for a test but it requires a lot of setup that isn't related to what I actually want to test, so adding an interface here seems simpler, quicker and less error prone approach to testing component behaviour.

pstibrany (approved):

PR #598, opened 2024-10-08
Fix flaky memberlist `TestRejoin` test

julienduchesne:
It randomly failed here: https://github.com/grafana/dskit/pull/525 and in some of the test runs here: https://github.com/grafana/dskit/pull/596

This test fails due to the probe interval being 5s, which is the same time we poll for

### DIFF
--- a/.github/workflows/find-flaky-tests.yml
+++ b/.github/workflows/find-flaky-tests.yml
@@ -8,6 +8,7 @@ jobs:
   find-flaky-tests:
     # Only run this workflow when the comment contains '/find-flaky-tests' or it's manually triggered
     if: contains(github.event.comment.body, '/find-flaky-tests') || github.event_name == 'workflow_dispatch'
+    fail-fast: false
     strategy:
       matrix:
         runs: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20] # Run the build workflow 20 times
--- a/kv/memberlist/memberlist_client_test.go
+++ b/kv/memberlist/memberlist_client_test.go
@@ -1240,20 +1240,20 @@ func TestRejoin(t *testing.T) {
 		return mkv2.memberlist.NumMembers()
 	}
 
-	poll(t, 5*time.Second, 2, membersFunc)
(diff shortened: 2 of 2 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -201,6 +203,12 @@ func (cfg *KVConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
...
 	f.DurationVar(&cfg.BroadcastTimeoutForLocalUpdatesOnShutdown, prefix+"memberlist.broadcast-timeout-for-local-updates-on-shutdown", 10*time.Second, "Timeout for broadcasting all remaining locally-generated updates to other nodes when shutting down. Only used if there are nodes left in the memberlist cluster, and only applies to locally-generated updates, not to broadcast messages that are result of incoming gossip updates. 0 = no timeout, wait until all locally-generated updates are sent.")
 
+	// For our use cases, we don't need a very fast detection of dead nodes. Since we use a TCP transport
+	// and we open a new TCP connection for each packet, we prefer to reduce the probe frequency and increase
+	// the timeout compared to defaults.
+	f.DurationVar(&cfg.ProbeInterval, prefix+"memberlist.probe-interval", 5*time.Second, "Probe a random node every this interval. This setting is also the total timeout for the direct + indirect probes.")
+	f.DurationVar(&cfg.ProbeTimeout, prefix+"memberlist.probe-timeout", 2*time.Second, "Timeout for the direct probe.")
```
pstibrany, line 210: If we want to change it for tests only, let's not expose this as configurable fields.

julienduchesne: You mean put the new configs in `KVConfig` but not exposed as flags?

julienduchesne: 🤔. I could just bump the polling time to 7s. That would be an easier fix, if we don't want to expose the config at all

julienduchesne: Here, this should be better 😄

pstibrany: :) Thanks.

pstibrany: > You mean put the new configs in KVConfig but not exposed as flags?

Yes, this is what I meant. Also hide from yaml by using `yaml:"-"` tag.

pstibrany (approved):

Issue #572, opened 2024-08-22
Flaky test `TestPartitionInstanceLifecycler`


colega:
Failed here: https://drone.grafana.net/grafana/dskit/4690

```
--- FAIL: TestPartitionInstanceLifecycler (0.00s)
    --- FAIL: TestPartitionInstanceLifecycler/should_wait_for_the_configured_minimum_number_of_owners_before_switching_a_pending_partition_to_active (1.03s)
        partition_instance_lifecycler_test.go:64: 
            	Error Trace:	/drone/src/ring/partition_instance_lifecycler_test.go:64
            	Error:      	Condition never satisfied
            	Test:       	TestPartitionInstanceLifecycler/should_wait_for_the_configured_minimum_number_of_owners_before_switching_a_pending_partition_to_active
FAIL
FAIL	github.com/grafana/dskit/ring	554.854s
```

PR #600, opened 2024-10-08
Fix all flaky tests

julienduchesne:
- `TestPartitionInstanceLifecycler`: Increase test polling timeout. 1s seems tight, as it sometimes fails with `-race`. Fixes https://github.com/grafana/dskit/issues/572
- `TestBasicLifecycler_HeartbeatWhileRunning`: Same as above
- `TestRejoin`: Converted the test to use `require.Eventually` instead of a custom func + increased polling timeout a bit
- `TestTLSServerWithLocalhostCertWithClientCertificateEnforcementUsingClientCA1`: Wait until the server is ready before running tests

Mostly increasing timeouts, but I believe `-race` is making those tests fail. Can't rely on time working correctly

### DIFF
--- a/crypto/tls/test/tls_integration_test.go
+++ b/crypto/tls/test/tls_integration_test.go
@@ -115,6 +115,23 @@ func newIntegrationClientServer(
 		require.NoError(t, err)
 	}()
 
+	// Wait until the server is up and running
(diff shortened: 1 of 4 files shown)


seizethedave (approved):
LGTM nice!

PR #603, opened 2024-10-09
Really fix flaky `TestMultipleClients` tests

julienduchesne:
https://github.com/grafana/dskit/pull/599 went in a wrong direction, so I revert it here: it makes all `TestMultipleClients` tests flaky since not all members are necessarily joined after the updates

Instead, to fix the `TestMultipleClientsWithMixedLabelsAndExpectFailure` test, I add a couple more labeled members, that way, even if we reach 3 updates (which happened some times, hence the flakes), we'll never get to 5 with a single member

Also, try to fix the `TestTLSServerWithLocalhostCertWithClientCertificateEnforcementUsingClientCA1` test by re-doing requests that have been reset by the server, but only if we were expecting an error. Note that this is not newly flaky, this comment has been there since the beginning: https://github.com/grafana/dskit/blob/2e104a8053fa31d76ba7f568fdc7ef20e38a5bf7/crypto/tls/test/tls_integration_test.go#L381-L383

Flaky tests run: https://github.com/grafana/dskit/actions/runs/11257895128 🟢

### DIFF
--- a/crypto/tls/test/tls_integration_test.go
+++ b/crypto/tls/test/tls_integration_test.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


aldernero (approved):
LGTM

Issue #389, opened 2023-09-27
kv/memberlist: test failure on Mac 


kevinburkesegment:
On my Mac, if I run: 

```
$ GO111MODULE=on go test ./kv/memberlist
```

I get the following failure: 

```
--- FAIL: TestMultipleClientsWithSameLabelWithClusterLabelVerification (24.02s)
    memberlist_client_test.go:685: Waiting before start
    memberlist_client_test.go:693: Observing ring ...
    memberlist_client_test.go:707: Update 11.177042ms : Ring has 2 members, and 256 tokens, oldest timestamp: 93.133ms avg timestamp: 93.133ms youngest timestamp: 93.133ms
    memberlist_client_test.go:707: Update 250.462417ms : Ring has 3 members, and 384 tokens, oldest timestamp: 332.422ms avg timestamp: 332.422ms youngest timestamp: 332.422ms
    memberlist_client_test.go:707: Update 344.509542ms : Ring has 4 members, and 512 tokens, oldest timestamp: 426.47ms avg timestamp: 426.47ms youngest timestamp: 426.47ms
    memberlist_client_test.go:707: Update 405.829458ms : Ring has 5 members, and 640 tokens, oldest timestamp: 487.791ms avg timestamp: 487.791ms youngest timestamp: 487.791ms
    memberlist_client_test.go:707: Update 405.88675ms : Ring has 7 members, and 896 tokens, oldest timestamp: 1.487848s avg timestamp: 1.487848s youngest timestamp: 487.848ms
    memberlist_client_test.go:707: Update 580.5835ms : Ring has 8 members, and 1024 tokens, oldest timestamp: 1.662547s avg timestamp: 1.662547s youngest timestamp: 662.547ms
    memberlist_client_test.go:707: Update 771.606208ms : Ring has 9 members, and 1152 tokens, oldest timestamp: 1.853573s avg timestamp: 1.853573s youngest timestamp: 853.573ms
    memberlist_client_test.go:707: Update 771.712708ms : Ring has 10 members, and 1280 tokens, oldest timestamp: 1.853679s avg timestamp: 1.853679s youngest timestamp: 853.679ms
    memberlist_client_test.go:707: Update 1.001151042s : Ring has 10 members, and 1280 tokens, oldest timestamp: 2.083121s avg timestamp: 2.083121s youngest timestamp: 83.121ms
[...]

PR #608, opened 2024-10-10
`TestMultipleClients*`: Poll for the memberlist status for 10 seconds

julienduchesne:
Flaky test run: https://github.com/grafana/dskit/actions/runs/11281170696
Closes https://github.com/grafana/dskit/issues/389

These tests are highly dependant on timing, which makes them flaky. For example, they don't fail on my machine unless I cut down the initial update time a bit, but they seem to fail often on Github Actions. I think this is due to machine power. 

Solution: Poll for a bit longer at the end to make sure that the memberlist was updated


seizethedave (approved):
Looks good!

PR #607, opened 2024-10-10
`ReusableGoroutinesPool`: Fix datarace on `Close`

julienduchesne:
Fixes this datarace in Mimir: https://github.com/grafana/mimir/pull/9585

Without the mutex:
```
julienduchesne@triceratops dskit % go test ./concurrency -race
==================
WARNING: DATA RACE
Write at 0x00c00001e240 by goroutine 103:
  runtime.recvDirect()
      /opt/homebrew/opt/go/libexec/src/runtime/chan.go:388 +0x7c
  github.com/grafana/dskit/concurrency.(*ReusableGoroutinesPool).Close()
      /Users/julienduchesne/Repos/dskit/concurrency/worker.go:40 +0x1d4
  github.com/grafana/dskit/concurrency.TestReusableGoroutinesPool_Race()
      /Users/julienduchesne/Repos/dskit/concurrency/worker_test.go:89 +0x1c4
  testing.tRunner()
      /opt/homebrew/opt/go/libexec/src/testing/testing.go:1690 +0x184
  testing.(*T).Run.gowrap1()
      /opt/homebrew/opt/go/libexec/src/testing/testing.go:1743 +0x40

Previous read at 0x00c00001e240 by goroutine 106:
  runtime.chansend1()
      /opt/homebrew/opt/go/libexec/src/runtime/chan.go:157 +0x2c
  github.com/grafana/dskit/concurrency.(*ReusableGoroutinesPool).Go()
      /Users/julienduchesne/Repos/dskit/concurrency/worker.go:29 +0x48
  github.com/grafana/dskit/concurrency.TestReusableGoroutinesPool_Race.func1()
      /Users/julienduchesne/Repos/dskit/concurrency/worker_test.go:82 +0xfc

Goroutine 103 (running) created at:
  testing.(*T).Run()
      /opt/homebrew/opt/go/libexec/src/testing/testing.go:1743 +0x5e0
  testing.runTests.func1()
      /opt/homebrew/opt/go/libexec/src/testing/testing.go:2168 +0x80
  testing.tRunner()
      /opt/homebrew/opt/go/libexec/src/testing/testing.go:1690 +0x184
  testing.runTests()
      /opt/homebrew/opt/go/libexec/src/testing/testing.go:2166 +0x6e0
  testing.(*M).Run()
      /opt/homebrew/opt/go/libexec/src/testing/testing.go:2034 +0xb74
  go.uber.org/goleak.VerifyTestMain()
[...]

PR #610, opened 2024-10-14
Remove mutex from ReusableGoroutinesPool

colega:
**What this PR does**:

I find it unsettling that we need to acquire a mutex (a concurrency primitive) to send data to a channel (another concurrency primitive).

I think this achieves the same effect (we can stop the worker goroutines in a safe way) without having to mix them and leaving the main code path clear (just try to send a job, or run a new goroutine).

This is a follow up on https://github.com/grafana/dskit/pull/607

**Checklist**
- [ ] Tests updated

### DIFF
--- a/concurrency/worker.go
+++ b/concurrency/worker.go
@@ -1,20 +1,22 @@
 package concurrency
 
-import (
-	"sync"
-)
-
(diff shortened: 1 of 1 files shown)


colega:
If we push, we need to wait until all workloads finish, and that's not the contract of the method right now.

PR #612, opened 2024-10-21
Add support for deletes in the memberlist backed KV store

aldernero:
WIP: Add support for deletes in the memberlist kv store.

**Checklist**
- [X] Tests updated


kv/memberlist/memberlist_client.go
```diff
@@ -330,6 +337,9 @@ type ValueDesc struct {
 
 	// ID of codec used to write this value. Only used when sending full state.
 	CodecID string
+
+	Deleted    bool
```
NickAnge, line 341: nit: Maybe , add comments similar to the above variables to be consistent ?

kv/memberlist/memberlist_client.go
```diff
@@ -155,7 +160,8 @@ type KVConfig struct {
 	RejoinInterval   time.Duration       `yaml:"rejoin_interval" category:"advanced"`
 
 	// Remove LEFT ingesters from ring after this timeout.
-	LeftIngestersTimeout time.Duration `yaml:"left_ingesters_timeout" category:"advanced"`
+	LeftIngestersTimeout   time.Duration `yaml:"left_ingesters_timeout" category:"advanced"`
+	ObsoleteEntriesTimeout time.Duration `yaml:"obsolete_entries_timeout" category:"advanced"`
```
NickAnge, line 164: not sure if we should test it first in some cells before set it as advanced. Whatever you prefer though

aldernero: good catch, we should mark it as experimental until it's gone through more testing in dev

kv/memberlist/memberlist_client.go
```diff
@@ -1005,6 +1031,35 @@ func (m *KV) notifyWatchersSync(key string) {
...
 
+func (m *KV) Delete(key string) error {
+	m.storeMu.Lock()
+
+	val, ok := m.store[key]
+	if !ok || val.Deleted {
+		return nil
+	}
+
+	m.store[key] = val
```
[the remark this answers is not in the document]
aldernero: thanks for catching this, it was left over from when I was explicitly setting Deleted before putting it back in the store, but that is handled by `mergeValueForKey`

kv/memberlist/memberlist_client_test.go
```diff
@@ -574,6 +575,117 @@ func defaultKVConfig(i int) KVConfig {
...
+	err = kv1.Delete(context.Background(), key)
+	if err != nil {
+		t.Fatalf("Failed to delete key %s: %v", key, err)
+	}
+
+	time.Sleep(5 * time.Second) // wait for obsolete entries to be removed
+
+	val, err = kv1.Get(context.Background(), key)
+	require.NoError(t, err)
+	require.NotNil(t, val)
```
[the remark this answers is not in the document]
aldernero: thanks for catching the `NotNil`, should have been `Nil`. I changed it and realized that the 500ms for the second client to start might not be enough time. After increasing the time to 2s, the test is no longer flaky for me.

kv/memberlist/memberlist_client.go
```diff
@@ -1005,6 +1034,34 @@ func (m *KV) notifyWatchersSync(key string) {
 	}
 }
 
+func (m *KV) Delete(key string) error {
+	m.storeMu.Lock()
+
+	val, ok := m.store[key]
+	if !ok || val.Deleted {
```
NickAnge, line 1041: @aldernero , if we return here the mutex is never gonna be unlocked, no ? 🤔 

 ```suggestion
	m.storeMu.Lock()
	val, ok := m.store[key]
	m.storeMu.Unlock()
	    
	if !ok || val.Deleted {
	     return nil
	}
	
```

File: kv/memberlist/memberlist_client.go [hunk 6]
```diff
@@ -1465,16 +1525,25 @@ func (m *KV) mergeValueForKey(key string, incomingValue Mergeable, incomingValue
...
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, false, time.Time{}, err
+	}
+
+	newVersion = curr.Version + 1
+	newUpdated = curr.UpdateTime
+	newDeleted = curr.Deleted
+
+	if !updateTime.IsZero() && updateTime.After(newUpdated) {
```
NickAnge, line 1539: Could you please clarify why this part is needed.  Maybe also add a comment clarifying it. wdyt ?

Based on my understanding, here we check if the incoming value was updated more recently, use these values. 🤔

kv/memberlist/memberlist_client.go
aldernero: yes, nice catch, I updated the code

kv/memberlist/memberlist_client.go
aldernero: that's correct, I added a comment

NickAnge (approved):

PR #614, opened 2024-10-24
add a span tag for content type if the header is set

cstyan:
Sometimes queries are sent to our databases as post requests, which may have the query encoded into the request body. To parse it out properly via `r.ParseForm`, the http method must be POST and the content type must be `application/x-www-form-urlencoded`. It would be useful when looking at traces for requests which have not been able to log their query properly to see a span tag with the content type, to know if these POST queries were sent with the correct content type, or if we're unable to log the query for some other reason.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated


56quarters (approved):
LGTM

PR #616, opened 2024-11-06
flagext: add intstring type

titolins:
**What this PR does**:
- Adds a new `IntString` type to be used for configuration values

- As mentioned in the previous mimir PR (https://github.com/grafana/mimir/pull/9840), we're having issues with certain configuration parameters that take long integers (e.g. `2048000` gets converted to `2.048e+06`)
- Adding a type that supports both strings and integers gives us the flexibility to migrate long values and not break the existing setup

**Checklist**
- [x] Tests updated
- [x] `CHANGELOG.md` updated


titolins:
We've decided to use `flagext.Bytes` instead of introducing a new type: https://github.com/grafana/mimir/pull/9845

PR #618, opened 2024-11-14
test(kv): create mock client for counting func calls

NickAnge:
**What this PR does**:

This PR creates a `MockCountingClient`, which wraps the `kv.client` and can be use in order to count calls at specific functions of the interface.

This will be used [here](https://github.com/grafana/mimir/pull/9905) for testing purposes

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/kv/mock.go
+++ b/kv/mock.go
@@ -2,7 +2,6 @@ package kv
 
 import (
 	"context"
-	"sync"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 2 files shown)


kv/mock.go
```diff
@@ -37,3 +39,78 @@ func (m mockClient) WatchKey(_ context.Context, _ string, _ func(interface{}) bo
...
+	mc.client.WatchKey(ctx, key, f)
+}
+
+func (mc *MockCountingClient) WatchPrefix(ctx context.Context, key string, f func(string, interface{}) bool) {
+	mc.incCall("WatchPrefix")
+
+	mc.client.WatchPrefix(ctx, key, f)
+}
+
+func (mc *MockCountingClient) GetCalls(name string) int {
```
56quarters, line 94: I would just use a separate counter member for each type of call instead of a map of them since we know all the types of calls that will be made. There's no need to add new calls dynamically.

e.g.

```go

type MockCountingClient struct {
	client Client

	GetCalls *atomic.Int32
	ListCalls *atomic.Int32
	// ...
}

Comment (NickAnge) [reply kv/mock.go:94]: Yeah makes sense actually 😅 . Its much simpler that way. I will update it

Comment (NickAnge) [reply kv/mock.go]: Addressed here [77843bd](https://github.com/grafana/dskit/pull/618/commits/77843bddc2e67e27cf1f24df14727a2ab29d9789)

File: kv/mock.go [hunk 2]
```diff
@@ -37,3 +38,63 @@ func (m mockClient) WatchKey(_ context.Context, _ string, _ func(interface{}) bo
...
+
+// MockCountingClient is a wrapper around the Client interface that counts the number of times its functions are called.
+// This is used for testing only.
+type MockCountingClient struct {
+	client Client
+
+	ListCalls        *atomic.Int32
+	GetCalls         *atomic.Int32
+	DeleteCalls      *atomic.Int32
+	CASCalls         *atomic.Int32
```
56quarters, line 50: These should probably be Uint32 since they're counters.

56quarters (approved):
LGTM with one small change.

kv/mock.go
NickAnge: [9e8cce4](https://github.com/grafana/dskit/pull/618/commits/9e8cce4a8d07084a6407ba100f82f178306b6e08)

PR #619, opened 2024-11-14
Feature: Add new metric `request_throughput`

tinitiuset:
**What this PR does**:

This PR adds two configuration parameters `server.throughput.request-cutoff` and `server.throughput.unit`, exposes a new metric `request_server_throughput` and calculates throughput in `units/s` for the metric using information from header `Server-Timing`.

If `server.throughput.request-cutoff` is `0` no throughput will be calculated.

Implemented code is really similar to [this branch](https://github.com/grafana/dskit/tree/krajo/throughputmetric) by @krajorama. But adds some flexibility to measure throughput based on different signals. It will default to `total_samples` as _processed samples are easier to explain to users_. Discussed with @dimitarvdimitrov.

**Checklist**
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Changelog
 
-* [CHANGE] Add new metric `slow_request_server_throughput` to track the throughput of slow queries.
+* [CHANGE] Add new metric `slow_request_server_throughput` to track the throughput of slow queries. #619
 * [CHANGE] Log middleware updated to honor `logRequestHeaders` in all logging scenarios. #615
 * [CHANGE] Roll back the gRPC dependency to v1.65.0 to allow downstream projects to avoid a performance regression and maybe a bug in v1.66.0. #581
 * [CHANGE] Update the gRPC dependency to v1.66.0 and deprecate the `grpc_server_recv_buffer_pools_enabled` option that is no longer supported by it. #580
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] Add new metric `slow_request_server_throughput` to track the throughput of slow queries.
 * [CHANGE] Log middleware updated to honor `logRequestHeaders` in all logging scenarios. #615
 * [CHANGE] Roll back the gRPC dependency to v1.65.0 to allow downstream projects to avoid a performance regression and maybe a bug in v1.66.0. #581
 * [CHANGE] Update the gRPC dependency to v1.66.0 and deprecate the `grpc_server_recv_buffer_pools_enabled` option that is no longer supported by it. #580
--- a/middleware/instrument.go
+++ b/middleware/instrument.go
@@ -52,7 +52,7 @@ type Instrument struct {
 	RequestBodySize   *prometheus.HistogramVec
 	ResponseBodySize  *prometheus.HistogramVec
 	InflightRequests  *prometheus.GaugeVec
-	RequestCutoff     time.Duration
+	LatencyCutoff     time.Duration
 	ThroughputUnit    string
 	RequestThroughput *prometheus.HistogramVec
 }
@@ -110,7 +110,7 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 			labelValues = append(labelValues, tenantID)
 			instrument.ObserveWithExemplar(r.Context(), i.PerTenantDuration.WithLabelValues(labelValues...), respMetrics.Duration.Seconds())
 		}
-		if i.RequestCutoff > 0 && respMetrics.Duration > i.RequestCutoff {
+		if i.LatencyCutoff > 0 && respMetrics.Duration > i.LatencyCutoff {
 			volume, err := extractValueFromMultiValueHeader(w.Header().Get("Server-Timing"), i.ThroughputUnit, "val")
 			if err == nil {
 				instrument.ObserveWithExemplar(r.Context(), i.RequestThroughput.WithLabelValues(r.Method, route), volume/respMetrics.Duration.Seconds())
@@ -126,20 +126,20 @@ func extractValueFromMultiValueHeader(h, name string, key string) (float64, erro
 		return 0, fmt.Errorf("not a multi-value header")
 	}
 	for _, part := range parts {
-		if strings.HasPrefix(part, name) {
-			part = part[len(name):]
-			sparts := strings.Split(part, ";")
-			for _, spart := range sparts {
-				if strings.HasPrefix(spart, key) {
-					var value float64
-					_, err := fmt.Sscanf(spart, key+"=%f", &value)
-					if err != nil {
-						return 0, err
-					}
-					return value, nil
+		if part, found := strings.CutPrefix(part, name); found == true {
+			for _, spart := range strings.Split(part, ";") {
+				if !strings.HasPrefix(spart, key) {
+					continue
 				}
+				var value float64
+				_, err := fmt.Sscanf(spart, key+"=%f", &value)
+				if err != nil {
+					return 0, fmt.Errorf("failed to parse value from header: %w", err)
+				}
+				return value, nil
 			}
 		}
+
 	}
 	return 0, fmt.Errorf("desired name not found in header")
 }
--- a/middleware/instrument.go
+++ b/middleware/instrument.go
@@ -126,7 +126,7 @@ func extractValueFromMultiValueHeader(h, name string, key string) (float64, erro
 		return 0, fmt.Errorf("not a multi-value header")
 	}
 	for _, part := range parts {
-		if part, found := strings.CutPrefix(part, name); found == true {
+		if part, found := strings.CutPrefix(part, name); found {
 			for _, spart := range strings.Split(part, ";") {
 				if !strings.HasPrefix(spart, key) {
 					continue
--- a/middleware/instrument.go
+++ b/middleware/instrument.go
@@ -6,10 +6,12 @@ package middleware
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/felixge/httpsnoop"
 	"github.com/gorilla/mux"
@@ -44,12 +46,15 @@ func (f PerTenantCallback) shouldInstrument(ctx context.Context) (string, bool) 
 
 // Instrument is a Middleware which records timings for every HTTP request
 type Instrument struct {
-	Duration          *prometheus.HistogramVec
-	PerTenantDuration *prometheus.HistogramVec
-	PerTenantCallback PerTenantCallback
-	RequestBodySize   *prometheus.HistogramVec
-	ResponseBodySize  *prometheus.HistogramVec
-	InflightRequests  *prometheus.GaugeVec
+	Duration                    *prometheus.HistogramVec
+	PerTenantDuration           *prometheus.HistogramVec
+	PerTenantCallback           PerTenantCallback
+	RequestBodySize             *prometheus.HistogramVec
+	ResponseBodySize            *prometheus.HistogramVec
+	InflightRequests            *prometheus.GaugeVec
+	SlowRequestCutoff           time.Duration
+	ServerThroughputUnit        string
+	SlowRequestServerThroughput *prometheus.HistogramVec
 }
 
 // IsWSHandshakeRequest returns true if the given request is a websocket handshake request.
@@ -104,6 +109,17 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 		if tenantID, ok := i.PerTenantCallback.shouldInstrument(r.Context()); ok {
 			labelValues = append(labelValues, tenantID)
 			instrument.ObserveWithExemplar(r.Context(), i.PerTenantDuration.WithLabelValues(labelValues...), respMetrics.Duration.Seconds())
+		}
+		if i.SlowRequestCutoff > 0 && respMetrics.Duration > i.SlowRequestCutoff {
+			parts := strings.Split(w.Header().Get("Server-Timing"), ", ")
+			volume := int64(0)
+			for _, part := range parts {
+				if strings.HasPrefix(part, i.ServerThroughputUnit) {
+					_, _ = fmt.Sscanf(part, i.ServerThroughputUnit+"=%d", &volume)
+					break
+				}
+			}
+			i.SlowRequestServerThroughput.WithLabelValues(r.Method, route).Observe(float64(volume) / respMetrics.Duration.Seconds())
 		}
 	})
 }
--- a/middleware/instrument.go
+++ b/middleware/instrument.go
@@ -111,28 +111,37 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 			instrument.ObserveWithExemplar(r.Context(), i.PerTenantDuration.WithLabelValues(labelValues...), respMetrics.Duration.Seconds())
 		}
 		if i.RequestCutoff > 0 && respMetrics.Duration > i.RequestCutoff {
-			volume, err := extractValueFromMultiValueHeader(w.Header().Get("Server-Timing"), i.ThroughputUnit)
+			volume, err := extractValueFromMultiValueHeader(w.Header().Get("Server-Timing"), i.ThroughputUnit, "val")
 			if err == nil {
-				instrument.ObserveWithExemplar(r.Context(), i.RequestThroughput.WithLabelValues(r.Method, route), float64(volume)/respMetrics.Duration.Seconds())
+				instrument.ObserveWithExemplar(r.Context(), i.RequestThroughput.WithLabelValues(r.Method, route), volume/respMetrics.Duration.Seconds())
 			}
 		}
 	})
 }
 
-// Extracts a single value from a multi-value header, e.g. "throughput=500, duration=1000"
-func extractValueFromMultiValueHeader(h, key string) (int64, error) {
+// Extracts a single value from a multi-value header, e.g. "name0;key0=0.0;key1=1.1, name1;key0=1.1"
+func extractValueFromMultiValueHeader(h, name string, key string) (float64, error) {
 	parts := strings.Split(h, ", ")
 	if len(parts) == 0 {
-		return 0, fmt.Errorf("no a multi-value header")
+		return 0, fmt.Errorf("not a multi-value header")
 	}
-	value := int64(0)
 	for _, part := range parts {
-		if strings.HasPrefix(part, key) {
-			_, err := fmt.Sscanf(part, key+"=%d", &value)
-			return value, err
+		if strings.HasPrefix(part, name) {
+			part = part[len(name):]
+			sparts := strings.Split(part, ";")
+			for _, spart := range sparts {
+				if strings.HasPrefix(spart, key) {
+					var value float64
+					_, err := fmt.Sscanf(spart, key+"=%f", &value)
+					if err != nil {
+						return 0, err
+					}
+					return value, nil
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] Add new metric `slow_request_server_throughput` to track the throughput of slow queries.
```
dimitarvdimitrov, line 3: can you add the PR number like it was done for other entries?

server/server.go
```diff
@@ -209,6 +216,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.LogRequestExcludeHeadersList, "server.log-request-headers-exclude-list", "", "Comma separated list of headers to exclude from loggin. Only used if server.log-request-headers is true.")
 	f.BoolVar(&cfg.LogRequestAtInfoLevel, "server.log-request-at-info-level-enabled", false, "Optionally log requests at info level instead of debug level. Applies to request headers as well if server.log-request-headers is enabled.")
 	f.BoolVar(&cfg.ProxyProtocolEnabled, "server.proxy-protocol-enabled", false, "Enables PROXY protocol.")
+	f.DurationVar(&cfg.ThroughputConfig.SlowRequestCutoff, "server.throughput-config.slow-request-cutoff", 0, "Duration after which a request is considered slow. For requests taking longer than this duration to finish, the throughput will be calculated. If set to 0, the throughput will not be calculated.")
+	f.StringVar(&cfg.ThroughputConfig.Unit, "server.throughput-config.unit", "total_samples", "Unit of the server throughput metric, for example 'processed_bytes' or 'total_samples'. If set, it is appended to the server_throughput metric name.")
```
dimitarvdimitrov, line 220:
```suggestion
	f.StringVar(&cfg.ThroughputConfig.Unit, "server.throughput-config.unit", "total_samples", "Unit of the server throughput metric, for example 'processed_bytes' or 'total_samples'. If set, it is appended to the `slow_request_server_throughput` metric name.")
```

server/metrics.go
```diff
@@ -73,5 +75,14 @@ func NewServerMetrics(cfg Config) *Metrics {
 			Name:      "inflight_requests",
 			Help:      "Current number of inflight requests.",
 		}, []string{"method", "route"}),
+		SlowRequestServerThroughput: reg.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace:                       cfg.MetricsNamespace,
+			Name:                            "slow_request_server_throughput_" + cfg.ThroughputConfig.Unit,
+			Help:                            "Server throughput of long running requests.",
+			ConstLabels:                     prometheus.Labels{"cutoff_ms": strconv.FormatInt(cfg.ThroughputConfig.SlowRequestCutoff.Milliseconds(), 10)},
+			NativeHistogramBucketFactor:     cfg.MetricsNativeHistogramFactor,
```
dimitarvdimitrov, line 83: let's add the default prometheus buckets too. I think without them we wouldn't export classic histograms. Because this is a library it would make sense to still support the old format

middleware/instrument.go
```diff
@@ -105,6 +110,17 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 			labelValues = append(labelValues, tenantID)
 			instrument.ObserveWithExemplar(r.Context(), i.PerTenantDuration.WithLabelValues(labelValues...), respMetrics.Duration.Seconds())
 		}
+		if i.SlowRequestCutoff > 0 && respMetrics.Duration > i.SlowRequestCutoff {
+			parts := strings.Split(w.Header().Get("Server-Timing"), ", ")
+			volume := int64(0)
+			for _, part := range parts {
+				if strings.HasPrefix(part, i.ServerThroughputUnit) {
+					_, _ = fmt.Sscanf(part, i.ServerThroughputUnit+"=%d", &volume)
```
dimitarvdimitrov, line 118: if the format doesn't comply, what do you think about not observing the throughput at all? I mean if upstream didn't properly communicate what volume it processed, then how can we measure the throughput?

middleware/instrument.go
```diff
@@ -105,6 +110,17 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
+					break
+				}
+			}
+			i.SlowRequestServerThroughput.WithLabelValues(r.Method, route).Observe(float64(volume) / respMetrics.Duration.Seconds())
```
dimitarvdimitrov, line 122: can we observe with exemplar like above?

middleware/instrument.go
```diff
@@ -105,6 +110,17 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 			labelValues = append(labelValues, tenantID)
 			instrument.ObserveWithExemplar(r.Context(), i.PerTenantDuration.WithLabelValues(labelValues...), respMetrics.Duration.Seconds())
 		}
+		if i.SlowRequestCutoff > 0 && respMetrics.Duration > i.SlowRequestCutoff {
```
dimitarvdimitrov, line 113: in the original (private :( ) issue it's discussed to have the cutoff be the volume ("N samples") instead of latency. Should this be the other way around?

server/metrics.go
```diff
@@ -73,5 +75,14 @@ func NewServerMetrics(cfg Config) *Metrics {
 			Name:      "inflight_requests",
 			Help:      "Current number of inflight requests.",
 		}, []string{"method", "route"}),
+		SlowRequestServerThroughput: reg.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace:                       cfg.MetricsNamespace,
+			Name:                            "slow_request_server_throughput_" + cfg.ThroughputConfig.Unit,
```
dimitarvdimitrov, line 80: I am not sure about the `_server_` name. Doesn't everything measured here refer to the server? (same goes for for the field name)

middleware/instrument.go
```diff
@@ -105,6 +110,17 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
+			parts := strings.Split(w.Header().Get("Server-Timing"), ", ")
```
dimitarvdimitrov, line 114: can you add some tests for this parsing? I see that there is no `instrument_test.go`, but I think how we handle some edge cases especially when parsing these headers is important to put in a test.

Since here the output is in Prometheus metrics, it's may not immediately obvious how to write tests against them (it def. wasn't to me when I started). Here's an example which asserts on the metrics the code generates.

https://github.com/grafana/dskit/blob/90d7ee0bed7c54bbc7f9792f1f1e687073b919b9/log/ratelimit_test.go#L17-L37

server/server.go
dimitarvdimitrov, line 220: can you mention that the unit is also expected in the `Server-Timing` response header?

server/server.go
```diff
@@ -153,6 +153,13 @@ type Config struct {
 
 	// This limiter is called for every started and finished gRPC request.
 	GrpcMethodLimiter GrpcInflightMethodLimiter `yaml:"-"`
+
+	ThroughputConfig ThroughputConfig `yaml:"throughput_config"`
```
dimitarvdimitrov, line 157: `throughput_config` is the name of the config. The `_config` suffix is a bit redundant.

dimitarvdimitrov (commented):
Thanks for opening!

The only thing I'm not sure about is the cutoff in milliseconds. I don't think that's what the original issue discussed, can you double-check?

It would also be nice to see some tests, the logic is getting a bit more complicated and shouldn't stay untested.

middleware/instrument.go
```diff
@@ -112,14 +112,16 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 		}
 		if i.SlowRequestCutoff > 0 && respMetrics.Duration > i.SlowRequestCutoff {
 			parts := strings.Split(w.Header().Get("Server-Timing"), ", ")
-			volume := int64(0)
-			for _, part := range parts {
-				if strings.HasPrefix(part, i.ServerThroughputUnit) {
-					_, _ = fmt.Sscanf(part, i.ServerThroughputUnit+"=%d", &volume)
-					break
+			if len(parts) == 0 {
```
dimitarvdimitrov, line 115:
```suggestion
			if len(parts) > 0 {
```

dimitarvdimitrov: but now I'm also not sure if we need this line at all. Do we?

middleware/instrument.go
```diff
@@ -112,14 +112,16 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
+				volume := int64(0)
+				for _, part := range parts {
+					if strings.HasPrefix(part, i.ThroughputUnit) {
+						_, _ = fmt.Sscanf(part, i.ThroughputUnit+"=%d", &volume)
```
dimitarvdimitrov, line 119: here if the format is wrong (e.g. `samples_total=foo`), we'd observe throughput of `0`. What do you think about not observing the throughput at all?

server/server.go
dimitarvdimitrov: I think this is still unresolved

middleware/instrument_test.go
```diff
@@ -0,0 +1,148 @@
...
+	tests := []struct {
+		name     string
+		sleep    bool
+		header   string
+		observed bool
+	}{
+		{"WithSleep", true, "unit=0, other_unit=2", true},
+		{"WithoutSleep", false, "unit=0, other_unit=2", false},
+		{"WithoutSleep", true, "", false},
+		{"WithoutSleep", false, "", false},
```
dimitarvdimitrov, line 29: can you update the names of these tests?

middleware/instrument_test.go
```diff
@@ -0,0 +1,148 @@
...
+		})
+	}
+}
+
+func NewInstrument(registry *prometheus.Registry) Instrument {
+	reg := promauto.With(registry)
+
+	const metricsNativeHistogramFactor = 1.1
+	const throughputUnit = "unit"
+	const SlowRequestCutoff = 100 * time.Millisecond
```
dimitarvdimitrov, line 74:
```suggestion
	const slowRequestCutoff = 100 * time.Millisecond
```

capital letters mean exported members in go. Since this is a constant in a function, then it doesn't make sense to export it.

middleware/instrument_test.go
```diff
@@ -0,0 +1,148 @@
...
+				request_throughput_unit_count{cutoff_ms="100",method="GET",route="other"} 1
+			`
+			}
+
+			require.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(output), "slow_request_throughput_"+i.ThroughputUnit))
+		})
+	}
+}
+
+func NewInstrument(registry *prometheus.Registry) Instrument {
```
dimitarvdimitrov, line 69:
```suggestion
func newInstrument(registry *prometheus.Registry) Instrument {
```

this is only a test, so it doesn't make sense to export it

middleware/instrument_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func TestExtractValueFromMultiValueHeader(t *testing.T) {
+	tests := []struct {
+		header   string
+		key      string
+		expected int64
+		err      bool
+	}{
+		{"key0=0, key1=1", "key0", 0, false},
```
dimitarvdimitrov, line 128: can you add keys to each struct initialization? 
```suggestion
		{header: "key0=0, key1=1", key: "key0", expected: 0, err: false},
```

maybe putting them on separate lines might make this more readable


```suggestion
		{
		   header: "key0=0, key1=1",
		   key: "key0",
		   expected: 0,
		   err: false,
		 },
```

server/server.go
```diff
@@ -153,6 +153,13 @@ type Config struct {
 
 	// This limiter is called for every started and finished gRPC request.
 	GrpcMethodLimiter GrpcInflightMethodLimiter `yaml:"-"`
+
+	Throughput Throughput `yaml:"throughput"`
```
dimitarvdimitrov, line 157: i also prefer nested config, but i noticed that most of the other configs in server.Config are flattened (LogLevel, LogFormat instead of Log.Level, Log.Format). I think it's better to stay consistent because this is a shared library

so can you just move the two fields and prefix them with `Throughput`?

server/server.go
```diff
@@ -209,6 +216,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.LogRequestExcludeHeadersList, "server.log-request-headers-exclude-list", "", "Comma separated list of headers to exclude from loggin. Only used if server.log-request-headers is true.")
 	f.BoolVar(&cfg.LogRequestAtInfoLevel, "server.log-request-at-info-level-enabled", false, "Optionally log requests at info level instead of debug level. Applies to request headers as well if server.log-request-headers is enabled.")
 	f.BoolVar(&cfg.ProxyProtocolEnabled, "server.proxy-protocol-enabled", false, "Enables PROXY protocol.")
+	f.DurationVar(&cfg.Throughput.RequestCutoff, "server.throughput.request-cutoff", 0, "Duration after which a request will be observed. For requests taking longer than this duration to finish, the throughput will be calculated. If set to 0, the throughput will not be calculated.")
```
dimitarvdimitrov, line 219: when reading this in the context of all the config in `server.Config` I am not sure what request cutoff does. Can you clarify how it interacts with the `Server-Timing` header, the unit, and the `request_throughput` histogram?

dimitarvdimitrov (commented):
mostly LGTM, although looking at the Server-Timing spec, i think we're missing some cases in how it can be formatted.

middleware/instrument_test.go
```diff
@@ -0,0 +1,148 @@
...
+		{"key0=0, key1=1", "key1", 1, false},
+		{"key0=0, key1=1", "key2", 0, true},
```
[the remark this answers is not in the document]
tinitiuset: This will be refactored to follow the standard `Server-Timing` syntax.

middleware/instrument.go
```diff
@@ -111,28 +111,37 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
 	if len(parts) == 0 {
-		return 0, fmt.Errorf("no a multi-value header")
+		return 0, fmt.Errorf("not a multi-value header")
 	}
-	value := int64(0)
 	for _, part := range parts {
-		if strings.HasPrefix(part, key) {
-			_, err := fmt.Sscanf(part, key+"=%d", &value)
-			return value, err
+		if strings.HasPrefix(part, name) {
```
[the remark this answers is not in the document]
dimitarvdimitrov: the same applies to the `if` 4 lines down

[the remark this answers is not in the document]
dimitarvdimitrov: also, did you considering using `strings.CutPrefix`. It might be a better fit for this use case

middleware/instrument_test.go
```diff
@@ -18,22 +18,42 @@ import (
...
+			sleep:    true,
+			header:   "",
+			observed: false,
+		},
+		{
+			testName: "WithoutSleepEmptyHeader",
+			sleep:    false,
+			header:   "",
+			observed: false,
+		},
```
dimitarvdimitrov, line 49: let's add one where the header doesn't contain the unit too; that should be a common case

server/server.go
```diff
@@ -216,7 +216,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.LogRequestExcludeHeadersList, "server.log-request-headers-exclude-list", "", "Comma separated list of headers to exclude from loggin. Only used if server.log-request-headers is true.")
 	f.BoolVar(&cfg.LogRequestAtInfoLevel, "server.log-request-at-info-level-enabled", false, "Optionally log requests at info level instead of debug level. Applies to request headers as well if server.log-request-headers is enabled.")
 	f.BoolVar(&cfg.ProxyProtocolEnabled, "server.proxy-protocol-enabled", false, "Enables PROXY protocol.")
-	f.DurationVar(&cfg.Throughput.RequestCutoff, "server.throughput.request-cutoff", 0, "Duration after which a request will be observed. For requests taking longer than this duration to finish, the throughput will be calculated. If set to 0, the throughput will not be calculated.")
+	f.DurationVar(&cfg.Throughput.RequestCutoff, "server.throughput.request-cutoff", 0, "Requests taking over cutoff will be observed to measure Throughput. Server-Timing header will be used with specified unit as the indicator, for example 'Server-Timing: unit;val=8.2'. If set to 0, the throughput will not be calculated.")
```
dimitarvdimitrov, line 219: the example shows a float value, but the implementation supports only integer values

middleware/instrument_test.go
```diff
@@ -104,45 +117,89 @@ func NewInstrument(registry *prometheus.Registry) Instrument {
...
+		{
+			testName: "ExistantKeyInName1",
+			header:   "name0;key0=0.0;key1=1.1, name1;key0=1.1",
+			name:     "name0",
+			key:      "key0",
+			expected: 0.0,
+			err:      false,
+		},
+		{
+			testName: "NonExistantName1",
```
dimitarvdimitrov, line 149: did you mean `NonExistantName2`?

middleware/instrument_test.go
```diff
@@ -104,45 +117,89 @@ func NewInstrument(registry *prometheus.Registry) Instrument {
...
+			header:   "name0;key0=0.0;key1=1.1, name1;key0=1.1",
+			name:     "name2",
+			key:      "key0",
+			expected: 0.0,
+			err:      true,
+		},
+		{
+			testName: "ExistantKeyInName2",
```
dimitarvdimitrov, line 157: did you mean `*Name1`?

maybe i'm missing the naming convention 🙈

server/server.go
dimitarvdimitrov, line 219:
```suggestion
f.DurationVar(&cfg.Throughput.RequestCutoff, "server.throughput.request-cutoff", 0, "Requests taking over the cutoff are be observed to measure throughput. Server-Timing header is used with specified unit as the indicator, for example 'Server-Timing: unit;val=8.2'. If set to 0, the throughput is not calculated.")
```

also re `RequestCutoff` - do we want this to be `LatencyCutoff` instead? what do you think?

server/server.go
```diff
@@ -216,7 +216,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 	f.StringVar(&cfg.Throughput.Unit, "server.throughput.unit", "total_samples", "Unit of the server throughput metric, for example 'processed_bytes' or 'total_samples'. Observed values will be gathered from Server-Timing header with defined key. If set, it is appended to the request_server_throughput metric name.")
```
dimitarvdimitrov, line 220: can you use present tense in the helpo text?


```suggestion
	f.StringVar(&cfg.Throughput.Unit, "server.throughput.unit", "total_samples", "Unit of the server throughput metric, for example 'processed_bytes' or 'total_samples'. Observed values are gathered from the 'Server-Timing' header with the 'val' key. If set, it is appended to the request_server_throughput metric name.")
```

middleware/instrument_test.go
tinitiuset: There are two tests for `NonExistantName`, `NonExistantName1` and `NonExistantName2`

middleware/instrument_test.go
tinitiuset: There are two tests for `ExistantKeyInName`, `ExistantKeyInName1` and `ExistantKeyInName2`

middleware/instrument_test.go
dimitarvdimitrov: i see. maybe it will be less confusing if the keys and the test names used different strings

dimitarvdimitrov (approved):
LGTM, thanks for going through all the rounds of reviews 💪 Kudos for working on this!

PR #620, opened 2024-11-18
ring: add GetWithOptions method to adjust per call behavior

56quarters:
**What this PR does**:

This change adds a new method that accepts 0 or more `Option` instances that modify the behavior of the call. These options can (currently) be used to adjust the replication factor for a particular key or use buffers to avoid excessive allocation.

Part of grafana/mimir#9944

**Checklist**
- [ ] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -34,13 +34,50 @@ const (
 	GetBufferSize = 5
 )
 
+type Options struct {
+	ReplicationFactor int
+	BufDescs          []InstanceDesc
+	BufHosts          []string
+	BufZones          []string
+}
+
+type Option func(opts *Options)
+
+func WithBuffers(bufDescs []InstanceDesc, bufHosts, bufZones []string) Option {
+	return func(opts *Options) {
+		opts.BufDescs = bufDescs
+		opts.BufHosts = bufHosts
+		opts.BufZones = bufZones
+	}
+}
+
+func WithReplicationFactor(replication int) Option {
+	return func(opts *Options) {
+		opts.ReplicationFactor = replication
+	}
+}
+
+func collectOptions(opts ...Option) *Options {
+	final := &Options{}
+	for _, opt := range opts {
+		opt(final)
+	}
+	return final
+}
+
 // ReadRing represents the read interface to the ring.
 // Support for read-only instances requires use of ShuffleShard or ShuffleShardWithLookback prior to getting a ReplicationSet.
 type ReadRing interface {
 	// Get returns n (or more) instances which form the replicas for the given key.
+	//
 	// bufDescs, bufHosts and bufZones are slices to be overwritten for the return value
 	// to avoid memory allocation; can be nil, or created with ring.MakeBuffersForGet().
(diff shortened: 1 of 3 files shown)


ring/ring.go
```diff
@@ -34,21 +34,58 @@ const (
...
+	}
+}
+
+func collectOptions(opts ...Option) *Options {
+	final := &Options{}
+	for _, opt := range opts {
+		opt(final)
+	}
+	return final
+}
```
dimitarvdimitrov, line 66: we had some benchmarks about not allocating on `Get`. The plan is not to use this GetWithOptions on the hot path, right? (but also maybe just returning an `Options` might do away with the allocation)

56quarters: I'll see if I can improve this since it'd be nice to clean it up:

```
ring/ring.go:46:18: leaking param: bufDescs
ring/ring.go:46:43: leaking param: bufHosts
ring/ring.go:46:53: leaking param: bufZones
ring/ring.go:47:14: opts does not escape
ring/ring.go:47:9: func literal escapes to heap
ring/ring.go:55:14: opts does not escape
ring/ring.go:55:9: func literal escapes to heap
ring/ring.go:60:21: opts does not escape
ring/ring.go:61:11: &Options{} escapes to heap
```

56quarters:
I wasn't able to avoid these functions escaping to the heap. Benchmarking the two methods shows that `GetWithOptions` is about 14% slower. I'm going to leave these as mostly separate methods and work on adding tests to `GetWithOptions`. After that I'll mark this as ready for review and we can start experimenting with higher replication factors in Mimir.

```
$ benchstat old.txt new.txt 
goos: linux
goarch: amd64
                                                         │   old.txt    │                new.txt                 │
                                                         │    sec/op    │    sec/op      vs base                 │
Ring_Get/without_zone_awareness-16                         863.2n ± ∞ ¹   1009.0n ± ∞ ¹        ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness,_not_enough_instances-16   654.2n ± ∞ ¹    758.2n ± ∞ ¹        ~ (p=1.000 n=1) ²
Ring_Get/with_zone_awareness-16                            1.138µ ± ∞ ¹    1.252µ ± ∞ ¹        ~ (p=1.000 n=1) ²
Ring_Get/one_excluded_zone-16                              855.0n ± ∞ ¹    962.0n ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                                    861.0n          979.7n        +13.80%
```

56quarters:
After adding unit tests for this change, it turns out to not be compatible with zone-awareness. There's an assumption in the ring that only a single instance per zone is returned during quorum checks. The result of this is that there are `Ring.Get()` calls that _should_ fail when not enough zones are available that would now succeed since we return multiple instances per zone and the existing code only checks the number of instances, not zones.

56quarters:
I'm going to close this PR. The changes to the ring required for https://github.com/grafana/mimir/issues/9944 are more complicated than expected. I'll put the various concerns into a design doc for some discussion with the Mimir team and other database teams as well.

PR #624, opened 2024-12-03
Faster tests using the mocked consul in-memory backend

pracucci:
**What this PR does**:

I was investigating some slow unit tests in Mimir and I noticed they're slow because of the slow long polling done by `WatchKey()` on the mocked consul in-memory backend. In https://github.com/cortexproject/cortex/pull/2078 I introduced `mockedMaxWaitTime()` to have faster tests. In this PR I propose to further reduce the max wait interval to 100ms. To get faster tests we also have to increase the `m.cond.Broadcast()` frequency too (done in this PR too).

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/kv/consul/mock.go
+++ b/kv/consul/mock.go
@@ -15,6 +15,9 @@ import (
 	"github.com/grafana/dskit/kv/codec"
 )
 
+// The max wait time allowed for mockKV operations, in order to have faster tests.
+const maxWaitTime = 100 * time.Millisecond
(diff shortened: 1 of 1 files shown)


56quarters (approved):

dimitarvdimitrov:
❤️ i wanted to do something like this in https://github.com/grafana/dskit/pull/519, but eventually gave up

PR #540, opened 2024-07-03
Update minimum go version to 1.21

pracucci:
**What this PR does**:

In dskit we aim to keep support for the last 2 stable go version. Currently dskit requires 1.20. I propose to upgrade the requirement to 1.21. In this PR I'm also updating drone to run CI against latest 1.21 minor release.

**Which issue(s) this PR fixes**:

N/A

**Checklist**
- [ ] Tests updated
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #626, opened 2024-12-16
Replace `exp/slices` with `slices` from standard library

Juneezee:
**What this PR does**:

https://github.com/grafana/dskit/blob/e27df29220ea69cf7335f653da18b8ea51ab27e4/ring/token_range.go#L7

We upgraded the minimum Go version to 1.21 in PR https://github.com/grafana/dskit/pull/540.

The experimental functions in `golang.org/x/exp/slices` are now included in the standard library since Go 1.21.

Reference: https://go.dev/doc/go1.21#slices

**Checklist**
- [N/A] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## Changelog
 
-* [CHANGE] Replace `exp/slices` package with `slices` from standard library. #626
 * [CHANGE] Add new metric `slow_request_server_throughput` to track the throughput of slow queries. #619
 * [CHANGE] Log middleware updated to honor `logRequestHeaders` in all logging scenarios. #615
 * [CHANGE] Roll back the gRPC dependency to v1.65.0 to allow downstream projects to avoid a performance regression and maybe a bug in v1.66.0. #581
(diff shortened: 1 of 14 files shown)


loser/loser.go
```diff
@@ -2,9 +2,9 @@
 
 package loser
 
-import "golang.org/x/exp/constraints"
+import "cmp"
 
-func New[E constraints.Ordered](lists [][]E, maxVal E) *Tree[E] {
+func New[E cmp.Ordered](lists [][]E, maxVal E) *Tree[E] {
```
Juneezee, line 7: The `constraints.Ordered` from `golang.org/x/exp/constraints` is also included into Go 1.21 as `cmp.Ordered`. See https://pkg.go.dev/cmp#Ordered

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 
+* [CHANGE] Replace `exp/slices` package with `slices` from standard library. #626
```
aknuds1, line 3: There's no need for a changelog item, as the PR doesn't have any user visible effects, it's pure refactoring.
```suggestion
```

aknuds1 (commented):
LGTM except there shouldn't be a changelog item.

aknuds1 (approved):
LGTM, thanks!

PR #630, opened 2024-12-30
Chore: log last heartbeat timestamp along with "instance found in the ring"

pracucci:
**What this PR does**:

Few weeks ago I was investigating an issue and I needed to know how long an instance didn't heartbeat the ring during a rolling update (the instance didn't leave the ring at shutdown). This is an information that's missing. In this PR I propose to log it.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/ring/basic_lifecycler.go
+++ b/ring/basic_lifecycler.go
@@ -294,7 +294,7 @@ func (l *BasicLifecycler) registerInstance(ctx context.Context) error {
 		var exists bool
 		instanceDesc, exists = ringDesc.Ingesters[l.cfg.ID]
 		if exists {
-			level.Info(l.logger).Log("msg", "instance found in the ring", "instance", l.cfg.ID, "ring", l.ringName, "state", instanceDesc.GetState(), "tokens", len(instanceDesc.GetTokens()), "registered_at", instanceDesc.GetRegisteredAt().String())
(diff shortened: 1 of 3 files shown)


ring/basic_lifecycler.go
```diff
@@ -294,7 +294,7 @@ func (l *BasicLifecycler) registerInstance(ctx context.Context) error {
 		var exists bool
 		instanceDesc, exists = ringDesc.Ingesters[l.cfg.ID]
 		if exists {
-			level.Info(l.logger).Log("msg", "instance found in the ring", "instance", l.cfg.ID, "ring", l.ringName, "state", instanceDesc.GetState(), "tokens", len(instanceDesc.GetTokens()), "registered_at", instanceDesc.GetRegisteredAt().String())
+			level.Info(l.logger).Log("msg", "instance found in the ring", "instance", l.cfg.ID, "ring", l.ringName, "state", instanceDesc.GetState(), "tokens", len(instanceDesc.GetTokens()), "registered_at", instanceDesc.GetRegisteredAt().String(), "last_heartbeat_at", instanceDesc.GetLastHeartbeatAt().String())
```
colega, line 297: No need to send a commit, but a future nit: logger will always call the Stringer's `String()` method if there's such, so no need to call it here.

colega (approved):

PR #631, opened 2025-01-02
chore: Display token information in partition ring status page

alexweav:
**What this PR does**:

The partition ring has a concept of tokens, but its tokens aren't visible in the partition ring status UI.

This PR ports the token-related affordances (count, ownership%, view all) from the ingester ring status page to the partition ring status page.

Screenshot of new columns:
2025-01-02-165849_2553x822_scrot

When "Show Tokens" at the bottom is clicked:
2025-01-02-165908_2553x819_scrot

Contrib https://github.com/grafana/mimir-squad/issues/2350

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_ring_model.go
+++ b/ring/partition_ring_model.go
@@ -94,6 +94,34 @@ func (m *PartitionRingDesc) partitionByToken() map[Token]int32 {
 	return out
 }
 
+// CountTokens returns the summed token distance of all tokens in each partition.
+func (m *PartitionRingDesc) countTokens() map[int32]int64 {
+	owned := make(map[int32]int64, len(m.Partitions))
+	sortedTokens := m.tokens()
+	tokensToPartitions := m.partitionByToken()
+
+	for i, token := range sortedTokens {
+		partition := tokensToPartitions[Token(token)]
+
+		var prevToken uint32
+		if i == 0 {
+			prevToken = sortedTokens[len(sortedTokens)-1]
+		} else {
+			prevToken = sortedTokens[i-1]
+		}
+		diff := tokenDistance(prevToken, token)
+		owned[partition] = owned[partition] + diff
+	}
+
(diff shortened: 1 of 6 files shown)


ring/partition_ring_status.gohtml
```diff
@@ -15,6 +15,8 @@
                 <th>State</th>
                 <th>State updated at</th>
                 <th>Owners</th>
+                <th>Tokens</th>
```
alexweav, line 18: Since we usually [only show the Partition Ring page in ingest-storage mode](https://github.com/grafana/mimir/blob/27a7389510b3758812dfd7cb652147a22881ac58/pkg/mimir/modules.go#L378), I chose not to add any extra toggle to show or hide these values. In contexts where these values don't make sense, the entire page is usually inaccessible.

ring/partition_ring_model.go
```diff
@@ -94,6 +94,34 @@ func (m *PartitionRingDesc) partitionByToken() map[Token]int32 {
 	return out
 }
 
+// CountTokens returns the summed token distance of all tokens in each partition.
+func (m *PartitionRingDesc) countTokens() map[int32]int64 {
```
alexweav, line 98: Chose this name for consistency with its corresponding operation on the instance ring: https://github.com/grafana/dskit/blob/9935aca9d266be5de05450afd8278d99637a24b6/ring/ring.go#L625

jhesketh (approved):

pracucci (approved):
Nice work, thanks!

alexweav:
I noticed in the second screenshot the header was misleading, it reads `Instance 0` when it would be better to read `Partition 0`.
2025-01-03-095825_95x78_scrot

Pushed a commit changing that. It's tiny, so i don't think that warrants resetting reviews. I'll leave this open for a bit longer in case of any objections, then m erge.

PR #632, opened 2025-01-03
ring: add GetWithOptions method to adjust per call behavior

56quarters:
**What this PR does**:

This change adds a new method that accepts 0 or more `Option` instances that modify the behavior of the call. These options can (currently) be used to adjust the replication factor for a particular key or use buffers to avoid excessive allocation.

The most notable changes are in the `Ring.findInstancesForKey` method which is the core of the `Ring.Get` method. Instead of keeping track of distinct zones and assuming that only a single instance per zone would ever be returned, we keep a map of the number of instances found in each zone.

Part of grafana/mimir#9944

Previous attempt https://github.com/grafana/dskit/pull/620

**Checklist**
- [X] Tests updated

### DIFF
--- a/ring/replication_strategy.go
+++ b/ring/replication_strategy.go
@@ -7,10 +7,15 @@ import (
 )
 
 type ReplicationStrategy interface {
-	// Filter out unhealthy instances and checks if there're enough instances
+	// Filter out unhealthy instances and checks if there are enough instances
 	// for an operation to succeed. Returns an error if there are not enough
 	// instances.
 	Filter(instances []InstanceDesc, op Operation, replicationFactor int, heartbeatTimeout time.Duration, zoneAwarenessEnabled bool) (healthy []InstanceDesc, maxFailures int, err error)
+
+	// SupportsExpandedReplication returns true for replication strategies that
+	// support increasing the replication factor beyond a single instance per zone,
+	// false otherwise.
+	SupportsExpandedReplication() bool
 }
 
 type defaultReplicationStrategy struct{}
@@ -70,6 +75,14 @@ func (s *defaultReplicationStrategy) Filter(instances []InstanceDesc, op Operati
 	return instances, len(instances) - minSuccess, nil
 }
 
+func (s *defaultReplicationStrategy) SupportsExpandedReplication() bool {
+	// defaultReplicationStrategy assumes that a single instance per zone is returned and that
+	// it can treat replication factor as equivalent to the number of zones. This doesn't work
+	// when a per-call replication factor increases it beyond the configured replication factor
+	// and the number of zones.
+	return false
+}
+
 type ignoreUnhealthyInstancesReplicationStrategy struct{}
 
 func NewIgnoreUnhealthyInstancesReplicationStrategy() ReplicationStrategy {
@@ -99,6 +112,10 @@ func (r *ignoreUnhealthyInstancesReplicationStrategy) Filter(instances []Instanc
 	}
 
 	return instances, len(instances) - 1, nil
+}
+
+func (r *ignoreUnhealthyInstancesReplicationStrategy) SupportsExpandedReplication() bool {
+	return true
 }
 
 func (r *Ring) IsHealthy(instance *InstanceDesc, op Operation, now time.Time) bool {
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -34,6 +34,41 @@ const (
 	GetBufferSize = 5
 )
 
+// Options are the result of Option instances that can be used to modify Ring.GetWithOptions behavior.
+type Options struct {
+	ReplicationFactor int
+	BufDescs          []InstanceDesc
+	BufHosts          []string
+	BufZones          []string
+}
+
+// Option can be used to modify Ring behavior when calling Ring.GetWithOptions
+type Option func(opts *Options)
+
+// WithBuffers creates an Option that will cause the given buffers to be used, avoiding allocations.
+func WithBuffers(bufDescs []InstanceDesc, bufHosts, bufZones []string) Option {
+	return func(opts *Options) {
+		opts.BufDescs = bufDescs
+		opts.BufHosts = bufHosts
+		opts.BufZones = bufZones
+	}
+}
+
+// WithReplicationFactor creates an Option that overrides the default replication factor for a single call.
+func WithReplicationFactor(replication int) Option {
+	return func(opts *Options) {
+		opts.ReplicationFactor = replication
+	}
+}
+
+func collectOptions(opts ...Option) Options {
+	final := Options{}
+	for _, opt := range opts {
+		opt(&final)
+	}
+	return final
+}
+
 // ReadRing represents the read interface to the ring.
 // Support for read-only instances requires use of ShuffleShard or ShuffleShardWithLookback prior to getting a ReplicationSet.
 type ReadRing interface {
@@ -42,13 +77,17 @@ type ReadRing interface {
 	// to avoid memory allocation; can be nil, or created with ring.MakeBuffersForGet().
 	Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts, bufZones []string) (ReplicationSet, error)
 
+	// GetWithOptions returns n (or more) instances which form the replicas for the given key
+	// with 0 or more Option instances to change the behavior of the method call.
+	GetWithOptions(key uint32, op Operation, opts ...Option) (ReplicationSet, error)
+
 	// GetAllHealthy returns all healthy instances in the ring, for the given operation.
 	// This function doesn't check if the quorum is honored, so doesn't fail if the number
 	// of unhealthy instances is greater than the tolerated max unavailable.
 	GetAllHealthy(op Operation) (ReplicationSet, error)
 
 	// GetReplicationSetForOperation returns all instances where the input operation should be executed.
-	// The resulting ReplicationSet doesn't necessarily contains all healthy instances
+	// The resulting ReplicationSet doesn't necessarily contain all healthy instances
 	// in the ring, but could contain the minimum set of instances required to execute
 	// the input operation.
 	GetReplicationSetForOperation(op Operation) (ReplicationSet, error)
@@ -424,19 +463,44 @@ func (r *Ring) setRingStateFromDesc(ringDesc *Desc, updateMetrics, updateRegiste
 }
 
 // Get returns n (or more) instances which form the replicas for the given key.
-func (r *Ring) Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts, bufZones []string) (ReplicationSet, error) {
+func (r *Ring) Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts, _ []string) (ReplicationSet, error) {
+	// Note that we purposefully aren't calling GetWithOptions here since the closures it
+	// uses result in heap allocations which we specifically avoid in this method since it's
+	// called in hot loops.
+	return r.getReplicationSetForKey(key, op, bufDescs, bufHosts, r.cfg.ReplicationFactor)
+}
+
+// GetWithOptions returns n (or more) instances which form the replicas for the given key
+// with 0 or more options to change the behavior of the method call.
+func (r *Ring) GetWithOptions(key uint32, op Operation, opts ...Option) (ReplicationSet, error) {
+	options := collectOptions(opts...)
+	return r.getReplicationSetForKey(key, op, options.BufDescs, options.BufHosts, options.ReplicationFactor)
+}
+
+func (r *Ring) getReplicationSetForKey(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts []string, replicationFactor int) (ReplicationSet, error) {
 	r.mtx.RLock()
 	defer r.mtx.RUnlock()
 	if r.ringDesc == nil || len(r.ringTokens) == 0 {
 		return ReplicationSet{}, ErrEmptyRing
 	}
 
-	instances, err := r.findInstancesForKey(key, op, bufDescs, bufHosts, bufZones, nil)
+	if replicationFactor <= 0 || replicationFactor < r.cfg.ReplicationFactor {
+		replicationFactor = r.cfg.ReplicationFactor
+	}
+
+	// Not all replication strategies support increasing the replication factor beyond
+	// the number of zones available. Return an error unless a ReplicationStrategy has
+	// explicitly opted into supporting this.
+	if replicationFactor > r.cfg.ReplicationFactor && !r.strategy.SupportsExpandedReplication() {
+		return ReplicationSet{}, fmt.Errorf("per-call replication factor %d cannot exceed the configured replication factor %d with this replication strategy", replicationFactor, r.cfg.ReplicationFactor)
+	}
+
+	instances, err := r.findInstancesForKey(key, op, bufDescs, bufHosts, replicationFactor, nil)
 	if err != nil {
 		return ReplicationSet{}, err
 	}
 
-	healthyInstances, maxFailure, err := r.strategy.Filter(instances, op, r.cfg.ReplicationFactor, r.cfg.HeartbeatTimeout, r.cfg.ZoneAwarenessEnabled)
+	healthyInstances, maxFailure, err := r.strategy.Filter(instances, op, replicationFactor, r.cfg.HeartbeatTimeout, r.cfg.ZoneAwarenessEnabled)
 	if err != nil {
 		return ReplicationSet{}, err
 	}
@@ -450,9 +514,9 @@ func (r *Ring) Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts, 
 // Returns instances for given key and operation. Instances are not filtered through ReplicationStrategy.
 // InstanceFilter can ignore uninteresting instances that would otherwise be part of the output, and can also stop search early.
 // This function needs to be called with read lock on the ring.
-func (r *Ring) findInstancesForKey(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts []string, bufZones []string, instanceFilter func(instanceID string) (include, keepGoing bool)) ([]InstanceDesc, error) {
+func (r *Ring) findInstancesForKey(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts []string, replicationFactor int, instanceFilter func(instanceID string) (include, keepGoing bool)) ([]InstanceDesc, error) {
 	var (
-		n            = r.cfg.ReplicationFactor
+		n            = replicationFactor
 		instances    = bufDescs[:0]
 		start        = searchToken(r.ringTokens, key)
 		iterations   = 0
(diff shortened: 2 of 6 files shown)


ring/ring.go
```diff
@@ -447,21 +511,31 @@ func (r *Ring) Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts,
 // Returns instances for given key and operation. Instances are not filtered through ReplicationStrategy.
 // InstanceFilter can ignore uninteresting instances that would otherwise be part of the output, and can also stop search early.
 // This function needs to be called with read lock on the ring.
-func (r *Ring) findInstancesForKey(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts []string, bufZones []string, instanceFilter func(instanceID string) (include, keepGoing bool)) ([]InstanceDesc, error) {
+func (r *Ring) findInstancesForKey(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts []string, bufZones []string, replicationFactor int, instanceFilter func(instanceID string) (include, keepGoing bool)) ([]InstanceDesc, error) {
```
56quarters, line 514: `bufZones` is no longer used. However, it's part of the method signature of `Ring.Get()`. I'd like to replace all the buffer related parameters with a single `Buffers` struct and pass that around. That would avoid the need to change the `Ring.Get` method signature every time the implementation changes.

ring/ring.go
```diff
@@ -1332,36 +1420,3 @@ func (op Operation) ShouldExtendReplicaSetOnState(s InstanceState) bool {
 
 // All states are healthy, no states extend replica set.
 var allStatesRingOperation = Operation(0x0000ffff)
-
-// numberOfKeysOwnedByInstance returns how many of the supplied keys are owned by given instance.
-func (r *Ring) numberOfKeysOwnedByInstance(keys []uint32, op Operation, instanceID string, bufDescs []InstanceDesc, bufHosts []string, bufZones []string) (int, error) {
```
56quarters, line 1337: This was only called from tests. I've moved most of this logic to the test in question (`token_range_test.go`).

ring/ring_test.go
```diff
@@ -627,27 +628,31 @@ func TestRing_Get_ZoneAwareness(t *testing.T) {
 		"should succeed if there are enough instances per zone on RF = 3": {
 			numInstances:         16,
 			numZones:             3,
+			totalZones:           3,
 			replicationFactor:    3,
 			zoneAwarenessEnabled: true,
 			expectedInstances:    3,
 		},
 		"should fail if there are instances in 1 zone only on RF = 3": {
```
56quarters, line 636: This test will fail with the code changes in this PR _unless_ we create at least a single instance in each zone (without tokens) below. The code previously assumed that replication factor was equivalent to the total number of zones. However, that's not the case anymore since replication factor can be higher than the number of zones. Thus, we need to create an instance in each zone so that `ringTokensByZone` gets updated and has the correct number of zones.

56quarters, line 636: Without the change to add an instance in each zone below, this test would fail.

Previously, `findInstancesForKey` ignored the actual number of zones, `len(r.ringTokensByZone)`, and assumed that replication factor was equivalent. Now that we've changed things so that replication factor is not necessarily the same as the number of zones, we need to add at least one instance in each zone so that `findInstancesForKey` is aware of them and only picks a single instance from each zone.

ring/ring.go
```diff
@@ -447,21 +511,31 @@ func (r *Ring) Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts,
...
 		maxInstances = len(r.ringDesc.Ingesters)
 
 		// We use a slice instead of a map because it's faster to search within a
-		// slice than lookup a map for a very low number of items.
+		// slice than lookup a map for a very low number of items, we only expect
+		// to have low single-digit number of hosts.
 		distinctHosts = bufHosts[:0]
-		distinctZones = bufZones[:0]
+
+		// TODO: Do we need to pass this in to avoid allocations?
```
56quarters, line 528: I'd like to replace passing `bufDescs, bufHosts, bufZones` around everywhere with a single struct `Buffers` which could include this map but that was a pretty invasive and non-backwards compatible change (not that we have any guarantees in dskit).

dimitarvdimitrov (commented):
only reviewed the prod changes so far

ring/replication_strategy.go
```diff
@@ -7,10 +7,15 @@ import (
...
-	// Filter out unhealthy instances and checks if there're enough instances
+	// Filter out unhealthy instances and checks if there are enough instances
 	// for an operation to succeed. Returns an error if there are not enough
 	// instances.
 	Filter(instances []InstanceDesc, op Operation, replicationFactor int, heartbeatTimeout time.Duration, zoneAwarenessEnabled bool) (healthy []InstanceDesc, maxFailures int, err error)
+
+	// SupportsExpandedReplication returns true for replication strategies that
+	// support increasing the replication factor beyond a single instance per zone,
+	// false otherwise.
+	SupportsExpandedReplication() bool
```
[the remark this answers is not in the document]
56quarters: Ah, I see what you mean. I'll leave it as-is for now and we can revisit if we end up having more than just this feature we need to test against.

ring/ring.go
```diff
@@ -447,21 +511,31 @@ func (r *Ring) Get(key uint32, op Operation, bufDescs []InstanceDesc, bufHosts,
...
+		hostsPerZone       = make(map[string]int)
```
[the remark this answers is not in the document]
56quarters: Seems like we don't need to pass this in to avoid an allocation :shrug: . In the new version, `Get` ends up being ~7% slower than the old version

```
$ benchstat old.txt new.txt 
goos: linux
goarch: amd64
                                                         │   old.txt    │               new.txt                │
                                                         │    sec/op    │    sec/op     vs base                │
Ring_Get/with_zone_awareness-16                            1.145µ ± ∞ ¹   1.233µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/one_excluded_zone-16                              835.5n ± ∞ ¹   943.5n ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness-16                         881.8n ± ∞ ¹   934.8n ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness,_not_enough_instances-16   647.6n ± ∞ ¹   676.1n ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                                    859.7n         926.0n        +7.71%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                                         │   old.txt   │               new.txt               │
                                                         │    B/op     │    B/op      vs base                │
Ring_Get/with_zone_awareness-16                            0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/one_excluded_zone-16                              0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness-16                         0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness,_not_enough_instances-16   0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                                              ³                +0.00%               ³
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
³ summaries must be >0 to compute geomean

                                                         │   old.txt   │               new.txt               │
                                                         │  allocs/op  │  allocs/op   vs base                │
Ring_Get/with_zone_awareness-16                            0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/one_excluded_zone-16                              0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness-16                         0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness,_not_enough_instances-16   0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                                              ³                +0.00%               ³
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
³ summaries must be >0 to compute geomean
```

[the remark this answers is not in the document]
dimitarvdimitrov: i was more concerned about the added CPU time. 7% isn't a huge deal. In a random cluster at GL, `ring.Get` is only consuming 1.5% of distributor CPU. So this increase won't be material. Is doing a slice with a `struct{ ... }` complicated?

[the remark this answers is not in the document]
dimitarvdimitrov: yes. So previously we had a buffer with zones `[]string`. That was in place of a map because iterating it is faster than accessing a map.

But now we also care about how many instances per zone we have, so you went with a map. To keep the lookup speed of a slice we can have something like this

```go
type zoneInstanceCount struct {
  zone string
  numInstances int
}

hostsPerZone := make([]zoneInstanceCount, 0, 0)

findHostsPerZone := func(zone string) int {
  for _, c := range hostsPerZone {
    if c.zone == zone {
      return c.numInstances
    }
  }
  return 0
}

addInstanceToZone := func(zone string) {
  for i, c := range hostsPerZone {
    if c.zone == zone {
      hostsPerZone[i].numInstances++
    }
  }
}
```

the zoneInstanceCount can be a package type and the funcs can also be proper methods.

`hostsPerZone` can also be taken from a buffer after you make the change you mentioned with a  `Buffers` type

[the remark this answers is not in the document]
56quarters: Pushed a hacky version of this [here](https://github.com/grafana/dskit/commit/cf9fc4e22f7cb08109b913da0002a5e4a1263cf3). It doesn't seem like using slices is any faster so I'm going to leave this as-is.

```
$ benchstat map.txt slice.txt 
goos: linux
goarch: amd64
                                                         │   map.txt    │               slice.txt               │
                                                         │    sec/op    │    sec/op      vs base                │
Ring_Get/without_zone_awareness-16                         893.2n ± ∞ ¹    932.0n ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness,_not_enough_instances-16   670.4n ± ∞ ¹    735.1n ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/with_zone_awareness-16                            1.238µ ± ∞ ¹    1.401µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/one_excluded_zone-16                              943.8n ± ∞ ¹   1023.0n ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                                    914.6n          995.4n        +8.84%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                                         │   map.txt   │              slice.txt              │
                                                         │    B/op     │    B/op      vs base                │
Ring_Get/without_zone_awareness-16                         0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness,_not_enough_instances-16   0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/with_zone_awareness-16                            0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/one_excluded_zone-16                              0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                                              ³                +0.00%               ³
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
³ summaries must be >0 to compute geomean

                                                         │   map.txt   │              slice.txt              │
                                                         │  allocs/op  │  allocs/op   vs base                │
Ring_Get/without_zone_awareness-16                         0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/without_zone_awareness,_not_enough_instances-16   0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/with_zone_awareness-16                            0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
Ring_Get/one_excluded_zone-16                              0.000 ± ∞ ¹   0.000 ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                                              ³                +0.00%               ³
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
³ summaries must be >0 to compute geomean
```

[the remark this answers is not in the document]
dimitarvdimitrov: i had to see this with my own eyes so i ran the benchmarks too. I also ran them multiple times because benchstat advises at least 7 runs. If anything slices are slower 🤯 


```
goos: darwin
goarch: arm64
                                                         │   maps.txt   │             slices.txt              │
                                                         │    sec/op    │   sec/op     vs base                │
Ring_Get/with_zone_awareness-10                            746.2n ± 11%   832.5n ± 7%  +11.55% (p=0.004 n=10)
Ring_Get/one_excluded_zone-10                              589.1n ±  1%   587.0n ± 1%        ~ (p=0.325 n=10)
Ring_Get/without_zone_awareness-10                         482.6n ±  0%   537.3n ± 0%  +11.32% (p=0.000 n=10)
Ring_Get/without_zone_awareness,_not_enough_instances-10   364.7n ±  1%   427.6n ± 3%  +17.25% (p=0.000 n=10)
geomean                                                    527.4n         578.8n        +9.75%

                                                         │   maps.txt   │             slices.txt              │
                                                         │     B/op     │    B/op     vs base                 │
Ring_Get/with_zone_awareness-10                            0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
Ring_Get/one_excluded_zone-10                              0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
Ring_Get/without_zone_awareness-10                         0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
Ring_Get/without_zone_awareness,_not_enough_instances-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                               ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                                                         │   maps.txt   │             slices.txt              │
                                                         │  allocs/op   │ allocs/op   vs base                 │
Ring_Get/with_zone_awareness-10                            0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
Ring_Get/one_excluded_zone-10                              0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
Ring_Get/without_zone_awareness-10                         0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
Ring_Get/without_zone_awareness,_not_enough_instances-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                               ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

ring/ring_test.go
dimitarvdimitrov: hm, in this case then dynamic replication only works for zone-aware replication. Should we add this to the docs and validation code?

ring/ring.go
dimitarvdimitrov: for some reason there are no allocations now; do you want to remove this TODO (i still like the idea of passing `Buffers`)

ring/ring_test.go
```diff
@@ -721,6 +732,199 @@ func TestRing_Get_ZoneAwareness(t *testing.T) {
...
+
+	type testCase struct {
+		name            string
+		ringInstances   map[string]InstanceDesc
+		strategy        ReplicationStrategy
+		options         []Option
+		expectedSetSize int
+		expectError     bool
+	}
+	cases := []testCase{
```
dimitarvdimitrov, line 749: i think this could use some more test cases. The ones I can think of right now are 

- the expanded replication is less than the number of instances, so not all instances are included in the resulting set
- increased replication factor is 6 and there are 6 healthy instances
- increased RF is lower than 6 and there are some unhealthy instances (not sure if that's a novel case; maybe `higher replication factor, ignore unhealthy strategy, some unhealthy` already covers this)
- number of instances per zone aren't the same. e.g. RF=6, we can only select one instance from zone-a and 7 instances from zone-b. Would we compensate and select 3 instances from zone-b because zone-a had only one healthy?

ring/token_range_test.go
```diff
@@ -190,7 +190,7 @@ func TestCheckingOfKeyOwnership(t *testing.T) {
 }
 
 func testCheckingOfKeyOwnership(t *testing.T, randomizeInstanceStates bool) {
```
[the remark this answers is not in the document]
56quarters: tl;dr - to make the test dramatically faster than it was before. 

There was an unexported method added to the `Ring` that was only called from this test. I removed it from the `Ring` and moved to the test file since that's the only place it was used. As part of this I changed it to only call public `Ring` methods. However, going through the public methods which involve locking and doing some other things that aren't needed for the test case made the tests about ~30% slower. So I sped things up by reducing the scale here.

ring/ring_test.go
56quarters: > hm, in this case then dynamic replication only works for zone-aware replication. Should we add this to the docs and validation code?

It still works for non zone-aware replication. It just needs instances with tokens to determine the number of zones, even if the number of zones is one.

ring/ring.go
56quarters: Yeah, I'll remove the TODO. I'm going to handle the `Buffers` change in a separate PR because it's going to be a hassle.

ring/ring_test.go
```diff
@@ -670,6 +675,12 @@ func TestRing_Get_ZoneAwareness(t *testing.T) {
 				prevTokens = append(prevTokens, ingTokens...)
 			}
 
+			// Add instances to the ring that don't own any tokens so that the ring is aware of all zones.
+			for i := testData.numInstances; i < testData.numInstances+testData.totalZones; i++ {
```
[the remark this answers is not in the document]
56quarters: Nope, `zone-76` is definitely wrong. I'll fix. 

In general, this is necessary so that the ring can tell how many zones there are, even when the test only adds instances with tokens in a single zone. The ring previously and currently uses `len(r.ringTokensByZone)` to determine the number of zones.

ring/ring_test.go
56quarters: I'll try to add these today.

56quarters: > i think this could use some more test cases. The ones I can think of right now are
> 
>     * the expanded replication is less than the number of instances, so not all instances are included in the resulting set

Added

>     * increased replication factor is 6 and there are 6 healthy instances

Added

>     * increased RF is lower than 6 and there are some unhealthy instances (not sure if that's a novel case; maybe `higher replication factor, ignore unhealthy strategy, some unhealthy` already covers this)

I think it does.

>     * number of instances per zone aren't the same. e.g. RF=6, we can only select one instance from zone-a and 7 instances from zone-b. Would we compensate and select 3 instances from zone-b because zone-a had only one healthy?

In this case, it would attempt to get two instances from each zone. It wouldn't try to compensate for one of the zones being mostly unhealthy. I'll add a test to demonstrate / prove this.

ring/ring_test.go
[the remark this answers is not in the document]
56quarters: > The ring previously and currently uses `len(r.ringTokensByZone)` to determine the number of zones.

The more I think about this, I think it's not correct. Reading the admittedly terse documentation for the `replication_factor` and `zone_awareness_enabled` configuration, it seems like the configured (_not_ per-call) replication factor should be treated as the number of zones. I'm going to see how this affects our existing tests.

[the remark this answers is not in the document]
56quarters: Tests all pass when treating `maxZones = r.cfg.ReplicationFactor` and I feel better about not needing to hack up the test to get it passing. Going to further test this change with Mimir locally.

ring/ring_test.go
56quarters: I've since changed how `findInstancesForKey` works such that this isn't required anymore. Now, we explicitly treat the configured replication factor as the number of zones when zone-awareness is enabled.

ring/ring_test.go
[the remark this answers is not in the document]
dimitarvdimitrov: yeah, i remember going through this code at some point and hitting the fact that it assumed number of zones == replication factor and it was using the two interchangeably. The fact that tests pass is reassuring  👍

dimitarvdimitrov (approved):
LGTM 🚀

PR #633, opened 2025-01-03
chore: Add option to hide tokens from ring statuspage

alexweav:
**What this PR does**:

This PR is the inverse of https://github.com/grafana/dskit/pull/631.
There are contexts in which the instance ring is used, but tokens are ignored. Such contexts can now configure `HideTokens` on the ring or lifecycler in order to hide the concept of tokens completely from the ring status page.

Here is what the page looks like with the toggle left at default (no visible changes):
2025-01-03-154402_2545x814_scrot

Hiding the tokens changes the page to look like:
2025-01-03-155551_2546x721_scrot

Similarly, no "Show Tokens" button at the bottom:
2025-01-03-155604_507x180_scrot

Contrib https://github.com/grafana/mimir-squad/issues/2350

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/basic_lifecycler.go
+++ b/ring/basic_lifecycler.go
@@ -53,6 +53,8 @@ type BasicLifecyclerConfig struct {
 	HeartbeatTimeout    time.Duration
 	TokensObservePeriod time.Duration
 	NumTokens           int
+	// HideTokens allows tokens to be hidden from e.g. the status page, for use in contexts which do not utilize tokens.
+	HideTokens bool
 
 	// If true lifecycler doesn't unregister instance from the ring when it's stopping. Default value is false,
 	// which means unregistering.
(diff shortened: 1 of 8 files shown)


ring/basic_lifecycler.go
```diff
@@ -53,6 +53,8 @@ type BasicLifecyclerConfig struct {
 	HeartbeatTimeout    time.Duration
 	TokensObservePeriod time.Duration
 	NumTokens           int
+	// HideTokens allows tokens to be hidden from e.g. the status page, for use in contexts which do not utilize tokens.
+	HideTokens bool
```
56quarters, line 57: nit: Maybe we can name this to indicate it's just for the status page? `HideTokens` at the top-level config sounds scary and behavior-changing.

56quarters (approved):
LGTM

ring/basic_lifecycler.go
alexweav: Good call - made the name more specific.

56quarters (approved):

PR #635, opened 2025-01-07
Update default values for server.throughput.unit

tinitiuset:
**What this PR does**:

Based on comment https://github.com/grafana/mimir/pull/10103#pullrequestreview-2476439373 this change is intended for consistency. Changes default value from `total_samples` to `samples_processed`.

**Checklist**
- [ ] Tests updated


dimitarvdimitrov (approved):
I think no one is relying on this value, so it's ok to change 👍

PR #640, opened 2025-01-28
Adding grpc_cluster client and server middleware

duricanikolic:
**What this PR does**:
This PR introduces 2 gRPC middlewares:
- A client middleware of type `grpc.UnaryClientInterceptor`, that propagates a given cluster info to the outgoing gRPC metadata.
- A server middleware of type `grpc.UnaryServerInterceptor`, that checks if the incoming gRPC metadata contain any cluster information and if so, checks if the latter corresponds to the given cluster info. If it is the case, the request is further propagated. Otherwise, an error is returned.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/middleware/grpc_cluster.go
+++ b/middleware/grpc_cluster.go
@@ -27,16 +27,17 @@ func ClusterUnaryClientInterceptor(cluster string) grpc.UnaryClientInterceptor {
 	}
 }
 
-// ClusterUnaryServerInterceptor checks if the incoming gRPC metadata contain any cluster information and if so,
-// checks if the latter corresponds to the given info. If it is the case, the request is further propagated.
+// ClusterUnaryServerInterceptor checks if the incoming gRPC metadata contains any cluster information and if so,
+// checks if the latter corresponds to the given cluster. If it is the case, the request is further propagated.
 // Otherwise, an error is returned.
 func ClusterUnaryServerInterceptor(cluster string, logger log.Logger) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-		reqCluster, ok := getClusterFromIncomingContext(ctx)
-		if ok {
+		reqCluster, requestClusterFound := getClusterFromIncomingContext(ctx)
+		clustersConsistent := (cluster == "" && !requestClusterFound) || cluster == reqCluster
+		if !clustersConsistent {
 			if reqCluster != cluster {
 				msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
-				level.Error(logger).Log("msg", msg)
+				level.Warn(logger).Log("msg", msg)
 				return nil, status.Error(codes.FailedPrecondition, msg)
 			}
 		}
--- a/middleware/grpc_cluster.go
+++ b/middleware/grpc_cluster.go
@@ -32,9 +32,8 @@ func ClusterUnaryClientInterceptor(cluster string) grpc.UnaryClientInterceptor {
 // Otherwise, an error is returned.
 func ClusterUnaryServerInterceptor(cluster string, logger log.Logger) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-		reqCluster, requestClusterFound := getClusterFromIncomingContext(ctx)
-		clustersConsistent := (cluster == "" && !requestClusterFound) || cluster == reqCluster
-		if !clustersConsistent {
+		reqCluster := getClusterFromIncomingContext(ctx)
+		if cluster != reqCluster {
 			if reqCluster != cluster {
 				msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
 				level.Warn(logger).Log("msg", msg)
@@ -45,10 +44,10 @@ func ClusterUnaryServerInterceptor(cluster string, logger log.Logger) grpc.Unary
 	}
 }
 
-func getClusterFromIncomingContext(ctx context.Context) (string, bool) {
+func getClusterFromIncomingContext(ctx context.Context) string {
 	clusterIDs := metadata.ValueFromIncomingContext(ctx, MetadataClusterKey)
 	if len(clusterIDs) != 1 {
-		return "", false
+		return ""
 	}
-	return clusterIDs[0], true
+	return clusterIDs[0]
 }
--- a/middleware/grpc_cluster.go
+++ b/middleware/grpc_cluster.go
@@ -32,21 +32,21 @@ func ClusterUnaryClientInterceptor(cluster string) grpc.UnaryClientInterceptor {
 // Otherwise, an error is returned.
 func ClusterUnaryServerInterceptor(cluster string, logger log.Logger) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-		reqCluster := getClusterFromIncomingContext(ctx)
+		reqCluster := getClusterFromIncomingContext(ctx, logger)
 		if cluster != reqCluster {
-			if reqCluster != cluster {
-				msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
-				level.Warn(logger).Log("msg", msg)
-				return nil, status.Error(codes.FailedPrecondition, msg)
-			}
+			msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
+			level.Warn(logger).Log("msg", msg)
+			return nil, status.Error(codes.FailedPrecondition, msg)
 		}
 		return handler(ctx, req)
 	}
 }
 
-func getClusterFromIncomingContext(ctx context.Context) string {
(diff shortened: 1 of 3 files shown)


middleware/grpc_cluster.go
```diff
@@ -0,0 +1,53 @@
...
+			ctx = metadata.AppendToOutgoingContext(ctx, MetadataClusterKey, cluster)
+		}
+
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
+// ClusterUnaryServerInterceptor checks if the incoming gRPC metadata contain any cluster information and if so,
+// checks if the latter corresponds to the given info. If it is the case, the request is further propagated.
+// Otherwise, an error is returned.
```
aknuds1, line 32: [Nit] I personally think "info" is too vague here, why not just refer to the "cluster" argument by name?
```suggestion
// ClusterUnaryServerInterceptor checks if the incoming gRPC metadata contains any cluster information and if so,
// checks if the latter corresponds to the given cluster. If it is the case, the request is further propagated.
// Otherwise, an error is returned.
```

middleware/grpc_cluster.go
```diff
@@ -0,0 +1,53 @@
...
+func ClusterUnaryServerInterceptor(cluster string, logger log.Logger) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		reqCluster, ok := getClusterFromIncomingContext(ctx)
+		if ok {
+			if reqCluster != cluster {
+				msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
+				level.Error(logger).Log("msg", msg)
```
aknuds1, line 39: Does perhaps `level.Warn` make more sense here, as it's not a server error, but a bad request? On the HTTP side, I chose `level.Warn` for the same scenario. We should at least harmonize on this aspect.

middleware/grpc_cluster_test.go
```diff
@@ -0,0 +1,106 @@
...
+		require.Len(t, clusterIDs, 1)
+		require.Equal(t, expectedCluster, clusterIDs[0])
+	}
+	for testName, testCase := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			interceptor := ClusterUnaryClientInterceptor(testCase.cluster)
+			invoker := func(ctx context.Context, _ string, _, _ any, _ *grpc.ClientConn, _ ...grpc.CallOption) error {
+				if testCase.expectedClusterFromContext != "" {
+					verify(ctx, testCase.expectedClusterFromContext)
+				}
```
aknuds1, line 49: Shouldn't we also verify the inverse, i.e. that no cluster is defined in the metadata when none is expected?

aknuds1 (commented):
Generally LGTM, but please see comments.

middleware/grpc_cluster.go
duricanikolic: It's OK for me to log this at `warn` level.

middleware/grpc_cluster_test.go
```diff
@@ -0,0 +1,106 @@
...
+		"equal request and server clusters give no error": {
+			requestCluster: "cluster",
+			serverCluster:  "cluster",
+			expectedError:  nil,
+		},
+		"different request and server clusters give rise to an error": {
+			requestCluster: "wrong-cluster",
+			serverCluster:  "cluster",
+			expectedError:  status.Error(codes.FailedPrecondition, "request intended for cluster \"wrong-cluster\" - this is cluster \"cluster\""),
+		},
```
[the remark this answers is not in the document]
duricanikolic: I have improved the test.

middleware/grpc_cluster.go
```diff
@@ -0,0 +1,54 @@
...
+	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		reqCluster, requestClusterFound := getClusterFromIncomingContext(ctx)
+		clustersConsistent := (cluster == "" && !requestClusterFound) || cluster == reqCluster
+		if !clustersConsistent {
+			if reqCluster != cluster {
+				msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
+				level.Warn(logger).Log("msg", msg)
+				return nil, status.Error(codes.FailedPrecondition, msg)
+			}
+		}
```
aknuds1, line 43: I don't understand why we have two `ifs`, and find the logic difficult to decipher. I would simplify as follows (just let `getClusterFromIncomingContext` return an empty string when no cluster is defined in the request metadata, it should have the same effect as defining an empty cluster name in the metadata):

```suggestion
		reqCluster := getClusterFromIncomingContext(ctx)
		if reqCluster != cluster {
			msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
			level.Warn(logger).Log("msg", msg)
			return nil, status.Error(codes.FailedPrecondition, msg)
	}
```

middleware/grpc_cluster.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func getClusterFromIncomingContext(ctx context.Context) (string, bool) {
+	clusterIDs := metadata.ValueFromIncomingContext(ctx, MetadataClusterKey)
+	if len(clusterIDs) != 1 {
+		return "", false
+	}
+	return clusterIDs[0], true
+}
```
aknuds1, line 54: Why differentiate between empty and not defined cluster in metadata?
```suggestion
func getClusterFromIncomingContext(ctx context.Context) string {
	clusterIDs := metadata.ValueFromIncomingContext(ctx, MetadataClusterKey)
	if len(clusterIDs) != 1 {
		return ""
	}
	return clusterIDs[0]
}
```

aknuds1 (commented):
I find the handling of empty and not defined cluster metadata unnecessarily complicated, I think the two cases can be handled as the same. WDYT?

middleware/grpc_cluster.go
```diff
@@ -0,0 +1,53 @@
...
+func ClusterUnaryServerInterceptor(cluster string, logger log.Logger) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		reqCluster := getClusterFromIncomingContext(ctx)
+		if cluster != reqCluster {
+			if reqCluster != cluster {
+				msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
+				level.Warn(logger).Log("msg", msg)
+				return nil, status.Error(codes.FailedPrecondition, msg)
+			}
+		}
```
aknuds1, line 42: I think an extra `if` must've been left behind:
```suggestion
		if reqCluster != cluster {
			msg := fmt.Sprintf("request intended for cluster %q - this is cluster %q", reqCluster, cluster)
			level.Warn(logger).Log("msg", msg)
			return nil, status.Error(codes.FailedPrecondition, msg)
		}
```

middleware/grpc_cluster_test.go
```diff
@@ -0,0 +1,139 @@
...
+			serverCluster:   "cluster",
+			expectedError:   nil,
+		},
+		"different request and server clusters give rise to an error": {
+			incomingContext: createIncomingContext(true, "wrong-cluster"),
+			requestCluster:  "wrong-cluster",
+			serverCluster:   "cluster",
+			expectedError:   status.Error(codes.FailedPrecondition, "request intended for cluster \"wrong-cluster\" - this is cluster \"cluster\""),
+		},
+		"empty request cluster and non-empty server custer give rise to an error": {
```
aknuds1, line 78:
```suggestion
		"empty request cluster and non-empty server cluster give rise to an error": {
```

middleware/grpc_cluster_test.go
```diff
@@ -0,0 +1,139 @@
...
+			incomingContext: createIncomingContext(true, ""),
+			requestCluster:  "",
+			serverCluster:   "cluster",
+			expectedError:   status.Error(codes.FailedPrecondition, "request intended for cluster \"\" - this is cluster \"cluster\""),
+		},
+		"no request cluster and non-empty server custer give rise to an error": {
```
aknuds1, line 84:
```suggestion
		"no request cluster and non-empty server cluster give rise to an error": {
```

middleware/grpc_cluster_test.go
```diff
@@ -0,0 +1,139 @@
...
+			incomingContext: createIncomingContext(false, ""),
+			requestCluster:  "",
+			serverCluster:   "cluster",
+			expectedError:   status.Error(codes.FailedPrecondition, "request intended for cluster \"\" - this is cluster \"cluster\""),
+		},
+		"empty request cluster and empty server custer give no error": {
```
aknuds1, line 90:
```suggestion
		"empty request cluster and empty server cluster give no error": {
```

middleware/grpc_cluster_test.go
```diff
@@ -0,0 +1,139 @@
...
+			incomingContext: createIncomingContext(true, ""),
+			requestCluster:  "",
+			serverCluster:   "",
+			expectedError:   nil,
+		},
+		"no request cluster and empty server custer give no error": {
```
aknuds1, line 96:
```suggestion
		"no request cluster and empty server cluster give no error": {
```

aknuds1 (commented):
Mostly fine, but there's an extra `if` left behind.

middleware/grpc_cluster.go
duricanikolic: Thank you for finding this out

aknuds1 (approved):
LGTM, thanks!

PR #641, opened 2025-01-30
WIP: Add support for validating cluster name in HTTP requests

aknuds1:
**What this PR does**:

**Checklist**
- [ ] Tests updated


clusterutil/clusterutil.go
```diff
@@ -0,0 +1,71 @@
...
+	return context.WithValue(ctx, interface{}(MetadataClusterKey), cluster)
+}
+
+// ExtractCluster gets the cluster from the context.
+func ExtractCluster(ctx context.Context) (string, error) {
+	userID, ok := ctx.Value(MetadataClusterKey).(string)
+	if !ok {
+		return "", ErrNoCluster
+	}
+	return userID, nil
```
aknuds1, line 56: @duricanikolic Can we instead just check whether `userID` is empty, so also that case is covered?
```suggestion
	userID := ctx.Value(MetadataClusterKey).(string)
	if userID == "" {
		return "", ErrNoCluster
	}
	return userID, nil
```

grpcutil/status.go
```diff
@@ -60,6 +61,17 @@ func ErrorToStatusCode(err error) codes.Code {
 	return codes.Unknown
 }
 
+func Status(errCode codes.Code, errMessage string, details ...proto.Message) *status.Status {
+	stat := status.New(errCode, errMessage)
+	if details != nil {
```
aknuds1, line 66: Less confusing/fragile:
```suggestion
	if len(details) {
```

middleware/cluster.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/grafana/dskit/clusterutil"
+)
+
+// ClusterValidationMiddleware validates that requests have the correct cluster verification label.
+func ClusterValidationMiddleware(cluster string, auxPaths []string, invalidClusters *prometheus.CounterVec, logger log.Logger) Interface {
+	var reB strings.Builder
+	// Allow for a potential path prefix being configured.
+	reB.WriteString(".*/(metrics|debug/pprof.*|ready")
```
colega, line 20: I think this MUST be configurable. Each downstream will have some paths that may not need the cluster verification, for example the admin frontend for Mimir.

colega: Maybe that's what `auxPaths` is? I think this method needs more docs.

aknuds1: `auxPaths` is for additional paths, yes. I kept a base set of hard-coded paths here since they are common (`/metrics`, `/debug/pprof`, `/ready`).

middleware/cluster.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, p := range auxPaths {
+		reB.WriteString("|" + regexp.QuoteMeta(p))
+	}
+	reB.WriteString(")")
+	reAuxPath := regexp.MustCompile(reB.String())
+
+	return Func(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			reqCluster := r.Header.Get(clusterutil.ClusterVerificationLabelHeader)
+			if !reAuxPath.MatchString(r.URL.Path) && reqCluster != cluster {
```
colega, line 30:
```suggestion
			if reqCluster != cluster && !reAuxPath.MatchString(r.URL.Path) {
```
Most of the requests are correct, don't waste time doing regexp matching here.

middleware/cluster.go
```diff
@@ -0,0 +1,45 @@
...
+				level.Warn(logger).Log("msg", "rejecting request with wrong cluster verification label",
+					"cluster_verification_label", cluster, "request_cluster_verification_label", reqCluster,
+					"header", clusterutil.ClusterVerificationLabelHeader, "url", r.URL, "path", r.URL.Path)
+				if invalidClusters != nil {
+					invalidClusters.WithLabelValues("http", r.URL.Path, reqCluster).Inc()
```
colega, line 35: I think this is very fragile, the's no contract on which labels `invalidClusters` has, and if those change, this will panic when it should fail the request. I would suggest passing here a closure that would increment the correct metric (and let the caller pass an empty closure, not nil), so the responsibility of declaring labels and using them would stay on the initialization side.

Shameless plug, this is why I wrote [cabify/gotoprom](https://github.com/cabify/gotoprom) 5 years ago :D

aknuds1: Interesting, will look into this when I have more time.

PR #642, opened 2025-01-30
feat(server): Add option to count per-tenant requests

julienduchesne:
In https://github.com/grafana/dskit/pull/490, the option was added to record durations in an histogram but this adds too many metrics to enable it for a large number of tenants

This PR extends the feature to have a counter instead. We can have a counter for most tenants, and an histogram for select tenants

From my code search, the only place where this option is used is in our own private repo (backend-enterprise)

**Checklist**
- [x] Tests updated


julienduchesne:
> When moving tenants from just counter to histogram, you would loose the counter (or would have to rebuild it from the histogram), not too sure you agree but that feels quite disruptive for saving 1 active series. This not blocking but I would prefer before [7d33881](https://github.com/grafana/dskit/commit/7d33881bbd392960dce0c7199a589b765a797c33)

Good point. I'll take off the last commit 😄

PR #643, opened 2025-01-31
Update GrpcInflightMethodLimiter to support interceptors

jhalterman:
**What this PR does**:

This PR adds server interceptor support to the `GrpcInflightMethodLimiter` via a `RPCCallProcessing` callback. This is in addition to the existing tap handle support via `RPCCallStarting`.

Interceptors are needed for inflight limiting where we want request blocking to be able to happen, since tap handles are not meant to have any blocking occur. In particular, adaptive limiters will use tap handles for most rejection, but interceptors for any blocking of requests that might be needed since they'll generally keep concurrency limits low.

This will require downstream users of `GrpcInflightMethodLimiter` to implement `RPCCallProcessing`, which can be a noop if it's not used:

```go
func (m *methodLimiter) RPCCallProcessing(_ context.Context, _ string) (func(error), error) {
	return nil, nil
}
```

Fixes #<issue number>

**Checklist**
- [x] Tests updated


56quarters (approved):
LGTM

PR #645, opened 2025-02-10
fix memberlist delete operation

aldernero:
Fixes the scenario where a previously deleted key in the kv store is "undeleted" by a re-joining member getting the value from another member that has not yet marked it as delete.

The PR ensures that the delete can only go one way, from `false` to `true`.

## Tests
- Introduces `checkMemberlistEntry`, that will constantly checks the memberlist entries up to specific amount of time before considers the test failure. It utilizes test.poll so we wait the maximum time only in the worst case scenario in which the delayed operation has been delayed
- extend `TestDelete` to check if there are any `encoding` errors during the Delete flow.


kv/memberlist/memberlist_client.go
```diff
@@ -1496,6 +1505,10 @@ func (m *KV) MergeRemoteState(data []byte, _ bool) {
 }
 
 func (m *KV) mergeBytesValueForKey(key string, incomingData []byte, codec codec.Codec, deleted bool, updateTime time.Time) (Mergeable, uint, bool, time.Time, error) {
+	// Even if there is no change to the Mergeable, we still may need to update the timestamp and deleted state.
+	if len(incomingData) == 0 {
```
NickAnge, line 1509: This makes sense as a protective mechanism in case there is an empty slice generated which has not been encoded first, so I am in favor of having this

However, I still believe that there shouldn't be any empty slices/(empty changes) generated, apart from the delete operation before the fix.

NickAnge: What do you think of having a UT for this one ? Something like:
- pass an empty slice
- ensure that it returns no error

NickAnge (approved):
I have added a small comment about the `len(incomingData) == 0` but change looks good. Maybe we can cover the testing in a next pr

PR #647, opened 2025-02-24
Make Memcached soft-dependency

tinitiuset:
**What this PR does**: This PR makes Memcached a soft-dependency on startup.

**Which issue(s) this PR fixes**: If Memcached is not available on cache initialization, `MemcachedClient` would return `nil` with `error` which would not permit for a recovery if Memcached gets available.

**Checklist**
- [x] Tests updated

### DIFF
--- a/cache/memcached_client.go
+++ b/cache/memcached_client.go
@@ -291,12 +291,11 @@ func newMemcachedClient(
 		func() float64 { return 1 },
 	)
 
-	// As soon as the client is created it must ensure that memcached server
-	// addresses are resolved, so we're going to trigger an initial addresses
-	// resolution here.
+	// Resolve Memcached addresses on startup.
 	if err := c.resolveAddrs(); err != nil {
 		return nil, err
 	}
+	// Additionally, resolve addresses periodically.
 	go c.resolveAddrsLoop()
 
 	return c, nil
@@ -740,10 +739,10 @@ func (c *MemcachedClient) resolveAddrs() error {
 	if err := c.addressProvider.Resolve(ctx, c.config.Addresses); err != nil {
 		level.Error(c.logger).Log("msg", "failed to resolve addresses for memcached", "addresses", strings.Join(c.config.Addresses, ","), "err", err)
 	}
-	// Fail in case no server address is resolved.
+	// Don't fail in case no server address is resolved as it might just be temporarily unavailable.
 	servers := c.addressProvider.Addresses()
 	if len(servers) == 0 {
-		return fmt.Errorf("no server address resolved for %s", c.name)
+		level.Warn(c.logger).Log("msg", "no memcached server addresses were resolved")
 	}
 
 	return c.selector.SetServers(servers...)
--- a/cache/memcached_client.go
+++ b/cache/memcached_client.go
@@ -311,6 +311,7 @@ func newMemcachedClient(
 }
 
 // Initialize will start the background DNS resolution periodically and do an initial DNS resolution.
+// The background resolution goroutine is started even if the initial resolution fails.
 func (c *MemcachedClient) Initialize() error {
 	// Start the background DNS resolution
 	go c.resolveAddrsLoop()
--- a/cache/memcached_client.go
+++ b/cache/memcached_client.go
@@ -124,6 +124,10 @@ type MemcachedClientConfig struct {
 
 	// TLS to use to connect to the Memcached server.
 	TLS dstls.ClientConfig `yaml:",inline"`
+
+	// DNSInitializationEnabled enables initial DNS lookup and background resolution on client creation.
+	// When false, Initialize() must be called manually.
+	DNSInitializationEnabled bool `yaml:"dns_initialization_enabled" category:"advanced"`
 }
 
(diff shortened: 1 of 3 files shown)


56quarters:
I understand the intent here but when is this actually a problem?

tinitiuset:
For example, this can be a problem in Mimir when Store-gateways or Queriers spin up with Memcached configured but Memcached is not available. 

In my opinion it's best to initialize Memcached anyways and let requests fail until its available over disabling if not available on startup or not starting the component at all.

tinitiuset:
Also for reference, I have decided not to put this behind a feature flag as I have not seen this code used in Loki or Tempo, just Mimir, but I'm open to other opinions.

56quarters (commented):
I'm not a huge fan of this change but I don't feel strongly enough to block it.

* We should be running enough of all types of cache that k8s DNS can always return at least one answer. That may mean increasing replicas of the metadata cache. Even for a rarely used cache, a single replica is not sufficient.
* We're preventing all callers from handling this error how they want and instead mandating that these errors can never be fatal. Instead, I'd rather see limited retry logic around creating a cache client in Mimir if we want it.

tinitiuset:
The thing that I'm not sure about within current implementation is, why enforce for at least a Memcached address **only** on startup?

We're already assuming that addresses are variable, as we update them every `dnsProviderUpdateInterval`.

If we want to keep treating the absence of a Memcached target on startup as an error I'm fine with it, but I would still return `MemcachedClient`.

56quarters:
> If we want to keep treating the absence of a Memcached target on startup as an error I'm fine with it, but I would still return `MemcachedClient`.

This change prevents anyone from doing that by not returning an error and giving no indication that anything has happened besides logging which cannot be checked programmatically.

tinitiuset:
I have modified the proposal to return `error` but also `MemcachedClient`. It would be up to the caller to decide how to treat the error.

dimitarvdimitrov:
Isn't it very uncommon in go to return both an `error` and a _thing_? WDYT about making this behaviour configurable via the `MemcachedClientConfig`?

56quarters:
> Isn't it very uncommon in go to return both an `error` and a _thing_? WDYT about making this behaviour configurable via the `MemcachedClientConfig`?

I think a variation of this would be workable:

* Move the initial DNS lookup and creation of lookup goroutine to some sort of `Initialize()` method that returns an error
* Default to calling this method from the constructor based on the config setting. This keeps the existing behavior
* Allow callers to disable the call to `Initialize()` from the constructor. This allows callers to ignore the `Initialize()` error and continue with startup if they want. 

We don't even have to add the configuration to enable/disable the call to `Initialize()` if we want. I'm not sure how many consumers of this lib there are in Grafana. 

Splitting object construction from initialization is pretty common in Java with DI frameworks.

cache/memcached_client.go
```diff
@@ -291,15 +307,16 @@ func newMemcachedClient(
...
-	// As soon as the client is created it must ensure that memcached server
-	// addresses are resolved, so we're going to trigger an initial addresses
-	// resolution here.
-	if err := c.resolveAddrs(); err != nil {
-		return nil, err
-	}
+	return c, nil
+}
+
+// Initialize will start the background DNS resolution periodically and do an initial DNS resolution.
```
56quarters, line 313: Can you note here that the background goroutine is started even if the initial resolution fails?

cache/memcached_client_test.go
```diff
@@ -334,6 +334,59 @@ func BenchmarkMemcachedClient_sortKeysByServer(b *testing.B) {
...
+			name:                     "with DNS initialization disabled",
+			dnsInitializationEnabled: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := &MemcachedClientConfig{}
+			memcachedClientConfigDefaultValues(cfg)
+			cfg.Addresses = flagext.StringSliceCSV{"dns+test:11211"}
```
56quarters, line 356: Since this needs to always not exist, can we use `invalid.` from https://www.rfc-editor.org/rfc/rfc6761 instead of `test`

56quarters (approved):
LGTM with a few comments

cache/memcached_client.go
```diff
@@ -124,6 +124,10 @@ type MemcachedClientConfig struct {
 
 	// TLS to use to connect to the Memcached server.
 	TLS dstls.ClientConfig `yaml:",inline"`
+
+	// DNSInitializationEnabled enables initial DNS lookup and background resolution on client creation.
+	// When false, Initialize() must be called manually.
+	DNSInitializationEnabled bool `yaml:"dns_initialization_enabled" category:"advanced"`
```
56quarters, line 130: Can you make this experimental for now incase we decide to just change the behavior with respect to `Initialize()` entirely?

cache/memcached_client_test.go
tinitiuset: Good catch, thank you.

cache/memcached_client.go
tinitiuset: Sure!

dimitarvdimitrov (approved):

PR #648, opened 2025-02-25
Improving gRPC client and server interceptor for cluster label validation

duricanikolic:
**What this PR does**:
In #640 we introduced `ClusterUnaryClientInterceptor` and `ClusterUnaryServerInterceptor`, gRPC interceptors intended for the cluster label verification. In this PR we improve their implementation.

`ClusterUnaryClientInterceptor` gets a non-empty cluster label, a non-nil counter and a non-nil logger, and propagates the given cluster label to gRPC metadata, before calling the next invoker. If no cluster label is provided, an `errNoClusterProvided` is returned. In case of an error related to the cluster label validation, the given counter is incremented. 

`ClusterUnaryServerInterceptor` gets a non-empty cluster label, and a non-nil logger, and checks if the incoming gRPC metadata contains any cluster label. If affirmative case, and if the former corresponds to the given cluster label, the request is propagated. Otherwise, an error is returned.

Cluster label validation-related errors, exchanged between `ClusterUnaryServerInterceptor` and `ClusterUnaryClientInterceptor`, are gRPC errors, built by the newly introduced `grpcutil.Status()` method, that contain an error detail of type `grpcutil.ErrorDetails` with cause `grpcutil.WRONG_CLUSTER_VERIFICATION_LABEL`, both introduced by this PR.

**Checklist**
- [x] Tests updated


grpcutil/status.go
```diff
@@ -60,6 +61,19 @@ func ErrorToStatusCode(err error) codes.Code {
...
 
+// Status creates a new a *github.com/gogo/status.Status with the
+// given error code, error message and error details.
+func Status(errCode codes.Code, errMessage string, details ...proto.Message) *status.Status {
+	stat := status.New(errCode, errMessage)
+	if len(details) > 0 {
+		statWithDetails, err := stat.WithDetails(details...)
+		if err == nil {
+			return statWithDetails
+		}
```
colega, line 72: This silently ignores the error if `WithDetails` fails. When can it fail?

clusterutil/clusterutil_test.go
```diff
@@ -0,0 +1,87 @@
...
+				require.Equal(t, testCase.expectedValue, value)
+			}
+		})
+	}
+}
+
+func checkSingleClusterInOutgoingCtx(ctx context.Context, t *testing.T, shouldExist bool, expectedValue string) {
+	md, ok := metadata.FromOutgoingContext(ctx)
+	if !ok {
+		require.False(t, ok)
```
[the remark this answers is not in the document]
colega: I think the above should be `shouldExist`

grpcutil/status.go
duricanikolic: Internally, `status.Status.WithDetails()` when it calls `types.MarshalAny(detail)` on each of the provided details.

In that case, we just return `stat`, i.e., the `status.Status` object without details.

middleware/grpc_cluster.go
```diff
@@ -6,48 +6,118 @@ import (
...
-		return handler(ctx, req)
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		// We skip the gRPC health check.
+		if _, ok := info.Server.(healthpb.HealthServer); ok {
+			return handler(ctx, req)
+		}
+
+		if cluster == "" {
+			if logger != nil {
+				level.Warn(logger).Log("msg", "no cluster verification label sent to the interceptor", "method", info.FullMethod)
```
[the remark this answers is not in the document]
duricanikolic: Thank you for this useful hint. I will apply it

middleware/grpc_cluster.go
```diff
@@ -6,48 +6,118 @@ import (
...
 	}
 }
 
-func getClusterFromIncomingContext(ctx context.Context, logger log.Logger) string {
-	clusterIDs := metadata.ValueFromIncomingContext(ctx, MetadataClusterKey)
-	if len(clusterIDs) != 1 {
-		msg := fmt.Sprintf("gRPC metadata should contain exactly 1 value for key \"%s\", but the current set of values is %v. Returning an empty string.", MetadataClusterKey, clusterIDs)
-		level.Warn(logger).Log("msg", msg)
-		return ""
+func getClusterFromIncomingContext(ctx context.Context, method string, expectedCluster string, failOnEmpty bool) ([]any, error) {
```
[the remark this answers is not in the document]
duricanikolic: A good catch, initially it was also returning the cluster, but this has been changed. I will improve this

clusterutil/clusterutil.go
```diff
@@ -0,0 +1,67 @@
...
+
+type errDifferentClusterVerificationLabels string
+
+func NewErrDifferentClusterVerificationLabels(clusterIDs []string) error {
+	return errDifferentClusterVerificationLabels(fmt.Sprintf("gRPC metadata should contain exactly 1 value for key %q, but it contains %v", MetadataClusterVerificationLabelKey, clusterIDs))
+}
+
+func (e errDifferentClusterVerificationLabels) Error() string {
+	return string(e)
+}
```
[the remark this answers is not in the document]
duricanikolic: I have changed this implementation

middleware/grpc_cluster.go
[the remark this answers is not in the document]
duricanikolic: I have applied this change both in `ClusterUnaryClientInterceptor` and `ClusterUnaryServerInterceptor`

grpcutil/status.go
colega: Should we panic in that case? We should handle the error somehow: if it can happen, we should return it to the caller, if it can't, we should panic.

duricanikolic: I will rework this implementation...
I am quite sure we have a similar code in Mimir too, where this error is silently ignored. But this is a separate story

middleware/grpc_cluster.go
```diff
@@ -6,48 +6,123 @@ import (
...
+		if reqCluster == expectedCluster {
+			return nil, nil
+		}
+		return []any{"msg", "rejecting request with wrong cluster verification label", "method", method, "clusterVerificationLabel", expectedCluster, "requestClusterVerificationLabel", reqCluster}, fmt.Errorf("rejected request with wrong cluster verification label %q - it should be %q", reqCluster, expectedCluster)
+	}
+	if errors.Is(err, clusterutil.ErrNoClusterVerificationLabel) {
+		if logOnEmpty {
+			return []any{"msg", "processing request with no cluster verification label", "method", method, "clusterVerificationLabel", expectedCluster}, nil
+		}
+		return nil, nil
```
colega, line 125: I'm confused about this, shouldn't it return an error here when we're doing hard validation?

grpcutil/status.go
duricanikolic: I have changed this implementation @colega

middleware/grpc_cluster.go
duricanikolic: I will change this implementation by following what we agreed on.

clusterutil/clusterutil.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+var (
+	ErrNoClusterVerificationLabel         = fmt.Errorf("no cluster verification label in context")
+	errDifferentClusterVerificationLabels = func(clusterIDs []string) error {
+		return fmt.Errorf("gRPC metadata should contain exactly 1 value for key %q, but it contains %v", MetadataClusterVerificationLabelKey, clusterIDs)
+	}
+)
+
+func NewIncomingContext(containsRequestCluster bool, requestCluster string) context.Context {
```
colega, line 25: I don't see where is this used, and why we need a separate flag, can we just check `requestCluster != ""`?

middleware/grpc_cluster.go
```diff
@@ -6,48 +6,139 @@ import (
...
 
-		return invoker(ctx, method, req, reply, cc, opts...)
+		msgs, err := checkClusterFromIncomingContext(ctx, method, cluster, true, softValidation)
+		if len(msgs) > 0 {
+			level.Warn(logger).Log(msgs...)
+		}
+		if err != nil {
+			invalidCluster.WithLabelValues(method, clusterutil.ReasonClient).Inc()
+			return grpcutil.Status(codes.Internal, err.Error()).Err()
+		}
```
colega, line 39: I'm confused about this, this looks like the client is checking the `cluster` _before_ sending the request? Did I understand correctly?

clusterutil/clusterutil.go
colega: Discussed offline: it's used in tests. Then lets move it to the tests. If it's used in multiple packages, then just copy it, as [a little copying is better than a little dependency](https://go-proverbs.github.io/).

If we keep it here, next person who comes will not know whether it's used by some dependency of `dskit`.

middleware/grpc_cluster.go
```diff
@@ -6,48 +6,122 @@ import (
...
+	}
+	if invalidCluster == nil {
+		panic("no invalid cluster counter provided")
+	}
+	if logger == nil {
+		panic("no logger provided")
+	}
+}
+
+func handleError(err error, cluster string, method string, invalidCluster *prometheus.CounterVec, logger log.Logger) error {
```
colega, line 48:
```suggestion
func handleClusterValidationError(err error, cluster string, method string, invalidCluster *prometheus.CounterVec, logger log.Logger) error {
```

We're in the `middleware` package, in such a broad context the method name should be very specific.

middleware/grpc_cluster.go
```diff
@@ -6,48 +6,122 @@ import (
...
+// ClusterUnaryClientInterceptor propagates the given cluster label to gRPC metadata, before calling the next invoker.
+// If an empty cluster label, nil invalidCounter or nil logger are provided, ClusterUnaryClientInterceptor panics.
+// In case of an error related to the cluster label validation, the error is returned, and invalidCounter is incremented.
+func ClusterUnaryClientInterceptor(cluster string, invalidCluster *prometheus.CounterVec, logger log.Logger) grpc.UnaryClientInterceptor {
+	validateClusterClientInterceptorInputParameters(cluster, invalidCluster, logger)
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-		if cluster != "" {
-			ctx = metadata.AppendToOutgoingContext(ctx, MetadataClusterKey, cluster)
+		// We skip the gRPC health check.
+		if method == healthpb.Health_Check_FullMethodName {
```
colega, line 27: I would remove this check, I think the client should always inject the cluster, it's up to the server whether it wants to check it or not.

middleware/grpc_cluster.go
```diff
@@ -6,48 +6,122 @@ import (
...
+
+		msgs, err := checkClusterFromIncomingContext(ctx, info.FullMethod, cluster, softValidation)
+		if len(msgs) > 0 {
+			level.Warn(logger).Log(msgs...)
+		}
+		if err == nil {
+			return handler(ctx, req)
+		}
+		stat := grpcutil.Status(codes.FailedPrecondition, err.Error(), &grpcutil.ErrorDetails{Cause: grpcutil.WRONG_CLUSTER_VERIFICATION_LABEL})
+		return nil, stat.Err()
```
colega, line 89: Nit just because that's the usual order.
```suggestion
		if err != nil {
			stat := grpcutil.Status(codes.FailedPrecondition, err.Error(), &grpcutil.ErrorDetails{Cause: grpcutil.WRONG_CLUSTER_VERIFICATION_LABEL})
			return nil, stat.Err()
		}
		return handler(ctx, req)
```

colega (approved):
Logic LGTM, I left some nitpicks that I'd like to see addressed before merging, but I'm leaving the approval.

middleware/grpc_cluster.go
duricanikolic: I have renamed the function. Thank you.

middleware/grpc_cluster.go
duricanikolic: That's true.
This check was needed for the previous implementations, since the client interceptor was also doing some checks. Now it actually makes no sense to have this check.

PR #649, opened 2025-02-27
gRPC cluster validation: remove unnecessary metric labels

duricanikolic:
**What this PR does**:
In #648 we introduced `ClusterUnaryClientInterceptor`, a gRPC interceptor that enriches the outgoing context with a  cluster label, propagates further the request, and once a response is returned, if it corresponds to a cluster validation-related error, tracks it into a metrics given as input to the interceptor, and returns an error. This PR removes the `reason` label from the metrics, because it adds no additional value, since the only available reason was `server_check_failed`.

Fixes #<issue number>

**Checklist**
- [x] Tests updated


dimitarvdimitrov (approved):
SGTM

PR #650, opened 2025-02-27
Allow gRPC cluster validation configuration

duricanikolic:
**What this PR does**:
In #648 we introduced `ClusterClientUnaryInterceptor` and `ClusterServerUnaryInterceptor`, gRPC client and server interceptors that allow request cluster validation in gRPC calls.
This PR introduces configuration options for enabling cluster validation.
The following configuration options are introduced:
- `-server.cluster-validation.label`: specifies the cluster validation label of the server, that is used for request validation.
- `-server.cluster-validation.grpc.enabled`: when enabled, allows the server to execute `ClusterServerUnaryInterceptor`'s cluster validation.
-`-server.cluster-validation.grpc.soft-validation`: when enabled, allows the server to execute `ClusterServerUnaryInterceptor`'s soft validation, i.e., the server tries to go ahead with the request execution, even if the request's cluster validation label doesn't match with `-server.cluster-validation.label`. It can be enabled only if `-server.cluster-validation.grpc.enabled` is enabled too.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/clusterutil/cluster_validation_config.go
+++ b/clusterutil/cluster_validation_config.go
@@ -16,7 +16,7 @@ type ClusterValidationProtocolConfig struct {
 }
 
 func (cfg *ClusterValidationConfig) Validate() error {
-	return cfg.GRPC.Validate("gRPC", cfg.Label)
+	return cfg.GRPC.Validate("grpc", cfg.Label)
 }
 
 func (cfg *ClusterValidationConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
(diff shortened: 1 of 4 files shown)


clusterutil/cluster_validation_config.go
```diff
@@ -0,0 +1,51 @@
...
+		return fmt.Errorf("%s: hard validation and soft validation cannot be enabled at the same time", prefix)
+	}
+	return nil
+}
+
+func (cfg *ClusterValidationProtocolConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	softValidationFlag := prefix + ".soft-validation"
+	enabledFlag := prefix + ".enabled"
+	f.BoolVar(&cfg.SoftValidation, softValidationFlag, false, fmt.Sprintf("When enabled, soft cluster label validation will be executed. Cannot be enabled together with %s", enabledFlag))
+	f.BoolVar(&cfg.Enabled, enabledFlag, false, fmt.Sprintf("When enabled, cluster label validation will be executed. Cannot be enabled together with %s", softValidationFlag))
```
colega, line 50: I find this config confusing, I'd expect `Enabled` to enable the validation, and `SoftValidation` switch to a soft-validation mode. I wouldn't expect `SoftValidation` to enable feature at all, and I would expect that if `Enabled` is `false` then validation is not enabled.

duricanikolic: Hi @colega 
My idea was that `enabled` means hard validation, and `soft-validation` means soft validation.
But yours is more intuitive. I will adapt the code. Then I can also simplify the validation

colega (approved):

PR #651, opened 2025-02-27
flagext: Add LimitsMap from Mimir

alexweav:
**What this PR does**:

This PR moves the `LimitsMap` type used in a few Mimir configuration options to flagext.

It's a flag type that represents a map with string keys, similar to the other types in this package that represent lists of strings or similar, thus it seems to belong here. Specifically, I'm moving it now so that packages depending on this type don't have to depend on the whole `github.com/grafana/mimir/pkg/util/validation`.

The mimir implementation: https://github.com/grafana/mimir/blob/9d6eb2312b070892152aab1c36566407bcb411a3/pkg/util/validation/limits_map.go

It's copied verbatim along with its test file except for one line which is called out below.

n/a

**Checklist**
- [x] Tests updated

### DIFF
--- a/flagext/map.go
+++ b/flagext/map.go
@@ -1,0 +1,110 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package flagext
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+// LimitsMap is a flag.Value implementation that looks like a generic map, holding float64s, ints, or strings as values.
+type LimitsMap[T float64 | int | string] struct {
+	data      map[string]T
+	validator func(k string, v T) error
+}
+
+func NewLimitsMap[T float64 | int | string](validator func(k string, v T) error) LimitsMap[T] {
+	return NewLimitsMapWithData(make(map[string]T), validator)
+}
+
+func NewLimitsMapWithData[T float64 | int | string](data map[string]T, validator func(k string, v T) error) LimitsMap[T] {
+	return LimitsMap[T]{
+		data:      data,
+		validator: validator,
+	}
+}
+
+// IsInitialized returns true if the map is initialized.
+func (m LimitsMap[T]) IsInitialized() bool {
+	return m.data != nil
+}
+
+// String implements flag.Value
+func (m LimitsMap[T]) String() string {
+	out, err := json.Marshal(m.data)
+	if err != nil {
+		return fmt.Sprintf("failed to marshal: %v", err)
+	}
+	return string(out)
+}
+
+// Set implements flag.Value
+func (m LimitsMap[T]) Set(s string) error {
+	newMap := make(map[string]T)
+	if err := json.Unmarshal([]byte(s), &newMap); err != nil {
+		return err
+	}
+	return m.updateMap(newMap)
+}
+
+func (m LimitsMap[T]) Read() map[string]T {
+	return m.data
+}
+
+// Clone returns a copy of the LimitsMap.
+func (m LimitsMap[T]) Clone() LimitsMap[T] {
+	newMap := make(map[string]T, len(m.data))
+	for k, v := range m.data {
+		newMap[k] = v
+	}
+	return LimitsMap[T]{data: newMap, validator: m.validator}
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (m LimitsMap[T]) UnmarshalYAML(value *yaml.Node) error {
+	newMap := make(map[string]T)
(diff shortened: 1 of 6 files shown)


flagext/map.go
```diff
@@ -0,0 +1,110 @@
...
+	for k, v := range m.data {
+		newMap[k] = v
+	}
+	return LimitsMap[T]{data: newMap, validator: m.validator}
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (m LimitsMap[T]) UnmarshalYAML(value *yaml.Node) error {
+	newMap := make(map[string]T)
+	if err := value.Decode(newMap); err != nil {
```
alexweav, line 68: Here's the one change; the Mimir implementation uses:
```
if err := value.DecodeWithOptions(newMap, yaml.DecodeOptions{KnownFields: true}); err != nil {
```

DecodeWithOptions is an unmerged contribution, that comes from a fork and is not part of core yaml.v3:
https://github.com/grafana/mimir/blob/9d6eb2312b070892152aab1c36566407bcb411a3/go.mod#L308-L311

`KnownFields: true` has no effect here, since we're decoding into a map. That flag only makes sense when decoding into structs. That means, we can avoid the fork and stick to only the only the core `yaml.v3` API surface, which is more portable.

flagext/map.go
```diff
@@ -0,0 +1,110 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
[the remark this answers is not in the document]
alexweav: good catch! updated

flagext/map_test.go
```diff
@@ -0,0 +1,431 @@
...
+			"Equal empty maps": {
+				map1:     LimitsMap[float64]{data: map[string]float64{}},
+				map2:     LimitsMap[float64]{data: map[string]float64{}},
+				expected: true,
+			},
+		}
+
+		for name, tt := range tc {
+			t.Run(name, func(t *testing.T) {
+				require.Equal(t, tt.expected, tt.map1.Equal(LimitsMap[float64]{data: tt.map2.data}))
```
56quarters, line 314: I'm not a huge fan of calling `require.Equal()` with booleans since the message if it fails is not helpful. Maybe split the tests into two groups and use the appropriate `require.Equal()` / `require.NotEqual()` on the values themselves?

flagext/map_test.go
```diff
@@ -0,0 +1,431 @@
...
+		require.True(t, original.Equal(cloned), "expected cloned LimitsMap to be different from original")
+
+		// Modify the original LimitsMap and ensure the cloned map is not affected.
+		original.data["limit1"] = 10.0
+		require.False(t, cloned.data["limit1"] == 10.0, "expected cloned LimitsMap to be unaffected by changes to original")
+
+		// Modify the cloned LimitsMap and ensure the original map is not affected.
+		cloned.data["limit3"] = 3.0
+		_, exists := original.data["limit3"]
+		require.False(t, exists, "expected original LimitsMap to be unaffected by changes to cloned")
```
56quarters, line 378: IIRC, there's a `require.NotContains` for this kind of test

flagext/map.go
```diff
@@ -0,0 +1,108 @@
...
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (m LimitsMap[T]) UnmarshalYAML(value *yaml.Node) error {
+	newMap := make(map[string]T)
+	if err := value.Decode(newMap); err != nil {
+		return err
+	}
+	return m.updateMap(newMap)
+}
+
+func (m LimitsMap[T]) updateMap(newMap map[string]T) error {
```
56quarters, line 72: Does this need to have a pointer receiver?

alexweav: Not strictly (though I see an argument for doing it to prevent footguns in future changes)

It doesn't change a field in the LimitsMap struct directly; it only changes the contents of the map pointed to by `m.data`. You can copy the outer struct freely and the copies will still point to the same underlying map.

56quarters (approved):

flagext/map_test.go
alexweav: I've pushed a commit splitting them - however it's not quite as clean as I'd hoped.

Part of the reason it's written this way is to specifically test the `cmp.Equal` output and the custom equality function. `require.Equal()` internally only checks the same thing as the `==` operator, which doesn't necessarily map to the the Equal() function and can in theory differ.

Since `go-cmp` equality is a requirement, I still need a `require.True(t, cmp.Equal(...))` call as this test is trying to specifically cover the custom `Equal()` function.

PR #652, opened 2025-02-27
cache: Improve jump hash server selector performance

56quarters:
**What this PR does**:

Improve the performance of the jump hash Memcached ServerSelector implementation by _not_ delegating to the default implementation in `grafana/gomemcache`. This avoids the need to iterate through the entire list of servers to pick a single address.

I've tested the new implementation along side the original and the results (which server was picked for which key) were identical. Performance is about 3-4X better.

See also https://github.com/grafana/gomemcache/pull/22

**Checklist**
- [x] Tests updated

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -46,7 +46,7 @@ jobs:
       matrix:
         # We want to make sure dskit can support multiple golang versions
         # by ensuring the test would pass using all these supported versions.
-        go-version: ['1.21.x', '1.22.x', '1.23.x']
+        go-version: ['1.22.x', '1.23.x', '1.24.x']
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


56quarters:
```
$ benchstat old.txt new.txt 
goos: linux
goarch: amd64
pkg: github.com/grafana/dskit/cache
                                        │   old.txt    │               new.txt               │
                                        │    sec/op    │   sec/op     vs base                │
MemcachedJumpHashSelector_PickServer-16   162.75n ± 1%   31.65n ± 1%  -80.55% (p=0.000 n=20)

                                        │  old.txt   │              new.txt               │
                                        │    B/op    │   B/op     vs base                 │
MemcachedJumpHashSelector_PickServer-16   24.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=20)

                                        │  old.txt   │               new.txt               │
                                        │ allocs/op  │ allocs/op   vs base                 │
MemcachedJumpHashSelector_PickServer-16   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=20)
```

PR #657, opened 2025-03-03
grpcclient.Config: add support for injecting middleware.ClusterUnaryClientInterceptor

duricanikolic:
**What this PR does**:
This PR adds the following experimental configuration options for enabling the cluster validation of both gRPC clients and servers. 
On the gRPC client side, the `grpcclient.Config` configuration can be extended with `cluster-validation.label`, and is used to define the cluster validation label to be sent together with the requests by the gRPC client.

On the gRPC server side, cluster validation-related configuration options are the following:
- `-server.cluster-validation.label`, that defines the server's cluster validation label. When the validation is enabled, this value will be compared with the cluster validation label received through the requests.
- `-server.cluster-validation.grpc.enabled`, used for enabling the cluster validation.
- `-server.cluster-validation.grpc.soft-validation`, that can be enabled only together with the cluster validation, and that implies a soft validation.

Moreover, this PR changes the signature of `grpcclient.Config.DialOption()` by adding an additional parameter of type `middleware.InvalidClusterValidationReporter`, which is a callback used for reporting the cluster validation issues back to the caller.

Fixes #<issue number>

**Checklist**
- [x] Tests updated


flagext/register.go
```diff
@@ -53,3 +54,41 @@ func DefaultValues(rs ...interface{}) {
 	}
 	_ = fs.Parse([]string{})
 }
+
+// RegisteredFlagsTracker is an interface that allows to extract RegisteredFlags.
+type RegisteredFlagsTracker interface {
+	GetRegisteredFlags() RegisteredFlags
```
colega, line 60: Idiomatic Go is to drop the `Get` prefix on the getters.
```suggestion
	RegisteredFlags() RegisteredFlags
```

duricanikolic: Thank you, I have renamed the method.

grpcclient/grpcclient.go
```diff
@@ -45,6 +46,8 @@ type Config struct {
 
 	// CustomCompressors allows configuring custom compressors.
 	CustomCompressors []string `yaml:"-"`
+
+	ClusterValidation clusterutil.ClientClusterValidationConfig `yaml:"cluster_validation" category:"experimental"`
```
colega, line 50: I just realized that by configuring this, nothing is configured.

I think that `Config.DialOption()` below should inject the client interceptor if this is configured, otherwise as a user of this library this looks weird.

clusterutil/cluster_validation_config.go
```diff
@@ -3,26 +3,57 @@ package clusterutil
...
+	cfg.registeredFlags = flagext.TrackRegisteredFlags(prefix, f, func(prefix string, f *flag.FlagSet) {
+		cfg.RegisterFlagsWithPrefix(prefix, f)
+	})
+}
+
+func (cfg *ClientClusterValidationConfig) RegisteredFlags() flagext.RegisteredFlags {
+	return cfg.registeredFlags
+}
+
+type ClusterValidationConfig struct {
```
colega, line 29: This looks like "ServerClusterValidationConfig" and the one above is "ClusterValidationConfig" instead of being "ClientClusterValidationConfig" and I think that this one should just embed that one, shouldn't it?

duricanikolic: I have addressed this in 5f81d6ac6ce3efbed28dc737a78cba756205a175

grpcclient/grpcclient.go
duricanikolic: I have improved this implementation, as discussed in a Slack [thread](https://raintank-corp.slack.com/archives/C089NNDHR45/p1741540392530969).

grpcclient/grpcclient.go
```diff
@@ -129,6 +135,10 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{NewRateLimiter(cfg)}, unaryClientInterceptors...)
 	}
 
+	if cfg.ClusterValidation.Label != "" {
+		unaryClientInterceptors = append([]grpc.UnaryClientInterceptor{middleware.ClusterUnaryClientInterceptor(cfg.ClusterValidation.Label, o.InvalidClusterValidations, o.Logger)}, unaryClientInterceptors...)
```
colega, line 139: I think you should check here that `Logger` and `InvalidClusterValidations` are not nil, and return an error otherwise.

clusterutil/cluster_validation_config.go
```diff
@@ -3,26 +3,49 @@ package clusterutil
...
-func (cfg *ClusterValidationConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
-	clusterValidationPrefix := prefix + ".cluster-validation"
-	f.StringVar(&cfg.Label, clusterValidationPrefix+".label", "", "Optionally define server's cluster validation label.")
-	cfg.GRPC.RegisterFlagsWithPrefix(clusterValidationPrefix+".grpc", f)
+func (cfg *ServerClusterValidationConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	cfg.registeredFlags = flagext.TrackRegisteredFlags(prefix, f, func(prefix string, f *flag.FlagSet) {
+		f.StringVar(&cfg.Label, prefix+"label", "", "Optionally define the server's cluster validation label. When the validation is enabled, this value will be compared with the cluster validation label received through the requests.")
+		cfg.GRPC.RegisterFlagsWithPrefix(prefix+"grpc.", f)
+	})
+}
```
colega, line 40: Shouldn't this register ClusterValidationConfig's flags too?

grpcclient/grpcclient.go
duricanikolic: I have added the validation check.

grpcclient/grpcclient.go
```diff
@@ -108,9 +115,17 @@ func (cfg *Config) CallOptions() []grpc.CallOption {
...
-func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) ([]grpc.DialOption, error) {
+// DialOption returns the config as a grpc.DialOptions. The passed interceptors wrap around the configured middleware.
+// invalidClusterValidation and logger are required to be non-nil, and they are used for reporting the results of the
+// cluster validation, when it is enabled.
+func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor, invalidClusterValidations *prometheus.CounterVec, logger log.Logger) ([]grpc.DialOption, error) {
+	if invalidClusterValidations == nil {
+		return nil, fmt.Errorf("invalidClusterValidations must not be null")
+	}
+	if logger == nil {
+		return nil, fmt.Errorf("logger must not be null")
```
colega, line 126: Being nitpicky, but we don't have nulls in go.

```suggestion
		return nil, fmt.Errorf("invalidClusterValidations must not be nil")
	}
	if logger == nil {
		return nil, fmt.Errorf("logger must not be nil")
```

grpcclient/grpcclient.go
```diff
@@ -108,9 +115,17 @@ func (cfg *Config) CallOptions() []grpc.CallOption {
 	return opts
 }
 
-// DialOption returns the config as a grpc.DialOptions. The passed inceptors
-// wrap around the configured middleware.
-func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor) ([]grpc.DialOption, error) {
+// DialOption returns the config as a grpc.DialOptions. The passed interceptors wrap around the configured middleware.
+// invalidClusterValidation and logger are required to be non-nil, and they are used for reporting the results of the
+// cluster validation, when it is enabled.
+func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientInterceptor, streamClientInterceptors []grpc.StreamClientInterceptor, invalidClusterValidations *prometheus.CounterVec, logger log.Logger) ([]grpc.DialOption, error) {
```
colega, line 121: These docs should state what labels `invalidClusterValidations` will accept.

I did a similar comment on the original PR that @aknuds1 shared about this cluster validation, and I still think the same: I would remove the dependency on Prometheus here and would just accept a function `increaseInvalidClusterValidations func(cluster string)`, so if someone really wants to opt out from this, they can just provide an empty function, or maybe they don't want to use Prometheus metrics? But since this param is going more and more widespread, I thought it's time to bring this topic again.

duricanikolic: I have implemented this idea in 236f35dc1dde2e9da78524ddffb3ce07524c3ae4.

middleware/grpc_cluster.go
```diff
@@ -50,8 +55,7 @@ func handleClusterValidationError(err error, cluster string, method string, inva
 			if errDetails, ok := details[0].(*grpcutil.ErrorDetails); ok {
 				if errDetails.GetCause() == grpcutil.WRONG_CLUSTER_VALIDATION_LABEL {
 					msg := fmt.Sprintf("request rejected by the server: %s", stat.Message())
-					level.Warn(logger).Log("msg", msg, "method", method, "clusterValidationLabel", cluster)
-					invalidCluster.WithLabelValues(method).Inc()
+					invalidClusterValidationReporter(msg, cluster, method)
```
colega, line 58: Does it make sense to pass the `cluster` here? What is the reporter doing to do with it, it's quite obvious where we run this (in this cluster), what we really know is which cluster responded that?

If we keep it, can we at least update the docs to say that it's "this cluster" no "that cluster"?

middleware/grpc_cluster.go
```diff
@@ -11,36 +11,41 @@ import (
...
-	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 )
 
+// InvalidClusterValidationReporter is called by ClusterUnaryClientInterceptor to report the cluster validation issues
+// back to the caller. Its parameters are the error message explaining the reason for a bad cluster validation,
+// the involved cluster, and the method that triggered the validation.
+type InvalidClusterValidationReporter func(errorMsg string, cluster string, method string)
```
colega, line 22: Commented below, I think we need to improve the docs on _the involved cluster_, I assumed this would be the cluster that rejected the request, but it's just this cluster from the config, so I'd say "configured cluster in this client"

middleware/grpc_cluster.go
```diff
@@ -11,36 +11,41 @@ import (
...
+
+// NoOpInvalidClusterValidationReporter in an InvalidClusterValidationReporter that reports nothing.
+var NoOpInvalidClusterValidationReporter InvalidClusterValidationReporter = func(string, string, string) {}
```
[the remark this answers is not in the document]
colega: (This is opinionated though, feel free to ignore)

colega (approved):
LGTM, please consider my comment on the `cluster` docs of the callback function.

middleware/grpc_cluster.go
duricanikolic: I have decided to remove the cluster parameter, because it can actually be retrieved directly from the caller.

middleware/grpc_cluster.go
duricanikolic: I have removed the cluster parameter.

PR #658, opened 2025-03-05
Memcached: Change `dns_initialization_enabled` for `dns_ignore_startup_failures`

tinitiuset:
**What this PR does**:

This PR is a followup of #647, it tries to simplify usability of the client by using the config flag to ignore DNS startup failures or not.

- Cache is an interface, so whenever we create the cache,  I would argue that we need to return something directly usable.
- Using a config variable to give the user the freedom to call a method right after creating the client seemed a little bit redundant, using the config variable to decide preferred behavior of error handling looks cleaner to me.

**Checklist**
- [x] Tests updated


dimitarvdimitrov (approved):
SGTM, I had something like that in mind at first. I'll defer to @56quarters for a final review though

56quarters (approved):

PR #659, opened 2025-03-05
flagext: Go-cmp support for Secret

alexweav:
**What this PR does**:

Go-cmp on any type containing a flagext.Secret panics due to the unexported field, without supplying specific options that may not be generally desirable.

Mimir expects its configuration objects to be comparable with go-cmp. The other flagext types generally implicitly support it, by using either exported inner fields or type aliases.

n/a

**Checklist**
- [ ] Tests updated

### DIFF
--- a/flagext/secret_test.go
+++ b/flagext/secret_test.go
@@ -3,6 +3,7 @@ package flagext
 import (
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


alexweav:
IMO this is not important enough to deserve changelog, please let me know if you disagree.

flagext/secret_test.go
```diff
@@ -75,3 +76,58 @@ func TestSecretdYAML(t *testing.T) {
 		assert.Equal(t, testStruct, actualStruct)
 	}
 }
+
+func TestSecretEquals(t *testing.T) {
```
aknuds1, line 80: [Nit] I personally prefer the test naming convention `Test$TYPE_$METHOD`, makes it easier to understand what's being tested:
```suggestion
func TestSecret_Equals(t *testing.T) {
```

aknuds1 (approved):
LGTM modulo nit (non-blocker).

flagext/secret_test.go
alexweav: SGTM!

PR #660, opened 2025-03-05
dns: New resolver backend "miekgdns2" with reliability improvements

56quarters:
**What this PR does**:

Create a new version of the `dns/miekgdns` package with changes to improve the reliability of queries since they are typically used for service discovery.

The notable changes:
* Search domains are not supported. We don't use them, when they are used they put more load on DNS servers, and they make the behavior of resolution harder to understand.
* Always use TCP for queries, don't start with UDP. We're using DNS for service discovery and the result size is usually large enough that TCP must be used anyway. Instead of starting with UDP and needing to wait N seconds to know if the packet was dropped - just start with TCP. [TCP support for DNS is required as of RFC 7766](https://datatracker.ietf.org/doc/html/rfc7766.txt).
* Pool connections to DNS servers. Instead of creating new UDP and TCP sockets for every query, we keep TCP connections to DNS servers open and reuse them if possible.
* Use the "attempts" option from /etc/resolv.conf. Instead of failing the query immediately due to network issues, we retry based on the "attempts" setting which is 2 by default.

This change also makes the resolver backend used by the Memcached client configurable so that we can test the new backend before making it the default implementation.

N/A

**Notes for reviewers**:

* `Resolver` is the majority of this change. `miekgdns2.Resolver` serves the same purpose as `miekgdns.Resolver` - doing A, AAAA, and SRV DNS requests to discover a list of servers.
* `PoolingClient` is a thin wrapper over the `dns.Client` type that supports keeping connections to each nameserver open. It maintains a `Pool` instance for each nameserver.
* `Pool` is a pool of connections to a single server that stores them in a channel, [roughly based on the `filePool` type](https://github.com/grafana/mimir/blob/5126fa342b148c70fcb7385e993f561515a67bbb/pkg/storage/indexheader/encoding/factory.go#L168-L182) in Mimir.
 

**Checklist**
- [x] Tests updated

### DIFF
--- a/dns/miekgdns2/resolver.go
+++ b/dns/miekgdns2/resolver.go
@@ -12,14 +12,17 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/miekg/dns"
 
 	"github.com/grafana/dskit/multierror"
 )
 
 const (
-	DefaultResolvConfPath  = "/etc/resolv.conf"
-	defaultMaxConnsPerHost = 2
+	DefaultResolvConfPath   = "/etc/resolv.conf"
+	defaultResolvConfReload = time.Second * 5
+	defaultMaxConnsPerHost  = 2
 )
 
 type Client interface {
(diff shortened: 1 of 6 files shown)


dns/miekgdns2/resolver.go
```diff
@@ -0,0 +1,356 @@
...
+		default:
+			return "", nil, fmt.Errorf("invalid SRV response record %s", record)
+		}
+	}
+
+	return "", addrs, err
+}
+
+func (r *Resolver) LookupIPAddr(ctx context.Context, host string) ([]net.IPAddr, error) {
+	servers, conf, err := r.loadConfiguration()
```
charleskorn, line 108: Should we cache the config? Seems expensive to parse it for every DNS query, although maybe those are infrequent enough that it doesn't matter.

dns/miekgdns2/resolver_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+var resolvConfContents = `
+nameserver 127.0.0.1
+nameserver 127.0.0.2
+option attempts:2
+option timeout:1
+`
+
+func TestResolver_LookupSRV(t *testing.T) {
+	t.Run("bad resolv.conf", func(t *testing.T) {
```
charleskorn, line 25: [nit] This might be a clearer name for the test:

```suggestion
	t.Run("missing resolv.conf", func(t *testing.T) {
```

dns/miekgdns2/resolver_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+		resolver := NewResolverWithClient(cfgPath, client)
+		_, res, err := resolver.LookupSRV(context.Background(), "cache", "tcp", "example.com")
+
+		require.Nil(t, res)
+		require.Error(t, err)
+	})
+}
+
+func TestResolver_LookupIP(t *testing.T) {
```
charleskorn, line 102: Same comments as above apply here

dns/miekgdns2/resolver_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+	"github.com/miekg/dns"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
+)
+
+var resolvConfContents = `
+nameserver 127.0.0.1
+nameserver 127.0.0.2
+option attempts:2
```
charleskorn, line 20: It'd be good to see a test that verifies that the retry behaviour is as expected (ie. that each server is attempted `attempts` times), and another that verifies what happens if `option attempts` is not present in `resolv.conf`.

dns/miekgdns2/resolver.go
```diff
@@ -0,0 +1,356 @@
...
+		for _, server := range servers {
+			response, _, err := r.client.Exchange(ctx, msg, server)
+			if err != nil {
+				merr.Add(fmt.Errorf("resolution against server %s for %s: %w", server, name, err))
+				continue
+			}
+
+			if response.Truncated {
+				merr.Add(fmt.Errorf("resolution against server %s for %s: response truncated", server, name))
+				continue
```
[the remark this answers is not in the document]
56quarters: Yes, good point.

dns/miekgdns2/resolver.go
56quarters: This is what the existing resolver did but it does seem expensive. I don't think it'll be an issue but I can look into caching it for a short amount of time.

dns/miekgdns2/resolver.go
```diff
@@ -0,0 +1,356 @@
...
+
+	return servers, conf, nil
+}
+
+// PoolingClient is a DNS client that pools TCP connections to each nameserver.
+type PoolingClient struct {
+	network string
+	maxOpen int
+
+	mtx   sync.Mutex
```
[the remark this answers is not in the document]
56quarters: Both places where the lock is held modify `pools` so there's nowhere that we could get away with a read lock.

dns/miekgdns2/resolver_test.go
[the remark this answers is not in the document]
56quarters: Good idea.

dns/miekgdns2/resolver_test.go
```diff
@@ -0,0 +1,323 @@
...
+option timeout:1
```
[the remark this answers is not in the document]
56quarters: It's not. I decided not to implement this since it's redundant with `context.Context` and the underlying `dns` library applies its own defaults as well. I'll remove this from the tests and note that this setting is ignored.

dns/miekgdns2/resolver_test.go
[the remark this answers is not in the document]
56quarters: Actually, we can't really test a malformed configuration. The underlying `dns` library ignores anything in resolv.conf that it doesn't understand.

dns/miekgdns2/resolver_test.go
56quarters: Done.

dns/miekgdns2/resolver.go
56quarters: I've added some logic to reload the configuration and clean up defunct connections in a separate goroutine

dns/miekgdns2/resolver_test.go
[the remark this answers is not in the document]
charleskorn: In that case, should we add a test that confirms that config we understand is used, but other things are ignored?

dns/miekgdns2/resolver.go
```diff
@@ -185,21 +210,71 @@ func (r *Resolver) query(ctx context.Context, servers []string, attempts int, na
...
 	servers := make([]string, len(conf.Servers))
-	for i, nameserver := range conf.Servers {
-		servers[i] = net.JoinHostPort(nameserver, conf.Port)
+	for i, ip := range conf.Servers {
+		servers[i] = net.JoinHostPort(ip, conf.Port)
 	}
 
-	return servers, conf, nil
+	// Close connections to any servers that are no longer in resolv.conf.
+	r.client.Clean(servers)
```
charleskorn, line 246: Should we do this after updating `r.conf` below? Otherwise we might end up cleaning up some servers that will be immediately recreated before we have a chance to update `r.conf`.

(They'll be cleaned up the next time `loadConfig()` runs, but this seems tidier.)

charleskorn (approved):
LGTM modulo suggestion below

dns/miekgdns2/resolver.go
56quarters: Sure, I'll move this after the update.

charleskorn (approved):

dns/miekgdns2/resolver_test.go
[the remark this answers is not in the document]
56quarters: Added a test that ensures we can load configuration even when there are unknown settings.

PR #662, opened 2025-03-06
ObserveWithExemplar support other spancontext types

mdisibio:
**What this PR does**:
Was researching that Tempo no longer exported exemplars and identified the root cause:   The dskit server instrumentation to export `request_duration_seconds` histogram and exemplars no longer works for apps using the Otel or other (non-jaeger) SDK, because we are only checking for `jaeger.SpanContext`.  This updates it to also support other, including Otel [SpanContext](https://github.com/open-telemetry/opentelemetry-go/blob/main/trace/trace.go#L194) and [bridgeSpanContext](https://github.com/open-telemetry/opentelemetry-go/blob/7a63439b15acf4b4cf1b88f9e2001e8ced82ab77/bridge/opentracing/bridge.go#L36)

Because this repo doesn't have a dependency on otel, some work is needed to get the data without access to the underlying types, and a simple interface conversion doesn't seem to be possible.

Therefore this is **Draft** so we can decide on the approach.  It's possible that were is a better method that doesn't use reflection.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -6,7 +6,9 @@ package tracing
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"reflect"
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


colega:
@mdisibio here's what I'm doing: https://github.com/grafana/dskit/pull/681

pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

PR #664, opened 2025-03-10
Usage tracker: support ownership of multiple partitions, read-only in basic lifecycler and write replica selection

colega:
This merges into `usage-tracker` hackathon branch.

There are three main features implemented here:

### Support multiple partition ownership in PartitionInstanceRing. 

Since PartitionInstanceRing doesn't allow owning multiple partitions by a single instance out of the box, we workaround that by adding a well known suffix to the instance IDs owning the partitions, this suffix is built by adding a slash `/` and the partition that is being owned. I.e. `foo-zone-a-1/0` is `foo-zone-a-1` owning the partition `0`, while `foo-zone-a-1/2` is the same instance owning partition 2. This is being taken into account when calculating the replication set.

### Support selecting highest replica from each zone

By specifying a flag we can select only the highest replica from each zone: this is useful when a partition is being transferred to a different instance when scaling out. There's also a flag introduced to attempt select an non-readonly instance: this allows distributors to transfer the writes to the lower replica when scaling down the instance number and highest replicas become read-only.

### Add read-only state to BasicLifecycler

This is just a port of same functionality from Lifecycler, as we need it for Usage-Tracker.

### DIFF
--- a/ring/basic_lifecycler.go
+++ b/ring/basic_lifecycler.go
@@ -153,6 +153,20 @@ func (l *BasicLifecycler) GetState() InstanceState {
 	return l.currInstanceDesc.GetState()
 }
 
+func (l *BasicLifecycler) GetReadOnlyState() (bool, time.Time) {
+	l.currState.RLock()
+	defer l.currState.RUnlock()
+
+	if l.currInstanceDesc == nil {
+		return false, time.Time{}
+	}
+
+	if l.currInstanceDesc.ReadOnly {
+		return true, time.Unix(l.currInstanceDesc.ReadOnlyUpdatedTimestamp, 0)
+	}
+	return false, time.Time{}
+}
+
 func (l *BasicLifecycler) GetTokens() Tokens {
 	l.currState.RLock()
 	defer l.currState.RUnlock()
@@ -189,6 +203,12 @@ func (l *BasicLifecycler) IsRegistered() bool {
 func (l *BasicLifecycler) ChangeState(ctx context.Context, state InstanceState) error {
 	return l.run(func() error {
 		return l.changeState(ctx, state)
+	})
+}
+
+func (l *BasicLifecycler) ChangeReadOnlyState(ctx context.Context, readOnly bool) error {
+	return l.run(func() error {
+		return l.changeReadOnlyState(ctx, readOnly)
 	})
 }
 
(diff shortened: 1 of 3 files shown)


ring/basic_lifecycler.go
```diff
@@ -514,6 +534,31 @@ func (l *BasicLifecycler) changeState(ctx context.Context, state InstanceState)
...
+		if i.ReadOnly == readOnly {
+			return false
+		}
+
+		if readOnly {
+			i.ReadOnly = true
+			i.ReadOnlyUpdatedTimestamp = time.Now().Unix()
+		} else {
+			i.ReadOnly = false
+			i.ReadOnlyUpdatedTimestamp = 0
```
pracucci, line 549: In the `Lifecycler` implementation, this is also set to the `time.Now()` when we set `ReadOnly = false`. Reason is that its meaning is the timestamp of when the flag change the last time (if it ever changed). I think we should keep it consistent here.

ring/partition_instance_ring.go
```diff
@@ -74,15 +90,24 @@ func (r *PartitionInstanceRing) GetReplicationSetsForOperation(op Operation) ([]
 
 // GetReplicationSetForPartitionAndOperation returns a ReplicationSet for the input partition. If the partition doesn't
 // exist or there are no healthy owners for the partition, an error is returned.
-func (r *PartitionInstanceRing) GetReplicationSetForPartitionAndOperation(partitionID int32, op Operation) (ReplicationSet, error) {
+// pickHighestPreferrablyNonReadOnlyOwnerFromEachZone instructs to pick the highest instance from each zone when building the replication set,
+// non-read-only instances are preferred over read-only instances (this is more important than picking the highest instance).
+func (r *PartitionInstanceRing) GetReplicationSetForPartitionAndOperation(partitionID int32, op Operation, pickHighestPreferrablyNonReadOnlyOwnerFromEachZone bool) (ReplicationSet, error) {
```
pracucci, line 95: [thought] When we'll merge this stuff to `main`, I think we should consider not breaking `GetReplicationSetForPartitionAndOperation()` to not clutter it with options that only make sense when `multiplePartitions = true`. I think an option may be creating a new data type `MultiPartitionInstanceRing` that embeds `PartitionInstanceRing` data structure, where we override the `GetReplicationSetForPartitionAndOperation()` function and it calls the "underlying" `getReplicationSetForPartitionAndOperation()` (lower case one).

pracucci (approved):
Nice work! LGTM. I just left a comment.

ring/basic_lifecycler.go
colega:
```suggestion
			i.ReadOnlyUpdatedTimestamp = time.Now().Unix()
```

ring/partition_instance_ring.go
colega: This is a function you added just for usage-tracker in the `usage-tracker` branch. But yes, maybe not cluttering and having a different structure would be great.

colega:
I enabled rebase merging temporarily to avoid squash-merging this

PR #665, opened 2025-03-14
Upgrade github.com/sercand/kuberesolver to v6

aknuds1:
**What this PR does**:

Upgrade github.com/sercand/kuberesolver from v5 to v6, to unblock this upgrade in dependent projects.

This also requires bumping the minimum Go version to v1.22, which should be OK since the latest Go version is v1.24.

**Checklist**
- [x] Tests updated


narqo (approved):
🔥

PR #666, opened 2025-03-14
Build with Go v1.24 / upgrade golangci-lint to v1.64.8

aknuds1:
**What this PR does**:

Build with Go v1.24 in CI. This also requires upgrading golangci-lint to v1.64.8 and faillint to v1.15.0.

Dependency updates are blocked until we upgrade the Go version compatibility.

**Checklist**
- [x] Tests updated

### DIFF
--- a/crypto/tls/test/tls_integration_test.go
+++ b/crypto/tls/test/tls_integration_test.go
@@ -145,14 +145,14 @@ func newIntegrationClientServer(
 				if err == nil {
 					return false
 				}
-				return strings.Contains(err.Error(), "connection reset by peer") || strings.Contains(err.Error(), "broken pipe")
+				return strings.Contains(err.Error(), "connection reset by peer") ||
+					strings.Contains(err.Error(), "broken pipe") ||
+					strings.Contains(err.Error(), "client conn could not be established")
 			}
-			for i := 0; i < 3 && isRST(err) && tc.httpExpectError != nil; i++ {
+			for i := 0; i < 5 && isRST(err) && tc.httpExpectError != nil; i++ {
+				t.Logf("Sleeping before retry #%d, due to RST error: %s", i+1, err)
 				time.Sleep(100 * time.Millisecond)
 				resp, err = client.Do(req)
-				if err == nil {
-					defer resp.Body.Close()
-				}
 			}
 			if err == nil {
 				defer resp.Body.Close()
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.22.x'
+          go-version: '1.24.x'
       - name: Go mod check
         run: make mod-check
       - name: Lint
@@ -46,7 +46,7 @@ jobs:
       matrix:
         # We want to make sure dskit can support multiple golang versions
         # by ensuring the test would pass using all these supported versions.
-        go-version: ['1.21.x', '1.22.x', '1.23.x']
+        go-version: ['1.22.x', '1.23.x', '1.24.x']
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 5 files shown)


colega (commented):
Is it really necessary to modify our code just become someone decided in some linter that it's a good idea? I don't think this is making the code better.

colega:
> Dependency updates are blocked until we upgrade the Go version compatibility.

Which dependencies? We probably don't want to update those dependencies anyway because we don't want to force people to update to 1.24 if they don't want yet, right?

aknuds1:
> Is it really necessary to modify our code just become someone decided in some linter that it's a good idea? I don't think this is making the code better.

I don't agree that there's any point in fighting the linter on this. I subjectively actually prefer not shadowing built-in names. Let's discuss this with the Mimir team, since we also have to decide for our Mimir etc?

> Which dependencies? We probably don't want to update those dependencies anyway because we don't want to force people to update to 1.24 if they don't want yet, right?

@colega There's no forcing people to upgrade to v1.24, as evidenced by the PR still building as far back as v1.22. The PR just bumps the version we build with in CI to v1.24 (and effectively adds support for v1.24). 

There is one dependency update I want to do, which requires the minimum Go version to be bumped to v1.22. It won't build due to the old golangci-lint version. I don't see why it would be a problem to bump the minimum Go version to v1.22 (which doesn't happen in this PR in any case). E.g. Grafana Mimir has the policy of keeping the minimum Go version one version behind the latest (i.e., its minimum version is at v1.23 currently). Prometheus has the same policy.

aknuds1:
A test seems to be flaking a lot with v1.24. I'll have to investigate.

crypto/tls/test/tls_integration_test.go
```diff
@@ -145,14 +145,14 @@ func newIntegrationClientServer(
 				if err == nil {
 					return false
 				}
-				return strings.Contains(err.Error(), "connection reset by peer") || strings.Contains(err.Error(), "broken pipe")
+				return strings.Contains(err.Error(), "connection reset by peer") ||
+					strings.Contains(err.Error(), "broken pipe") ||
+					strings.Contains(err.Error(), "client conn could not be established")
```
aknuds1, line 150: Adding a new error message that tests would flake _a lot_ with under v1.24.

crypto/tls/test/tls_integration_test.go
```diff
@@ -145,14 +145,14 @@ func newIntegrationClientServer(
...
 			}
-			for i := 0; i < 3 && isRST(err) && tc.httpExpectError != nil; i++ {
+			for i := 0; i < 5 && isRST(err) && tc.httpExpectError != nil; i++ {
```
aknuds1, line 152: Let's try a few more attempts, since this test code was flaking a ton under v1.24.

crypto/tls/test/tls_integration_test.go
```diff
@@ -145,14 +145,14 @@ func newIntegrationClientServer(
...
+				t.Logf("Sleeping before retry #%d, due to RST error: %s", i+1, err)
 				time.Sleep(100 * time.Millisecond)
 				resp, err = client.Do(req)
-				if err == nil {
-					defer resp.Body.Close()
-				}
```
pracucci, line 155: Why this change? `resp` get overwritten each loop cycle. I think it was correct being here. What I think that can be removed is the `defer resp.Body.Close()` below, not this one.

pracucci (approved):
Please take a look at my comment before merging.

crypto/tls/test/tls_integration_test.go
aknuds1: I actually thought this through, and I'm pretty certain I've got it right. If `err == nil`, the loop breaks because the `isRST(err)` condition returns false. Ergo, in that case execution continues on to line 157 and executes `defer resp.Body.Close()`.

aknuds1: If I should turn out to be wrong in my analysis (I don't see how, as I've looked it over again), we can fix in a follow-up PR :)

PR #669, opened 2025-03-19
Make WatchPrefix buffer size configurable

aldernero:
This pull request introduces several changes to the `kv/memberlist/memberlist_client.go` file to improve the functionality and monitoring of the `WatchPrefix` function. The most important changes include adding a buffer size configuration for the `WatchPrefix` function, updating metrics, and ensuring proper default values for the buffer size.

### Configuration and Buffer Size:

* Added a new constant `watchPrefixBufferSize` to define the default buffer size for the `WatchPrefix` function.
* Introduced a new configuration field `WatchPrefixBufferSize` in the `KVConfig` struct to allow customization of the buffer size.
* Registered the new `WatchPrefixBufferSize` configuration field with flags to enable command-line configuration.
* Sets the default buffer size to 128. This is an increase from the previous static value of 16. By reviewing memberlist clusters of various sizes in production, 128 seems like a good default that should be sufficient for most environments without too much of an increase in memory.

### Metrics and Monitoring:

* Added a new gauge metric `numberOfKeyNotifications` to track the number of pending key notifications in the `WatchPrefix` function.
* Updated the `WatchPrefix` function to use the configured buffer size for its channel.
* Ensured the `WatchPrefixBufferSize` has a default value if not set in the configuration.
* Updated the `sendKeyNotifications` method to set the new `numberOfKeyNotifications` metric.

**Checklist**
- [x] Tests updated

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -31,6 +31,7 @@ const (
 	maxCasRetries              = 10          // max retries in CAS operation
 	noChangeDetectedRetrySleep = time.Second // how long to sleep after no change was detected in CAS
 	notifyMsgQueueSize         = 1024        // size of buffered channels to handle memberlist messages
+	watchPrefixBufferSize      = 128         // size of buffered channel for the WatchPrefix function
 )
 
 // Client implements kv.Client interface, by using memberlist.KV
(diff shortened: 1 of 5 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -171,6 +172,9 @@ type KVConfig struct {
 	// How much space to use to keep received and sent messages in memory (for troubleshooting).
 	MessageHistoryBufferBytes int `yaml:"message_history_buffer_bytes" category:"advanced"`
 
+	// Size of the buffer for key watchers.
+	WatchPrefixBufferSize int `yaml:"watch_prefix_buffer_size" category:"advanced"`
```
NickAnge, line 176: Can we add a Unit test  also to verify that the config option works as expected ?

aldernero: done

NickAnge (approved):
LGTM. Thanks a lot for addressing my comments

PR #671, opened 2025-03-24
Allow support for the cluster validation of HTTP requests

duricanikolic:
**What this PR does**:

This PR adds the following HTTP middlewares:
- on HTTP client side, `middleware.ClusterValidationRoundTripper`, an `http.RoundTripper` that enriches HTTP request headers with the cluster validation labels, and 
- on HTTP server side, `middleware.ClusterValidationMiddleware`, an implementation of `middleware.Interface` that checks if HTTP requests contain a correct cluster validation label, and returns an error if it is not the case.

Moreover, this PR introduces support for the configuration of the cluster validation on HTTP server side, via the following experimental CLI flags:
- `-server.cluster-validation.label`, that defines the server's cluster validation label. When the validation is enabled, this value will be compared with the cluster validation label received through the requests.
- `-server.cluster-validation.http.enabled`, used for enabling the cluster validation.
- `-server.cluster-validation.http.soft-validation`, that can be enabled only together with the cluster validation, and that implies a soft validation.
- `-server.cluster-validation.http.exclude-paths`, a comma-separated list of url paths that are excluded from the cluster validation check.

Fixes #<issue number>

**Checklist**
- [x] Tests updated


middleware/http_cluster.go
```diff
@@ -0,0 +1,140 @@
...
+	w.Header().Set("Content-Type", "application/json")
+
+	data, err := json.Marshal(e)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusNetworkAuthenticationRequired)
+	// We ignore errors here, because we cannot do anything about them.
```
duricanikolic, line 35: Note to reviewers: this comment was copied from a similar usage of `Write()` in Mimir.

middleware/http_cluster_test.go
```diff
@@ -0,0 +1,307 @@
...
+
+			recorder := httptest.NewRecorder()
+			req, err := http.NewRequest("GET", fmt.Sprintf("http://localhost:8080/%s", testCase.requestPath), nil)
+			require.NoError(t, err)
+			req.Header[clusterutil.ClusterVerificationLabelHeader] = []string{"client-cluster"}
+
+			handler.ServeHTTP(recorder, req)
+			require.Equal(t, testCase.expectedStatusCode, recorder.Code)
+			if recorder.Code != http.StatusOK {
+				var clusterValidationErr clusterValidationError
```
aknuds1, line 293: Could you also test that the server responds with `Content-Type` `application/json`?

middleware/http_cluster_test.go
```diff
@@ -0,0 +1,307 @@
...
+				if testCase.header != nil {
+					testCase.header(req)
+				}
+				handler.ServeHTTP(recorder, req)
+				if softValidation {
+					require.Equal(t, http.StatusOK, recorder.Code)
+				} else {
+					require.Equal(t, testCase.expectedStatusCode, recorder.Code)
+					if recorder.Code != http.StatusOK {
+						var clusterValidationErr clusterValidationError
```
aknuds1, line 220: Could you also test that the server responds with `Content-Type` `application/json`?

aknuds1 (changes requested):
The changelog changes aren't entirely correct, please see other comments as well.

CHANGELOG.md
```diff
@@ -103,12 +103,19 @@
...
+* [FEATURE] Add `middleware.ClusterUnaryServerInterceptor`, a `grpc.UnaryServerInterceptor` that checks if the incoming gRPC metadata contains a correct cluster validation label, and returns an error if it is not the case. #640 #648 #649 #655
 * [FEATURE] Server: Add support for adding `middleware.ClusterUnaryServerInterceptor` as `server.Server` unary interceptor via the following experimental configuration options: #650 #657
   * `-server.cluster-validation.label`
   * `-server.cluster-validation.grpc.soft-validation`
   * `-server.cluster-validation.grpc.enabled`
+* [FEATURE] Add `middleware.ClusterValidationRoundTripper`, an `http.RoundTripper` that enriches HTTP request headers with the cluster validation labels. #671
+* [FEATURE] Add `middleware.ClusterValidationMiddleware`, an implementation of `middleware.Interface` that checks if HTTP requests contain a correct cluster validation label, and returns an error if it is not the case. #671
+* [FEATURE] Server: Add support for adding `middleware.ClusterValidationMiddleware` as `server.Server` HTTP middleware via the following experimental configuration options: #671
+  * `-server.cluster-validation.label`
```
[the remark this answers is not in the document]
aknuds1: On second thought, the `-server.cluster-validation.label` flag is necessary. Maybe one could refer to it as necessary (although added by a previous PR)? I think I got confused by it looking as if it's added by this PR.

clusterutil/cluster_validation_config.go
```diff
@@ -68,3 +74,18 @@ func (cfg *ClusterValidationProtocolConfig) RegisterFlagsWithPrefix(prefix strin
...
+}
+
+func (cfg *ClusterValidationProtocolWithExcludedPathsConfig) Validate(prefix string, label string) error {
+	return cfg.ClusterValidationProtocolConfig.Validate(prefix, label)
+}
+
+func (cfg *ClusterValidationProtocolWithExcludedPathsConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	cfg.ClusterValidationProtocolConfig.RegisterFlagsWithPrefix(prefix, f)
+	excludedPathsFlag := prefix + "excluded-paths"
+	f.StringVar(&cfg.ExcludedPaths, excludedPathsFlag, "", "Comma-separated list of url paths that are excluded from the cluster validation check.")
```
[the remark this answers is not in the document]
duricanikolic: I will remove this variable

middleware/http_cluster.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/grafana/dskit/httpgrpc"
+)
+
+type clusterValidationError struct {
+	Message string `json:"message"`
+}
+
+// writeAsJSON writes this error as JSON to the HTTP response.
+func (e *clusterValidationError) writeAsJSON(w http.ResponseWriter) {
+	w.Header().Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
duricanikolic: Thank you. I've done it.

clusterutil/clusterutil.go
```diff
@@ -42,3 +50,30 @@ func GetClusterFromIncomingContext(ctx context.Context) (string, error) {
 	}
 	return clusterIDs[0], nil
 }
+
+// PutClusterIntoHeader enriches the given request's header with the provided value
+// for the ClusterVerificationLabelHeader key.
+// Empty values are ignored.
+func PutClusterIntoHeader(req *http.Request, cluster string) {
```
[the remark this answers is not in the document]
duricanikolic: @aknuds1 this is the same approach we used for the gRPC part. So, if you don't have strong feelings about this, I'd keep it. Otherwise, we'll have to remove the gRPC-related methods too.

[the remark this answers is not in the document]
aknuds1: OK, let's keep them in that case.

middleware/http_cluster.go
```diff
@@ -0,0 +1,140 @@
...
+
+func ClusterValidationRoundTripper(cluster string, invalidClusterValidationReporter InvalidClusterValidationReporter, next http.RoundTripper) RoundTripperFunc {
+	validateClusterValidationRoundTripperInputParameters(cluster, invalidClusterValidationReporter)
+	return func(req *http.Request) (*http.Response, error) {
+		clusterutil.PutClusterIntoHeader(req, cluster)
+		resp, err := next.RoundTrip(req)
+		if err != nil {
+			return nil, err
+		}
+		if body, err := io.ReadAll(resp.Body); err == nil {
```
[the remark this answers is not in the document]
duricanikolic: > Also, I think resp.Body should be closed if reading it was successful?
@aknuds1 IMO this is not correct. My idea is to try to convert a response `resp` into `clusterValidationError`. If this was possible, I then process the error, we close the response body, and return another one. 
Otherwise, I want to propagate the response `resp`. In this case we don't need to close it. In this case, I need to reset the response body content by calling `io.NopCloser()`, because once we call `io.ReadAll()`, the response body becomes empty.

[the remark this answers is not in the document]
aknuds1: > Otherwise, I want to propagate the response resp. In this case we don't need to close it.

Why do you not need to close `resp.Body`, considering you are replacing it with `io.NopCloser(bytes.NewReader(body))` before returning?

clusterutil/clusterutil.go
```diff
@@ -42,3 +50,30 @@ func GetClusterFromIncomingContext(ctx context.Context) (string, error) {
...
+	if req.Header == nil {
+		req.Header = make(http.Header)
+	}
+	req.Header.Set(ClusterVerificationLabelHeader, cluster)
+}
+
+// GetClusterFromRequest returns a single value corresponding to the ClusterVerificationLabelHeader key
+// from the header of the given request, if it exists.
+// In all other cases an error is returned.
+func GetClusterFromRequest(req *http.Request) (string, error) {
```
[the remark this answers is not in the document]
duricanikolic: This is the same approach we used for the gRPC part. So, if you don't have strong feelings about this, I'd keep it. Otherwise, we'll have to remove the gRPC-related methods too.

middleware/http_cluster.go
```diff
@@ -0,0 +1,142 @@
...
+	validateClusterValidationRoundTripperInputParameters(cluster, invalidClusterValidationReporter)
+	return func(req *http.Request) (*http.Response, error) {
+		clusterutil.PutClusterIntoHeader(req, cluster)
+		resp, err := next.RoundTrip(req)
+		if err != nil {
+			return nil, err
+		}
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return resp, err
```
aknuds1, line 50: I don't think the caller should deal with `resp` if this function returns an error. We should close `resp.Body`, the caller shouldn't have to care about that when receiving an error.
```suggestion
			return nil, err
```

middleware/http_cluster.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		var clusterValidationErr clusterValidationError
+		err = json.Unmarshal(body, &clusterValidationErr)
+		if err != nil {
+			resp.Body = io.NopCloser(bytes.NewReader(body))
+			return resp, nil
+		}
+		defer resp.Body.Close()
```
aknuds1, line 58: The problem now is that `resp.Body` gets closed in only one of the three possible code branches. On reading the code again, I think that `defer resp.Body.Close()` should be moved to line 48. This is so `resp.Body` gets closed for all code branches, as there is no branch where the caller should receive the original `resp.Body`:

* If this function returns an error, the caller shouldn't have to close `resp.Body` if `resp` is non-nil, but actually ignore `resp`.
* If `io.ReadAll` succeeds, we have exhausted `resp.Body` and it should be closed.

aknuds1 (commented):
Please see my explanation.

middleware/http_cluster.go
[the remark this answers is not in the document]
duricanikolic: @aknuds1 it did. The problem is probably the fact that I replied your comments before pushing a fix.

middleware/http_cluster.go
duricanikolic: I have fixed this.

middleware/http_cluster.go
duricanikolic: I have put resp.Body.Close() at line 48, as suggested

middleware/http_cluster.go
```diff
@@ -0,0 +1,142 @@
...
+func ClusterValidationRoundTripper(cluster string, invalidClusterValidationReporter InvalidClusterValidationReporter, next http.RoundTripper) RoundTripperFunc {
+	validateClusterValidationRoundTripperInputParameters(cluster, invalidClusterValidationReporter)
+	return func(req *http.Request) (*http.Response, error) {
+		clusterutil.PutClusterIntoHeader(req, cluster)
+		resp, err := next.RoundTrip(req)
+		if err != nil {
+			return nil, err
+		}
+		body, err := io.ReadAll(resp.Body)
+		defer resp.Body.Close()
```
aknuds1, line 49: Nit, but I meant the following, which is more logical:
```suggestion
		defer resp.Body.Close()

		body, err := io.ReadAll(resp.Body)
```

middleware/http_cluster.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return nil, err
+		}
+		var clusterValidationErr clusterValidationError
+		err = json.Unmarshal(body, &clusterValidationErr)
```
aknuds1, line 54: I'm still wondering if you can't check the response's `Content-Type` header and/or the response status code, for whether a cluster validation error should be expected?

aknuds1 (commented):
Please see comment about checking response content-type header/status code, for whether to expect a cluster validation error. It seems more robust to me.

server/server.go
```diff
@@ -569,6 +569,10 @@ func BuildHTTPMiddleware(cfg Config, router *mux.Router, metrics *Metrics, logge
 	} else {
 		httpMiddleware = append(defaultHTTPMiddleware, cfg.HTTPMiddleware...)
 	}
+	if cfg.ClusterValidation.HTTP.Enabled {
+		excludedPaths := strings.Split(cfg.ClusterValidation.HTTP.ExcludedPaths, ",")
```
colega, line 573: I think that instead of doing this we should just define `ExcludedPaths` as `flagext.StringSlice` or `flagext.CommaSeparatedStringSlice` (don't remember the exact name).

middleware/http_cluster.go
duricanikolic: As discussed via Slack, I didn't understand what you meant. That was my bad, and I added the checks you mentioned.

middleware/http_cluster.go
```diff
@@ -0,0 +1,148 @@
...
+		if resp.Header.Get("Content-Type") != "application/json" {
+			return resp, nil
+		}
+		defer resp.Body.Close()
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+		var clusterValidationErr clusterValidationError
+		err = json.Unmarshal(body, &clusterValidationErr)
```
aknuds1, line 60: Come to think of it, we probably want to protect ourselves from a too long response body:
```suggestion
		var clusterValidationErr clusterValidationError
		err = json.NewDecoder(io.LimitReader(resp.Body, 1024)).Decode(&clusterValidationErr)
```

server/server.go
duricanikolic: I have used `flagext.StringSliceCSV`

middleware/http_cluster.go
```diff
@@ -0,0 +1,148 @@
...
+		if err != nil {
+			resp.Body = io.NopCloser(bytes.NewReader(body))
+			return resp, nil
```
aknuds1, line 63: Since the server returned an unexpected format (and we don't necessarily read the whole body), I think we should return internal server error (it's a scenario that should not happen):
```suggestion
			resp.Body = nil
			resp.StatusCode = http.InternalServerError
			resp.Header.Set("Content-Type", "text/plain")
			return resp, nil
```

aknuds1 (commented):
I think we should limit the read from the body, to protect ourselves.

middleware/http_cluster.go
colega: I think this would be unexpected and hard to debug if happens. 

Imagine that you're seeing lots of internal errors on the client but server doesn't send any?

Let's just return something like `fmt.Errorf("server returned 511 but the result wasn't a valid json: %w", err)`,

middleware/http_cluster.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	reB.WriteString(")")
+	reExcludedPath := regexp.MustCompile(reB.String())
+
+	return Func(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			msgs, err := checkClusterFromRequest(r, cluster, softValidation, reExcludedPath)
+			if len(msgs) > 0 {
+				level.Warn(logger).Log(msgs...)
+			}
```
colega, line 101: Why do we want to return the msgs here instead of passing the logger to `checkClusterFromRequest`?

colega (approved):
LGTM but let's see Mimir's changes before merging please.

middleware/http_cluster.go
aknuds1: I agree w/ Oleg, I think that's a more readable approach.

clusterutil/cluster_validation_config_test.go
```diff
@@ -23,6 +23,7 @@ func TestClusterValidationConfig_RegisteredFlags(t *testing.T) {
 	require.Len(t, registeredFlags.Flags, 1)
 	_, ok := registeredFlags.Flags["label"]
 	require.True(t, ok)
+
```
aknuds1, line 26: This change must be unintentional, please revert it.
```suggestion
```

middleware/http_cluster_test.go
```diff
@@ -0,0 +1,307 @@
...
+			expectedStatusCode: http.StatusNetworkAuthenticationRequired,
+			expectedErrorMsg:   `rejected request with empty cluster validation label - it should be "cluster"`,
+		},
+		"no request cluster and non-empty server cluster give an error if soft validation disabled": {
+			serverCluster:      "cluster",
+			expectedLogs:       `level=warn msg="request with no cluster validation label" path=/Test/Me cluster_validation_label=cluster soft_validation=%v`,
+			expectedStatusCode: http.StatusNetworkAuthenticationRequired,
+			expectedErrorMsg:   `rejected request with empty cluster validation label - it should be "cluster"`,
+		},
+		"if the incoming request contains more than one cluster labels and soft validation is disabled an error is returned": {
```
aknuds1, line 183: [Nit]
```suggestion
		"if the incoming request contains more than one cluster label and soft validation is disabled an error is returned": {
```

aknuds1 (approved):
LGTM, but there's an unrelated change that should be reverted. I also agree with @colega's comment about logging. Plus, I think we should implement limited reading in a quick follow-up.

middleware/http_cluster.go
duricanikolic: As agreed, this will be done in a follow-up.

middleware/http_cluster.go
duricanikolic: Good idea. Thank you. I have applied this.

aknuds1 (approved):
I looked over the draft Mimir PR making use of this, thanks. LGTM.

PR #672, opened 2025-03-25
ring: Fix pathological case when an entire zone leaves

56quarters:
**What this PR does**:

This change improves performance in a the case where an entire zone is not ACTIVE and the replication set is meant to be extended. Previously, when an entire zone was unavailable, the ring kept searching for instances by looking at every single token trying to find an instance in the required zone that was ACTIVE. This meant thousands of iterations to find a host that would never work.

This change keeps track of the number of hosts that we have examined in each zone. It returns early once we have either found the hosts in each zone we need _OR_ we have examined all hosts in the zone and so know that we won't find one.

N/A

**Checklist**
- [X] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -535,13 +535,14 @@ func (r *Ring) findInstancesForKey(key uint32, op Operation, bufDescs []Instance
 		// to have low single-digit number of hosts.
 		distinctHosts = bufHosts[:0]
 
-		hostsPerZone       = make(map[string]int)
-		targetHostsPerZone = max(1, replicationFactor/maxZones)
+		examinedHostsPerZone = make(map[string]int)
+		foundHostsPerZone    = make(map[string]int)
+		targetHostsPerZone   = max(1, replicationFactor/maxZones)
 	)
 
 	for i := start; len(distinctHosts) < min(maxInstances, n) && iterations < len(r.ringTokens); i++ {
-		// If we have the target number of instances in all zones, stop looking.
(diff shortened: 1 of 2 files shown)


56quarters:
Comparison against [a commit from early January](https://github.com/grafana/dskit/tree/56quarters/ring-prev) which predates my changes in #632:

```
$ benchstat prev.txt current.txt 
goos: linux
goarch: amd64
                           │   prev.txt    │             current.txt             │
                           │    sec/op     │   sec/op     vs base                │
Ring_Get_OneZoneLeaving-16   8329.3µ ± 13%   114.6µ ± 5%  -98.62% (p=0.000 n=10)

                           │   prev.txt   │           current.txt            │
                           │     B/op     │     B/op      vs base            │
Ring_Get_OneZoneLeaving-16   10.27Ki ± 0%   10.27Ki ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                           │  prev.txt  │          current.txt           │
                           │ allocs/op  │ allocs/op   vs base            │
Ring_Get_OneZoneLeaving-16   12.00 ± 0%   12.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

pstibrany (commented):
I haven't seen ring code in a while, and originally I thought I've found a bug in the PR, but staring at it for some more time, I'm wrong and the change seems legit and makes sense to me.

julienduchesne (approved):
It makes sense to me, not super familiar with ring code though

PR #673, opened 2025-03-26
Memberlist: switch a noisy info to debug

colega:
In cells with thousands of memberlist component this just says the same thing all the time, and it's not really helpful.


kv/memberlist/memberlist_client.go
```diff
@@ -555,7 +555,7 @@ func (m *KV) running(ctx context.Context) error {
 		case <-obsoleteEntriesTickerChan:
 			// cleanupObsoleteEntries is normally called during push/pull, but if there are no other
 			// nodes to push/pull with, we can call it periodically to make sure we remove unused entries from memory.
-			level.Info(m.logger).Log("msg", "initiating cleanup of obsolete entries")
+			level.Debug(m.logger).Log("msg", "initiating cleanup of obsolete entries")
```
56quarters, line 558: I don't even think it makes sense to log this at debug, this is better as a metric since it's not logging anything interesting.

56quarters (approved):
LGTM with a non-blocking suggestion

PR #674, opened 2025-03-31
Memberlist: add -memberlist.abort-if-fast-join-fails support and retries on DNS resolution

pracucci:
**What this PR does**:

In Mimir we have an issue (https://github.com/grafana/mimir/issues/10866) caused by memberlist cluster not being initialised when other services, depending memberlist, are started. As of today, there's no way to enforce that memberlist successfully runs a fast-join before switching to the Running state.

In this PR I propose to:

1. Add a new CLI flag, named `-memberlist.abort-if-fast-join-fails` (disabled by default), that allows to have memberlist service failing when the fast-join fails
2. Add retries to DNS resolution used by the cluster join procedure (both fast-join and full join), given this is the actual error we get, due to unreliable networking issues right after a new K8S node is created

How will be used in Mimir? My plan is to enable the new `-memberlist.abort-if-fast-join-fails` only for ingesters, because in our context it's critical to have an updated view of the memberlist cluster when the ingester service startup (otherwise it can't correctly compute local limits).

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -154,12 +154,13 @@ type KVConfig struct {
 	ClusterLabelVerificationDisabled bool   `yaml:"cluster_label_verification_disabled" category:"advanced"`
 
 	// List of members to join
-	JoinMembers      flagext.StringSlice `yaml:"join_members"`
-	MinJoinBackoff   time.Duration       `yaml:"min_join_backoff" category:"advanced"`
-	MaxJoinBackoff   time.Duration       `yaml:"max_join_backoff" category:"advanced"`
-	MaxJoinRetries   int                 `yaml:"max_join_retries" category:"advanced"`
-	AbortIfJoinFails bool                `yaml:"abort_if_cluster_join_fails"`
-	RejoinInterval   time.Duration       `yaml:"rejoin_interval" category:"advanced"`
+	JoinMembers          flagext.StringSlice `yaml:"join_members"`
+	MinJoinBackoff       time.Duration       `yaml:"min_join_backoff" category:"advanced"`
+	MaxJoinBackoff       time.Duration       `yaml:"max_join_backoff" category:"advanced"`
+	MaxJoinRetries       int                 `yaml:"max_join_retries" category:"advanced"`
+	AbortIfFastJoinFails bool                `yaml:"abort_if_cluster_fast_join_fails" category:"advanced"`
+	AbortIfJoinFails     bool                `yaml:"abort_if_cluster_join_fails"`
+	RejoinInterval       time.Duration       `yaml:"rejoin_interval" category:"advanced"`
 
 	// Remove LEFT ingesters from ring after this timeout.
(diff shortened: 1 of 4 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -196,7 +201,8 @@ func (cfg *KVConfig) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.DurationVar(&cfg.MinJoinBackoff, prefix+"memberlist.min-join-backoff", 1*time.Second, "Min backoff duration to join other cluster members.")
 	f.DurationVar(&cfg.MaxJoinBackoff, prefix+"memberlist.max-join-backoff", 1*time.Minute, "Max backoff duration to join other cluster members.")
 	f.IntVar(&cfg.MaxJoinRetries, prefix+"memberlist.max-join-retries", 10, "Max number of retries to join other cluster members.")
-	f.BoolVar(&cfg.AbortIfJoinFails, prefix+"memberlist.abort-if-join-fails", cfg.AbortIfJoinFails, "If this node fails to join memberlist cluster, abort.")
+	f.BoolVar(&cfg.AbortIfFastJoinFails, prefix+"memberlist.abort-if-fast-join-fails", false, "Abort if this node fails the fast memberlist cluster joining procedure at startup. When enabled, it's guaranteed that other services, depending on memberlist, have an updated view over the cluster state when they're started.")
```
NickAnge, line 204: > When enabled, it's guaranteed that other services, depending on memberlist, have an updated view over the cluster state when they're started.

I am not sure I understand this ⬆️  🤔  If this is true, then its not guaranteed that other services will have an updated view no ? Cause with one failure we skip the retries 🤔  Please correct me if I misunderstood.

kv/memberlist/memberlist_client_test.go
```diff
@@ -1114,7 +1115,91 @@ func TestMemberlistFailsToJoin(t *testing.T) {
...
+		"first resolution succeed": {
+			numResolveFailures: 0,
+			expectedAddrs:      addrsMocked,
+		},
+		"first 3 resolutions fails": {
+			numResolveFailures: 3,
+			expectedAddrs:      addrsMocked,
+		},
+		"all resolutions fail": {
+			numResolveFailures: math.MaxInt,
```
NickAnge, line 1173: Are we trying to show that it will stop attempting after 10 retries? If so, should we also check the number of retries for each test, or is it too much?

NickAnge (commented):
LGTM

kv/memberlist/memberlist_client.go
NickAnge: Amazing. Makes total sense , thanks for the explanation.

kv/memberlist/memberlist_client_test.go
pracucci: Correct. Added an assertion on max retries in https://github.com/grafana/dskit/pull/674/commits/163670c8a03c1906099afdaddec14e60f9921165

PR #675, opened 2025-04-02
Fix WatchKey() not sending out notifications before the full-join has been completed, when -memberlist.notify-interval is configured

pracucci:
**What this PR does**:

When `-memberlist.notify-interval` is configured, `WatchKey()` notifications are sent out asynchronously by a goroutine. However, the goroutine is started only after the full-join is completed, which may take few minutes on large clusters. This cause any other component (e.g. ring client) not getting any update until the full-join is completed.

This PR fixes it, simply starting the goroutine before running the full-join. The new test I've added fails on `main` for the test case where `-memberlist.notify-interval` is configured.

N/A

**Checklist**
- [x] Tests updated


seizethedave (approved):
Looks good. Thank you for fixing it.

PR #679, opened 2025-04-11
SpanLogger: don't embed Span

colega:
**What this PR does**:

Instead of embedding the span and letting everyone access it, expose methods explicitly.

This would allow us to replace the implementation with OTEL without having to modify all usages of the Spanlogger.

Users of this library will have to replace the `ext.Error.Set(spanlog, true)` by `spanlog.SetError`.

Ref: https://github.com/grafana/mimir-squad/issues/1511#event-16121302090

**Checklist**
- [x] Tests updated


56quarters (approved):
Nice

PR #680, opened 2025-04-16
Server: Add metric server_invalid_cluster_validation_label_requests_total

aknuds1:
**What this PR does**:

Add server metric `server_invalid_cluster_validation_label_requests_total`, that is incremented whenever the gRPC or HTTP server receives a request with an invalid cluster validation label. The metric is incremented even when validation is in soft mode.

Please note that I also had to fix some tests, so they won't register the new metric multiple times (causes a panic).

**Checklist**
- [x] Tests updated

### DIFF
--- a/middleware/grpc_cluster.go
+++ b/middleware/grpc_cluster.go
@@ -70,7 +70,7 @@ func handleClusterValidationError(err error, method string, invalidClusterValida
 // If an empty cluster label or nil logger are provided, ClusterUnaryServerInterceptor panics.
 // If the softValidation parameter is true, errors related to the cluster label validation are logged, but not returned.
 // Otherwise, an error is returned.
-func ClusterUnaryServerInterceptor(cluster string, softValidation bool, invalidClusterValidations *prometheus.CounterVec, logger log.Logger) grpc.UnaryServerInterceptor {
+func ClusterUnaryServerInterceptor(cluster string, softValidation bool, invalidClusterRequests *prometheus.CounterVec, logger log.Logger) grpc.UnaryServerInterceptor {
 	validateClusterServerInterceptorInputParameters(cluster, logger)
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 		// We skip the gRPC health check.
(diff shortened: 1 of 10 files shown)


middleware/metrics.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// NewInvalidClusterValidations registers and returns a new counter metric server_request_invalid_cluster_validation_labels_total.
+func NewInvalidClusterValidations(reg prometheus.Registerer) *prometheus.CounterVec {
+	return promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Name: "server_request_invalid_cluster_validation_labels_total",
```
colega, line 11:
```suggestion
		Name: "server_invalid_cluster_validation_label_requests_total",
```

☝️  We count requests, not labels.

middleware/grpc_cluster.go
```diff
@@ -69,15 +70,15 @@ func handleClusterValidationError(err error, method string, invalidClusterValida
 // If an empty cluster label or nil logger are provided, ClusterUnaryServerInterceptor panics.
 // If the softValidation parameter is true, errors related to the cluster label validation are logged, but not returned.
 // Otherwise, an error is returned.
-func ClusterUnaryServerInterceptor(cluster string, softValidation bool, logger log.Logger) grpc.UnaryServerInterceptor {
+func ClusterUnaryServerInterceptor(cluster string, softValidation bool, invalidClusterValidations *prometheus.CounterVec, logger log.Logger) grpc.UnaryServerInterceptor {
```
56quarters, line 73: nit: it's a bit awkward to call this "invalid ... validations". Maybe just `invalidCluster`?

aknuds1: Now that you say it, you've got a point. I didn't really name it myself, just copied from a corresponding client side metric in Mimir.

aknuds1: I'll rename it as above, since it will match the metric name pretty well.

56quarters: Makes sense, thanks!

PR #385, opened 2023-09-21
[WIP] Mimimum changes in dskit in order to migrate to opentelemetry

ying-jeanne:
**What this PR does**:

This PR showcases minimal code adjustments in dskit to transition from using Mimir/GEM to open-telemetry instrumentation. While this PR isn't meant for immediate merging, it serves as a basis for reviewing the changes and creating a design document. This document will ensure dskit's backward compatibility for teams not yet prepared for migration.

Key Inclusions in this PR:

- Introduction of oteltracing.go: This file initializes the tracerProvider, mirroring the functionality of opentracing.SetGlobalTracer(). It preserves the Jaeger environment tracing settings, mainly concerning the following environment variables: 
```
JAEGER_AGENT_HOST,JAEGER_TAGS,JAEGER_SAMPLER_MANAGER_HOST_PORT, JAEGER_SERVICE_NAME, JAEGER_SAMPLER_TYPE, and JAEGER_SAMPLER_PARAM
```
It also maintains compatibility for HTTP requests by decoding both opentracing (Jaeger trace header) and opentelemetry (W3C header).

- Incorporation of spanlogger.go: This file combines a spanLogger object, which bundles a span and logger together. Many projects use this object, even if they don't employ dskit's tracing instrumentation. It exposes a public opentracing.Span accessible to libraries that import it.

- grpcclient/instrumentation.go: This section addresses the different interceptors used by opentracing and opentelemetry.

- use of otelhttp and otelhttptrace: These modules are used whenever there is a need to inject context into request headers or extract request headers into context. They replace the previous methods since they manage the injection and extraction of corresponding headers.

- Utilization of opentelemetry's test tracerprovider with an in-memory exporter: All traces in unit tests have been adapted to use the tracetest tracerprovider provided by opentelemetry.

- Modification in extracting the traceID from context: Instead of directly casting with jaeger.SpanContext from the context, this PR replaces it with otelSpan.SpanContext().TraceID().

This PR is being testing on dev environment.

**Which issue(s) this PR fixes**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #681, opened 2025-04-17
Support for OTel tracing

colega:
## What this PR does

This adds support for OTel everywhere. It is similar to https://github.com/grafana/dskit/pull/385 which migrated to OTel, and takes pieces of code from there, but it does **not** remove support for OpenTracing.

## `spanlogger` package changes

The new `spanlogger.NewOTel` will create a new OTel span, using the tracer provided. We're asking for a tracer instead of using one from `spanlogger` because that's how [Tempo did it with per-package Tracer](https://github.com/grafana/tempo/pull/3646).

`spanlogger.FromContext` will find the existing span in the context, either OpenTracing or OTel one, and will return a `SpanLogger` that abstracts both.

In order to achieve that `spanlogger.SpanLogger` now contains either an `opentelemetry.Span` or an `trace.Span` (OTel). If there's an OTel one, it will do the operations with it, if not it will fall back to OpenTracing implementation. Some calls that were cheaper before, like `LogFields` now are more expensive since we have to translate them to OTel attributes.

## `httpgrpc/server` package changes

When serving an HTTP request, we'll use OpenTracing to start the span if `opentracing.IsGlobalTracerRegistered()`. If there's an OTel span in the request context, we'll inject that one to the request.

## `instrument.CollectedRequest()` changes

This function is widely used, so I migrated it from using `opentracing.StartSpanFromContext` to `tracing.StartSpanFromContext`: a new wrapper in the `tracing` package that will use the configured tracing library.

## `tracing` package changes

The new `tracing.StartSpanFromContext` creates an adapter `tracing.Span` that acts like an `opentracing.Span` or `trace.Span` depending on the library used in the parent span from the `context.Context` provided to `StartSpanFromContext`.

This abstraction should disappear once we've migrated our dependencies to OTel. 🤞 (Famous last words)

The `ExtractSampledTraceID`, `ExtractTraceID` and `ExtractTraceSpanID` now support both opentracing and OTel libraries. This should solve https://github.com/grafana/dskit/pull/662 as well.

Finally, I brought from https://github.com/grafana/dskit/pull/385 the `NewOTelFromJaegerEnv` function that will configure OTel tracing using Jaeger env vars, with Jaeger exporter.

## `server` package changes

Server now has the `otelgrpc.NewServerHandler()` which will inject the OTel Spans to the context.

## `middleware` package changes

When propagating an HTTP or HTTPGRPC requests, we'll use OpenTracing to start the span if `opentracing.IsGlobalTracerRegistered()`. If there's an OTel span in the request context, we'll inject that one to the request.

## Which issue(s) this PR fixes

Ref: https://github.com/grafana/mimir/issues/2708

**Checklist**
- [x] Tests updated


colega:
Still draft because I need to fix tests, add tests, and double check the server configuration of middlewares.

colega:
I assume it's inevitable with my task to add the `otel/trace` dependency. The ultimate goal, once Loki & Pyroscope are migrated too, is to remove the `opentracing-go` one.

tracing/otel.go
```diff
@@ -0,0 +1,253 @@
...
+	cfg := config{}
+	var err error
+
+	// Parse reporting agent configuration
+	if e := os.Getenv(envJaegerEndpoint); e != "" {
+		u, err := url.ParseRequestURI(e)
+		if err != nil {
+			return cfg, errors.Wrapf(err, "cannot parse env var %s=%s", envJaegerEndpoint, e)
+		}
+		cfg.collectorEndpoint = u.String()
```
krajorama, line 111: Kind of weird to call it collectorEndpoint as we're using Jaeger exporter to write to it and also doesn't come from OTEL env vars: https://opentelemetry.io/docs/specs/otel/protocol/exporter/#endpoint-urls-for-otlphttp

Let's call it cfg.jaegerEndpoint

tracing/otel_test.go
```diff
@@ -0,0 +1,146 @@
...
+				expectedOutput: []attribute.KeyValue{
+					attribute.String("key1", "value1"),
+					attribute.String("key2", "value2"),
+				},
+				expectedError: nil,
+			},
+			{
+				input: "key1=${EXISTENT_ENV_KEY},key2=${NON_EXISTENT_ENV_KEY:default_value}",
+				expectedOutput: []attribute.KeyValue{
+					attribute.String("key1", os.Getenv("EXISTENT_ENV_KEY")),
```
krajorama, line 37: nit: let's just hardcode expected value
```suggestion
					attribute.String("key1", "env_value"),
```

tracing/otel.go
```diff
@@ -0,0 +1,253 @@
...
+	*tracesdk.TracerProvider
+}
+
+func (c Closer) Close() error {
+	return c.Shutdown(context.Background())
+}
+
+// NewResource creates a new OpenTelemetry resource using the provided service name and custom attributes.
+// This resource will be used for creating both tracers and meters, enriching telemetry data with context.
+func NewResource(serviceName string, customAttributes []attribute.KeyValue) (*resource.Resource, error) {
```
krajorama, line 240: It's probably fine for first iteration that service.namespace and service.version  aren't here. https://opentelemetry.io/docs/specs/semconv/resource/#service  I haven't seen semconf.ServiceNamespace used in grafana code. (Something to be aware of anyway)

tracing/span.go
```diff
@@ -0,0 +1,140 @@
...
+type Span struct {
+	opentracingSpan opentracing.Span
+	otelSpan        trace.Span
+}
+
+// StartSpanFromContext starts a new span from the context using the parent tracing library.
+// If parent span is not sampled, it returns a noop span.
+func StartSpanFromContext(ctx context.Context, operation string, options ...SpanOption) (*Span, context.Context) {
+	otelSpan, opentracingSpan, sampled := SpanFromContext(ctx)
+	if sampled {
```
krajorama, line 26: nit: let's do an early return on `!sampled` to reduce indent

tracing/span_test.go
```diff
@@ -0,0 +1,167 @@
...
+		{
+			name:     "int64 slice",
+			kv:       kv{"key", []int64{1, 2, 3}},
+			expected: attribute.Int64Slice("key", []int64{1, 2, 3}),
+		},
+		{
+			name:     "stringer",
+			kv:       kv{"key", testStringer{}},
+			expected: attribute.Stringer("key", testStringer{}),
+		},
```
krajorama, line 67: nit: does not test `[]byte` and fallback.

tracing/tracing.go
```diff
@@ -6,58 +6,26 @@ package tracing
...
-	if cfg.Sampler.SamplingServerURL == "" && cfg.Reporter.LocalAgentHostPort == "" && cfg.Reporter.CollectorEndpoint == "" {
-		return nil, ErrBlankTraceConfiguration
-	}
-
-	return installJaeger(serviceName, cfg, options...)
-}
-
 // ExtractTraceID extracts the trace id, if any from the context.
 func ExtractTraceID(ctx context.Context) (string, bool) {
 	if tid, _, ok := extractJaegerContext(ctx); ok {
```
krajorama, line 23: nit: maybe we could reuse `SpanFromContext` in the this and following function? Isn't extraction from context now implemented separately for otrace and otel and then in `SpanFromContext` for both at the same time?

httpgrpc/server/server.go
```diff
@@ -177,13 +180,17 @@ func NewClient(address string) (*Client, error) {
...
+
 	dialOptions := []grpc.DialOption{
 		grpc.WithDefaultServiceConfig(grpcServiceConfig),
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
-		grpc.WithChainUnaryInterceptor(
-			otgrpc.OpenTracingClientInterceptor(opentracing.GlobalTracer()),
-			middleware.ClientUserHeaderInterceptor,
-		),
+		grpc.WithChainUnaryInterceptor(unaryInterceptors...),
+		grpc.WithStatsHandler(otelgrpc.NewClientHandler()),
```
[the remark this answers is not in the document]
colega: I wasn't going to... but ok, fair enough, let's add a check: https://github.com/grafana/dskit/pull/681/commits/c59312f51d3b0a7243ca848857f448538bead045

middleware/http_tracing.go
```diff
@@ -54,6 +69,38 @@ func (t Tracer) Wrap(next http.Handler) http.Handler {
 	return nethttp.Middleware(opentracing.GlobalTracer(), next, options...)
 }
 
+func (t Tracer) wrapWithOTel(next http.Handler) http.Handler {
+	// If no OpenTracing, let's do OTel.
```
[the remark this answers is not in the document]
colega: Right, fixed: https://github.com/grafana/dskit/pull/681/commits/cf8aa951952fe61e8ddebdd5d7ba8da6a01d2dfa

middleware/http_tracing.go
```diff
@@ -54,6 +69,38 @@ func (t Tracer) Wrap(next http.Handler) http.Handler {
...
+	tracingMiddleware := otelhttp.NewHandler(next, "http.tracing", otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
+		return httpOperationName(r)
+	}))
+
+	// Wrap the 'tracingMiddleware' to capture its execution
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if labeler, ok := otelhttp.LabelerFromContext(r.Context()); ok {
+			// add a tag with the client's user agent to the span
```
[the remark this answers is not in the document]
colega: Right, fixed: https://github.com/grafana/dskit/pull/681/commits/586d82e6188cf57d65023863fe79428c34b495c3

spanlogger/opentracing_spanlogger_test.go
```diff
@@ -0,0 +1,269 @@
...
+		t.Run(name, func(t *testing.T) {
+			t.Run("logging with Log()", func(t *testing.T) {
+				logs, spanLogger, span := setupTest(t, loggerFactory)
+
+				_, thisFile, lineNumberTwoLinesBeforeFirstLogCall, ok := runtime.Caller(0)
+				require.True(t, ok)
+				_ = spanLogger.Log("msg", "this is a test")
+
+				logged := logs.String()
+				require.Contains(t, logged, "caller="+toCallerInfo(thisFile, lineNumberTwoLinesBeforeFirstLogCall+2))
```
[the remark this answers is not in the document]
colega: Heh, it is, but I just moved these tests from the previous `spanlogger_test.go` file :)

spanlogger/otel_spanlogger_test.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/grafana/dskit/user"
+)
+
+var (
+	tracer       = otel.Tracer("dskit/spanlogger.test")
+	spanExporter = tracetest.NewInMemoryExporter()
+)
+
+func init() {
+	otel.SetTracerProvider(tracesdk.NewTracerProvider(tracesdk.WithSpanProcessor(tracesdk.NewSimpleSpanProcessor(spanExporter))))
```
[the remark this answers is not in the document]
colega: Oh... well this comment made me go deep into the can of worms.
I tried moving this to each test, but then all tests started failing because the `var tracer = otel.Tracer()` was no longer using the correct tracer provider. This was a question I was asking myself for a long time, [I saw that in Tempo's switch](https://github.com/grafana/tempo/pull/3646) they were creating a global `var tracer = otel.Tracer(...)` but that's executed before we configure the tracer provider in `main()`, so is that tracer even valid? It should be, right? Because they are using that code, so I had to look deep into otel and found out that _the first time_ you set a tracer provider, it switches the delegate on already created tracers, which makes it work.

So here I should either keep this init here or move the tracer creation to each test too, but that's also not what OTel recommends, so I'd rather keep it like this as it's just less changes.

tracing/otel.go
colega: Fair enough, renamed in https://github.com/grafana/dskit/pull/681/commits/4243e6da74529685c531cb27c80ae411b1f77f1a

tracing/otel.go
colega: [Tempo also doesn't have it](https://github.com/grafana/tempo/blob/a4289d96f02f88b42ee168265d17dab3b1efe7e5/cmd/tempo/main.go#L233-L239), and they're the tracing experts for me.

tracing/otel_test.go
colega: Done in https://github.com/grafana/dskit/pull/681/commits/b8e1ea2bcc3e176d9517bcb861fd1bba2b0a5e99

tracing/tracing.go
colega: It would actually look longer:
```go
// ExtractTraceID extracts the trace id, if any from the context.
func ExtractTraceID(ctx context.Context) (string, bool) {
	otelSpan, opentracingSpan, _ := SpanFromContext(ctx)
	if otelSpan != nil {
		if sc := otelSpan.SpanContext(); sc.IsValid() {
			return sc.TraceID().String(), true
		}
	}
	if opentracingSpan != nil {
		if sctx, ok := opentracingSpan.Context().(jaeger.SpanContext); ok {
			return sctx.TraceID().String(), true
		}
	}
	return "", false
}
```
I'd rather leave it as is, and make everything cleaner once we get rid of opentracing.

tracing/span.go
colega: Done in https://github.com/grafana/dskit/pull/681/commits/80751c9791753d3adf368830cfa28c8bf8d72067

tracing/span_test.go
colega: Done in https://github.com/grafana/dskit/pull/681/commits/0b10eb1182aaa1ec3b1edb3bcfcdeb24018d3d41

tracing/span.go
```diff
@@ -0,0 +1,141 @@
...
+			otelOptions = append(otelOptions, opt.otelSpanOptions()...)
+		}
+		ctx, span := otelSpan.TracerProvider().Tracer("dskit/tracing").Start(ctx, operation, otelOptions...)
+		s := &Span{otelSpan: span}
+		for _, opt := range options {
+			opt.apply(s)
+		}
+		return s, ctx
+	}
+	return &Span{}, ctx
```
krajorama, line 55: This is a different behavior compared to before the PR, since `opentracing.StartSpanFromContext` would create a real span if there was no parent: see https://github.com/opentracing/opentracing-go/blob/10b1cf09e00bdc84234b8c7a4b4d4e4afe64de87/gocontext.go#L63

colega: You're right, fixed here and added some tests: https://github.com/grafana/dskit/pull/681/commits/51ca0c0e327048a78dcd78bb56ae54b6cbfe054c

colega:
I'm testing this PR on a local dev environment, to make sure that we didn't break any existing OpenTracing features before merging.

colega:
I checked the traces and they still exist.

middleware/http_tracing.go
```diff
@@ -54,6 +70,42 @@ func (t Tracer) Wrap(next http.Handler) http.Handler {
...
+
+			labeler.Add(attribute.String("http.url", r.URL.Path))
+			labeler.Add(attribute.String("http.method", r.Method))
+
+			// add the content type, useful when query requests are sent as POST
+			if ct := r.Header.Get("Content-Type"); ct != "" {
+				labeler.Add(attribute.String("http.content_type", ct))
+			}
+
+			labeler.Add(attribute.String("headers", fmt.Sprintf("%v", r.Header)))
```
bboreham, line 95: Should we worry about sensitive headers here, e.g. authorization tokens?

See [here](https://github.com/grafana/dskit/blob/77dd0d353b2151b6b56703e891c118ab9536e227/middleware/logging.go#L29) for instance.

colega: Good catch:
https://github.com/grafana/dskit/pull/707

tracing/tracing.go
```diff
@@ -81,17 +52,42 @@ func extractJaegerContext(ctx context.Context) (tid jaeger.TraceID, sid jaeger.S
...
-		return "", false
+		return
 	}
 	sctx, ok := sp.Context().(jaeger.SpanContext)
 	if !ok {
-		return "", false
+		return
 	}
 
-	return sctx.TraceID().String(), sctx.IsSampled()
```
slim-bean, line 96: The previous behavior of ExtractSampledTraceID would return the traceID regardless if it was sampled or not, the new behavior only returns the traceID if it was sampled.

Unfortunately this is a breaking change for Loki where we would use this method to append the log line with additional information if the trace was sampled or not, but we relied on it returning the traceID to use in the log line regardless of being sampled.

PR #685, opened 2025-04-29
feat(grpc): Record max streams per connection

julienduchesne:
This will be used to alert if we're approaching the `-server.grpc-max-concurrent-streams` limit 

It uses a b-tree to keep track of the highest number of streams per connection

### DIFF
--- a/middleware/grpc_stats.go
+++ b/middleware/grpc_stats.go
@@ -163,6 +163,6 @@ func (st *StreamTracker) MaxStreams() int {
 	if st.tree.Len() == 0 {
 		return 0
 	}
-	entry := st.tree.Max().(streamEntry)
+	entry := st.tree.Min().(streamEntry)
 	return entry.streamCount
 }
--- a/middleware/grpc_stats.go
+++ b/middleware/grpc_stats.go
@@ -6,17 +6,22 @@ package middleware
 
 import (
 	"context"
+	"sync"
 
+	"github.com/google/btree"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


middleware/grpc_stats.go
```diff
@@ -65,10 +84,85 @@ func (g *grpcStatsHandler) HandleRPC(ctx context.Context, rpcStats stats.RPCStat
...
+}
+
+func NewStreamTracker() *StreamTracker {
+	return &StreamTracker{
+		tree:    btree.New(2), // degree 2 is fine for small datasets
+		connMap: make(map[string]streamEntry),
+	}
+}
+
+func (st *StreamTracker) OpenStream(connID string) {
```
pracucci, line 124: Could you quickly benchmark `OpenStream()` and `CloseStream()` to see how long it takes? I just want to make sure we're not introducing a scaling limit with this global lock (which effectively serialise the opening / closing of streams). I think it's pretty quick, so nothing to worry about, but measuring it in a benchmark may be worth it.

pracucci (approved):
Nice!

julienduchesne:
Added tests and benchmarks

goos: darwin
goarch: arm64
cpu: Apple M3 Pro
BenchmarkStreamTracker_OpenStream-11     	  652555	      1790 ns/op	     480 B/op	      20 allocs/op
BenchmarkStreamTracker_CloseStream-11    	30788110	        39.98 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/grafana/dskit/middleware	3.584s

PR #687, opened 2025-05-02
Add PartitionRingWatcher delegate

jhalterman:
**What this PR does**:

Adds a PartitionRingWatcher delegate so users of that ring can observe state changes.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_ring_watcher.go
+++ b/ring/partition_ring_watcher.go
@@ -51,8 +51,11 @@ func NewPartitionRingWatcher(name, key string, kv kv.Client, logger log.Logger, 
 	return r
 }
 
-func (w *PartitionRingWatcher) WithDelegate(d PartitionRingWatcherDelegate) *PartitionRingWatcher {
-	w.delegate = d
+// WithDelegate adds the delegate to be called when the partition ring changes.
+//
+// Not concurrency safe.
+func (w *PartitionRingWatcher) WithDelegate(delegate PartitionRingWatcherDelegate) *PartitionRingWatcher {
+	w.delegate = delegate
 	return w
 }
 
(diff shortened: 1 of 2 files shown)


ring/partition_ring_watcher.go
```diff
@@ -81,6 +94,9 @@ func (w *PartitionRingWatcher) updatePartitionRing(desc *PartitionRingDesc) {
 	newRing := NewPartitionRing(*desc)
 
 	w.ringMx.Lock()
+	if w.delegate != nil {
+		w.delegate.OnPartitionRingChanged(&w.ring.desc, desc)
```
56quarters, line 98: Do we need to hold the lock while this runs? Does `newRing := NewPartitionRing(*desc)` make a copy of `desc`?

ring/partition_ring_watcher.go
```diff
@@ -29,6 +30,10 @@ type PartitionRingWatcher struct {
 	numPartitionsGaugeVec *prometheus.GaugeVec
 }
 
+type PartitionRingWatcherDelegate interface {
+	OnPartitionRingChanged(oldRing, newRing *PartitionRingDesc)
```
56quarters, line 34: Can you add some comments here that explain what implementations should or should not do? e.g. don't modify the ring descriptions, must be concurrency safe, etc.

56quarters (approved):
LGTM but I think we might be able to avoid locking.

ring/partition_ring_watcher.go
jhalterman: Moved the callback outside of the lock

PR #689, opened 2025-05-05
Update grpc, maintain health check interface; requries go update

zalegrala:
**What this PR does**:

Here we update the `grpc` version, which requires two changes.  First a Go update to `1.23` and second an interface update to include `List()` method.  The implementation simply wraps the `Check` method.

I invented a key here, which seemed to suit the change, but perhaps "default" could be as well.  I also notice that we have an individual check for all services wrapped in a closure, but perhaps a future change might modify the `List` method to include the status of each of the manager's services.

Dependabot PRs updating dependencies which include newer grpc.  For an example: https://github.com/grafana/tempo/pull/5097

Fixes #<issue number>

**Checklist**
- [x] Tests updated


aknuds1 (commented):
Any idea why tests fail?

zalegrala:
No I'm a little confused why the tests are failing.  They timeout locally, but I don't PR this repo often enough to know what's normal.

zalegrala:
Note that if I run the test individually which are failing in this PR, they pass locally.

aknuds1:
It's not flakiness is it @zalegrala?

zalegrala:
Hmm, possibly yes.  I do get a failure if I run with `-count=100 -failfast`.
```
go test -race -cover ./crypto/tls/test/ -count=100 -failfast -run=".*"
```

aknuds1:
I guess failing tests need some un-flaking then :)

zalegrala:
At least one of the failures looks like `localhost` meaning different TCP networks being resolved between the server and client.  I'm taking a look at what we can do here and trying to not break existing tests in the process.

zalegrala:
I've raised https://github.com/grafana/dskit/pull/691.  Let's see what happens in CI over there.

PR #695, opened 2025-05-21
memberlist: WatchPrefix, handle gracefully nil generated value

NickAnge:
**What this PR does**:

In this PR, we modify memberlist.WatchPrefix to ignore nil values, which can occur when an entry has already been deleted from the Memberlist KV store.

### Test

In the test scenario, we reproduced the issue by increasing the NotifyInterval, so that the notification is sent after the cleanupObsoleteTimeout has been triggered.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -957,8 +957,8 @@ func (m *KV) WatchKey(ctx context.Context, key string, codec codec.Codec, f func
 
 // WatchPrefix watches for any change of values stored under keys with given prefix. When change occurs,
 // function 'f' is called with key and current value.
-// Each change of the key results in one notification. If there are too many pending notifications ('f' is slow),
-// some notifications may be lost.
+// Each change of the key results in one notification. If the key has already been deleted, a nil value is generated that it is ignored.
+// If there are too many pending notifications ('f' is slow), some notifications may be lost.
 //
 // Watching ends when 'f' returns false, context is done, or this client is shut down.
 func (m *KV) WatchPrefix(ctx context.Context, prefix string, codec codec.Codec, f func(string, interface{}) bool) {
@@ -986,7 +986,10 @@ func (m *KV) WatchPrefix(ctx context.Context, prefix string, codec codec.Codec, 
 				level.Warn(m.logger).Log("msg", "failed to decode value while watching for changes", "key", key, "err", err)
 				continue
 			}
-
+			if val == nil {
+				// Skip nil values that can be generated if the notification is received after the entry has been deleted
+				continue
+			}
 			if !f(key, val) {
 				return
 			}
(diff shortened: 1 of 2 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -957,8 +957,8 @@ func (m *KV) WatchKey(ctx context.Context, key string, codec codec.Codec, f func
 
 // WatchPrefix watches for any change of values stored under keys with given prefix. When change occurs,
 // function 'f' is called with key and current value.
-// Each change of the key results in one notification. If there are too many pending notifications ('f' is slow),
-// some notifications may be lost.
+// Each change of the key results in one notification. If the key has already been deleted, a nil value is generated that it is ignored.
```
colega, line 960: > If the key has already been deleted, a nil value is generated that it is ignored.

This is an implementation detail. The comment should say something like "Key deletions are not notified.".

Also: this comment should also be on the interface, because that's what we use.

Also: can the value be actually `nil` without being deleted? I'm not sure what's the contract here.

kv/memberlist/memberlist_client_test.go
```diff
@@ -2365,3 +2365,77 @@ func TestMemberlist_WatchKey_ShouldStartNotifyingChangesBeforeFullJoinIsComplete
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go func() {
+		kv.WatchPrefix(ctx, "test", func(s string, i interface{}) bool {
+			if i == nil {
+				mu.Lock()
+				nilNotifications = append(nilNotifications, s)
+				mu.Unlock()
+			}
```
colega, line 2417: This could be just `t.Error("got nil notification)`

kv/memberlist/memberlist_client.go
NickAnge: > This is an implementation detail. The comment should say something like "Key deletions are not notified.".

Ok. 

> Also: can the value be actually nil without being deleted? I'm not sure what's the contract here.

I checked the code, and couldn't find a way to generate a nil value at the KV Store. I tried when I was thinking about the test use case. The only way that it can happen (at least from what I have found) is when this race condition happens

kv/memberlist/memberlist_client_test.go
NickAnge: Thanks, fixed [c523a86](https://github.com/grafana/dskit/pull/695/commits/c523a86d291b5609eca4fcbd1c0fde6c469f7e77)

kv/memberlist/memberlist_client.go
NickAnge: Addressed [c523a86](https://github.com/grafana/dskit/pull/695/commits/c523a86d291b5609eca4fcbd1c0fde6c469f7e77)

colega (approved):

PR #697, opened 2025-05-21
middleware: log trace_id from cluster validation failures

narqo:
**What this PR does**:

Following #692 this PR updates the internals of `ClusterUnaryServerInterceptor` and `ClusterValidationMiddleware` to log the request's `trace_id` — if the request's trace wasn't sampled.

The idea is that this should make it easier to investigate, where the concerning requests come from.

**Checklist**
- [x] Tests updated

### DIFF
--- a/middleware/grpc_cluster.go
+++ b/middleware/grpc_cluster.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/peer"
 
 	"github.com/grafana/dskit/clusterutil"
 	"github.com/grafana/dskit/grpcutil"
@@ -113,6 +114,9 @@ func checkClusterFromIncomingContext(
 	)
 	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
 		logger = log.With(logger, "trace_id", traceID)
+	}
+	if p, ok := peer.FromContext(ctx); ok {
+		logger = log.With(logger, "peer_address", p.Addr.String())
 	}
 
 	if err == nil {
--- a/middleware/grpc_cluster.go
+++ b/middleware/grpc_cluster.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/dskit/clusterutil"
 	"github.com/grafana/dskit/grpcutil"
 	"github.com/grafana/dskit/tracing"
+	"github.com/grafana/dskit/user"
 
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
@@ -112,11 +113,14 @@ func checkClusterFromIncomingContext(
 		"cluster_validation_label", expectedCluster,
 		"soft_validation", softValidationEnabled,
 	)
-	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
-		logger = log.With(logger, "trace_id", traceID)
+	if tenantID, err := user.ExtractOrgID(ctx); err == nil {
+		logger = log.With(logger, "tenant", tenantID)
 	}
 	if p, ok := peer.FromContext(ctx); ok {
-		logger = log.With(logger, "peer_address", p.Addr.String())
+		logger = log.With(logger, "client_address", p.Addr.String())
+	}
+	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
+		logger = log.With(logger, "trace_id", traceID)
 	}
 
 	if err == nil {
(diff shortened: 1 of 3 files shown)


middleware/grpc_cluster_test.go
```diff
@@ -211,7 +211,7 @@ func TestClusterUnaryServerInterceptor(t *testing.T) {
 				if testCase.expectedLogs == "" {
 					require.Empty(t, buf.Bytes())
 				} else {
-					require.True(t, bytes.Contains(buf.Bytes(), []byte(fmt.Sprintf(testCase.expectedLogs, softValidation))))
+					require.True(t, bytes.Contains(buf.Bytes(), []byte(fmt.Sprintf(testCase.expectedLogs, softValidation))), "got %q", buf.String())
```
aknuds1, line 214: Better like this:
```suggestion
					require.Contains(t, buf.String(), fmt.Sprintf(testCase.expectedLogs, softValidation))
```

duricanikolic (changes requested):
@narqo could you please add the changes that https://github.com/grafana/dskit/pull/692 introduced to `http_cluster.go` in `grpc_cluster.go` too, so that both log the same info?

middleware/http_cluster.go
```diff
@@ -136,13 +142,12 @@ func checkClusterFromRequest(
...
+		logger = log.With(logger, "trace_id", traceID)
+	}
 
-	reqCluster, err := clusterutil.GetClusterFromRequest(r)
 	if err == nil {
-		if reqCluster == expectedCluster {
-			return nil
-		}
-
+		// No error, but the requested cluster didn't match the expectation.
```
duricanikolic, line 150:
```suggestion
               // No error, but request's and server's cluster validation labels  didn't match.
```

middleware/grpc_cluster.go
```diff
@@ -100,18 +102,32 @@ func checkClusterFromIncomingContext(
...
+	)
+	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
+		logger = log.With(logger, "trace_id", traceID)
+	}
+	if p, ok := peer.FromContext(ctx); ok {
+		logger = log.With(logger, "peer_address", p.Addr.String())
+	}
+
+	if err == nil {
+		// No error, but the requested cluster didn't match the expectation.
```
duricanikolic, line 123:
```suggestion
               // No error, but request's and server's cluster validation labels  didn't match.
```

middleware/grpc_cluster.go
```diff
@@ -100,18 +102,32 @@ func checkClusterFromIncomingContext(
...
+		"method", method,
+		"cluster_validation_label", expectedCluster,
+		"soft_validation", softValidationEnabled,
+	)
+	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
+		logger = log.With(logger, "trace_id", traceID)
+	}
+	if p, ok := peer.FromContext(ctx); ok {
+		logger = log.With(logger, "peer_address", p.Addr.String())
+	}
```
duricanikolic, line 120: I would suggest to replace
```
	logger = log.With(
		logger,
		"method", method,
		"cluster_validation_label", expectedCluster,
		"soft_validation", softValidationEnabled,
	)
	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
		logger = log.With(logger, "trace_id", traceID)
	}
	if p, ok := peer.FromContext(ctx); ok {
		logger = log.With(logger, "peer_address", p.Addr.String())
	}
```
with
```
	logger = log.With(
		logger,
		"method", method,
		"cluster_validation_label", expectedCluster,
		"soft_validation", softValidationEnabled,
	)
        if tenantID, err := user.ExtractOrgID(ctx); err == nil {
	  logger = log.With(logger, "tenant", tenantID)
        }
	if p, ok := peer.FromContext(ctx); ok {
		logger = log.With(logger, "client_address", p.Addr.String())
	}
	if traceID, ok := tracing.ExtractSampledTraceID(ctx); ok {
		logger = log.With(logger, "trace_id", traceID)
	}
```
I would do it this way in order to make the log messages between the gRPC interceptor and the HTTP middleware as much consistent as possible. In the HTTP middleware we log `tenant` info immediately after the `soft_validation` info, then we have `client_address` (and not `peer_address`), and at the end we have `trace_id`. WDYT?

duricanikolic (commented):
I have left a couple of suggestions.

aknuds1:
Taking another look.

aknuds1 (approved):
LGTM, thanks for applying my feedback!

middleware/grpc_cluster.go
narqo: That's a good one, indeed. I've updated this chuck to match your suggestion

duricanikolic (approved):
LGTM
Thank you @narqo

PR #703, opened 2025-06-02
fix(grpc): Make recording max streams per connection more performant

julienduchesne:
Improves https://github.com/grafana/dskit/pull/685 

- Expose the option under `-server.grpc-collect-max-streams-by-conn=true|false`
- Instead of computing the maximum every time a stream is closed or opened, defer to when metrics are collected
- Use a read mutex instead of a write mutex on all operations

Details on the performance improvement:

Instead of fully locking on every open/close, it will, do a read-level lock, grab an atomic integer from a map of connections, and unlock immediately after. If it's just a matter of incrementing or decrementing the count, then we don't need to lock again. If we need to create or delete the connection in the map, then we need to lock (this shouldn't happen too often)

New benchmark:
```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
BenchmarkStreamTracker/OpenStreamOnExistingConn-11          	62238513	        18.91 ns/op	       0 B/op	       0 allocs/op
BenchmarkStreamTracker/OpenStreamOnNewConn-11               	62958250	        18.89 ns/op	       0 B/op	       0 allocs/op
BenchmarkStreamTracker/MaxStreams-11                        	   98572	     12306 ns/op	       0 B/op	       0 allocs/op
BenchmarkStreamTracker/CloseStream-11                       	52800052	        25.49 ns/op	       0 B/op	       0 allocs/op
PASS
coverage: 3.7% of statements
ok  	github.com/grafana/dskit/middleware	7.421s
```

It's over 10x faster (compare with https://github.com/grafana/dskit/pull/685#issuecomment-2847088263) on each request and the locking is now a RLock in all functions. 

`MaxStreams` now takes much more time but it's still a fraction of a millisecond, and it will be done also on a read lock once for every collection

### DIFF
--- a/middleware/grpc_stats.go
+++ b/middleware/grpc_stats.go
@@ -143,7 +143,13 @@ func (st *StreamTracker) CloseStream(connID string) {
 	if res := conn.Dec(); res == 0 {
 		// Delete the entry if it's empty.
 		st.mu.Lock()
-		delete(st.connMap, connID)
+
+		// Get the entry again to avoid race condition.
+		conn = st.connMap[connID]
+		if conn == nil || conn.Load() == 0 {
+			delete(st.connMap, connID)
+		}
+
 		st.mu.Unlock()
 	}
 }
--- a/middleware/grpc_stats.go
+++ b/middleware/grpc_stats.go
@@ -8,20 +8,25 @@ import (
 	"context"
 	"sync"
 
-	"github.com/google/btree"
 	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
 	"google.golang.org/grpc/stats"
 )
 
 // NewStatsHandler creates handler that can be added to gRPC server options to track received and sent message sizes.
-func NewStatsHandler(receivedPayloadSize, sentPayloadSize *prometheus.HistogramVec, inflightRequests *prometheus.GaugeVec, grpcConcurrentStreamsByConnMax *prometheus.GaugeVec) stats.Handler {
+func NewStatsHandler(reg prometheus.Registerer, receivedPayloadSize, sentPayloadSize *prometheus.HistogramVec, inflightRequests *prometheus.GaugeVec, grpcConcurrentStreamsByConnMax *prometheus.Desc) stats.Handler {
+	var streamTracker *StreamTracker
+	if grpcConcurrentStreamsByConnMax != nil {
+		streamTracker = NewStreamTracker(grpcConcurrentStreamsByConnMax)
+		reg.MustRegister(streamTracker)
+	}
+
 	return &grpcStatsHandler{
 		receivedPayloadSize: receivedPayloadSize,
 		sentPayloadSize:     sentPayloadSize,
 		inflightRequests:    inflightRequests,
 
-		grpcConcurrentStreamsTracker:        NewStreamTracker(),
-		grpcConcurrentStreamsByConnMaxGauge: grpcConcurrentStreamsByConnMax,
+		grpcConcurrentStreamsTracker: streamTracker,
 	}
 }
 
@@ -30,8 +35,7 @@ type grpcStatsHandler struct {
 	sentPayloadSize     *prometheus.HistogramVec
 	inflightRequests    *prometheus.GaugeVec
 
-	grpcConcurrentStreamsTracker        *StreamTracker
-	grpcConcurrentStreamsByConnMaxGauge *prometheus.GaugeVec
+	grpcConcurrentStreamsTracker *StreamTracker
 }
 
 // Custom type to hide it from other packages.
@@ -61,13 +65,11 @@ func (g *grpcStatsHandler) HandleRPC(ctx context.Context, rpcStats stats.RPCStat
 		g.inflightRequests.WithLabelValues(gRPC, fullMethodName).Inc()
 		if hasConnID {
 			g.grpcConcurrentStreamsTracker.OpenStream(connID)
-			g.grpcConcurrentStreamsByConnMaxGauge.WithLabelValues().Set(float64(g.grpcConcurrentStreamsTracker.MaxStreams()))
 		}
 	case *stats.End:
 		g.inflightRequests.WithLabelValues(gRPC, fullMethodName).Dec()
 		if hasConnID {
 			g.grpcConcurrentStreamsTracker.CloseStream(connID)
-			g.grpcConcurrentStreamsByConnMaxGauge.WithLabelValues().Set(float64(g.grpcConcurrentStreamsTracker.MaxStreams()))
 		}
 	case *stats.InHeader:
 		// Ignore incoming headers.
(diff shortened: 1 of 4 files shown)


middleware/grpc_stats.go
```diff
@@ -85,84 +87,84 @@ func (g *grpcStatsHandler) HandleRPC(ctx context.Context, rpcStats stats.RPCStat
...
-	entry.streamCount--
-	if entry.streamCount == 0 {
+	if res := conn.Dec(); res == 0 {
+		// Delete the entry if it's empty.
+		st.mu.Lock()
 		delete(st.connMap, connID)
-	} else {
-		st.connMap[connID] = entry
-		st.tree.ReplaceOrInsert(entry)
+		st.mu.Unlock()
```
colega: There's a race condition here. `OpenStream` might have a reference to `conn` at this point and it might do `conn.Inc()`. You need to check again `conn.Load() == 0` while `Lock`-ed.

All of this, of course, if we really care about that. If we don't care about this race-condition that makes the result off-by-one, you can just document it.

middleware/grpc_stats.go
```diff
@@ -85,84 +87,84 @@ func (g *grpcStatsHandler) HandleRPC(ctx context.Context, rpcStats stats.RPCStat
...
 
 // MaxStreams returns the number of streams in the connection with the most streams.
 func (st *StreamTracker) MaxStreams() int {
+	var max int32 = 0
 	st.mu.RLock()
-	defer st.mu.RUnlock()
-
-	if st.tree.Len() == 0 {
-		return 0
+	for _, conn := range st.connMap {
```
colega, line 167: I would consider doing `maps.Clone(st.connMap)` while RLocked, and then traverse the loop while not locked. After all you're doing a ton of atomic operations here and preventing the map from being modified. I expect a single non-synchronized map clone to be faster than reading its values in an atomic way one by one.

middleware/grpc_stats.go
julienduchesne: I went at it thinking that it doesn't matter, but I just tested it and the impact is minimal, so I fixed it

middleware/grpc_stats.go
julienduchesne: Didn't seem that much faster 🤔. The only gain would be from the fact that the "greater than" comparisons would be outside of the lock, right? Offset by more memory

narqo: Note that `maps.Clone` is implemented in the runtime ([src](https://github.com/golang/go/blob/49f63047245a50f4ec38044fc6fa2bc75719726a/src/maps/maps.go#L43-L56)). So it _might_ be faster than iterating over the map inside a lock.

julienduchesne: With this diff: 
```diff
diff --git a/middleware/grpc_stats.go b/middleware/grpc_stats.go
index 420fed9..f23ef59 100644
--- a/middleware/grpc_stats.go
+++ b/middleware/grpc_stats.go
@@ -6,6 +6,7 @@ package middleware
 
 import (
 	"context"
+	"maps"
 	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -158,12 +159,13 @@ func (st *StreamTracker) CloseStream(connID string) {
 func (st *StreamTracker) MaxStreams() int {
 	var max int32 = 0
 	st.mu.RLock()
-	for _, conn := range st.connMap {
+	connCopy := maps.Clone(st.connMap)
+	st.mu.RUnlock()
+	for _, conn := range connCopy {
 		if conn.Load() > max {
 			max = conn.Load()
 		}
 	}
-	st.mu.RUnlock()
 	return int(max)
 }
 
diff --git a/middleware/grpc_stats_test.go b/middleware/grpc_stats_test.go
index cf50194..3953bf8 100644
--- a/middleware/grpc_stats_test.go
+++ b/middleware/grpc_stats_test.go
@@ -651,13 +651,15 @@ func BenchmarkStreamTrackerConnectionsSimulation(b *testing.B) {
 			}(i)
 		}
 
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			for i := 0; i < 1000; i++ {
-				tracker.MaxStreams()
-			}
-		}()
+		for i := 0; i < 10; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				for i := 0; i < 1000; i++ {
+					tracker.MaxStreams()
+				}
+			}()
+		}
 
 		wg.Wait()
 ```
 
 I get a pretty terrible result:
 ```
 julienduchesne@triceratops dskit % benchstat test-old.txt test-new.txt
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                                      │ test-old.txt │            test-new.txt            │
                                      │    sec/op    │   sec/op     vs base               │
StreamTrackerConnectionsSimulation-11    16.96m ± 9%   18.06m ± 2%  +6.53% (p=0.005 n=10)

                                      │ test-old.txt │             test-new.txt              │
                                      │     B/op     │     B/op      vs base                 │
StreamTrackerConnectionsSimulation-11   1.457Mi ± 1%   6.320Mi ± 1%  +333.79% (p=0.000 n=10)

                                      │ test-old.txt │             test-new.txt             │
                                      │  allocs/op   │  allocs/op   vs base                 │
StreamTrackerConnectionsSimulation-11    40.46k ± 0%   81.03k ± 1%  +100.28% (p=0.000 n=10)
```

And MaxStream is running _way_ more than it would run under normal circumstances

server/server.go
```diff
@@ -82,15 +82,16 @@ type Config struct {
...
-	ProxyProtocolEnabled bool   `yaml:"proxy_protocol_enabled"`
+	HTTPListenNetwork           string `yaml:"http_listen_network"`
+	HTTPListenAddress           string `yaml:"http_listen_address"`
+	HTTPListenPort              int    `yaml:"http_listen_port"`
+	HTTPConnLimit               int    `yaml:"http_listen_conn_limit"`
+	GRPCListenNetwork           string `yaml:"grpc_listen_network"`
+	GRPCListenAddress           string `yaml:"grpc_listen_address"`
+	GRPCListenPort              int    `yaml:"grpc_listen_port"`
+	GRPCConnLimit               int    `yaml:"grpc_listen_conn_limit"`
+	GRPCCollectMaxStreamsByConn bool   `yaml:"grpc_collect_max_streams_by_conn"`
```
narqo, line 93: I've looked through the changes multiple times, but I can't see where this `GRPCCollectMaxStreamsByConn` flag is being checked. The code below only check `cfg.GRPCServerStatsTrackingEnabled`

middleware/grpc_stats.go
colega: > Didn't seem that much faster 🤔. The only gain would be from the fact that the "greater than" comparisons would be outside of the lock, right? Offset by more memory

My concern isn't the "greater than" operation, but the `.Load()` operation which is atomic and has to be synchronized between different cores.

julienduchesne: If I do `maps.Clone`, won't I get pointers to the same atomic numbers and do those atomic operations anyways

colega: Yes but you won't be holding the mutex while atomic-reading each one of them.

Anyway, I was just brainstorming, this is not a blocker comment.

server/server.go
julienduchesne: Fixed, got confused in my refactors 🤦

narqo (approved):
🔥 Works for me

PR #362, opened 2023-08-14
Upgrade to latest github.com/grafana/memberlist

aknuds1:
**What this PR does**:
Upgrade to [github.com/grafana/memberlist@ f7d51a6](https://github.com/grafana/memberlist/commit/f7d51a6f67004af3d8f772baf42e33ff0bb144cf), which migrates from github.com/armon/go-metrics to github.com/hashicorp/go-metrics, which is the [new](https://github.com/hashicorp/go-metrics/pull/143) URL of the module.

**Which issue(s) this PR fixes**:

**Checklist**
- [ ] Tests updated
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #706, opened 2025-06-04
Use hashicorp/go-metrics instead of legacy armon package

aknuds1:
**What this PR does**:

Upgrade to [hashicorp/go-metrics](https://github.com/hashicorp/go-metrics) instead of legacy armon package (which is replaced by the former). hashicorp/go-metrics is already used by grafana/memberlist, but I forgot to switch dskit's usage of the module when I upgraded grafana/memberlist in #362.

**Checklist**
- [ ] Tests updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ toolchain go1.24.2
 require (
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137
 	github.com/alicebob/miniredis v2.5.0+incompatible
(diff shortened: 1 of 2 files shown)


NickAnge (approved):
LGTM apart the suggestion from copilot

aknuds1:
@NickAnge Copilot is wrong. I've run `go mod tidy` and armon/go-metrics is still an indirect dependency because hashicorp/go-metrics has a backwards compatibility package importing it.

PR #707, opened 2025-06-04
Tracing: stop adding HTTP headers as tracing attributes

colega:
When I added the OTel tracing in https://github.com/grafana/dskit/pull/681 I went too far and added headers as tracing attributes.

As @bboreham pointed out, this could log potentially sensitive headers like auth tokens.

This is a quick fix to remove that from traces, we can add an allowlist or a blocklist in the future as a feature.


colega:
Following up with a _decent fix_: https://github.com/grafana/dskit/pull/709

PR #708, opened 2025-06-05
tracing.NewOTelOrJaegerFromEnv should not fail if no env was provided

colega:
Follow up on https://github.com/grafana/dskit/pull/704

This function should not fail if no tracing env was provided, that's what the old tracing.NewFromEnv() did.

Previous behaviour was to just setup a Jaeger exporter pointing to localhost, we'll do the default OTEL exporter behaviour now.

### DIFF
--- a/tracing/internal/oteltest/otelorjaegerfromenv/config_test.go
+++ b/tracing/internal/oteltest/otelorjaegerfromenv/config_test.go
@@ -1,0 +1,50 @@
+package otelorjaegerfromenv
+
+import (
+	"os"
+	"testing"
+
+	"github.com/go-kit/log"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/dskit/tracing"
+)
+
+func TestTracingNewOTelOrJaegerFromEnvDoesNotFailWithNoEnv(t *testing.T) {
+	// Make sure no tracing env is set for the test.
+	defer unsetAndRestoreDeferred(
+		"JAEGER_AGENT_HOST",
+		"JAEGER_ENDPOINT",
+		"JAEGER_SAMPLER_MANAGER_HOST_PORT",
+		"OTEL_TRACES_EXPORTER",
+	)()
+
+	closer, err := tracing.NewOTelOrJaegerFromEnv("test-service", log.NewNopLogger())
+	require.NoError(t, err)
+	t.Cleanup(func() {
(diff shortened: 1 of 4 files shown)


tracing/tracing.go
```diff
@@ -14,8 +14,8 @@ import (
 )
 
 var (
-	// ErrBlankTraceConfiguration is an error to notify client to provide valid trace report agent or config server
-	ErrBlankTraceConfiguration = errors.New("no OTel or Jaeger tracing environment variables configured")
```
colega, line 18: I reverted the message of this error, and renamed it to make it explicit that it only applies to jaeger config.

tracing/internal/oteltest/otelorjaegerfromenv/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	return func() {
+		for _, v := range vars {
+			if originalValue, exists := originalValues[v]; exists {
+				if originalValue == "" {
+					os.Unsetenv(v)
+				} else {
+					os.Setenv(v, originalValue)
+				}
```
aknuds1: [Nit] This can be simplified:
```suggestion
			if originalValue := originalValues[v]; originalValue != "" {
				os.Setenv(v, originalValue)
```

tracing/otel_jaeger.go
```diff
@@ -45,23 +45,18 @@ const (
 // - JAEGER_AGENT_HOST
 // - JAEGER_ENDPOINT
 // - JAEGER_SAMPLER_MANAGER_HOST_PORT
-// Otherwise, it will initialize tracing with the OTel auto exporter, as per OTel docs, if any of the following environment variables are set:
-// - OTEL_EXPORTER_OTLP_ENDPOINT
-// - OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
-// - OTEL_TRACES_SAMPLER
-// - OTEL_TRACES_EXPORTER
+//
+// Otherwise, it will initialize tracing with the OTel auto exporter using the environment variables defined in the OTel docs
```
aknuds1: A period is missing.
```suggestion
// Otherwise, it will initialize tracing with the OTel auto exporter using the environment variables defined in the OTel docs.
```

aknuds1 (approved):
LGTM modulo nits.

PR #709, opened 2025-06-05
Fix OTel tracing of HTTP headers

colega:
OTel tracing of HTTP headers was broken, as we were using the LabelerFromContext which actually labels metrics, not adds tracing attributes.

I fixed headers tracing, introducing an exclusion list similar to the one we use in logging.

I also added tests for OTel tracing: I had to move these under server/internal/ because since both Jaeger and OTel tracing set up a global state, they need to run in different test packages.

I took a slightly different approach for excluded headers: instead of just silently skipping them, I've added a trace attribute saying they were present in the request: that will likely make debugging easier.

Ref: https://github.com/grafana/dskit/pull/707

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/middleware/http_tracing.go
+++ b/middleware/http_tracing.go
@@ -8,19 +8,16 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"slices"
-
-	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
-	"go.opentelemetry.io/otel"
-	"go.opentelemetry.io/otel/attribute"
-	semconv "go.opentelemetry.io/otel/semconv/v1.20.0" // otelhttp uses semconv v1.20.0 so we stick to the same version in order to produce consistent attributes on HTTP and HTTPGRPC spans.
-	"go.opentelemetry.io/otel/trace"
-	"golang.org/x/exp/maps"
 
 	"github.com/gorilla/mux"
 	"github.com/opentracing-contrib/go-stdlib/nethttp"
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.20.0" // otelhttp uses semconv v1.20.0 so we stick to the same version in order to produce consistent attributes on HTTP and HTTPGRPC spans.
+	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc"
 
 	"github.com/grafana/dskit/httpgrpc"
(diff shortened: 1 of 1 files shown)


middleware/http_tracing.go
```diff
@@ -30,8 +33,31 @@ var tracer = otel.Tracer("dskit/middleware")
...
 type Tracer struct {
 	SourceIPs *SourceIPExtractor
+
+	traceHeaders         bool
+	httpHeadersToExclude map[string]bool
+}
+
+// NewTracer creates a new tracer optionally configuring the tracing of HTTP headers.
+// The configuration for HTTP headers tracing only applies to OpenTelemetry spans.
+func NewTracer(sourceIPs *SourceIPExtractor, traceHeaders bool, excludeHeadersList []string) Tracer {
```
tcard: **nitpick:** `excludedHeaders` should be enough, the type already tells you it's a list(-like).

middleware/http_tracing.go
```diff
@@ -71,38 +99,57 @@ func (t Tracer) wrapWithOpenTracing(next http.Handler) http.Handler {
...
+		if route := ExtractRouteName(r.Context()); route != "" {
+			attributes = append(attributes, semconv.HTTPRoute(route))
+		}
 
-			// add the content type, useful when query requests are sent as POST
-			if ct := r.Header.Get("Content-Type"); ct != "" {
-				labeler.Add(attribute.String("http.content_type", ct))
+		// Add an attribute with the client's sourceIPs to the span, if a SourceIPExtractor is given.
+		if t.SourceIPs != nil {
+			attributes = append(attributes, attribute.String("sourceIPs", t.SourceIPs.Get(r)))
```
tcard: **nitpick:** This attribute is camel case while `headers_total` is snake case. (Maybe there is some convention I'm missing.)

middleware/http_tracing.go
```diff
@@ -71,38 +99,57 @@ func (t Tracer) wrapWithOpenTracing(next http.Handler) http.Handler {
...
+		}
+
+		if t.traceHeaders {
+			headers := maps.Keys(r.Header)
```
tcard: **suggestion:** `maps.Keys` has been in the stdlib for a few versions now, why keep using `x/exp/maps`?

middleware/http_tracing.go
```diff
@@ -30,8 +33,31 @@ var tracer = otel.Tracer("dskit/middleware")
 var _ = nethttp.MWURLTagFunc
 
 // Tracer is a middleware which traces incoming requests.
+// An empty Tracer is valid, but consider using NewTracer to access all its features.
 type Tracer struct {
 	SourceIPs *SourceIPExtractor
+
+	traceHeaders         bool
+	httpHeadersToExclude map[string]bool
```
tcard, line 38: **suggestion:** Using `struct{}` instead of `bool` for sets might be more idiomatic.

middleware/http_tracing.go
```diff
@@ -71,38 +99,57 @@ func (t Tracer) wrapWithOpenTracing(next http.Handler) http.Handler {
...
-			if ct := r.Header.Get("Content-Type"); ct != "" {
-				labeler.Add(attribute.String("http.content_type", ct))
+		// Add an attribute with the client's sourceIPs to the span, if a SourceIPExtractor is given.
+		if t.SourceIPs != nil {
+			attributes = append(attributes, attribute.String("sourceIPs", t.SourceIPs.Get(r)))
+		}
+
+		if t.traceHeaders {
+			headers := maps.Keys(r.Header)
+			const maxHeadersToAddAsSpanAttributes = 100
```
tcard, line 119: **suggestion:** Do `headers := make([]string, 0, maxHeadersToAddAsSpanAttributes)` and then copy into it with a for loop to avoid an allocation.

middleware/http_tracing.go
```diff
@@ -71,38 +99,57 @@ func (t Tracer) wrapWithOpenTracing(next http.Handler) http.Handler {
...
-			// SourceIPExtractor is given.
-			if t.SourceIPs != nil {
-				labeler.Add(attribute.String("sourceIPs", t.SourceIPs.Get(r)))
+			for _, header := range headers {
+				if _, ok := t.httpHeadersToExclude[header]; ok {
+					// Do not add excluded headers to the span attributes, but note that they were sent.
+					attributes = append(attributes, attribute.String("http.header."+header+".present", "true"))
+					continue
+				}
+				values := r.Header[header]
```
tcard: **comment:** It seems to me this is penalizing the common case (`len(headers) < maxHeadersToAddAsSpanAttributes`) for the edge case in which the request has too many headers. If the `for` iterated over `r.Header` instead, this lookup wouldn't be necessary.

Maybe it's worth specializing and only using a slice for the edge case? (Or just give up sorting.)

middleware/http_tracing.go
colega: I've spent too much time fighting myself about this. There's a precedent in OpenTracing for this being camelcase, although I'm quite sure nobody has metrics on this, so yeah, let's just switch it.

middleware/http_tracing.go
colega: Huh, I didn't even think about this much. I see that `maps.Keys` is an iterator, so it's quite useless for my purpose, and I don't see an alternative.

middleware/http_tracing.go
colega: I agree but there's `middleware.Log` that uses it this way, and I chose consistency over the standard.

middleware/http_tracing.go
colega: I wasn't focusing on performance, but since you asked.... 

https://github.com/grafana/dskit/pull/709/commits/658e1b361d688bb59ac058fd2fc0cd698290cb56

I sacrificed consistency because anyway, if someone's malicious and wants to spam us with requests with more than 100 headers, they can also just randomize them in every request so consistency won't help.

middleware/http_tracing.go
colega: I refactored this in https://github.com/grafana/dskit/pull/709/commits/658e1b361d688bb59ac058fd2fc0cd698290cb56 based on your comment below.

tcard (approved):

PR #714, opened 2025-06-13
fix(tracing): Make Jaeger-env OTel sampler parentbased

colega:
When configuring OTel tracing library with Jaeger environment variables we were not making the sampler parent-based, which means that each one of our components took it's own decision about sampling, and almost nothing is sampled.

This makes it parentbased by default, except if 'JAEGER_SAMPLER_NOT_PARENT_BASED=true' env is provided (in case someone needs this).

### DIFF
--- a/tracing/otel_jaeger.go
+++ b/tracing/otel_jaeger.go
@@ -208,10 +208,11 @@ func parseOTelJaegerConfig() (otelJaegerConfig, error) {
 	}
 
 	if e := os.Getenv(envJaegerSamplerNotParentBased); e != "" {
-		if e != "true" && e != "false" {
-			return cfg, errors.Errorf("invalid value for %s: %s, expected 'true' or 'false'", envJaegerSamplerNotParentBased, e)
+		if value, err := strconv.ParseBool(e); err == nil {
+			cfg.samplerNotParentBased = value
+		} else {
+			return cfg, errors.Wrapf(err, "cannot parse env var %s=%s", envJaegerSamplerNotParentBased, e)
 		}
-		cfg.samplerNotParentBased = e == "true"
 	}
 
 	return cfg, nil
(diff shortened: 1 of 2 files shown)


colega:
For some unknown reason I had to `go mod tidy` here.

tracing/otel_jaeger.go
```diff
@@ -31,6 +31,7 @@ const (
 	envJaegerTags                      = "JAEGER_TAGS"
 	envJaegerSamplerManagerHostPort    = "JAEGER_SAMPLER_MANAGER_HOST_PORT"
 	envJaegerSamplerParam              = "JAEGER_SAMPLER_PARAM"
+	envJaegerSamplerNotParentBased     = "JAEGER_SAMPLER_NOT_PARENT_BASED"
```
dimitarvdimitrov, line 34: don't we document these anywhere? i have no idea what this means now

tracing/otel_jaeger.go
```diff
@@ -204,6 +206,14 @@ func parseOTelJaegerConfig() (otelJaegerConfig, error) {
 			return cfg, errors.Wrapf(err, "cannot parse env var %s=%s", envJaegerReporterMaxQueueSize, e)
 		}
 	}
+
+	if e := os.Getenv(envJaegerSamplerNotParentBased); e != "" {
+		if e != "true" && e != "false" {
```
dimitarvdimitrov: nit to use strconv.ParseBool()

tracing/otel_jaeger.go
colega: We _could_ document them in Mimir, but all the Jaeger config is deprecated now, so I'm basically adding this in case something goes wrong.

tracing/otel_jaeger.go
colega: Fair enough: https://github.com/grafana/dskit/pull/714/commits/85e0795fb24957e72426b12e19b590644e999fd4

tracing/otel_jaeger.go
dimitarvdimitrov: and you have plans to remove them, right.. right?

dimitarvdimitrov (approved):

dimitarvdimitrov (approved):

PR #717, opened 2025-06-18
fix(tracing): Unset OTEL_TRACES_SAMPLER when it's jaeger_remote

colega:
After configuring our custom jaeger_remote or parentbased_jaeger_remote we should unset the OTEL_TRACES_SAMPLER, otherwise the SDK complains in the logs about 'unknown sampler: jaeger_remote', which is confusing for the users.

### DIFF
--- a/tracing/otel.go
+++ b/tracing/otel.go
@@ -56,7 +56,7 @@ func NewOTelFromEnv(serviceName string, logger log.Logger, opts ...OTelOption) (
 		tracesdk.WithBatcher(exp),
 		tracesdk.WithResource(resource),
 	}
-	if jaegerRemoteSampler, ok, err := MaybeJaegerRemoteSamplerFromEnv(serviceName); err != nil {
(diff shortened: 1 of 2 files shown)


tracing/otel.go
```diff
@@ -185,6 +185,10 @@ func MaybeJaegerRemoteSamplerFromEnv(serviceName string) (tracesdk.Sampler, bool
 		return nil, false, nil
 	}
 
+	// Unset the OTEL_TRACES_SAMPLER environment variable to the SDK's samplerFromEnv()
+	// function complaining about unknown sampler and logging confusing messages.
+	_ = os.Unsetenv("OTEL_TRACES_SAMPLER")
```
alexweav, line 192: Nit: Since this unsets the env var that it reads just prior, is it worth documenting that `MaybeJaegerRemoteSamplerFromEnv` might have different effects if it's called multiple times?

alexweav (approved):
LGTM

tracing/otel.go
colega: I documented and actually unexported the function: this is an internal detail, not intended to be called from anywhere else.

PR #718, opened 2025-06-25
feat: Rename experimental miekgdns2 resolver to miekgdns

56quarters:
**What this PR does**:

Remove the old miekgdns resolver and replace it with the experimental miekgdns2 resolver. The new resolver includes retry logic, connection pooling, only uses TCP, and does not support search domains.

Related #660

**Checklist**
- [X] Tests updated


dimitarvdimitrov:
have you spoken to loki and tempo about this? They're the biggest other users of the package that i know of. Will they be ok with the change?

56quarters:
> have you spoken to loki and tempo about this? They're the biggest other users of the package that i know of. Will they be ok with the change?

Neither Loki or Tempo use the cache code from dskit. The only consumers are Mimir and some internal adaptive telemetry and faro services.

PR #722, opened 2025-07-03
feat: concurrency.ReusableGoroutinesPool: create workers on demand

colega:
Instead of creating workers when instantiating the pool, create them when no workers are free to serve the function, up to the given size.

### DIFF
--- a/concurrency/worker.go
+++ b/concurrency/worker.go
@@ -1,31 +1,25 @@
 package concurrency
 
+import "go.uber.org/atomic"
+
 // NewReusableGoroutinesPool creates a new worker pool with the given size.
+// Workers are created on demand as they're needed up to the specified size.
+//
 // These workers will run the workloads passed through Go() calls.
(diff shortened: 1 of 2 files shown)


concurrency/worker.go
```diff
@@ -34,10 +28,30 @@ func (p *ReusableGoroutinesPool) Go(f func()) {
 	select {
 	case p.jobs <- f:
 	default:
+		if p.pending.Dec() >= 0 {
```
bboreham: This keeps on getting more negative?
Should we worry about overflow?

colega: Yes it keeps getting more negative. I was going to say that we shouldn't worry, but obviously those would be the _famous last words_ so let's reset it back to 0 if we're past halfway: https://github.com/grafana/dskit/pull/722/commits/03cf6fcb380799c414ecd7252827ac83d8317378

PR #723, opened 2025-07-14
fix(tracing): do not setup OTel tracing if environment variables were not provided

colega:
**What this PR does**:

OTel will setup the entire tracing stack by default, pointing at localhost:4318, and then it will try to export spans there, and it will fail, and it will log an error for each one of those failures. But maybe we just didn't want to setup tracing?

This changes the setup functions to skip setup if no enviroment variable was provided to export them, this is something that Tempo does in their main()

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/tracing/internal/notracingtest/no_tracing_test.go
+++ b/tracing/internal/notracingtest/no_tracing_test.go
@@ -42,11 +42,18 @@ func TestNoTracingConfigured(t *testing.T) {
 
 func ensureTracingDoesNotLogErrors(t *testing.T, logger mockLogger) {
 	tracer := otel.Tracer("test")
+	done := make(chan struct{})
+	defer close(done)
 	go func() {
-		for t.Context().Err() == nil {
-			// Generate A LOT of spans, so exporter will be forced to export them if configured.
-			_, sp := tracer.Start(context.Background(), "test-operation")
-			sp.End()
+		for {
+			select {
+			case <-done:
+				return
+			default:
+				// Generate A LOT of spans, so exporter will be forced to export them if configured.
+				_, sp := tracer.Start(context.Background(), "test-operation")
+				sp.End()
+			}
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


tracing/internal/notracingtest/no_tracing_test.go
```diff
@@ -0,0 +1,105 @@
...
+			default:
+				// Generate A LOT of spans, so exporter will be forced to export them if configured.
+				_, sp := tracer.Start(context.Background(), "test-operation")
+				sp.End()
+			}
+		}
+	}()
+
+	// There's no way to figure out whether OTel tracer provider is configured or not, so we'll just observe the logs.
+	timeout := time.After(time.Second)
```
tcard, line 61: Isn't a second a bit too much? If you undo your changes, how long does it take for this test to fail?

tracing/internal/notracingtest/no_tracing_test.go
```diff
@@ -0,0 +1,105 @@
...
+		originalValues[k] = os.Getenv(k)
+		os.Unsetenv(k)
+	}
+
+	return func() {
+		for _, v := range vars {
+			if originalValue := originalValues[v]; originalValue != "" {
+				os.Setenv(v, originalValue)
+			} else {
+				os.Unsetenv(v)
```
tcard, line 94: This whole `if` could just be the `os.Setenv` unconditionally.

tcard (approved):

tracing/internal/notracingtest/no_tracing_test.go
colega: GitHub CI is slow as hell, so one second should be good here. Since packages are tested in parallel in go this shouldn't increase the overall tests speed by a lot.

tracing/internal/notracingtest/no_tracing_test.go
colega: Probably true, or we could improve this to actually check whether the var was set initially. I keep copying this function from package to package without overthinking it so I'll just leave it like this for consistency.

PR #724, opened 2025-07-14
feat(ring): add ReadOnly state handling to BasicLifecycler

colega:
**What this PR does**:

This adds the ReadOnly property handling to BasicLifecycler.

Required for usage-tracker implementation in Mimir.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/basic_lifecycler.go
+++ b/ring/basic_lifecycler.go
@@ -337,10 +337,15 @@ func (l *BasicLifecycler) registerInstance(ctx context.Context) error {
 			registeredAt = time.Now()
 		}
 
+		readOnly, readOnlyUpdatedTimestamp := instanceDesc.GetReadOnlyState()
+		if readOnlyUpdatedTimestamp.IsZero() {
+			// Store the zero timestamp as 0 in the ring, not as -62135596800.
+			readOnlyUpdatedTimestamp = time.Unix(0, 0)
+		}
 		// Always overwrite the instance in the ring (even if already exists) because some properties
 		// may have changed (stated, tokens, zone, address) and even if they didn't the heartbeat at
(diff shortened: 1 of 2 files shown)


ring/basic_lifecycler.go
```diff
@@ -514,6 +534,26 @@ func (l *BasicLifecycler) changeState(ctx context.Context, state InstanceState)
...
 
+func (l *BasicLifecycler) changeReadOnlyState(ctx context.Context, readOnly bool) error {
+	err := l.updateInstance(ctx, func(_ *Desc, i *InstanceDesc) bool {
+		// No-op if the state hasn't changed.
+		if i.ReadOnly == readOnly {
+			return false
+		}
+
+		i.ReadOnly = readOnly
+		i.ReadOnlyUpdatedTimestamp = time.Now().Unix()
```
pracucci, line 551: Now that the BasicLifecycler can set the read-only state, the `ReadOnly` and `ReadOnlyUpdatedTimestamp` must be preserved in the `AddIngester` call done by `BasicLifecycler.updateInstance()` in case the instance is lost in the ring. Same in `BasicLifecycler.registerInstance()`. This will keep logic consistency with the legacy `Lifecycler`.

colega: Thanks, fixed this in https://github.com/grafana/dskit/pull/724/commits/690bafaf14c67a9ca83cce3a08ae17c963a44b8e

colega: Also updated registerInstance, as you pointed offline: https://github.com/grafana/dskit/pull/724/commits/30e63ad7c6e182346b77745aa1985201f7efbec1

PR #725, opened 2025-07-14
feat(ring): support for owning multiple partitions

colega:
**What this PR does**:

This adds a new `MultiPartitionInstanceRing` as an alternative to `PartitionInstanceRing` in which the instances can own more than one partition. It uses the same instance PartitionRing desc, so we leverage that by setting a different instance ID as owner of each partition.

The convention is enclosed in this implementation and consists of adding a suffix `/<partition>` to the instance ID when it owns `partition`. We initially added this as an option to `PartitionInstanceRing`, however most of the code actually differs, and the fact that you could instantiate a `PartitionInstanceRing` and use it in a multi-partition context introduced subtle bugs in the code. By adding a specific type to this we enforce that each implementation will consistently use one or another.

We do add this as an option to `PartitionInstanceLifecycler`, because there most of the code is shared. 

`PartitionRing` exposes a new method `MultiPartitionOwnerIDs` that will list the real (non-suffixed) instance IDs of partition owners ready to be used by any code.

Finally, we also add a method to `PartitionRingEditor` to remove an instance ID as an owner of a partition, required to perform the periodic cleanup of the stale instances in the ring by users of this library.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/multi_partition_instance_ring.go
+++ b/ring/multi_partition_instance_ring.go
@@ -73,7 +73,7 @@ func (r *MultiPartitionInstanceRing) GetReplicationSetForPartitionAndOperation(p
 	zonesBuffer = uniqueZonesFromInstances(instances, zonesBuffer[:0])
 	uniqueZones := len(zonesBuffer)
 
-	instances = highestPreferrablyNonReadOnlyFromEachZone(instances, ownerIDs, zonesBuffer)
+	instances = highestPreferablyNonReadOnlyFromEachZone(instances, ownerIDs, zonesBuffer)
 
 	return ReplicationSet{
 		Instances: instances,
@@ -91,7 +91,7 @@ func (r *MultiPartitionInstanceRing) GetReplicationSetForPartitionAndOperation(p
 
 // this method expects instanceIDs to be in the same order as instances.
 // instanceIDs should hold the parsed multi-partition owner IDs.
-func highestPreferrablyNonReadOnlyFromEachZone(instances []InstanceDesc, instanceIDs []string, instanceZones []string) []InstanceDesc {
+func highestPreferablyNonReadOnlyFromEachZone(instances []InstanceDesc, instanceIDs []string, instanceZones []string) []InstanceDesc {
 	var stackAllInstances [16]InstanceDesc
 	allInstances := append(stackAllInstances[:0], instances...)
 	instances = instances[:0] // Reset, this is what we're going to return.
--- a/ring/multi_partition_instance_ring.go
+++ b/ring/multi_partition_instance_ring.go
@@ -1,0 +1,158 @@
+package ring
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+	"time"
+)
+
+// MultiPartitionInstanceRing holds a partitions ring and a instances ring,
+// and provide functions to look up the intersection of the two (e.g. healthy instances by partition).
+// The difference between this and PartitionInstanceRing is that this ring supports multi-partition owners, i.e. an instance can own multiple partitions.
+type MultiPartitionInstanceRing struct {
+	partitionsRingReader PartitionRingReader
+	instancesRing        InstanceRingReader
+	heartbeatTimeout     time.Duration
+}
+
+func NewMultiPartitionInstanceRing(partitionsRingWatcher PartitionRingReader, instancesRing InstanceRingReader, heartbeatTimeout time.Duration) *MultiPartitionInstanceRing {
+	return &MultiPartitionInstanceRing{
+		partitionsRingReader: partitionsRingWatcher,
+		instancesRing:        instancesRing,
+		heartbeatTimeout:     heartbeatTimeout,
+	}
+}
+
+func (r *MultiPartitionInstanceRing) PartitionRing() *PartitionRing {
+	return r.partitionsRingReader.PartitionRing()
+}
+
+// GetReplicationSetForPartitionAndOperation returns a ReplicationSet for the input partition. If the partition doesn't
+// exist or there are no healthy owners for the partition, an error is returned.
+// It will pick the highest instance from each zone when building the replication set,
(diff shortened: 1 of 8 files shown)


ring/partition_instance_lifecycler_test.go
```diff
@@ -269,6 +272,86 @@ func TestPartitionInstanceLifecycler(t *testing.T) {
...
+
+		// Shutdown lifecyclerInstance1Partition1, this should remove the owner for partition 1.
+		require.NoError(t, services.StopAndAwaitTerminated(ctx, lifecyclerInstance1Partition1))
+
+		assertMapElementsMatch(t, map[int32][]string{
+			1: {multiPartitionOwnerInstanceID(instanceID2, 1)},
+			2: {multiPartitionOwnerInstanceID(instanceID1, 2)},
+		}, getPartitionRingFromStore(t, store, ringKey).ownersByPartition())
+
+		// Shutdown lifecyclerInstance1Partition1, this should *NOT* the owner for partition 2.
```
pracucci: [nit]
```suggestion
		// Shutdown lifecyclerInstance1Partition2, this should *NOT* remove the owner for partition 2, because "remove on shutdown" was disabled.
```

ring/multi_partition_instance_ring.go
```diff
@@ -0,0 +1,158 @@
...
+		partitionsRingReader: partitionsRingWatcher,
+		instancesRing:        instancesRing,
+		heartbeatTimeout:     heartbeatTimeout,
+	}
+}
+
+func (r *MultiPartitionInstanceRing) PartitionRing() *PartitionRing {
+	return r.partitionsRingReader.PartitionRing()
+}
+
```
pracucci, line 30: [nit] I suggest to add `InstanceRing()` like `PartitionInstanceRing`. It's legit to get the instances ring from `MultiPartitionInstanceRing`.

```suggestion

func (r * MultiPartitionInstanceRing) InstanceRing() InstanceRingReader {
	return r.instancesRing
}

```

ring/multi_partition_instance_ring.go
```diff
@@ -0,0 +1,158 @@
...
+			continue
+		}
+
+		if !instance.IsHealthy(op, r.heartbeatTimeout, now) {
+			continue
+		}
+
+		instances = append(instances, instance)
+		// Store this in the same position as the instance, so we can use it to compare later.
+		// We only need this when r.pickHighestZoneOwner=true, but it's probably cheaper to do it than to check it.
```
pracucci: [nit] There's no `pickHighestZoneOwner`.
```suggestion
```

ring/multi_partition_instance_ring.go
```diff
@@ -0,0 +1,158 @@
...
+		// We need response from at least 1 owner. The assumption is that we have 1 owner per zone
+		// but it's not guaranteed (depends on how the application was deployed). The safest thing
+		// we can do here is to just request a successful response from at least 1 zone.
+		MaxUnavailableZones: uniqueZones - 1,
+	}, nil
+}
+
+// this method expects instanceIDs to be in the same order as instances.
+// instanceIDs should hold the parsed multi-partition owner IDs.
+func highestPreferrablyNonReadOnlyFromEachZone(instances []InstanceDesc, instanceIDs []string, instanceZones []string) []InstanceDesc {
```
pracucci: [nit] Typo
```suggestion
func highestPreferablyNonReadOnlyFromEachZone(instances []InstanceDesc, instanceIDs []string, instanceZones []string) []InstanceDesc {
```

ring/multi_partition_instance_ring.go
```diff
@@ -0,0 +1,158 @@
...
+		ZoneAwarenessEnabled: true,
+
+		// We need response from at least 1 owner. The assumption is that we have 1 owner per zone
+		// but it's not guaranteed (depends on how the application was deployed). The safest thing
+		// we can do here is to just request a successful response from at least 1 zone.
+		MaxUnavailableZones: uniqueZones - 1,
+	}, nil
+}
+
+// this method expects instanceIDs to be in the same order as instances.
```
pracucci, line 92: [nit] Please also comment that `instances` is updated in-place and then returned.

ring/partition_ring.go
```diff
@@ -354,6 +355,31 @@ func (r *PartitionRing) PartitionOwnerIDsCopy(partitionID int32) []string {
...
+	}
+
+	if cap(buf) < len(ids) {
+		buf = make([]string, len(ids))
+	} else {
+		buf = buf[:len(ids)]
+	}
+
+	for i, ownerID := range ids {
+		if p := strings.IndexByte(ownerID, '/'); p != -1 {
```
pracucci: This search for the 1st `/` occurrence, but so far I don't a reason why `/` couldn't also be part of the instance ID. I think using `strings.LastIndexByte()` would more correct.

ring/multi_partition_instance_ring.go
colega: I explicitly left it unimplemented because we don't use it in the implementation, and if I add it, I'd feel to be forced to use it, but I think that could make unit-testing harder.

I'd rather leave it like this for now.

ring/multi_partition_instance_ring.go
colega: Right, that's a leftover, thanks.

ring/partition_ring.go
colega: Right, that's also more efficient. Fixed and added a test with an instance ID with a slash. https://github.com/grafana/dskit/pull/725/commits/1c09d6b5adb5e8f28d5d30615098da3d334675a1

ring/multi_partition_instance_ring.go
colega: Done in https://github.com/grafana/dskit/pull/725/commits/d2b489d9825699124e195869af52722d41289762

PR #726, opened 2025-07-14
feat(ring): add `Zones` as a way to get all the discovered zones

gotjosh:
**What this PR does**:

Exactly what it says in the tin, this is needed to know explicently which are the zones that we have discovered.

This is part of https://github.com/grafana/mimir/issues/12081

**Checklist**
- [ ] Tests updated

Signed-off-by: gotjosh <josue.abreu@gmail.com>

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -169,6 +169,7 @@ type Lifecycler struct {
 	healthyInstancesInZoneCount int
 	instancesInZoneCount        int
 	zonesCount                  int
+	zones                       []string
 
 	tokenGenerator TokenGenerator
 	// The maximum time allowed to wait on the CanJoin() condition.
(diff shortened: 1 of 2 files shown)


ring/lifecycler.go
```diff
@@ -522,6 +523,15 @@ func (i *Lifecycler) ZonesCount() int {
 	return i.zonesCount
 }
 
+// Zones return the list of zones for which there's at least 1 instance registered
+// in the ring.
```
pracucci: [nit] Make it clear that zones are guaranteed to be sorted, so the output is stable.

ring/lifecycler_test.go
```diff
@@ -445,6 +445,49 @@ func TestLifecycler_ZonesCount(t *testing.T) {
 	}
 }
 
+func TestLifecycler_Zones(t *testing.T) {
```
gotjosh, line 477: Feedback from Marco: Ensure we get more cases in the unit tests to test when instances are leaving, etc

ring/lifecycler.go
gotjosh: It now reads like this:

> // Zones return the list of zones for which there's at least 1 instance registered
> // in the ring. They are guaranteed to be sorted alphabetically.

ring/lifecycler_test.go
gotjosh: Done in [edb1f2b](https://github.com/grafana/dskit/pull/726/commits/edb1f2ba2607b24b4f00bdbde6b370d9740edc38)

pracucci (approved):

PR #727, opened 2025-07-17
feat(middleware): add excludedUserAgents support to ClusterValidationMiddleware

colega:
## Summary
Add support for excluding requests from cluster validation based on User-Agent header patterns. This allows requests from specific user agents (like curl, wget, etc.) to bypass cluster validation checks, similar to the existing excludedPaths functionality.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -607,7 +607,7 @@ func BuildHTTPMiddleware(cfg Config, router *mux.Router, metrics *Metrics, logge
 	}
 	if cfg.ClusterValidation.HTTP.Enabled {
 		httpMiddleware = append(httpMiddleware, middleware.ClusterValidationMiddleware(
-			cfg.ClusterValidation.Label, cfg.ClusterValidation.HTTP.ExcludedPaths,
+			cfg.ClusterValidation.Label, cfg.ClusterValidation.HTTP.ExcludedPaths, nil,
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -607,7 +607,7 @@ func BuildHTTPMiddleware(cfg Config, router *mux.Router, metrics *Metrics, logge
 	}
 	if cfg.ClusterValidation.HTTP.Enabled {
 		httpMiddleware = append(httpMiddleware, middleware.ClusterValidationMiddleware(
-			cfg.ClusterValidation.Label, cfg.ClusterValidation.HTTP.ExcludedPaths,
+			cfg.ClusterValidation.Label, cfg.ClusterValidation.HTTP.ExcludedPaths, nil,
```
duricanikolic: This shouldn’t be `nil` but `cfg.ExcludedUserAgents`, correct?

colega: Right, I changed the signature of `ClusterValidationMiddleware` to accept the entire HTTP config instead: https://github.com/grafana/dskit/pull/727/commits/65ea9ff910ffdca2a5c30e0bdfa85875f63b35b6

duricanikolic: Thank you

pr00se (approved):
Thank you!

PR #728, opened 2025-07-22
refactor: remove Redis cache backend

56quarters:
**What this PR does**:

This change removes the Redis client wrapper and the ability to create new instances from configuration.

Code search indicates that this client is not created directly in any Grafana code. It is only potentially created from configuration from Mimir and internal repositories that don't actually run Redis.

[Direct creation](https://github.com/search?q=org%3Agrafana%20%22cache.NewRedisClient%22&type=code)

[Creation from config](https://github.com/search?q=org%3Agrafana%20%22cache.CreateClient%22&type=code)

Per https://github.com/grafana/mimir/issues/11885

**Checklist**
- [X] Tests updated


andyasp (approved):
LGTM, thanks!

PR #733, opened 2025-08-16
feat(ring): add replica tracking for DoUntilQuorum calls

jhalterman:
**What this PR does**:

This PR adds an "available replicas" value to the context used for `DoUntilQuorum` calls when the replicaset contains only a single replica. This will allow the context to be propagated to server side recipients (separate PR) so they can take informed decisions about whether to reject a request, being the only replica available.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/replication_set.go
+++ b/ring/replication_set.go
@@ -15,8 +15,8 @@ import (
 	"github.com/grafana/dskit/spanlogger"
 )
 
-// Used to communicate, via a context, when an associated call was made for only a single replica.
-type onlyReplicaContextKey struct{}
+// Used to communicate, via a context, the number of replicas that are available for a DoUntilQuorum call.
+type availableReplicasContextKey struct{}
 
 // ReplicationSet describes the instances to talk to for a given key, and how
 // many errors to tolerate.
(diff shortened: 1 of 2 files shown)


ring/replication_set.go
```diff
@@ -136,6 +139,11 @@ type DoUntilQuorumConfig struct {
 	// This can be used to prioritise zones that are more likely to succeed, or are expected to complete
 	// faster, for example.
 	ZoneSorter ZoneSorter
+
+	// MarkOnlyReplica indicates when we should provide information in a DoUntilQuorum call, as part of the context, that
+	// indicates a replica is the only one in a set. This can be used by the replica to make informed decisions about load
+	// shedding.
+	MarkOnlyReplica bool
```
dimitarvdimitrov: wdyt about passing the number of eligible replicas in the set? it would be a tiny bit more reusable

jhalterman: That sounds fine - do you have any use cases in mind for it?

dimitarvdimitrov: spreading the load equally across all replicas? sending sharded requests to each replica and getting partial results so we can spread the load better and do more work in parallel

i just came up with this though 😄 if it's not compelling enough feel free to merge as-is

jhalterman: Added

PR #735, opened 2025-08-18
feat(metrics): add native histogram support to tenant registry aggregation

dimitarvdimitrov:
This PR adds support for native histogram aggregation in tenant registry functions.

## What

Implement native histogram support for `SendSumOfHistograms` and `SendSumOfHistogramsWithLabels` functions in the tenant registry system.

## Why  

Currently, the tenant registry only supports classic histogram aggregation. With Prometheus native histograms becoming more prevalent, we need to support aggregating these newer histogram formats across tenants.

## How

- Extended `HistogramData` struct with native histogram fields (schema, zero bucket, spans)
- Updated `AddHistogram` method to handle native histogram field aggregation


dimitarvdimitrov:
i've tested https://github.com/grafana/dskit/pull/736 (which is based on _this_ PR) with Mimir and I can see the per-tenant native histograms being exported along with additional labels. 

So this should be ready for a review

PR #737, opened 2025-08-28
chore: update gomemcache to support context in GetMulti

dimitarvdimitrov:
## Summary
Updates gomemcache dependency to support context.Context in the GetMulti method. This is a breaking change in the upstream library that adds context support for better cancellation.

The upstream gomemcache library merged https://github.com/grafana/gomemcache/pull/26 which adds context support to GetMulti to enable better timeout and cancellation handling. This change is required to use the latest version.

## Changes
- Update gomemcache to v0.0.0-20250828162811-a96f6acee2fe
- Update memcachedClientBackend interface to include context.Context parameter in GetMulti
- Update mock implementation to match new interface signature


cache/memcached_client.go
```diff
@@ -650,7 +650,7 @@ func (c *MemcachedClient) getMultiSingle(ctx context.Context, keys []string, opt
 		// cache client backend.
 		return nil, ctx.Err()
 	default:
-		items, err = c.client.GetMulti(keys, opts...)
```
56quarters, line 653: We can probably skip the check immediately above here now that we're checking the context in the client.

dimitarvdimitrov: yes! https://github.com/grafana/dskit/pull/737/commits/e0661199911c1ec24f18792faadbbea3d98d7baa

PR #742, opened 2025-09-18
feat: add backwards-compatible support for multiple cluster validation labels

periklis:
**What this PR does**:

This PR refines the cluster validation configuration to better distinguish between client-side cluster identity and server-side validation capabilities.

**Key changes**:
- **Kept** the `label` flag - it's now the recommended field for client cluster identity
- **Added** `additional-labels` for server-side multi-cluster validation
- **Updated** `GetEffectiveLabels()` to combine primary label with additional labels
- **Removed** validation preventing both fields from being set together
- **Enhanced** server validation to support multiple cluster labels while clients use single identity

**Architecture rationale**:
- **Clients** represent a single cluster and use the primary `label` for identity
- **Servers** can validate requests from multiple clusters using `label` + `additional-labels`
- This semantic separation clarifies the different use cases and simplifies configuration

**Backwards compatibility**:
- ✅ Existing containers using `-server.cluster-validation.label=cluster-a` continue working unchanged
- ✅ Flag validation no longer prevents using both label and additional-labels together
- ✅ Client behavior unchanged - still uses first effective label

**Example usage**:
```bash
# Client configuration (single cluster identity)
-server.cluster-validation.label=prod-cluster

# Server configuration (validate multiple clusters)
-server.cluster-validation.label=primary-cluster \
-server.cluster-validation.additional-labels=backup-cluster,staging-cluster
```

**Migration benefits**:
- Clear semantic distinction between client identity vs server validation
- No breaking changes - existing configurations work as before
- More intuitive configuration naming

**Checklist**
- [x] Tests updated for new field names and behavior
- [x] Backwards compatibility maintained
- [x] All existing tests pass
- [x] Documentation in code comments updated

### DIFF
--- a/clusterutil/cluster_validation_config.go
+++ b/clusterutil/cluster_validation_config.go
@@ -28,10 +28,7 @@ func (cfg *ClusterValidationConfig) RegisteredFlags() flagext.RegisteredFlags {
 // It combines the primary Label with any AdditionalLabels.
 // The primary Label is always first if present, followed by AdditionalLabels.
 func (cfg *ClusterValidationConfig) GetAllowedClusterLabels() []string {
-	var labels []string
-	if cfg.Label != "" {
-		labels = append(labels, cfg.Label)
-	}
+	labels := []string{cfg.Label}
 	labels = append(labels, cfg.AdditionalLabels...)
 	return labels
 }
--- a/clusterutil/cluster_validation_config.go
+++ b/clusterutil/cluster_validation_config.go
@@ -16,7 +16,7 @@ type ClusterValidationConfig struct {
 func (cfg *ClusterValidationConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	cfg.registeredFlags = flagext.TrackRegisteredFlags(prefix, f, func(prefix string, f *flag.FlagSet) {
 		f.StringVar(&cfg.Label, prefix+"label", "", "Primary cluster validation label.")
-		f.Var(&cfg.AdditionalLabels, prefix+"additional-labels", "Comma-separated list of additional cluster validation labels for server-side validation.")
+		f.Var(&cfg.AdditionalLabels, prefix+"additional-labels", "Comma-separated list of additional cluster validation labels that the server will accept from incoming requests.")
 	})
 }
 
@@ -24,10 +24,10 @@ func (cfg *ClusterValidationConfig) RegisteredFlags() flagext.RegisteredFlags {
 	return cfg.registeredFlags
 }
 
-// GetEffectiveLabels returns the effective cluster validation labels.
+// GetAllowedClusterLabels returns the effective cluster validation labels.
 // It combines the primary Label with any AdditionalLabels.
 // The primary Label is always first if present, followed by AdditionalLabels.
-func (cfg *ClusterValidationConfig) GetEffectiveLabels() []string {
+func (cfg *ClusterValidationConfig) GetAllowedClusterLabels() []string {
 	var labels []string
 	if cfg.Label != "" {
 		labels = append(labels, cfg.Label)
@@ -37,9 +37,11 @@ func (cfg *ClusterValidationConfig) GetEffectiveLabels() []string {
 }
 
 // Validate ensures the cluster validation configuration is valid.
-// Both Label and AdditionalLabels can be set together.
+// AdditionalLabels requires Label to be set.
 func (cfg *ClusterValidationConfig) Validate() error {
-	// Both fields can be set - no validation errors
+	if len(cfg.AdditionalLabels) > 0 && cfg.Label == "" {
+		return fmt.Errorf("additional cluster validation labels require primary label to be set")
+	}
 	return nil
 }
 
@@ -56,14 +58,12 @@ func (cfg *ServerClusterValidationConfig) Validate() error {
 		return err
 	}
 
-	// Get the effective labels (either from deprecated Label or new Labels)
-	labels := cfg.GetEffectiveLabels()
-
-	err := cfg.GRPC.Validate("grpc", labels)
+	// Protocol validation only checks against the primary label
+	err := cfg.GRPC.Validate("grpc", cfg.Label)
 	if err != nil {
 		return err
 	}
-	return cfg.HTTP.Validate("http", labels)
+	return cfg.HTTP.Validate("http", cfg.Label)
 }
 
 func (cfg *ServerClusterValidationConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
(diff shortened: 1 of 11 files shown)


grpcclient/grpcclient.go
```diff
@@ -146,8 +146,10 @@ func (cfg *Config) DialOption(unaryClientInterceptors []grpc.UnaryClientIntercep
 
 	// If cluster validation is enabled, ClusterUnaryClientInterceptor must be the last UnaryClientInterceptor
 	// to wrap the real call.
-	if cfg.ClusterValidation.Label != "" {
-		cfg.clusterUnaryClientInterceptor = middleware.ClusterUnaryClientInterceptor(cfg.ClusterValidation.Label, invalidClusterValidationReporter)
+	effectiveLabels := cfg.ClusterValidation.GetEffectiveLabels()
+	if len(effectiveLabels) > 0 {
+		// For client side, we use the first label as our cluster identity
+		cfg.clusterUnaryClientInterceptor = middleware.ClusterUnaryClientInterceptor(effectiveLabels[0], invalidClusterValidationReporter)
```
julienduchesne: Using the first label feels weird to me. Doesn't feel as clear as the current "label". How about we keep `label` as a setting, we don't deprecate it and we add a new `also_accept_requests_from_labels`. WDYT

periklis: You have a good point here. I've always thought this issues from the server-side validation. But it makes sense to keep the old label as the primary used by client and server interceptor and add an additional for server side validation. Will do! Thanks catching this.

clusterutil/cluster_validation_config.go
```diff
@@ -8,20 +8,41 @@ import (
...
+	Label            string                  `yaml:"label" category:"experimental"`
+	AdditionalLabels flagext.StringSliceCSV  `yaml:"additional_labels" category:"experimental"`
+	registeredFlags  flagext.RegisteredFlags `yaml:"-"`
 }
 
 func (cfg *ClusterValidationConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	cfg.registeredFlags = flagext.TrackRegisteredFlags(prefix, f, func(prefix string, f *flag.FlagSet) {
-		f.StringVar(&cfg.Label, prefix+"label", "", "Optionally define the cluster validation label.")
+		f.StringVar(&cfg.Label, prefix+"label", "", "Primary cluster validation label.")
+		f.Var(&cfg.AdditionalLabels, prefix+"additional-labels", "Comma-separated list of additional cluster validation labels for server-side validation.")
```
julienduchesne: I'd add a short explanation of what "additional labels" implies here: It's not the label that the cluster sends, but they're labels that it accepts from requests

clusterutil/cluster_validation_config.go
```diff
@@ -8,20 +8,41 @@ import (
...
+	labels = append(labels, cfg.AdditionalLabels...)
+	return labels
+}
+
+// Validate ensures the cluster validation configuration is valid.
+// Both Label and AdditionalLabels can be set together.
+func (cfg *ClusterValidationConfig) Validate() error {
+	// Both fields can be set - no validation errors
+	return nil
+}
```
julienduchesne, line 50: I believe this can be removed

clusterutil/cluster_validation_config.go
```diff
@@ -8,20 +8,41 @@ import (
...
 }
 
 func (cfg *ClusterValidationConfig) RegisteredFlags() flagext.RegisteredFlags {
 	return cfg.registeredFlags
 }
 
+// GetEffectiveLabels returns the effective cluster validation labels.
+// It combines the primary Label with any AdditionalLabels.
+// The primary Label is always first if present, followed by AdditionalLabels.
+func (cfg *ClusterValidationConfig) GetEffectiveLabels() []string {
```
julienduchesne:
```suggestion
// GetAllowedClusterLabels returns the effective cluster validation labels.
// It combines the primary Label with any AdditionalLabels.
// The primary Label is always first if present, followed by AdditionalLabels.
func (cfg *ClusterValidationConfig) GetAllowedClusterLabels() []string {
```

Maybe?

clusterutil/cluster_validation_config.go
```diff
@@ -54,10 +83,10 @@ type ClusterValidationProtocolConfig struct {
 	SoftValidation bool `yaml:"soft_validation" category:"experimental"`
 }
 
-func (cfg *ClusterValidationProtocolConfig) Validate(prefix string, label string) error {
-	if label == "" {
+func (cfg *ClusterValidationProtocolConfig) Validate(prefix string, labels []string) error {
+	if len(labels) == 0 {
```
julienduchesne: I think this is wrong. We want to validate that the "label" is set, we don't really care about the additional ones

clusterutil/cluster_validation_config.go
julienduchesne: Or better yet, validate that "label" is set if additional labels is set. Doesn't make sense to set "additional" labels but not the main one

clusterutil/cluster_validation_config_test.go
```diff
@@ -89,10 +96,94 @@ func TestServerClusterValidationConfig_RegisteredFlags(t *testing.T) {
...
+			label:            "",
+			additionalLabels: nil,
+			expectedLabels:   nil,
+		},
+		"only primary label set": {
+			label:            "cluster-a",
+			additionalLabels: nil,
+			expectedLabels:   []string{"cluster-a"},
+		},
+		"only additional labels set": {
```
julienduchesne, line 118: I don't think this case makes sense

grpcclient/grpcclient.go
julienduchesne: I would keep the same logic regardless, i.e.: keep using `cfg.ClusterValidation.Label`. I think the additional labels should only be referenced in the server logic (allowing requests)

clusterutil/cluster_validation_config.go
```diff
@@ -8,20 +8,43 @@ import (
...
+// It combines the primary Label with any AdditionalLabels.
+// The primary Label is always first if present, followed by AdditionalLabels.
+func (cfg *ClusterValidationConfig) GetAllowedClusterLabels() []string {
+	var labels []string
+	if cfg.Label != "" {
+		labels = append(labels, cfg.Label)
+	}
+	labels = append(labels, cfg.AdditionalLabels...)
+	return labels
+}
```
julienduchesne, line 41:
```suggestion
// The primary Label is always first if present, followed by AdditionalLabels.
func (cfg *ClusterValidationConfig) GetAllowedClusterLabels() []string {
	labels := []string{cfg.Label}
	labels = append(labels, cfg.AdditionalLabels...)
	return labels
}
```

grpcclient/grpcclient_test.go
```diff
@@ -70,6 +70,14 @@ func TestDialOptionWithClusterValidation(t *testing.T) {
...
 			expectedClusterUnaryClientInterceptor: true,
 		},
+		"if cluster validation label is set (deprecated field) and there is no input and an implicit UnaryClientInterceptor we expect ClusterUnaryClientInterceptor to be the last one": {
+			clusterValidation:      clusterutil.ClusterValidationConfig{Label: "cluster"},
+			inputUnaryInterceptors: nil,
+			// setting rateLimit creates an implicit UnaryClientInterceptor
+			rateLimit:                             10,
+			expectedUnaryInterceptors:             2,
+			expectedClusterUnaryClientInterceptor: true,
+		},
```
julienduchesne: I don't understand this test

julienduchesne (approved):
LGTM! Just some nits. I think this should work

grpcclient/grpcclient_test.go
periklis: This is obsolete since we don't deprecated the label field. I am going to remove this.

Issue #749, opened 2025-10-10
Bug: `grpcInflightLimitCheck` doesn't always call `RPCCallFinished`


colega:
_This is a child issue for https://github.com/grafana/mimir/issues/12966_

When updating `google.golang.org/grpc` to v1.75 it includes the change from https://github.com/grpc/grpc-go/pull/8439 which causes this statement to no longer be true:

https://github.com/grafana/dskit/blob/2b7008cbb8875df5add7ec480f9cc18ad2895271/server/limits.go#L46-L46

The new shortcut introduced after calling the InTapHandle causes request to be early-aborted after executing the tap handle, so `RPCCallStarting` was already called but `RPCCallFinished` will never be called.

PR #750, opened 2025-10-10
fix: grpcInflightLimitCheck should always call RPCCallFinished

colega:
Fixes: https://github.com/grafana/dskit/issues/749

This also updates grpc library to v1.75 where this bug is introduced, so the tests are consistent with the code.

This PR adds a timer on each request that will verify whether the request actually proceeded, and if it didn't, it will call RPCCallFinished to make sure that the method limiter provided has performed the necessary cleanup.

In the usual flow, we will cancel that timer as soon as we start processing the headers.


server/limits.go
```diff
@@ -3,13 +3,19 @@ package server
...
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/stats"
 	"google.golang.org/grpc/tap"
 )
 
+const unprocessedRequestCheckTimeout = 10 * time.Second
```
NickAnge, line 19: Curious — how was this value decided? Is it because the request timeout is 5 seconds?

Nit: Maybe we can add a comment mentioning that.

server/limits.go
```diff
@@ -26,32 +32,94 @@ type GrpcInflightMethodLimiter interface {
...
+	ctx, err := g.methodLimiter.RPCCallStarting(ctx, info.FullMethodName, info.Header)
+	if err != nil {
+		return ctx, err
+	}
+
+	// We called RPCCallStarting, so we need to ensure RPCCallFinished is called once the request is done.
+	// Because of a shortcut introduced in https://github.com/grpc/grpc-go/pull/8439 this may not happen.
+	// We could create a goroutine that would watch ctx.Done() and call RPCCallFinished if the context is done and we have not started processing the headers yet.
+	// However, that would mean paying the cost of an extra goroutine for every single gRPC request, just in case the request's context is cancelled before we start processing it.
+	// Instead of that we schedule a cheaper timer that we will cancel in the happy case, which will run after 10s and perform the cleanup only when
```
NickAnge: only when..  ? The request has timed out ?

colega: Only when ...neded 😅 
Does that make sense? Pushed.

server/limits.go
colega: Check https://github.com/grafana/dskit/pull/750/commits/9dc725b3adcef07cc696c8a64d286b1d2a7c31e4 to see if that convinces you :)

PR #757, opened 2025-10-15
feat(ring): add support for reporting component versions

charleskorn:
**What this PR does**:

This PR adds support for reporting component versions for a ring member.

The motivating scenario for this is to allow Mimir queriers to report the maximum query plan version they support, so that query-frontends don't generate query plans that queriers don't understand. See https://github.com/grafana/mimir/pull/13017 for more details.

(none)

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -55,17 +55,8 @@ type LifecyclerConfig struct {
 
 	// Injected internally
 	ListenPort int `yaml:"-"`
-	// HideTokensInStatusPage allows tokens to be hidden from management tools e.g. the status page, for use in contexts which do not utilize tokens.
-	HideTokensInStatusPage bool `yaml:"-"`
-	// ShowVersionsInStatusPage enables displaying versions on the status page.
-	ShowVersionsInStatusPage bool `yaml:"-"`
 
-	// ComponentNames are the names of the components in InstanceDesc.Versions, used only for display on the status page.
-	// If a component in Versions has no name in ComponentNames, then the version will be shown on the status page
-	// without a name.
-	// Given the values are only used for display on the status page, it is safe to leave this empty,
-	// make changes between releases etc.
-	ComponentNames map[uint64]string `yaml:"-"`
+	StatusPageConfig StatusPageConfig `yaml:"-"`
 
 	// If set, specifies the TokenGenerator implementation that will be used for generating tokens.
 	// Default value is nil, which means that RandomTokenGenerator is used.
@@ -1117,7 +1108,7 @@ func (i *Lifecycler) getRing(ctx context.Context) (*Desc, error) {
 }
 
 func (i *Lifecycler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
-	newRingPageHandler(i, i.cfg.HeartbeatTimeout, i.cfg.HideTokensInStatusPage, !i.cfg.ShowVersionsInStatusPage, i.cfg.ComponentNames).handle(w, req)
+	newRingPageHandler(i, i.cfg.HeartbeatTimeout, i.cfg.StatusPageConfig).handle(w, req)
 }
 
 // unregister removes our entry from consul.
--- a/ring/ring_http.go
+++ b/ring/ring_http.go
@@ -32,13 +32,9 @@ var defaultPageTemplate = template.Must(template.New("webpage").Funcs(template.F
 type httpResponse struct {
 	Ingesters []ingesterDesc `json:"shards"`
 	Now       time.Time      `json:"now"`
-	// ShowTokens indicates whether the Show Tokens button is clicked.
-	ShowTokens bool `json:"-"`
-	// DisableTokens hides the concept of tokens entirely in the page, across all elements.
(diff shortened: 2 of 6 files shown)


ring/ring.proto
```diff
@@ -55,6 +55,10 @@ message InstanceDesc {
 	// Read-only instances go through standard state changes, and special handling is applied to them
 	// during shuffle shards.
 	bool read_only = 11;
+
+  // Map of component / feature identifiers to version number.
```
56quarters, line 59: Can you expand some of the assumptions here? I.e. because we're using a `uint64` version numbers are expected to increase and can be ordered (I'm guessing)

ring/lifecycler.go
```diff
@@ -57,6 +57,13 @@ type LifecyclerConfig struct {
 	ListenPort int `yaml:"-"`
 	// HideTokensInStatusPage allows tokens to be hidden from management tools e.g. the status page, for use in contexts which do not utilize tokens.
 	HideTokensInStatusPage bool `yaml:"-"`
+	// ShowVersionsInStatusPage enables displaying versions on the status page.
+	ShowVersionsInStatusPage bool `yaml:"-"`
+
+	// ComponentNames are the names of the components in InstanceDesc.Versions, used for display on the status page.
+	// If a component in Versions has no name in ComponentNames, then the version will be shown on the status page
+	// without a name.
+	ComponentNames map[uint64]string `yaml:"-"`
```
56quarters: I notice that this mapping exists in a few different places in this PR. What are the expectations for how this gets defined in consumers of dskit? Are there consequences for changing it after it has already been set in a running cluster?

56quarters (approved):
LGTM with some clarifying questions

56quarters:
We probably want to have someone from Loki/Tempo/Pyroscope take a look at this too since they use the ring from dskit.

ring/ring.proto
charleskorn: Does what I've added in 1bfc588 cover it?

ring/lifecycler.go
charleskorn: Does what I've added in 63874f1 cover this?

I've also extracted `HideTokensInStatusPage`, `ShowVersionsInStatusPage` and `ComponentNames` to their own struct in d599603, which hopefully makes their purpose clearer.

ring/ring.proto
56quarters: Perfect, thank you.

ring/model.go
```diff
@@ -28,6 +28,8 @@ func (ts ByID) Len() int           { return len(ts) }
 func (ts ByID) Swap(i, j int)      { ts[i], ts[j] = ts[j], ts[i] }
 func (ts ByID) Less(i, j int) bool { return ts[i].Id < ts[j].Id }
 
+type InstanceVersions map[uint64]uint64
```
pstibrany, line 31: I don't understand this type. Can you please document what is this map trying to express? I assumed single version is a string, but here it's a map of ... numbers to numbers?

ring/ring.pb.go
```diff
@@ -137,6 +137,11 @@ type InstanceDesc struct {
 	// Read-only instances go through standard state changes, and special handling is applied to them
 	// during shuffle shards.
 	ReadOnly bool `protobuf:"varint,11,opt,name=read_only,json=readOnly,proto3" json:"read_only,omitempty"`
+	// Map of component / feature identifiers to version number.
+	// The component / feature identifiers are specific to the application (ie. aren't defined in dskit).
+	// We use a uint64 for the version number itself as versions are expected to increase and be
+	// sortable.
+	Versions map[uint64]uint64 `protobuf:"bytes,12,rep,name=versions,proto3" json:"versions,omitempty" protobuf_key:"varint,1,opt,name=key,proto3" protobuf_val:"varint,2,opt,name=value,proto3"`
```
pstibrany, line 144: Oh, there's documentation here.

ring/ring_http.go
```diff
@@ -32,24 +32,41 @@ var defaultPageTemplate = template.Must(template.New("webpage").Funcs(template.F
 type httpResponse struct {
 	Ingesters []ingesterDesc `json:"shards"`
 	Now       time.Time      `json:"now"`
-	// ShowTokens indicates whether the Show Tokens button is clicked.
-	ShowTokens bool `json:"-"`
-	// DisableTokens hides the concept of tokens entirely in the page, across all elements.
-	DisableTokens bool `json:"-"`
+	// TokensToggledOn indicates whether the Show Tokens button is clicked.
+	TokensToggledOn bool             `json:"-"`
```
pstibrany, line 36: super-nit: was there anything wrong with `ShowTokens`?

pstibrany (approved):
lgtm, thanks!

ring/ring_http.go
charleskorn: I was trying to make the difference between `TokensToggledOn` and `HideTokensUIElements` very clear (previously they were called `ShowTokens` and `disableTokens` and the difference wasn't super clear to me)

PR #760, opened 2025-10-22
feat: track all bytes sent and received in the memberlist TCP transport

pracucci:
**What this PR does**:

We currently don't have a single metric tracking all bytes sent/received using the memberlist TCP transport. Currently there are 2 distinct type of metrics:

- `memberlist_tcp_transport_packets_sent_bytes_total` and `memberlist_tcp_transport_packets_received_bytes_total`
- `memberlist_client_state_pulls_bytes_total` and `memberlist_client_state_pushes_bytes_total`

These metrics don't track pings, but to be honest pings are a small amount of data transferred.

That being said, to keep it easier to query, in this PR I propose to add a new set of metrics tracking all data transfer by memberlist TCP transport. I'm doing it by wrapping the `net.Conn` created by the `TCPTransport`. I think it's easier to reason about, and avoid asking questions like "is everything tracked if we sum packets + state pulls/pushes metrics or not?". I also understand that in practice the new metric is not much different than summing the other metrics we already have. I'm fine to close this PR if there are objections.

In the meanwhile I've tested it in a dev Mimir cluster and looks working:

Screenshot 2025-10-22 at 20 18 20

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/kv/memberlist/tcp_transport_conn.go
+++ b/kv/memberlist/tcp_transport_conn.go
@@ -1,0 +1,104 @@
+package memberlist
+
+import (
+	"net"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+)
+
+const (
+	// flushThreshold is the number of bytes after which we flush the local counters
+	// to the Prometheus counters. This reduces the frequency of expensive Prometheus
+	// counters updates.
+	flushThreshold = 64 * 1024 // 64KB
+)
+
+// meteredConn wraps a net.Conn to track the number of bytes sent and received.
+// It periodically flushes the byte counts to Prometheus counters when a threshold
+// is reached, and always flushes on Close() to ensure accurate tracking for
+// short-lived connections.
+type meteredConn struct {
+	net.Conn
+
+	sentBytesCounter     prometheus.Counter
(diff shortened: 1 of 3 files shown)


kv/memberlist/tcp_transport_conn.go
```diff
@@ -0,0 +1,104 @@
...
+
+// SetReadDeadline implements net.Conn.
+func (c *meteredConn) SetReadDeadline(t time.Time) error {
+	return c.Conn.SetReadDeadline(t)
+}
+
+// SetWriteDeadline implements net.Conn.
+func (c *meteredConn) SetWriteDeadline(t time.Time) error {
+	return c.Conn.SetWriteDeadline(t)
+}
```
seizethedave, line 78: nit: You should get these pass-through methods for free by embedding `net.Conn` in `meteredConn`?

pracucci: I thought about it. The reason why I typically don't use pass-through methods is because if the interface change I want the compilation to break, because it's a signal if the changes to the interface should be addressed or not (e.g. in this case, based on the actual change we may or may not need to change `meteredConn` logic too). Maybe unlikely that `net.Conn` will change, but I consider it a good practice to keep.

PR #762, opened 2025-10-23
feat: add zone-aware routing support to memberlist

pracucci:
**What this PR does**:

In this PR I'm adding zone-aware routing support to memberlist. This feature relies on `NodeSelectorDelegate` that I've introduced in [this PR](https://github.com/grafana/memberlist/pull/7).

By default, zone-aware routing is disabled and memberlist client behaviour does **not** change.

When zone-aware routing is enabled, memberlist client requires the following information to be set in the config:
- The availability zone
- The node role

There are two roles:
- member: a normal application instance. A member only gossip and push/pull to other nodes (members and bridges) in **the same zone**.
- bridge: a special application instance that act as a bridge between AZs. A bridge gossips and push/pull to other nodes in **the same zone**, plus to **the bridges in other zones**. The bridge is also special because it prefers to gossip to other bridges in different zones, to increase the likelihood that messages are actually propagated to the other zones. The way it works under the hood, is that for each message to gossip, the bridge always select at least 1 node in other zone (here comes the preference).

_Draft because I want to integrate it in Mimir and test in a dev cluster first._

Part of https://github.com/grafana/mimir-squad/issues/2634

**Checklist**
- [x] Tests updated

### DIFF
--- a/kv/memberlist/node_zone_aware_routing_test.go
+++ b/kv/memberlist/node_zone_aware_routing_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
 	"strconv"
 	"strings"
 	"testing"
@@ -250,7 +251,7 @@ func TestZoneAwareRouting_EndToEnd(t *testing.T) {
 			var (
 				c      = dataCodec{}
 				ctx    = context.Background()
-				logger = log.NewNopLogger()
+				logger = log.NewLogfmtLogger(os.Stderr)
 			)
 
 			// Helper function to create a node configuration.
@@ -282,7 +283,7 @@ func TestZoneAwareRouting_EndToEnd(t *testing.T) {
 			// Zone B: bridge-b, member-b-1, member-b-2
 
 			// Zone A bridge - create and start first.
-			bridgeA := NewKV(makeConfig(nil, "zone-a", "bridge"), logger, &staticDNSProviderMock{}, prometheus.NewPedanticRegistry())
+			bridgeA := NewKV(makeConfig(nil, "zone-a", "bridge"), log.WithPrefix(logger, "instance", "bridge-zone-a"), &staticDNSProviderMock{}, prometheus.NewPedanticRegistry())
 			require.NoError(t, services.StartAndAwaitRunning(ctx, bridgeA))
 			t.Cleanup(func() {
 				require.NoError(t, services.StopAndAwaitTerminated(ctx, bridgeA))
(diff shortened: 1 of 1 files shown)


kv/memberlist/node_zone_aware_routing.go
```diff
@@ -0,0 +1,146 @@
...
+	// Role configuration values.
+	roleConfigMember = "member"
+	roleConfigBridge = "bridge"
+)
+
+// ZoneAwareRoutingConfig holds configuration for zone-aware routing in memberlist.
+type ZoneAwareRoutingConfig struct {
+	Enabled bool   `yaml:"enabled" category:"experimental"`
+	Zone    string `yaml:"instance_availability_zone" category:"experimental"`
+	Role    string `yaml:"role" category:"experimental"`
```
tcard, line 47: Maybe not worth the complexity, but instead of pre-deciding the role of each node, maybe nodes in the same AZ could elect a bridge among themselves?

pracucci: We already have a Mimir component called `memberlist-seed`. That will be the bridge too, that's why I want to make it explicit.

kv/memberlist/memberlist_client.go
```diff
@@ -1295,9 +1348,9 @@ func (m *KV) broadcastNewValue(key string, change Mergeable, version uint, codec
 
 // NodeMeta is method from Memberlist Delegate interface
 func (m *KV) NodeMeta(_ int) []byte {
-	// we can send local state from here (512 bytes only)
-	// if state is updated, we need to tell memberlist to distribute it.
-	return nil
+	// Return the encoded node metadata if zone-aware routing is enabled.
+	// Otherwise, return nil (no metadata).
+	return m.nodeMeta
```
duricanikolic, line 1353: Who will call this method?
The content of `m.nodeMeta` is encoded combination of role and zone, and the caller has to know how to decode it.

kv/memberlist/node_zone_aware_routing.go
```diff
@@ -0,0 +1,146 @@
...
+	// Determine local role from config.
+	var localRole NodeRole
+	switch cfg.Role {
+	case NodeRoleMember.String():
+		localRole = NodeRoleMember
+	case NodeRoleBridge.String():
+		localRole = NodeRoleBridge
+	default:
+		// Default to member role if unknown (should never happen after validation).
+		localRole = NodeRoleMember
```
duricanikolic: [nit]: isn't it better if we return an error here? 

It is impossible that we have an unknown role, because we first call `ZoneAwareRoutingConfig.Validate()`, and then before calling `newZoneAwareNodeSelectionDelegate()` in `configureZoneAwareRouting()`, we check if the role is valid again. So we will check the role 3 times.

At this point, isn't it better if we pass `role` and `zone` to `newZoneAwareNodeSelectionDelegate()`?

kv/memberlist/memberlist_client.go
pracucci: It's called by the upstream memberlist library. `KV` implements `memberlist.Delegate` and this function is part of that interface.

kv/memberlist/node_zone_aware_routing.go
pracucci: Looks a very long comment for a case that's not expected to happen ;) I would just keep things easy.

duricanikolic: It was just a nit :)

pracucci: You know, I can afford to make such bold comments to you because we have such a good close relationship ;) A part from this, I will look at this nit too, when I will address the other comments. I'm currently working to test it in dev, to make sure it works, before working on final details.

duricanikolic: I know, of course :)
Anyway, TL;DR here is that we parse role 3 times before we create a `NodeSelectionDelegate`, and we can for sure improve this.

pracucci: You're right. Done in https://github.com/grafana/dskit/pull/762/commits/33982f0b4898c2a709257cd4835970085a6e1da3

kv/memberlist/node_meta.go
```diff
@@ -0,0 +1,82 @@
...
+// EncodeNodeMetadata encodes node metadata into a compact binary format.
+// The encoding is versioned to allow future extensions while maintaining backward compatibility.
+//
+// We use a custom compact encoding rather than protobuf because memberlist has strict size limits
+// for node metadata (typically 512 bytes) and this data is broadcast frequently in alive messages.
+//
+// Version 1 format:
+//   - Byte 0: Version (1 byte)
+//   - Byte 1: Role (1 byte): 1=member, 2=bridge
+//   - Byte 2: Zone length (1 byte): length of zone string (0-16)
```
colega, line 26: I don't think we need this: it ends where the bytes slice ends. You could argue that we may want to add more fields in the future, but that could just be a version 2, so YAGNI.

kv/memberlist/node_zone_aware_routing.go
```diff
@@ -0,0 +1,134 @@
...
+			return true, false
+		}
+		// Different zone: only select if it's a bridge node, and prefer it.
+		if remoteRole == NodeRoleBridge {
+			return true, true
+		}
+		return false, false
+
+	default:
+		// Unknown role: select but don't prefer (should never happen).
```
colega, line 137: I think it would be fair to log a warning here, given it's not much effort to have a logger on the delegate.

kv/memberlist/node_meta.go
```diff
@@ -0,0 +1,82 @@
...
+	if len(e) < 3+zoneLen {
+		return ""
+	}
+	if zoneLen == 0 {
+		return ""
+	}
+	// Zero-allocation conversion from []byte to string.
+	// Safe because we're not modifying the underlying data.
+	return unsafe.String(&e[3], zoneLen)
+}
```
colega, line 82: Ignorable nit, just talking aloud: we could have ignored the unsafe allocation by defining ZoneEquals:

```suggestion
// ZoneEquals returns true if the zone provided is the same as the encoded one.
// Returns false if the metadata is invalid.
func (e EncodedNodeMetadata) ZoneEquals(other string) bool {
	if len(e) < minMetadataSizeVersion1 || e[0] != 1 {
		return false
	}
	zoneLen := int(e[2])
	if len(e) < 3+zoneLen {
		return ""
	}
	return string(e[3:3+zoneLen]) == other
}
```

colega (approved):
LGTM, very elegant solution <3

kv/memberlist/node_meta.go
pracucci: I still prefer to keep the length, which is an indirect form of consistency check. It's just 1 byte

kv/memberlist/node_meta.go
pracucci: True, but there's still a case where I need to read the zone (UI). So the current function would have to exist anyway.

kv/memberlist/node_zone_aware_routing.go
pracucci: Done in https://github.com/grafana/dskit/pull/762/commits/625bdeccf3c91978c9dea7f079e891a3f9ad7290

duricanikolic (approved):
Thank you for addressing all my findings.
LGTM

PR #775, opened 2025-10-28
fix: dns: ignore non-SRV answers for SRV queries

56quarters:
**What this PR does**:

Some DNS servers include A or AAAA records as answers for SRV queries. This changes our resolver to ignore any non-SRV answers returned from a SRV query. This matches the behavior of the built-in go resolver.

See https://github.com/grafana/mimir/issues/12713

**Checklist**
- [X] Tests updated


dimitarvdimitrov (approved):
wild

PR #777, opened 2025-10-28
fix: dns: ignore non-SRV answers for SRV queries

56quarters:
Manual backport of #775

Some DNS servers include A or AAAA records as answers for SRV queries. This changes our resolver to ignore any non-SRV answers returned from a SRV query. This matches the behavior of the built-in go resolver.

See grafana/mimir#12713

### DIFF
--- a/dns/miekgdns/resolver.go
+++ b/dns/miekgdns/resolver.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"net"
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/miekg/dns"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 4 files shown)


dns/miekgdns/resolver.go
```diff
@@ -43,7 +46,13 @@ func (r *Resolver) LookupSRV(_ context.Context, service, proto, name string) (cn
 				Port:     addr.Port,
 			})
 		default:
-			return "", nil, errors.Errorf("invalid SRV response record %s", record)
+			// Ignore unexpected non-SRV responses. This matches the behavior of the
+			// built-in go resolver. See https://github.com/grafana/mimir/issues/12713
+			level.Debug(r.Logger).Log(
```
alexweav, line 51: nit: Will this crash if Logger is nil? There's no constructor in the old version and Resolver is exported, so if any databases init a Resolver directly without using the provider this might happen :thinking:

56quarters: This branch is specific to the Mimir 2.17 release so nobody else should be using it (and this resolver is removed in favor of "miekdns2" in later versions).

PR #779, opened 2025-10-30
fix(runtimeconfig): fail when can't merge multiple files

colega:
If one file has all string keys in a map, it's a `map[string]interface{}`, if it has a mix of strings and ints, then it's `map[interface{}]interface{}`. When two files with different types were merged, the conflicting keys of the first one were silently overwritten instead of merging them.

### DIFF
--- a/runtimeconfig/manager.go
+++ b/runtimeconfig/manager.go
@@ -278,7 +278,7 @@ func mergeConfigMaps(a, b map[string]interface{}, path string) (_ map[string]int
 		if v, ok := v.(map[string]interface{}); ok {
 			if bv, ok := out[k]; ok {
 				if bv, ok := bv.(map[string]interface{}); ok {
-					out[k], err = mergeConfigMaps(bv, v, "."+k)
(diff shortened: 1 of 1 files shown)


runtimeconfig/manager.go
```diff
@@ -258,23 +262,33 @@ func isGzip(data []byte) bool {
...
+		_, bIsMap := b[k].(map[string]interface{})
+		if aHasKey && aIsMap != bIsMap {
+			return nil, errors.Errorf("conflicting types for %q: %T != %T", path+"."+k, a[k], b[k])
+		}
+
 		if v, ok := v.(map[string]interface{}); ok {
 			if bv, ok := out[k]; ok {
 				if bv, ok := bv.(map[string]interface{}); ok {
-					out[k] = mergeConfigMaps(bv, v)
+					out[k], err = mergeConfigMaps(bv, v, "."+k)
```
colega:
```suggestion
					out[k], err = mergeConfigMaps(bv, v, path+"."+k)
```

PR #781, opened 2025-10-31
chore(metrics): {request,response}_message_bytes native histogram

korniltsev-grafanista:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -85,8 +85,9 @@ type Config struct {
 	// https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#HistogramOpts
 	// for details. A generally useful value is 1.1.
 	MetricsNativeHistogramFactor float64 `yaml:"-"`
-	// Same as MetricsNativeHistogramFactor, but for old metrics like response_message_bytes, request_message_bytes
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -85,6 +85,8 @@ type Config struct {
 	// https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#HistogramOpts
 	// for details. A generally useful value is 1.1.
 	MetricsNativeHistogramFactor float64 `yaml:"-"`
+	// Same as MetricsNativeHistogramFactor, but for old metrics like response_message_bytes, request_message_bytes
+	MetricsOldNativeHistogramFactor float64 `yaml:"-"`
```
simonswine: I do think this is ok, but I don't think it's really required to have different factor compared to the other ones. It is also quite hard to get rid of in the future.

Maybe just a boolean, default false. `MetricsMessageSizeNativeHistograms` or the likes.

If you want to keep the factor field, worth having what it controls in the field name `MetricsMessageSizeNativeHistogramFactor`

simonswine (approved):
LGTM

simonswine (approved):
LGTM

PR #782, opened 2025-10-31
chore: cache: Remove unneeded abstraction within Memcached client

56quarters:
**What this PR does**:

Now that we aren't sharing logic with a Redis client via baseClient, remove some cases where we were passing around closures and call methods directly.

This changes the behavior of the cache client in two ways but I don't think they are significant.

* Deletes are synchronous and don't use the job queue anymore. Previously, deletes went through the async job queue to make sure that they ran after any queued `SetAsync()` calls for the same key. This isn't  actually depended on anywhere except GEM which is not getting new releases anymore. Now, the same effect can be achieved by using the synchronous `Set()` method which didn't exist when the previous behavior was used in GEM.
* The `.CompareAndSwap()` method has been changed to use a shared method for store operations that means timing and payload size metrics are incremented even on error. I don't think this matters for any usecases at Grafana.

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/cache/memcached_client_test.go
+++ b/cache/memcached_client_test.go
@@ -130,6 +130,32 @@ func TestMemcachedClient_GetMulti(t *testing.T) {
 		require.Equal(t, map[string][]byte{"foo": []byte("bar")}, res)
 		require.Equal(t, 1, backend.allocations)
 	})
+
+	t.Run("multiple batches", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


cache/memcached_client.go
56quarters, line 606: I've elected to _not_ change this to only set these metrics on success. The client is currently inconsistent about this so I've tried to keep things largely the same.

PR #787, opened 2025-11-12
fix(runtimeconfig): treat nil values in config merge as zero maps when merging

chaudum:
**What this PR does**:

This PR fixes an issue when merging config files where one file has a `null` value and the other a zero or non-zero map.

Example:

`config1.yaml`:
```yaml
overrides:
  limits: null
```

`config2.yaml`:
```yaml
overrides:
  limits:
    limit_a: 1
```

**Checklist**
- [x] Tests updated

### DIFF
--- a/runtimeconfig/manager.go
+++ b/runtimeconfig/manager.go
@@ -268,9 +268,22 @@ func mergeConfigMaps(a, b map[string]interface{}, path string) (_ map[string]int
 		out[k] = v
 	}
 	for k, v := range b {
-		_, aHasKey := a[k]
+		aVal, aHasKey := a[k]
+		bVal, bHasKey := b[k]
+
 		_, aIsMap := a[k].(map[string]interface{})
 		_, bIsMap := b[k].(map[string]interface{})
+
+		if aHasKey && aVal == nil && bIsMap {
+			aIsMap = true
+			out[k] = make(map[string]interface{})
+		}
+
+		if bHasKey && bVal == nil && aIsMap {
+			bIsMap = true
+			v = make(map[string]interface{})
+		}
+
 		if aHasKey && aIsMap != bIsMap {
 			return nil, errors.Errorf("conflicting types for %q: %T != %T", path+"."+k, a[k], b[k])
 		}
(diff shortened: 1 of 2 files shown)


runtimeconfig/manager.go
```diff
@@ -268,9 +268,22 @@ func mergeConfigMaps(a, b map[string]interface{}, path string) (_ map[string]int
...
+
 		_, aIsMap := a[k].(map[string]interface{})
 		_, bIsMap := b[k].(map[string]interface{})
+
+		if aHasKey && aVal == nil && bIsMap {
+			aIsMap = true
+			out[k] = make(map[string]interface{})
+		}
+
+		if bHasKey && bVal == nil && aIsMap {
```
chaudum, line 282: `bHasKey` is always `true` and could be omitted.

runtimeconfig/manager_test.go
```diff
@@ -376,6 +376,38 @@ func TestOverridesManagerMultipleFilesWithEmptyFile(t *testing.T) {
...
+  "123": null
+  "456":
+    limit1: 100
+  "789": null
+`,
+			`overrides:
+  "123":
+    limit2: 100
+  "456": null
+  "789": {}
```
salvacorts, line 393: Can we add a third test where the first override is {} and the second is null?

runtimeconfig/manager_test.go
```diff
@@ -376,6 +376,38 @@ func TestOverridesManagerMultipleFilesWithEmptyFile(t *testing.T) {
...
+		Loader:       testLoadOverrides,
+	}
+
+	overridesManager, err := New(overridesManagerConfig, "overrides", nil, log.NewNopLogger())
+	require.NoError(t, err)
+	err = services.StartAndAwaitRunning(context.Background(), overridesManager)
+	require.NoError(t, err)
+
+	merged := overridesManager.GetConfig()
+	t.Log(merged)
```
salvacorts: I think we should check here the expected output

colega (approved):
Thanks

PR #788, opened 2025-11-17
feat(ring/client): add health check grace period to `Pool`

charleskorn:
**What this PR does**:

This PR adds an optional health check grace period to `Pool` in the `ring/client` package.

The motivating example for this change is in Mimir, where queriers use a `Pool` to manage connections back to the query-frontend, and monitor query-frontends' health to remove stale connections from the pool. Query-frontends are expected to gracefully complete requests on shutdown, and immediately change their health status as soon as they receive a shutdown signal (ie. while remaining requests are still inflight).

However, because query-frontends immediately change their health status when they receive a shutdown signal, queriers immediately start cleaning up connections to query-frontends that are shutting down. This means that any inflight responses from queriers to query-frontends are aborted, causing queries to fail. 

With this grace period in place, queriers can be configured to retain query-frontend connections until after query-frontends are expected to have shut down. 

Note to reviewers: the tests for this would have been a great candidate for `testing/synctest`, but we currently target Go 1.24 for this project, and `synctest` is only considered stable in Go 1.25.

(none)

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/client/pool.go
+++ b/ring/client/pool.go
@@ -279,14 +279,19 @@ func (p *Pool) cleanUnhealthy() {
 		}
 
 		if time.Since(member.firstFailedHealthCheck) >= p.cfg.HealthCheckGracePeriod {
-			level.Warn(p.logger).Log("msg", fmt.Sprintf("removing %s failing healthcheck", p.clientName), "addr", addr, "reason", err)
+			level.Warn(p.logger).Log(
+				"msg", fmt.Sprintf("removing %s failing healthcheck", p.clientName),
+				"addr", addr,
+				"reason", err,
+				"first_failed_at", member.firstFailedHealthCheck,
+			)
 			p.RemoveClientFor(addr)
 		} else {
 			level.Debug(p.logger).Log(
 				"msg", fmt.Sprintf("%s failed healthcheck within grace period, not removing", p.clientName),
 				"addr", addr,
 				"reason", err,
-				"first_failure", member.firstFailedHealthCheck,
+				"first_failed_at", member.firstFailedHealthCheck,
 			)
 		}
 


ring/client/pool.go
```diff
@@ -73,11 +82,16 @@ type Pool struct {
...
 	sync.RWMutex
-	clients map[string]PoolClient
+	members map[string]*poolMember
 
 	clientsMetric prometheus.Gauge
 }
 
+type poolMember struct {
+	client                 PoolClient
+	firstFailedHealthCheck time.Time
```
tcp13equals2, line 92: [nit] An reason not store the expiry time ... as then you don't have to do the extra arithmetic each time.

Calculate the expiry time when this is set, and then you only need to compare to now.

charleskorn: If we do that, then we can't log the first time the health check failed when removing the client (which I've just realised is inconsistently done - I'll push a change for this).

ring/client/pool.go
```diff
@@ -242,17 +262,41 @@ func (p *Pool) cleanUnhealthy() {
...
-			if err != nil {
-				level.Warn(p.logger).Log("msg", fmt.Sprintf("removing %s failing healthcheck", p.clientName), "addr", addr, "reason", err)
-				p.RemoveClientFor(addr)
-			}
+		if !ok {
+			return nil
 		}
+
+		err := healthCheck(ctx, member.client, p.cfg.HealthCheckTimeout)
+		if err == nil {
```
tcp13equals2, line 272: Should this be if err != nil && member. firstFailedHealthCheck.IsZero then set the firstFailedHealthCheck to now?

tcp13equals2: And does the firstFailedHealthCheck need to be cleared if the health check succeeds?

charleskorn: If there was no error, then we need to reset the last time the health check failed, and then there's nothing more to do, which is what this `if` covers.

Everything after this `if` is handling the case where the health check failed.

tcp13equals2: ah yes, I see. My mistake

PR #797, opened 2025-11-25
feat(ring): add ring_zone_member metric

juliusmh:
**What this PR does**:
Adds `ring_zone_members` metric, that is exposed when the ring runs in zone-aware mode. On startup, the metric is empty. If instances from a specific zone join the ring, the list of `allRingZones` will be populated. If the last and only instance of a zone leaves `ring_zone_members` is set to `0`.

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -10,6 +10,7 @@ import (
 	"math/rand"
 	"net/http"
 	"slices"
+	"strings"
 	"sync"
 	"time"
 
@@ -32,6 +33,8 @@ const (
 	// GetBufferSize is the suggested size of buffers passed to Ring.Get(). It's based on
 	// a typical replication factor 3, plus extra room for a JOINING + LEAVING instance.
 	GetBufferSize = 5
+
+	maxZonesForMetrics = 32
 )
 
 // Options are the result of Option instances that can be used to modify Ring.GetWithOptions behavior.
@@ -265,6 +268,10 @@ type Ring struct {
 	// to be sorted alphabetically.
 	ringZones []string
 
+	// Map containing all ring zones ever discovered, even if they have no instances,
+	// capped at 32 zones to limit cardinality.
+	trackedRingZones map[string]struct{}
+
 	// Number of registered instances with tokens.
 	instancesWithTokensCount int
 
@@ -286,6 +293,7 @@ type Ring struct {
 	shuffledSubringWithLookbackCache map[subringCacheKey]cachedSubringWithLookback[*Ring]
 
 	numMembersGaugeVec      *prometheus.GaugeVec
+	numZoneMembersGaugeVec  *prometheus.GaugeVec
 	totalTokensGauge        prometheus.Gauge
 	oldestTimestampGaugeVec *prometheus.GaugeVec
 
@@ -332,6 +340,7 @@ func NewWithStoreClientAndStrategy(cfg Config, name, key string, store kv.Client
 		KVClient:                         store,
 		strategy:                         strategy,
 		ringDesc:                         &Desc{},
+		trackedRingZones:                 map[string]struct{}{},
 		shuffledSubringCache:             map[subringCacheKey]*Ring{},
 		shuffledSubringWithLookbackCache: map[subringCacheKey]cachedSubringWithLookback[*Ring]{},
 		numMembersGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
@@ -340,6 +349,11 @@ func NewWithStoreClientAndStrategy(cfg Config, name, key string, store kv.Client
 			ConstLabels: map[string]string{"name": name},
 		},
 			[]string{"state"}),
+		numZoneMembersGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name:        "ring_zone_members",
+			Help:        "Number of ring members for each zone/state pair",
+			ConstLabels: map[string]string{"name": name},
+		}, []string{"zone", "state"}),
 		totalTokensGauge: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
 			Name:        "ring_tokens_total",
 			Help:        "Number of tokens in the ring",
@@ -448,7 +462,7 @@ func (r *Ring) setRingStateFromDesc(ringDesc *Desc, updateMetrics, updateRegiste
 	r.ringTokens = ringTokens
 	r.ringTokensByZone = ringTokensByZone
 	r.ringInstanceByToken = ringInstanceByToken
-	r.ringZones = ringZones
+	r.updateRingZones(ringZones)
 	r.instancesWithTokensCount = instancesWithTokensCount
 	r.instancesCountPerZone = instancesCountPerZone
 	r.instancesWithTokensCountPerZone = instancesWithTokensCountPerZone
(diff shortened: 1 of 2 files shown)


ring/ring.go
```diff
@@ -265,6 +265,9 @@ type Ring struct {
 	// to be sorted alphabetically.
 	ringZones []string
 
+	// List of all ring zones ever discovered, even if they have no instances.
```
aknuds1: The current is currently misleading.
```suggestion
	// Map of all ring zones ever discovered, even if they have no instances.
```

ring/ring.go
```diff
@@ -785,11 +795,29 @@ func (r *Desc) CountTokens() map[string]int64 {
 	return owned
 }
 
+func (r *Ring) updateRingZones(zones []string) {
+	r.ringZones = zones
+	for _, zone := range zones {
+		if _, ok := r.allRingZones[zone]; !ok {
+			r.allRingZones[zone] = struct{}{}
```
aknuds1: I wonder whether we should e.g. warn if the number of zones is higher than a relatively low number, since we don't want high cardinality from the zone label? WDYT?

ring/ring.go
```diff
@@ -785,11 +795,29 @@ func (r *Desc) CountTokens() map[string]int64 {
...
 	numByState := map[string]int{}
 	oldestTimestampByState := map[string]int64{}
 
+	// Initialize to zero so we emit zero-metrics. Will emit nothing if no zones were discovered.
+	var numByZone map[string]int
+	if r.cfg.ZoneAwarenessEnabled {
+		numByZone = map[string]int{}
+		for zone := range r.allRingZones {
+			numByZone[zone] = 0
+		}
```
aknuds1: [Nit] We don't need to explicitly initialize the map contents, right?
```suggestion
		numByZone = make(map[string]int, len(r.allRingZones))
```

ring/ring.go
```diff
@@ -813,6 +844,11 @@ func (r *Ring) updateRingMetrics() {
 	for state, timestamp := range oldestTimestampByState {
 		r.oldestTimestampGaugeVec.WithLabelValues(state).Set(float64(timestamp))
 	}
+	if r.cfg.ZoneAwarenessEnabled {
+		for zone, count := range numByZone {
+			r.numZoneMembersGaugeVec.WithLabelValues(zone).Set(float64(count))
```
aknuds1: [Nit] Isn't this simpler?
```suggestion
		for zone := range r.allRingZones {
			r.numZoneMembersGaugeVec.WithLabelValues(zone).Set(float64(numByZone[zone]))
```

ring/ring.go
```diff
@@ -1065,14 +1101,16 @@ func (r *Ring) buildRingForTheShard(shard map[string]InstanceDesc) *Ring {
...
 
 	return &Ring{
 		cfg:                                     r.cfg,
 		strategy:                                r.strategy,
 		ringDesc:                                shardDesc,
 		ringTokens:                              shardTokens,
 		ringTokensByZone:                        shardTokensByZone,
-		ringZones:                               getZones(shardTokensByZone),
+		ringZones:                               zones,
+		allRingZones:                            getZoneMap(zones),
```
aknuds1: [Nit] Can't we just inline `getZoneMap`? It's just a few lines of code.

ring/ring.go
juliusmh: If we don't do this, we have to track instance adding/removal and now that I've added state, also instance state transitions and update the gauge. This requires two-way-diff on instances.

ring/ring.go
juliusmh: I don't think there will be that many zones.

juliusmh: I added a check to stop tracking `allRingZones` if there are more than 32 distinct zones, e.g. due to a configuration mistake. This should cap the cardinality < 160 = 32 (zone) * 5 (state).

ring/ring.go
```diff
@@ -265,6 +266,9 @@ type Ring struct {
 	// to be sorted alphabetically.
 	ringZones []string
 
+	// Map containing all ring zones ever discovered, even if they have no instances.
+	allRingZones map[string]struct{}
```
aknuds1: Considering the cap, I think it should be renamed:
```suggestion
	// Map containing all ring zones ever discovered, even if they have no instances,
	// capped at 32 zones to limit cardinality.
	trackedRingZones map[string]struct{}
```

ring/ring.go
```diff
@@ -340,6 +346,11 @@ func NewWithStoreClientAndStrategy(cfg Config, name, key string, store kv.Client
 			ConstLabels: map[string]string{"name": name},
 		},
 			[]string{"state"}),
+		numZoneMembersGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name:        "ring_zone_members",
+			Help:        "Number of ring members in each zone",
```
aknuds1:
```suggestion
			Help:        "Number of ring members for each zone/state pair",
```

ring/ring.go
```diff
@@ -785,15 +796,49 @@ func (r *Desc) CountTokens() map[string]int64 {
 	return owned
 }
 
+func (r *Ring) updateRingZones(zones []string) {
+	r.ringZones = zones
+	var notAdded []string
+	for _, zone := range zones {
+		if _, ok := r.allRingZones[zone]; !ok {
+			if len(r.allRingZones) > 32 {
```
aknuds1: It should be `>=`. I also suggest extracting a constant.
```suggestion
			if len(r.trackedRingZones) >= maxZonesForMetrics {
```

ring/ring.go
```diff
@@ -1065,14 +1122,21 @@ func (r *Ring) buildRingForTheShard(shard map[string]InstanceDesc) *Ring {
 	shardDesc := &Desc{Ingesters: shard}
 	shardTokensByZone := shardDesc.getTokensByZone()
 	shardTokens := mergeTokenGroups(shardTokensByZone)
+	zones := getZones(shardTokensByZone)
+
+	ringZoneMap := make(map[string]struct{}, len(zones))
+	for _, zone := range zones {
+		ringZoneMap[zone] = struct{}{}
```
aknuds1: Can you implement the zone count limiting and warning here too?

ring/ring.go
```diff
@@ -785,15 +799,49 @@ func (r *Desc) CountTokens() map[string]int64 {
...
+		)
+	}
+}
+
 // updateRingMetrics updates ring metrics. Caller must be holding the Write lock!
 func (r *Ring) updateRingMetrics() {
 	numByState := map[string]int{}
 	oldestTimestampByState := map[string]int64{}
 
+	// Initialize to zero so we emit zero-metrics. Will emit nothing if no zones were discovered.
```
aknuds1: I think this comment doesn't make sense here, but instead the one on line 836 does.
```suggestion
```

juliusmh: Changed it. Thanks.

ring/ring.go
```diff
@@ -785,15 +799,49 @@ func (r *Desc) CountTokens() map[string]int64 {
...
+			}
+		}
+	}
+	if len(notAdded) > 0 {
+		level.Warn(r.logger).Log(
+			"msg", "not tracking metrics for zone(s) due to high cardinality",
+			"zones", strings.Join(notAdded, ","),
+		)
+	}
+}
```
colega, line 820: I'm not sure about this:
- Does the metric still make sense if it's incomplete? 
- Who has more than 32 zones?

IMO, instead of adding extra logic, we could've just added a new flag to disable this metric if someone's really worried about the cardinality.

juliusmh: The cardinality limits is mostly to prevent disaster if there are too many zones due to e.g. a configuration mistake. This will not happen under normal circumstances.

colega: That is why I think that maintaining a piece of code just to protect from that might be an overkill. I'm afraid of maintaining a software that does 10% business logic and 90% "protect from user configuration mistakes" :) 

Anyway, I'm not blocking on this, this is just a random Friday thought.

juliusmh: I see your point. But adding a flag to disable this metric, and then making sure `r.cfg.ZoneAwarenessEnabled` and `r.cfg.PerZoneMetricsEnable` don't collide or replacing checks with `r.cfg.ZoneAwarenessEnable && r.cfg.PerZoneMetricsEnabled` is also logic.

PR #799, opened 2025-11-25
feat: add state change lock mechanism

Segflow:
**What this PR does**:

This PR adds the ability to lock partition state change. If a state change is locked we cannot change the state of a partition ring. 

This is useful if we want to deactivate a partition ring and prevent it from being activated again.

We also add these 2 metrics:
- partition_ring_partition_state_change_locked{name, partition_id, state}
- partition_ring_partition_state_change_locked_timestamp_seconds{name, partition_id, state}

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> Introduce a lock to prevent partition state changes, updating proto/model, editor, lifecycler, HTTP API/UI, watcher logging, and tests.
> 
> - **Ring Model/Proto**:
>   - Add `PartitionDesc` fields: `stateChangeLocked` and `stateChangeLockedTimestamp` (`partition_ring_desc.proto` + regenerated `pb.go`).
>   - Enforce lock in `UpdatePartitionState()` (now returns `(bool, error)`) and add `UpdatePartitionStateChangeLock()`.
>   - Merge logic propagates lock fields based on timestamp.
> - **Editor/API**:
>   - Add `SetPartitionStateChangeLock()` and helper `SetPartitionStateChangeLock(...)`.
>   - `changePartitionState()` and callers adapted to new return signature.
> - **HTTP/UI**:
>   - Surface `stateChangeLocked` in page data; disable change buttons when locked.
>   - New POST actions: `state_change_lock` (lock/unlock) and `change_state_and_lock` (change then lock).
> - **Lifecycler**:
>   - Reconcile path updated to use new `UpdatePartitionState` signature.
> - **Watcher**:
>   - Log partitions with locked state changes on updates.
> - **Tests**:
>   - Add/extend tests for lock behavior across model, editor, HTTP handler, watcher, and ring tests; update existing calls for new signatures.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fd822d074759051e553bdd8e225e194805ab3c1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ring/partition_ring_watcher.go
+++ b/ring/partition_ring_watcher.go
@@ -28,9 +28,7 @@ type PartitionRingWatcher struct {
 	ring   *PartitionRing
 
 	// Metrics.
-	numPartitionsGaugeVec          *prometheus.GaugeVec
-	partitionLockStateGaugeVec     *prometheus.GaugeVec
-	partitionLockTimestampGaugeVec *prometheus.GaugeVec
+	numPartitionsGaugeVec *prometheus.GaugeVec
 }
 
 type PartitionRingWatcherDelegate interface {
@@ -49,16 +47,6 @@ func NewPartitionRingWatcher(name, key string, kv kv.Client, logger log.Logger, 
 			Help:        "Number of partitions by state in the partitions ring.",
 			ConstLabels: map[string]string{"name": name},
 		}, []string{"state"}),
-		partitionLockStateGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
-			Name:        "partition_ring_partition_state_change_locked",
-			Help:        "Whether the partition state change is locked. 1 if locked, 0 if unlocked.",
-			ConstLabels: map[string]string{"name": name},
-		}, []string{"partition_id", "state"}),
-		partitionLockTimestampGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
-			Name:        "partition_ring_partition_state_change_locked_timestamp_seconds",
-			Help:        "Unix timestamp (seconds) of when the partition state was locked the last time. 0 if unlocked.",
-			ConstLabels: map[string]string{"name": name},
-		}, []string{"partition_id", "state"}),
 	}
 
 	r.Service = services.NewBasicService(r.starting, r.loop, nil).WithName("partitions-ring-watcher")
@@ -120,26 +108,13 @@ func (w *PartitionRingWatcher) updatePartitionRing(desc *PartitionRingDesc) {
 		w.numPartitionsGaugeVec.WithLabelValues(state.CleanName()).Set(float64(count))
 	}
 
-	// Update partition lock metrics.
-	w.partitionLockStateGaugeVec.Reset()
-	w.partitionLockTimestampGaugeVec.Reset()
+	// Check partitions whose state change is locked and log them.
 	for partitionID, partition := range desc.Partitions {
 		state := partition.GetState().CleanName()
 		partitionIDStr := strconv.Itoa(int(partitionID))
-
-		// Lock state indicator (0 or 1)
-		lockState := 0.0
 		if partition.StateChangeLocked {
-			lockState = 1.0
+			level.Warn(w.logger).Log("msg", "partition state change is locked", "partition_id", partitionIDStr, "partition_state", state)
 		}
-		w.partitionLockStateGaugeVec.WithLabelValues(partitionIDStr, state).Set(lockState)
-
-		// Lock timestamp (0 if unlocked)
-		lockTimestamp := 0.0
-		if partition.StateChangeLocked {
-			lockTimestamp = float64(partition.StateChangeLockedTimestamp)
-		}
-		w.partitionLockTimestampGaugeVec.WithLabelValues(partitionIDStr, state).Set(lockTimestamp)
 	}
 }
 
--- a/ring/partition_ring_watcher_test.go
+++ b/ring/partition_ring_watcher_test.go
@@ -78,12 +78,6 @@ func TestPartitionRingWatcher_ShouldWatchUpdates(t *testing.T) {
 	}, time.Second, 10*time.Millisecond)
 
 	assert.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(`
-		# HELP partition_ring_partition_state_change_locked Whether the partition state change is locked. 1 if locked, 0 if unlocked.
-		# TYPE partition_ring_partition_state_change_locked gauge
-		partition_ring_partition_state_change_locked{name="test",partition_id="1",state="Active"} 0
-		# HELP partition_ring_partition_state_change_locked_timestamp_seconds Unix timestamp (seconds) of when the partition state change lock was last modified. 0 if unlocked.
-		# TYPE partition_ring_partition_state_change_locked_timestamp_seconds gauge
-		partition_ring_partition_state_change_locked_timestamp_seconds{name="test",partition_id="1",state="Active"} 0
 		# HELP partition_ring_partitions Number of partitions by state in the partitions ring.
 		# TYPE partition_ring_partitions gauge
 		partition_ring_partitions{name="test",state="Pending"} 0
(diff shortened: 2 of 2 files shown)


ring/partition_ring_model.go
```diff
@@ -344,6 +366,13 @@ func (m *PartitionRingDesc) mergeWithTime(mergeable memberlist.Mergeable, localC
 				thisPart.State = otherPart.State
 				thisPart.StateTimestamp = otherPart.StateTimestamp
 			}
+
```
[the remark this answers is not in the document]
Segflow: This is expected, Merge is used on gossip protocol, if another ring member have a more up to date information about `StateChangeLocked` we use that regardless of any local state.

ring/partition_ring_watcher.go
```diff
@@ -106,6 +119,26 @@ func (w *PartitionRingWatcher) updatePartitionRing(desc *PartitionRingDesc) {
...
+		}
+		w.partitionLockStateGaugeVec.WithLabelValues(partitionIDStr, state).Set(lockState)
+
+		// Lock timestamp (0 if unlocked)
+		lockTimestamp := 0.0
+		if partition.StateChangeLocked {
+			lockTimestamp = float64(partition.StateChangeLockedTimestamp)
+		}
+		w.partitionLockTimestampGaugeVec.WithLabelValues(partitionIDStr, state).Set(lockTimestamp)
+	}
```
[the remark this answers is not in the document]
Segflow: Check again, I reset the two metrics before the loop, should be OK for these metrics.

ring/partition_ring_editor.go
```diff
@@ -38,6 +38,12 @@ func (l *PartitionRingEditor) RemoveMultiPartitionOwner(ctx context.Context, ins
 	})
 }
 
+func (l *PartitionRingEditor) LockPartitionStateChange(ctx context.Context, partitionID int32, locked bool) error {
```
pracucci: [nit] The name `LockPartitionStateChange()` makes me think this is used for locking, but it's actually used to unlock too. I would change it something that doesn't imply the "lock" operation, like `SetPartitionStateChangeLock()` (just an example to give you the idea).

ring/partition_ring_http.go
```diff
@@ -29,6 +29,7 @@ var partitionRingPageTemplate = template.Must(template.New("webpage").Funcs(temp
 
 type PartitionRingUpdater interface {
 	ChangePartitionState(ctx context.Context, partitionID int32, toState PartitionState) error
+	LockPartitionStateChange(ctx context.Context, partitionID int32, locked bool) error
```
pracucci: [nit] Same comment about the function name here.

ring/partition_ring_http.go
```diff
@@ -147,6 +149,50 @@ func (h *PartitionRingPageHandler) handlePostRequest(w http.ResponseWriter, req
...
+		if err := h.updater.ChangePartitionState(req.Context(), int32(partitionID), PartitionState(toState)); err != nil {
+			http.Error(w, fmt.Sprintf("failed to change partition state: %s", err.Error()), http.StatusBadRequest)
+			return
+		}
+
+		if err := h.updater.LockPartitionStateChange(req.Context(), int32(partitionID), true); err != nil {
+			http.Error(w, fmt.Sprintf("failed to lock partition state change: %s", err.Error()), http.StatusBadRequest)
+			return
+		}
+	} else if req.FormValue("action") == "lock_state_change" {
```
pracucci: [nit] For the same reason, I would call this action `change_lock`

ring/partition_ring_http_test.go
```diff
@@ -246,4 +246,39 @@ func TestPartitionRingPageHandler_ChangePartitionState(t *testing.T) {
 		require.Equal(t, PartitionInactive, getPartitionStateFromStore(t, store, ringKey, 1))
 		require.Equal(t, PartitionActive, getPartitionStateFromStore(t, store, ringKey, 2))
 	})
+	t.Run("should also lock the state change if 'change_state_and_lock' is used", func(t *testing.T) {
```
pracucci, line 296: [nit] Do you mind adding a test case for `lock_state_change` too?

ring/partition_ring_status.gohtml
```diff
@@ -50,12 +50,21 @@
...
                     {{ if and (not .Corrupted) (ne (index $stateChanges .State) 0) }}
                         {{ $toState := index $stateChanges .State }}
-                        <form action="" method="POST" onsubmit="return confirm('Do you confirm you want to change the state to {{ $toState.CleanName }}?');">
+                        <form action="" method="POST">
                             <input type="hidden" name="csrf_token" value="$__CSRF_TOKEN_PLACEHOLDER__">
                             <input type="hidden" name="partition_id" value="{{ .ID }}">
                             <input type="hidden" name="partition_state" value="{{ $toState.String }}">
 
-                            <button name="action" value="change_state" type="submit">Change state to {{ $toState.CleanName }}</button>
+                            <button name="action" value="change_state" type="submit" {{ if .StateChangeLocked }}disabled title="State change is locked"{{ end }} onclick="return confirm('Do you confirm you want to change the state to {{ $toState.CleanName }}?');">Change state to {{ $toState.CleanName }}</button>
```
pracucci, line 58: Does the `onclick` here prevents the change if canceled? I'm wondering if the safest option would just be creating a `<form>` for each different action (in this case it would be 1 more form).

ring/partition_ring_watcher.go
```diff
@@ -46,6 +49,16 @@ func NewPartitionRingWatcher(name, key string, kv kv.Client, logger log.Logger,
...
+		partitionLockStateGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name:        "partition_ring_partition_state_change_locked",
+			Help:        "Whether the partition state change is locked. 1 if locked, 0 if unlocked.",
+			ConstLabels: map[string]string{"name": name},
+		}, []string{"partition_id", "state"}),
+		partitionLockTimestampGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name:        "partition_ring_partition_state_change_locked_timestamp_seconds",
+			Help:        "Unix timestamp (seconds) of when the partition state change lock was last modified. 0 if unlocked.",
+			ConstLabels: map[string]string{"name": name},
+		}, []string{"partition_id", "state"}),
```
pracucci: These metrics are going to be high cardinality because exposed for each partition by each pod. Are they really required? Looks like useful only once in a lifetime (maybe when investigating an incident) and maybe can be replaced by a log when the lock change? It could be logged here by the watcher.

ring/partition_ring_watcher.go
```diff
@@ -46,6 +49,16 @@ func NewPartitionRingWatcher(name, key string, kv kv.Client, logger log.Logger,
...
 			ConstLabels: map[string]string{"name": name},
 		}, []string{"state"}),
+		partitionLockStateGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name:        "partition_ring_partition_state_change_locked",
+			Help:        "Whether the partition state change is locked. 1 if locked, 0 if unlocked.",
+			ConstLabels: map[string]string{"name": name},
+		}, []string{"partition_id", "state"}),
+		partitionLockTimestampGaugeVec: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name:        "partition_ring_partition_state_change_locked_timestamp_seconds",
+			Help:        "Unix timestamp (seconds) of when the partition state change lock was last modified. 0 if unlocked.",
```
pracucci:
```suggestion
			Help:        "Unix timestamp (seconds) of when the partition state was locked the last time. 0 if unlocked.",
```

ring/partition_ring_status.gohtml
Segflow: Yes the onlick prevent the change if cancelled, the browser will just not send the request.

ring/partition_ring_model.go
```diff
@@ -344,6 +366,13 @@ func (m *PartitionRingDesc) mergeWithTime(mergeable memberlist.Mergeable, localC
...
+			if otherPart.StateChangeLockedTimestamp > thisPart.StateChangeLockedTimestamp {
+				changed = true
+
+				thisPart.StateChangeLocked = otherPart.StateChangeLocked
+				thisPart.StateChangeLockedTimestamp = otherPart.StateChangeLockedTimestamp
+			}
```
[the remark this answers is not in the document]
Segflow: I believe it's not needed in this case.

ring/partition_ring_watcher.go
Segflow: I preferred metrics so I can have a dashboard likes the one we have for flux [ignore dashboard](https://ops.grafana-ops.net/d/RUEClN37k/loki-team?orgId=1&from=now-24h&to=now&timezone=utc&var-Data_source_for_flux_ignore=2z9d6ElGk&viewPanel=panel-44) 
But since this is only an emergency lever, I agree that this high cardinality metric is not necessary, I replaced it with a simple warn log:
```
level.Warn(w.logger).Log("msg", "partition state change is locked", "partition_id", partitionIDStr, "partition_state", state)
```

ring/partition_ring_watcher.go
```diff
@@ -106,6 +107,15 @@ func (w *PartitionRingWatcher) updatePartitionRing(desc *PartitionRingDesc) {
 	for state, count := range desc.countPartitionsByState() {
 		w.numPartitionsGaugeVec.WithLabelValues(state.CleanName()).Set(float64(count))
 	}
+
+	// Check partitions whose state change is locked and log them.
+	for partitionID, partition := range desc.Partitions {
+		state := partition.GetState().CleanName()
+		partitionIDStr := strconv.Itoa(int(partitionID))
+		if partition.StateChangeLocked {
+			level.Warn(w.logger).Log("msg", "partition state change is locked", "partition_id", partitionIDStr, "partition_state", state)
```
pracucci, line 116: This is extremely verbose. Why not just logging when it has changed compared to the `oldRing`?

PR #800, opened 2025-11-26
feat: skip memberlist zone-aware routing if there is a zone with members but no alive bridges

pracucci:
**What this PR does**:

When zone-aware routing is enabled, if there is a zone with members but no alive bridges we end up in a network partitioning state. I think we can make the routing logic more resilient by checking for this specific condition, and internally disable the zone-aware routing if we detect a zone without alive bridges.

The new logic is definitely slower than the previous one, but in absolute terms I don't think it's prohibitive. `SelectNodes()` gets called once every gossip interval (200ms) and takes 0.17ms with 10K nodes and 2 zones (10K nodes is a pretty large scale):

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
BenchmarkZoneAwareNodeSelectionDelegate_SelectNodes-11    	    7033	    170166 ns/op	   81922 B/op	       1 allocs/op
BenchmarkZoneAwareNodeSelectionDelegate_SelectNodes-11    	    6940	    170834 ns/op	   81922 B/op	       1 allocs/op
BenchmarkZoneAwareNodeSelectionDelegate_SelectNodes-11    	    6925	    171649 ns/op	   81922 B/op	       1 allocs/op
BenchmarkZoneAwareNodeSelectionDelegate_SelectNodes-11    	    6906	    213222 ns/op	   81922 B/op	       1 allocs/op
BenchmarkZoneAwareNodeSelectionDelegate_SelectNodes-11    	    6957	    172906 ns/op	   81922 B/op	       1 allocs/op
PASS
ok  	github.com/grafana/dskit/kv/memberlist	9.339s
```

This PR is based on https://github.com/grafana/memberlist/pull/13.

N/A

**Checklist**
- [x] Tests updated

### DIFF
--- a/kv/memberlist/node_zone_aware_routing.go
+++ b/kv/memberlist/node_zone_aware_routing.go
@@ -97,21 +97,17 @@ type zoneAwareNodeSelectionDelegate struct {
 
 // newZoneAwareNodeSelectionDelegate creates a new zone-aware node selection delegate.
 func newZoneAwareNodeSelectionDelegate(localRole NodeRole, localZone string, logger log.Logger, registerer prometheus.Registerer) *zoneAwareNodeSelectionDelegate {
-	const subsystem = "memberlist_client_zone_aware_routing"
-
 	return &zoneAwareNodeSelectionDelegate{
 		localRole: localRole,
 		localZone: localZone,
 		logger:    logger,
 		selectNodesCalls: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
-			Subsystem: subsystem,
-			Name:      "select_nodes_total",
-			Help:      "Total number of times memberlist attempted to select node candidates for gossiping (tracked only when when zone-aware routing is enabled).",
+			Name: "memberlist_client_zone_aware_routing_select_nodes_total",
+			Help: "Total number of times memberlist attempted to select node candidates for gossiping (tracked only when when zone-aware routing is enabled).",
 		}),
 		selectNodesCallsSkipped: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
-			Subsystem: subsystem,
-			Name:      "select_nodes_skipped_total",
-			Help:      "Total number of times memberlist zone-aware routing was skipped because the local zone is unknown or a zone has no alive bridges.",
+			Name: "memberlist_client_zone_aware_routing_select_nodes_skipped_total",
+			Help: "Total number of times memberlist zone-aware routing was skipped because the local zone is unknown or a zone has no alive bridges.",
 		}),
 	}
 }


kv/memberlist/node_zone_aware_routing.go
```diff
@@ -96,18 +98,82 @@ func newZoneAwareNodeSelectionDelegate(localRole NodeRole, localZone string, log
...
+				}
+			}
+		}
+	}
+
+	// Skip zone-aware routing if any zone has members but no alive bridges.
+	// This prevents network partitioning when bridges are missing or dead.
+	for _, zone := range zonesWithMembers {
+		if !slices.Contains(zonesWithAliveBridges, zone) {
+			return nodes, nil
```
colega, line 179: Maybe log a warning here?

kv/memberlist/node_zone_aware_routing.go
```diff
@@ -96,18 +98,82 @@ func newZoneAwareNodeSelectionDelegate(localRole NodeRole, localZone string, log
...
+	if d.localZone == "" {
+		return nodes, nil
+	}
+
+	// Pre-allocate backing arrays on the stack for up to 5 zones (common case).
+	// Slices will grow on the heap only if more zones are encountered.
+	var zonesWithMembersBacking [5]string
+	zonesWithMembers := zonesWithMembersBacking[:0]
+	var zonesWithAliveBridgesBacking [5]string
+	zonesWithAliveBridges := zonesWithAliveBridgesBacking[:0]
```
colega: I was wondering why the compiler wouldn't optimise this... and I had to check, and turns out this stuff we do is unnecessary. [The generated code is the same](https://godbolt.org/z/zE8Thvozj)

```suggestion
	// Pre-allocate backing arrays on the stack for up to 5 zones (common case).
	zonesWithMembers := make([]string, 0, 5)
	zonesWithAliveBridges := make([]string, 0, 5)
```

colega (approved):

kv/memberlist/node_zone_aware_routing.go
pracucci: You're right, thanks.

kv/memberlist/node_zone_aware_routing.go
pracucci: Good point. Yes, I will add a log, but I think I should add a metric too, so we can alert on it.

kv/memberlist/node_zone_aware_routing.go
pracucci: Done in https://github.com/grafana/dskit/pull/800#:~:text=Verified-,f9d9c4a,-Merge%20info

kv/memberlist/node_zone_aware_routing.go
pracucci: Log added in https://github.com/grafana/dskit/pull/800#:~:text=Verified-,f9d9c4a,-Merge%20info

pracucci: And metrics added in https://github.com/grafana/dskit/pull/800/commits/59851035bbc81867369f27bc3839fa0f999c283d

kv/memberlist/node_zone_aware_routing.go
```diff
@@ -85,29 +89,111 @@ type zoneAwareNodeSelectionDelegate struct {
...
+		selectNodesCalls: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      "select_nodes_total",
+			Help:      "Total number of times memberlist attempted to select node candidates for gossiping (tracked only when when zone-aware routing is enabled).",
+		}),
+		selectNodesCallsSkipped: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      "select_nodes_skipped_total",
+			Help:      "Total number of times memberlist zone-aware routing was skipped because the local zone is unknown or a zone has no alive bridges.",
+		}),
```
colega: Nit: I would use the full metric name in `Name:` and [avoid using Namespace/Subsystem](https://github.com/prometheus/client_golang/issues/240#issuecomment-539743118).

(Not only because that comment says, but because I think that grepping for full metric name is much easier than having to find concatenated stuff).

pracucci: I didn't even notice claude did use Subsystem (I'm doing 2 project at the same time...). Fixing now.

pracucci: Manually fixed in https://github.com/grafana/dskit/pull/800/commits/b2f8a75a0799340496123a91e2b8a107a581714c

colega: Awesome thanks

PR #811, opened 2025-11-26
chore(memberlist): use a more clever zone lookup in SelectNodes

colega:
This is a follow up on #800

`slices.Contains` is the slowest part of the new `SelectNodes()`, so I thought: maybe we can do better?

Realistically, most of the setups will have zone-[abc], or similar, and after checking a couple of nodes we'll already have all three zones in the slice, so if we managed to keep the slice sorted, we can check the expected index in the slice first.

This does a 15% decrease in cpu time for `BenchmarkZoneAwareNodeSelectionDelegate_SelectNodes`

```
$ benchstat pr-baseline contains-zone
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                                              │ pr-baseline │           contains-zone            │
                                              │   sec/op    │   sec/op     vs base               │
ZoneAwareNodeSelectionDelegate_SelectNodes-12   178.2µ ± 4%   151.5µ ± 1%  -14.96% (p=0.002 n=6)

                                              │ pr-baseline  │         contains-zone         │
                                              │     B/op     │     B/op      vs base         │
ZoneAwareNodeSelectionDelegate_SelectNodes-12   80.00Ki ± 0%   80.00Ki ± 0%  ~ (p=1.000 n=6)

                                              │ pr-baseline │         contains-zone         │
                                              │  allocs/op  │ allocs/op   vs base           │
ZoneAwareNodeSelectionDelegate_SelectNodes-12    1.000 ± 0%   1.000 ± 0%  ~ (p=1.000 n=6) ¹
```

### DIFF
--- a/kv/memberlist/node_zone_aware_routing.go
+++ b/kv/memberlist/node_zone_aware_routing.go
@@ -145,13 +145,15 @@ func (d *zoneAwareNodeSelectionDelegate) SelectNodes(nodes []*memberlist.NodeSta
 			if remoteRole == NodeRoleBridge {
 				// Only count alive bridges.
 				if node.State == memberlist.StateAlive {
-					if !slices.Contains(zonesWithAliveBridges, remoteZone) {
+					if !containsZone(zonesWithAliveBridges, remoteZone) {
 						zonesWithAliveBridges = append(zonesWithAliveBridges, remoteZone)
+						slices.Sort(zonesWithAliveBridges)
 					}
 				}
 			} else {
-				if !slices.Contains(zonesWithMembers, remoteZone) {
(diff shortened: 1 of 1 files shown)


memberlist.test
pracucci, line 1: Please remove this file

cpu.pprof
pracucci, line 1: Remove this file

memberlist.test
colega: Oh 🤦 Removed and force-pushed.

cpu.pprof
colega: Oh 🤦 Removed and force-pushed.

PR #837, opened 2025-12-01
ci: configure Renovate more extensively

aknuds1:
**What this PR does**:

Configure Renovate more extensively, inspired by the Mimir one. Importantly, this increases the concurrent PR limit to 20, from our default of 5.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated


renovate.json
```diff
@@ -0,0 +1,55 @@
...
+      "description": "Only update dependencies digests on Monday mornings to avoid PR spam from repos not using semver",
+      "matchUpdateTypes": [
+        "digest"
+      ],
+      "schedule": ["before 5am on Monday"]
+    }
+  ],
+  "branchPrefix": "deps-update/",
+  "labels": [
+    "dependency-update"
```
juliusmh, line 39: I think in Mimir we use `dependencies` and `dependency-update` ?

aknuds1: Having a look.

aknuds1: I compared configurations, they are the same as in Mimir?

PR #845, opened 2025-12-04
fix: avoid memory leaks when soft-deleting tenant registries

### LINKED ISSUE #746: Memory leak when repeatedly creating and soft-removing tenant registries
#### Description

When [soft-removing a tenant registry](https://github.com/grafana/dskit/blob/d6a51ec2b8c9331eb59e29629661d94324c00a72/metrics/tenant_registries.go#L703-L737), we gather metrics, add them to the `lastGather` field (minus gauges and unknowns), and [keep the registry](https://github.com/grafana/dskit/blob/d6a51ec2b8c9331eb59e29629661d94324c00a72/metrics/tenant_registries.go#L695-L697) in the `TenantRegistries` struct.

These metrics are gathered and kept in memory each time we create and soft-remove a registry. Repeatedly creating and deleting registries for the same tenants will cause [`TenantRegistries.regs` field](https://github.com/grafana/dskit/blob/d6a51ec2b8c9331eb59e29629661d94324c00a72/metrics/tenant_registries.go#L642) in the `TenantRegistries` to grow indefinitely.

#### Testing

I used 4 Mimir Alertmanager instances:
- 2 with hard deletes
- 2 with soft deletes

I created a script to simulate 2.5K tenants being added and removed every few minutes.

<details>
<summary>Memory usage on each Alertmanager.</summary>

Image

</details>

#### Proposed solution

We can use a separate `TenantRegistry` to "archive" gathered metrics for all soft-removed tenants, merging/deduplicating them to avoid unnecessary memory usage (I'm not sure how feasible this is).

santihernandezc:
### Description

This PR fixes a [memory leak when soft-deleting tenant registries](https://github.com/grafana/dskit/issues/746) by creating a special `"archived"` registry in the `TenantRegistries` struct.

Instead of keeping user registries around indefinitely after soft deletion, the `TenantRegistries` struct now copies counters, histograms, and summaries into the `archived` registry. This way, memory growth is not unbounded if we need to delete and recreate the same user registry multiple times.

### Testing

I tested this PR with two Mimir Alertmanager instances:
- One from the latest `main`
- Another one importing this `dskit` branch

Both Instances had _strict initialization_ enabled with a grace period of 10 seconds to allow for fast cycles of creating and deleting tenant registries.

I created a script to send bursts of requests for 2.5K tenants to both instances and sleep for 60 seconds. This resulted in registries for these 2.5K tenants being created and deleted every minute.

Results:
- The instance vendoring this fix maintained a steady memory usage
- The instance built from Mimir's latest `main` accumulated more memory on each soft deletion

fix

Fixes https://github.com/grafana/dskit/issues/746

### 

---

> [!NOTE]
> **Medium Risk**
> Changes how soft-deleted tenant registries are retained and aggregated, which can affect scrape-time metric totals and label-series merging if the deduplication/merge logic is incorrect. Concurrency and Prometheus metric-type semantics (histogram buckets/summary quantiles) make this moderately risky despite added tests.
> 
> **Overview**
> Prevents memory growth from soft-deleted tenant registries by **removing** the old per-tenant registry objects and **archiving** only cumulative metric types (counters/histograms/summaries) into a shared `archived` metric map (gauges are intentionally excluded).
> 
> When archiving, metrics are **deduplicated by label set** and merged so values accumulate across tenant lifecycles, and `BuildMetricFamiliesPerTenant()` now includes the archived metrics as an empty-tenant entry while snapshotting `regs`/`archived` under one lock to avoid scrape races.
> 
> Adds end-to-end tests to ensure archived series don’t grow when tenants are repeatedly cycled and that aggregated totals never decrease across add/observe/soft-remove cycles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81951b26f2e75a19859a3d68742c60fb9d8d5eb2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/metrics/tenant_registries.go
+++ b/metrics/tenant_registries.go
@@ -640,12 +640,16 @@ type TenantRegistries struct {
 
 	regsMu sync.Mutex
 	regs   []TenantRegistry
+
+	// archived holds metrics for soft-deleted registries.
+	archived MetricFamilyMap
 }
 
 // NewTenantRegistries makes new TenantRegistries.
 func NewTenantRegistries(logger log.Logger) *TenantRegistries {
 	return &TenantRegistries{
-		logger: logger,
+		logger:   logger,
+		archived: MetricFamilyMap{},
 	}
 }
 
@@ -663,13 +667,9 @@ func (r *TenantRegistries) AddTenantRegistry(tenant string, reg *prometheus.Regi
 			continue
 		}
 
-		if r.softRemoveTenantRegistry(&r.regs[idx]) {
-			// Keep it.
-			idx++
-		} else {
-			// Remove it.
-			r.regs = append(r.regs[:idx], r.regs[idx+1:]...)
-		}
+		// Archive and remove the registry.
+		r.archiveTenantRegistry(&r.regs[idx])
+		r.regs = append(r.regs[:idx], r.regs[idx+1:]...)
 	}
 
 	// New registries must be added to the end of the list, to guarantee stability.
@@ -682,7 +682,7 @@ func (r *TenantRegistries) AddTenantRegistry(tenant string, reg *prometheus.Regi
 // RemoveTenantRegistry removes all Prometheus registries for a given tenant.
 // If hard is true, registry is removed completely.
 // If hard is false, the latest registry values are preserved for future aggregations.
-func (r *TenantRegistries) RemoveTenantRegistry(tenant string, hard bool) {
+func (r *TenantRegistries) RemoveTenantRegistry_Old(tenant string, hard bool) {
 	r.regsMu.Lock()
 	defer r.regsMu.Unlock()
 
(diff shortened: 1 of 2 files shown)


pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

metrics/tenant_registries.go
```diff
@@ -692,21 +689,21 @@ func (r *TenantRegistries) RemoveTenantRegistry(tenant string, hard bool) {
...
 	}
 }
 
-// Returns true, if we should keep latest metrics. Returns false if we failed to gather latest metrics,
-// and this can be removed completely.
-func (r *TenantRegistries) softRemoveTenantRegistry(ur *TenantRegistry) bool {
+// archiveTenantRegistry gathers and stores counters/histograms/summaries from a tenant
+// registry being removed. Gauges are excluded since they represent current state, not
+// cumulative values. Archived metrics are aggregated to avoid counter resets.
+func (r *TenantRegistries) archiveTenantRegistry(ur *TenantRegistry) {
```
moustafab, line 702: The change here to make this essentially infallible might result in some strange behaviors if we fail to gather the metrics. I'm not sure that is completely safe. In the old model we'd remove if we failed to soft remove so perhaps this is still equivalent. Have you considered this?

moustafab (commented):
I only really have one concern. This looks like a sane and useful change to me and if we can confirm the "fail to gather on archive" behavior then I think we're good to go.

metrics/tenant_registries.go
santihernandezc: Yep, this is still the same behavior as before.
If we fail to gather the metrics in the [current code](https://github.com/grafana/dskit/blob/ce4dec6f437abd02d37be0403af259d28c6f6ae7/metrics/tenant_registries.go#L706-L710), we return `false`. This causes the registry to be [removed](https://github.com/grafana/dskit/blob/ce4dec6f437abd02d37be0403af259d28c6f6ae7/metrics/tenant_registries.go#L669-L672) (same as hard-deleted), so it would cause counter restarts.

moustafab (approved):

PR #853, opened 2025-12-09
feat(cache): add optional error-returning for GetMulti

t00mas:
**What this PR does**:

Adds a new `WithErrorOut(*error)` option for `GetMulti` that allows callers to observe errors that are otherwise only logged/tracked internally. This is useful when you need visibility into cache failures (e.g., memcached unreachable) without changing the existing return signature.

Usage:

```go
var cacheErr error
results := client.GetMulti(ctx, keys, cache.WithErrorOut(&cacheErr))
if cacheErr != nil {
    // handle error
}
```

Added this way so as to not make it a breaking change.

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> Adds a `WithErrorOut(*error)` option and plumbs it through memcached `GetMulti` to allow callers to observe backend errors without changing the return type.
> 
> - **Cache API**:
>   - Add `Options.ErrOut` and `WithErrorOut(*error)` to capture errors from cache operations.
> - **Memcached client**:
>   - Update `toMemcacheOptions` to return `(*Options, []memcache.Option)`.
>   - `GetMulti(...)` clears `ErrOut` at start and sets it on error; passes memcache options downstream.
> - **Tests**:
>   - Add tests covering `WithErrorOut` behavior (error propagation, no-error case, clearing after success).
>   - Extend mock backend to simulate errors via `mockMemcachedClientBackend.err`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d1fabf038b57c93de07985bf18a62466a0d2c168. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cache/memcached_client.go
+++ b/cache/memcached_client.go
@@ -417,6 +417,9 @@ func (c *MemcachedClient) GetMulti(ctx context.Context, keys []string, opts ...O
 	}
 
 	options, mcOpts := toMemcacheOptions(opts...)
+	if options.ErrOut != nil {
+		*options.ErrOut = nil
+	}
 
 	c.metrics.requests.Add(float64(len(keys)))
 	batches, err := c.getMultiBatched(ctx, keys, mcOpts...)
--- a/cache/memcached_client_test.go
+++ b/cache/memcached_client_test.go
@@ -182,6 +182,25 @@ func TestMemcachedClient_GetMulti(t *testing.T) {
 		require.Equal(t, map[string][]byte{"foo": []byte("bar")}, res)
 		require.NoError(t, cacheErr)
 	})
+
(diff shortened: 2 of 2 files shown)


t00mas:
> > Main use case is to know if the backend is temporarily unavailable (like a pod restart).
> 
> Right, I get that but to what end? Is there circuit breaking going on or something?

Not 100% but sort of, for us it sits in the critical path and knowing the difference between a key not found or an unreachable backend enables decision making.

In the same critical path of GetMulti we have Delete and that actually return errors, so we're using those errors as a workaround at the moment.

PR #856, opened 2025-12-12
feat: add LRU cache to partition ring shuffle shard cache

grobinson-grafana:
This prevents unbounded growth of the shuffle shard cache.

**What this PR does**:

This PR adds optional LRU-based caching to the partition ring shuffle shard cache.

The cache type is selected based on size parameter:
- size <= 0: Uses unbounded map-based cache (existing behavior)
- size > 0: Uses LRU cache with the specified maximum size

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> Adds an optional LRU-backed shuffle-shard cache to PartitionRing, introduces configurable options, and updates constructors to return errors with corresponding changes across watcher and tests.
> 
> - **Ring core**:
>   - Introduces `PartitionRingOptions`/`DefaultPartitionRingOptions` and `NewPartitionRingWithOptions()`; `NewPartitionRing()` now returns `(*PartitionRing, error)`.
>   - Refactors shuffle-shard cache to pluggable storage with generic interface; supports unbounded map or LRU (`github.com/hashicorp/golang-lru/v2`).
>   - Updates cache APIs and callers (`ShuffleShard`, `ShuffleShardWithLookback`) to use new cache layer.
> - **Watcher**:
>   - `PartitionRingWatcher` now handles constructor failures; `updatePartitionRing()` returns error; `starting()`/`loop()` propagate errors.
> - **Tests/benchmarks**:
>   - Update usages to handle new constructor signatures and options; add tests for map vs LRU cache and storage behavior; adjust concurrency/cache assertions and helpers.
> - **Dependencies**:
>   - Adds `github.com/hashicorp/golang-lru/v2`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cd1e6a02d273452ea89c397a289977ae2e0ba5d6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ring/partition_ring_test.go
+++ b/ring/partition_ring_test.go
@@ -1011,8 +1011,8 @@ func TestPartitionRing_ShuffleShardWithLookback_CachingConcurrency(t *testing.T)
 			wg.Wait()
 
 			// Ensure the cache was populated.
-			assert.Greater(t, ring.shuffleShardCache.lenWithLookback(), 0)
-			assert.Equal(t, 0, ring.shuffleShardCache.lenWithoutLookback())
+			assert.Greater(t, ring.shuffleShardCache.cacheWithLookback.len(), 0)
+			assert.Equal(t, 0, ring.shuffleShardCache.cacheWithoutLookback.len())
 		})
 	}
 }
--- a/ring/partition_ring_watcher.go
+++ b/ring/partition_ring_watcher.go
@@ -38,7 +38,7 @@ type PartitionRingWatcherDelegate interface {
 func NewPartitionRingWatcher(name, key string, kv kv.Client, logger log.Logger, reg prometheus.Registerer) *PartitionRingWatcher {
 	emptyRing, err := NewPartitionRing(*NewPartitionRingDesc())
 	if err != nil {
-		panic(err) // This should never executes since we're using an empty descriptor.
+		panic(err) // This should never executes.
 	}
 	r := &PartitionRingWatcher{
 		key:    key,
--- a/ring/partitions_ring_shuffle_shard_cache.go
+++ b/ring/partitions_ring_shuffle_shard_cache.go
@@ -130,20 +130,6 @@ func (r *partitionRingShuffleShardCache) getSubringWithLookback(identifier strin
 	return cached.subring
 }
 
-func (r *partitionRingShuffleShardCache) lenWithLookback() int {
-	r.mtx.RLock()
-	defer r.mtx.RUnlock()
-
-	return r.cacheWithLookback.len()
-}
-
-func (r *partitionRingShuffleShardCache) lenWithoutLookback() int {
-	r.mtx.RLock()
-	defer r.mtx.RUnlock()
-
-	return r.cacheWithoutLookback.len()
-}
-
 // mapCacheStorage is a generic map-based implementation of shuffleShardCacheStorage.
 // Note: This implementation does not have its own mutex because thread-safety is guaranteed
 // by the mutex in partitionRingShuffleShardCache.


ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -1,22 +1,33 @@
...
+	cacheWithoutLookback *lru.Cache[subringCacheKey, *PartitionRing]
+	cacheWithLookback    *lru.Cache[subringCacheKey, cachedSubringWithLookback[*PartitionRing]]
 }
 
-func newPartitionRingShuffleShardCache() *partitionRingShuffleShardCache {
-	return &partitionRingShuffleShardCache{
-		cacheWithoutLookback: map[subringCacheKey]*PartitionRing{},
-		cacheWithLookback:    map[subringCacheKey]cachedSubringWithLookback[*PartitionRing]{},
+func newPartitionRingShuffleShardCache() (*partitionRingShuffleShardCache, error) {
+	cacheWithoutLookback, err := lru.New[subringCacheKey, *PartitionRing](128)
```
grobinson-grafana: `128` chosen for no reason other than to give it a value. I suspect we want this to be configurable via flags.

ring/partition_instance_ring_test.go
```diff
@@ -204,7 +205,8 @@ func BenchmarkPartitionInstanceRing_GetReplicationSetsForOperation(b *testing.B)
 		}
 	}
 
-	r := NewPartitionInstanceRing(newStaticPartitionRingReader(NewPartitionRing(*partitionsRing)), instancesRing, time.Hour)
+	partRing, _ := NewPartitionRing(*partitionsRing)
```
pracucci: [nit] Can you check the error here and below too, please?

ring/partition_ring.go
```diff
@@ -43,15 +43,46 @@ type PartitionRing struct {
...
+
+// DefaultPartitionRingOptions returns the default options for creating a PartitionRing.
+func DefaultPartitionRingOptions() PartitionRingOptions {
+	return PartitionRingOptions{
+		ShuffleShardCacheSize: DefaultShuffleShardCacheSize,
+	}
+}
+
+// NewPartitionRing creates a new PartitionRing with default options.
+func NewPartitionRing(desc PartitionRingDesc) (*PartitionRing, error) {
```
pracucci, line 62: This is a silent breaking change.

The previous behaviour was "unlimited cache", and the new behaviour is "limit to 128 items". A breaking change should come with breaking the API (function parameter) too, so that people that will integrate the change in their own repos (e.g. Mimir, Tempo, ...) will have a clear signal that a breaking change has been introduced and they need to address it.

I think this design suffered too much overthinking. I think we should just have 1 function `NewPartitionRing()` taking the shuffle shard cache size too. No built-in default should exist in a general purpose library like dskit, because the default value really depends on the use case.

I also think we need to support a value <= 0 to configure an unlimited cache. When the cache is unlimited, the previous behaviour could be restored (just use the old good map instead of the LRU). We can just have the cache being an interface, and 2 implementation: LRU and map. It shouldn't take longer than few lines of code changes.

ring/partition_ring_watcher.go
```diff
@@ -37,11 +37,13 @@ type PartitionRingWatcherDelegate interface {
 }
 
 func NewPartitionRingWatcher(name, key string, kv kv.Client, logger log.Logger, reg prometheus.Registerer) *PartitionRingWatcher {
+	// Create an empty partition ring. This should never fail since we're using an empty descriptor.
+	emptyRing, _ := NewPartitionRing(*NewPartitionRingDesc())
```
pracucci: It should never fail but we should ensure it. If you never expect it to fail, then check the error and `panic` if the unexpected happens.

ring/partition_ring_watcher.go
```diff
@@ -93,7 +95,11 @@ func (w *PartitionRingWatcher) loop(ctx context.Context) error {
 }
 
 func (w *PartitionRingWatcher) updatePartitionRing(desc *PartitionRingDesc) {
-	newRing := NewPartitionRing(*desc)
+	newRing, err := NewPartitionRing(*desc)
+	if err != nil {
+		level.Error(w.logger).Log("msg", "failed to create partition ring from descriptor", "err", err)
+		return
+	}
```
pracucci, line 105: We don't expect this error, but silently stopping to update the ring is a bad behaviour. What we should do instead is failing the `PartitionRingWatcher` service, so:

- Return error from `updatePartitionRing()`
- In `PartitionRingWatcher.loop()` exit the function with the error if `updatePartitionRing()` returns error
- In `PartitionRingWatcher.starting()` return the error if  `updatePartitionRing()` returns error

ring/partition_ring.go
pracucci: > I think this design suffered too much overthinking. I think we should just have 1 function NewPartitionRing() taking the shuffle shard cache size too. No built-in default should exist in a general purpose library like dskit, because the default value really depends on the use case.

Thinking more about this, the design you propose with functional options is actually fine if by default you keep the unlimited cache, because there will be no behavioural change if downstream projects will just update dskit without doing any code change.

Segflow: I just pushed your suggestion, I agree it's cleaner and does not introduce a breaking change silently.

ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -60,28 +63,22 @@ func (r *partitionRingShuffleShardCache) setSubringWithLookback(identifier strin
...
-	if existingEntry, haveCached := r.cacheWithLookback[key]; !haveCached || existingEntry.validForLookbackWindowsStartingAfter < lookbackWindowStart {
-		r.cacheWithLookback[key] = cachedSubringWithLookback[*PartitionRing]{
+	if existingEntry, haveCached := r.cacheWithLookback.Get(key); !haveCached || existingEntry.validForLookbackWindowsStartingAfter < lookbackWindowStart {
+		r.cacheWithLookback.Add(key, cachedSubringWithLookback[*PartitionRing]{
 			subring:                               subring,
 			validForLookbackWindowsStartingAfter:  lookbackWindowStart,
 			validForLookbackWindowsStartingBefore: validForLookbackWindowsStartingBefore,
-		}
+		})
 	}
```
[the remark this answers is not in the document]
Segflow: addressed

ring/partition_ring.go
```diff
@@ -255,7 +281,12 @@ func (r *PartitionRing) shuffleShard(identifier string, size int, lookbackPeriod
 		}
 	}
 
-	return NewPartitionRing(r.desc.WithPartitions(result)), nil
+	ring, err := NewPartitionRing(r.desc.WithPartitions(result))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create partition ring: %w", err)
+	}
+
+	return ring, nil
```
pracucci: [nit] You can keep it simple and trust the error coming from `NewPartitionRing()`

```suggestion
	return NewPartitionRing(r.desc.WithPartitions(result))
```

ring/partition_ring_test.go
```diff
@@ -137,7 +138,7 @@ func TestPartitionRing_ActivePartitionForKey_NoMemoryAllocations(t *testing.T) {
 		numInactivePartitions = 10
 	)
 
-	ring := createPartitionRingWithPartitions(numActivePartitions, numInactivePartitions, 0)
+	ring := createPartitionRingWithPartitions(PartitionRingOptions{}, numActivePartitions, numInactivePartitions, 0)
```
pracucci: Using `PartitionRingOptions{}` is not future proof. Right now the zero value struct is equal to the default one, but in the future it may not be true (e.g. the default of a new field could be a non-zero value). Always use `DefaultPartitionRingOptions()` instead. This comment applied to all other places in the test.

Note: this issue is why options as functional parameters are a stronger way to pass options, because you can't unintentionally pass a zero value. I see you decided to not use functional params and I will not block on this (there's nothing wrong in using a struct too, it's just more error prone).

ring/partition_ring_watcher.go
```diff
@@ -36,11 +36,15 @@ type PartitionRingWatcherDelegate interface {
 }
 
 func NewPartitionRingWatcher(name, key string, kv kv.Client, logger log.Logger, reg prometheus.Registerer) *PartitionRingWatcher {
+	emptyRing, err := NewPartitionRing(*NewPartitionRingDesc())
+	if err != nil {
+		panic(err) // This should never executes since we're using an empty descriptor.
```
pracucci: Whether  `NewPartitionRing()` can return error or not has nothing to do with "since we're using an empty descriptor".

ring/partition_ring_watcher.go
```diff
@@ -74,25 +78,34 @@ func (w *PartitionRingWatcher) starting(ctx context.Context) error {
 		value = NewPartitionRingDesc()
 	}
 
-	w.updatePartitionRing(value.(*PartitionRingDesc))
+	if err := w.updatePartitionRing(value.(*PartitionRingDesc)); err != nil {
+		return err
+	}
 	return nil
```
pracucci: [nit] Keep it simple, when it can be simple.

```suggestion
	return w.updatePartitionRing(value.(*PartitionRingDesc))
```

ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -1,22 +1,50 @@
...
 	"math"
 	"sync"
 	"time"
+
+	lru "github.com/hashicorp/golang-lru/v2"
 )
 
+// partitionRingShuffleShardCacheStorage is the interface for the underlying storage mechanism.
+// This abstracts the difference between map-based and LRU-based storage.
+type partitionRingShuffleShardCacheStorage interface {
```
pracucci: I don't think this interface has the right design. I don't understand why you need two versions: with and without lookback. The cache should be generic, then you instantiate two caches in the caller: one for the lookback and one for the case without the lookback (like it was happening in your original version with the LRU cache, there were 2 LRU cache instances).

Segflow: Updated the interface and the struct definition to these generic definitions:

```go
type shuffleShardCacheStorage[V any] interface {
	get(key subringCacheKey) (V, bool)
	set(key subringCacheKey, value V)
	len() int
}

type partitionRingShuffleShardCache struct {
	mtx                  sync.RWMutex
	cacheWithoutLookback shuffleShardCacheStorage[*PartitionRing]
	cacheWithLookback    shuffleShardCacheStorage[cachedSubringWithLookback[*PartitionRing]]
}
```

ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -1,22 +1,57 @@
...
+	get(key subringCacheKey) (V, bool)
+	set(key subringCacheKey, value V)
+	len() int
+}
+
+// partitionRingShuffleShardCache delegates storage operations to two generic cache storage implementations.
+// All cache operations are protected by a mutex to ensure thread-safety for compound operations
+// like check-then-act patterns (e.g., in setSubringWithLookback).
 type partitionRingShuffleShardCache struct {
 	mtx                  sync.RWMutex
```
pracucci, line 24: This lock should be moved to `mapCacheStorage`. The LRU is already thread safe so we don't need an extra lock for it.

ring/partitions_ring_shuffle_shard_cache.go
```diff
@@ -94,3 +129,72 @@ func (r *partitionRingShuffleShardCache) getSubringWithLookback(identifier strin
...
+
+	return r.cacheWithLookback.len()
+}
+
+func (r *partitionRingShuffleShardCache) lenWithoutLookback() int {
+	r.mtx.RLock()
+	defer r.mtx.RUnlock()
+
+	return r.cacheWithoutLookback.len()
+}
```
pracucci: What are these functions used for?

ring/partitions_ring_shuffle_shard_cache.go
pracucci: Discussed offline. Let's keep it here.

PR #876, opened 2026-01-14
feat(ring): lifecycler service flow: propagate context for canceling run func

francoposa:
**What this PR does**:

Without this PR, the lifecycler service cannot be stopped while it is blocked in `waitBeforeJoining` for the non-configurable 5-minute `canJoinTimeout`.
When any errors or context cancel/deadlines occur during ingester startup, this allows us to shutdown this service and clean up goroutines without waiting 5 minutes.

See: https://github.com/grafana/mimir/pull/14025

Fixes #<issue number>

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> Improves lifecycler cancellation by threading `ctx` through core run-loop operations, enabling early shutdown during startup waits.
> 
> - In `Lifecycler.loop`, call `initRing`, `autoJoin`, `verifyTokens`, `changeState`, and `updateConsul` with the run `ctx` instead of `context.Background()`
> - Enables cancellation to interrupt `waitBeforeJoining` and heartbeat updates during service shutdown or startup failures
> - Refactor `TestWaitBeforeJoining` into subtests using `t.Run`; emulate `loop()` by calling `initRing` directly, remove service start/stop, and assert `context.Canceled` when appropriate
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f6c5b85bdd71228cda035500601ec45f61ba0945. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ring/lifecycler.go
+++ b/ring/lifecycler.go
@@ -541,7 +541,7 @@ func (i *Lifecycler) Zones() []string {
 func (i *Lifecycler) loop(ctx context.Context) error {
 	// First, see if we exist in the cluster, update our state to match if we do,
 	// and add ourselves (without tokens) if we don't.
-	if err := i.initRing(context.Background()); err != nil {
+	if err := i.initRing(ctx); err != nil {
 		return errors.Wrapf(err, "failed to join the ring %s", i.RingName)
 	}
 
--- a/ring/lifecycler_test.go
+++ b/ring/lifecycler_test.go
@@ -1773,33 +1773,29 @@ func TestWaitBeforeJoining(t *testing.T) {
 			l, err := NewLifecycler(cfg, &nopFlushTransferer{}, "ingester", ringKey, true, log.NewNopLogger(), nil)
 			require.NoError(t, err)
 			l.canJoinTimeout = canJoinTimeout
-			require.NoError(t, services.StartAndAwaitRunning(ctx, l))
+
+			// Emulate running loop() method control flow;
+			// there is no starting method and all stages of the service happens in running method.
+			// The different stages cannot be tested without race conditions if using StartAndAwaitRunning.
+			require.NoError(t, l.initRing(context.Background()))
+
+			start := time.Now()
 
 			if testData.errorRequired {
 				cancel()
 			}
-			start := time.Now()
 			err = l.waitBeforeJoining(ctx)
-			if testData.errorRequired {
-				require.ErrorIs(t, err, context.Canceled)
-			} else {
-				require.NoError(t, err)
 
-				if testData.timeoutRequired {
-					require.GreaterOrEqual(t, time.Since(start), canJoinTimeout)
-				} else {
-					require.Less(t, time.Since(start), canJoinTimeout)
-				}
-			}
-			err = services.StopAndAwaitTerminated(context.Background(), l)
 			if testData.errorRequired {
-				// Error from canceled running func propagates as the failure case
-				// when stopping and awaiting service state.
-				require.ErrorContains(t, err, "failed to pick tokens")
-				// Error should be wrapped in the top-level cause, which is context cancellation.
 				require.ErrorIs(t, err, context.Canceled)
 			} else {
 				require.NoError(t, err)
+			}
+
+			if testData.timeoutRequired {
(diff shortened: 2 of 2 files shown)


ring/lifecycler.go
```diff
@@ -538,13 +538,16 @@ func (i *Lifecycler) Zones() []string {
 	return i.zones
 }
 
-func (i *Lifecycler) loop(ctx context.Context) error {
+func (i *Lifecycler) starting(ctx context.Context) error {
 	// First, see if we exist in the cluster, update our state to match if we do,
 	// and add ourselves (without tokens) if we don't.
-	if err := i.initRing(context.Background()); err != nil {
+	if err := i.initRing(ctx); err != nil {
 		return errors.Wrapf(err, "failed to join the ring %s", i.RingName)
```
francoposa, line 545: Moved `initRing` to `starting`. It doesn't change any actual behavior in terms of how it's used, but seems more in the spirit of the `Service` paradigm here.

ring/lifecycler.go
```diff
@@ -538,13 +538,16 @@ func (i *Lifecycler) Zones() []string {
 	return i.zones
 }
 
-func (i *Lifecycler) loop(ctx context.Context) error {
+func (i *Lifecycler) starting(ctx context.Context) error {
 	// First, see if we exist in the cluster, update our state to match if we do,
 	// and add ourselves (without tokens) if we don't.
-	if err := i.initRing(context.Background()); err != nil {
+	if err := i.initRing(ctx); err != nil {
```
francoposa, line 544: CAS operation on ring - do we want to be allowed to be canceled with the service context passed in, or give it a background context to proceed until the CAS operation succeeds or times out?

ring/lifecycler.go
```diff
@@ -601,7 +604,7 @@ func (i *Lifecycler) loop(ctx context.Context) error {
 
 		case <-heartbeatTickerChan:
 			i.lifecyclerMetrics.consulHeartbeats.Inc()
-			if err := i.updateConsul(context.Background()); err != nil {
+			if err := i.updateConsul(ctx); err != nil {
```
francoposa, line 604: These all do CAS operations on the ring - same as above, we may want to keep these as background contexts to give them the best change to propagate ring state?

They won't block for 5 minutes like `autoJoin` did, just for however long the CAS retries take to exhaust.

ring/lifecycler_test.go
```diff
@@ -1763,36 +1763,47 @@ func TestWaitBeforeJoining(t *testing.T) {
...
+			err = services.StopAndAwaitTerminated(context.Background(), l)
+			if testData.errorRequired {
+				// Error from canceled running func propagates as the failure case
+				// when stopping and awaiting service state.
+				require.ErrorContains(t, err, "failed to pick tokens")
+				// Error should be wrapped in the top-level cause, which is context cancellation.
+				require.ErrorIs(t, err, context.Canceled)
+			} else {
+				require.NoError(t, err)
+			}
```
francoposa: This is the actual test change - rest is indentation change from running with t.Run() to report the actual test names.

ring/lifecycler.go
pracucci: I don't think it's important for the `loop()` function. What's really important is to ensure the final state propagation in the `stopping()` and the functions there should keep using `context.Background()` (but you're not changing them in this PR).

pracucci: The key point is that today you may skip the updates done in `loop()` anyway if `case <-ctx.Done()` happens first. That's why I don't think there's much difference in practice.

ring/lifecycler.go
pracucci: I think it's just safer to move it back to `loop()`. We're basically changing the Lifecycler behaviour from the caller perspective, because callers typically do "start and await started". I prefer to get this PR stick to what was intended, that is cancel operations if the service gets stopped.

pracucci (approved):
Approving if you move `initRing()` back to `loop()`, otherwise please don't consider this PR approved by me.

PR #886, opened 2026-01-27
fix(kv): wrap and propagate version mismatch error in memberlist kv CAS

francoposa:
**What this PR does**:

We were overwriting/dropping this error and did not know why CAS updates were failing - this propagates the error instead of clobbering.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

---

> [!NOTE]
> **Low Risk**
> Small error-handling change limited to CAS failure messages plus a test update; no behavior change to the CAS retry logic or data path.
> 
> **Overview**
> Fixes CAS failure reporting in memberlist KV by **preserving the underlying `errVersionMismatch` cause** when the retry budget is exhausted, instead of collapsing it into a generic “too many retries” error.
> 
> This switches to `github.com/pkg/errors` for wrapping and updates the relevant unit test to assert the new error message (`too many retries: version mismatch`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 315a85aa7daa7c46af84ddd10d4d5c440f629c80. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -424,7 +424,6 @@ var (
 	// if merge fails because of CAS version mismatch, this error is returned. CAS operation reacts on it
 	errVersionMismatch     = errors.New("version mismatch")
 	errNoChangeDetected    = errors.New("no change detected")
-	errTooManyRetries      = errors.New("too many retries")
 	emptySnappyEncodedData = snappy.Encode(nil, []byte{})
 )
 
@@ -1302,10 +1301,9 @@ outer:
 
 		return nil
 	}
-
 	if errors.Is(lastError, errVersionMismatch) {
-		// this is more likely error than version mismatch.
-		lastError = errors.Wrap(lastError, errTooManyRetries.Error())
+		// Version mismatch err on CAS would have been retried up to the limit
+		lastError = errors.Wrap(lastError, "too many retries")
 	}
 
 	m.casFailures.Inc()
--- a/kv/memberlist/memberlist_client_test.go
+++ b/kv/memberlist/memberlist_client_test.go
@@ -487,7 +487,7 @@ func TestCASFailedBecauseOfVersionChanges(t *testing.T) {
 			return d, true, nil
 		})
 
-		require.EqualError(t, err, "failed to CAS-update key test: too many retries")
+		require.EqualError(t, err, "failed to CAS-update key test: too many retries: version mismatch")
 		require.Equal(t, maxCasRetries, calls)
 	})
 }


kv/memberlist/memberlist_client.go
```diff
@@ -1254,7 +1254,7 @@ outer:
 
 	if errors.Is(lastError, errVersionMismatch) {
 		// this is more likely error than version mismatch.
-		lastError = errTooManyRetries
+		lastError = errors.Wrap(lastError, errTooManyRetries.Error())
```
pracucci: I would remove the `errTooManyRetries` variable at all, and just reference the string here, given an `error.Is(errTooManyRetries)` wouldn't work anymore (but we never checked it, so it shouldn't be a regression).

```suggestion
		lastError = errors.Wrap(lastError, "too many retries")
```

kv/memberlist/memberlist_client.go
```diff
@@ -1251,10 +1250,9 @@ outer:
 
 		return nil
 	}
-
 	if errors.Is(lastError, errVersionMismatch) {
-		// this is more likely error than version mismatch.
-		lastError = errTooManyRetries
+		// Version mismatch err on CAS would have been retried up to the limit
+		lastError = errors.Wrap(lastError, "too many retries")
```
karlskewes, line 1306: Was it intended to pull in `github.com/pkg/errors`?

```suggestion
		// Version mismatch err on CAS would have been retried up to the limit
		lastError = fmt.Errorf("too many retries: %w", lastError)
```

kv/memberlist/memberlist_client.go
```diff
@@ -18,6 +17,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/golang/snappy"
 	"github.com/hashicorp/memberlist"
+	"github.com/pkg/errors"
```
karlskewes, line 20:
```suggestion
```

kv/memberlist/memberlist_client.go
```diff
@@ -5,7 +5,6 @@ import (
 	"context"
 	crypto_rand "crypto/rand"
 	"encoding/binary"
-	"errors"
```
karlskewes, line 8: Accidentally removed? Can't add suggestion to re-add.

karlskewes (changes requested):

kv/memberlist/memberlist_client.go
francoposa: yes, we use it everywhere else in this package

karlskewes: Ok, I see it is used everywhere. Let's ship this change then.

Caught my eye as pkg/errors was archived in 2021 and it has been getting removed in favour of the standard library errors package across the ecosystem.
Replacing in dskit should be a separate PR though.

PR #892, opened 2026-01-30
feat: distribute memberlist rejoins over time for different processes starting at the same time

pracucci:
**What this PR does**:

When multiple processes start at the same time (e.g., during a rolling restart or scale-up), they would all attempt to rejoin seed nodes nearly at the same intervals, creating a thundering herd effect. This PR introduces a random initial delay (between 0 and `RejoinInterval`) before the first rejoin attempt, uniformly distributing rejoins across time.

The variable ticker implementation has been copied from Mimir. Once this PR will be merged, I will update dskit in Mimir and remove the variable ticker implementation from there.

Here periodic push/pull rate caused by rejoin before and after the jitter rollout in a test Mimir cluster (before the rollout you can see it spikes every 5m, which is the configured rejoin interval):

542848503-e36a8611-7caf-41d0-8423-2b5a8c5f9624

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/timeutil/variable_ticker_test.go
+++ b/timeutil/variable_ticker_test.go
@@ -23,9 +23,9 @@ func TestVariableTicker(t *testing.T) {
 		}
 
 		tolerance := 250 * time.Millisecond
-		assert.InDelta(t, ticks[0].Sub(startTime).Seconds(), 1*time.Second.Seconds(), float64(tolerance))
-		assert.InDelta(t, ticks[1].Sub(startTime).Seconds(), 3*time.Second.Seconds(), float64(tolerance))
-		assert.InDelta(t, ticks[2].Sub(startTime).Seconds(), 5*time.Second.Seconds(), float64(tolerance))
+		assert.InDelta(t, ticks[0].Sub(startTime).Seconds(), 1*time.Second.Seconds(), tolerance.Seconds())
+		assert.InDelta(t, ticks[1].Sub(startTime).Seconds(), 3*time.Second.Seconds(), tolerance.Seconds())
+		assert.InDelta(t, ticks[2].Sub(startTime).Seconds(), 5*time.Second.Seconds(), tolerance.Seconds())
 	})
 
 	t.Run("should not close the channel on stop function called", func(t *testing.T) {


timeutil/variable_ticker_test.go
```diff
@@ -0,0 +1,77 @@
...
+		// Capture the timing of 3 ticks.
+		var ticks []time.Time
+		for len(ticks) < 3 {
+			ticks = append(ticks, <-tickerChan)
+		}
+
+		tolerance := 250 * time.Millisecond
+		assert.InDelta(t, ticks[0].Sub(startTime).Seconds(), 1*time.Second.Seconds(), float64(tolerance))
+		assert.InDelta(t, ticks[1].Sub(startTime).Seconds(), 3*time.Second.Seconds(), float64(tolerance))
+		assert.InDelta(t, ticks[2].Sub(startTime).Seconds(), 5*time.Second.Seconds(), float64(tolerance))
```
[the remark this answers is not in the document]
pracucci: Good point. This bug was already existing in Mimir. Fixed in https://github.com/grafana/dskit/pull/892/commits/030073ebdc66662ef8870def75bd337016766f15

timeutil/variable_ticker.go
```diff
@@ -0,0 +1,65 @@
+package timeutil
+
+import (
+	"sync"
+	"time"
+)
+
+// NewVariableTicker wrap time.Ticker to Reset() the ticker with the next duration (picked from
```
56quarters, line 8: Is it necessary to create this type with goroutines and channels to have an initial delay and a periodic delay? Could we combine `time.NewTimer` and `time.NewTicker` instead?

kv/memberlist/memberlist_client.go
```diff
@@ -624,10 +625,12 @@ func (m *KV) running(ctx context.Context) error {
 
 	var tickerChan <-chan time.Time
 	if m.cfg.RejoinInterval > 0 && len(m.cfg.GetRejoinSeedNodes()) > 0 {
-		t := time.NewTicker(m.cfg.RejoinInterval)
-		defer t.Stop()
-
-		tickerChan = t.C
+		// Use a random initial delay between 0 and RejoinInterval to uniformly
+		// distribute rejoins across time when multiple processes start simultaneously.
+		initialDelay := 1 + time.Duration(math_rand.Int63n(int64(m.cfg.RejoinInterval)))
```
56quarters, line 630: We could also do the Prometheus rule-eval thing here and do a hash of the node name modulo the RejoinInterval for the initial delay.

pracucci: What would be the benefit? Looks a complication to me.

timeutil/variable_ticker.go
pracucci: Maybe we can do it in a follow up PR. This code is what we're already using since a long time in Mimir, so looks low risk to just move it from Mimir to dskit.

kv/memberlist/memberlist_client.go
56quarters: It would be deterministic. It doesn't seem complicated to me, it'd just be a hash of the node name from config.

pracucci: Not sure I get the point. The node name is random (we add a random suffix). For example see:

Screenshot 2026-01-30 at 16 03 01

The last part of the node random is just a random suffix. It changes on each startup. We could hash without the suffix, but pod IDs are non deterministic too (change with each rollout), so it would be deterministic just for statefulsets.

timeutil/variable_ticker.go
pracucci: Merging this PR because I need to get this stuff in by EOW (need to fix an high inter-AZ data transfer caused by memberlist join) but I'm not ignoring your feedback. Added it to my todo list and will re-iterate in a follow up PR.

PR #901, opened 2026-02-11
feat(tenant): use tenant metadata instead of subtenant

juliusmh:
> [!NOTE]
> **Medium Risk**
> Touches core tenant identification/parsing logic and removes the prior subtenant API surface, which could affect callers and request routing if any edge cases or integrations relied on the old format.
> 
> **Overview**
> Adds a new `Metadata` type with parsing/validation utilities (including length limits and deterministic encoding via `WithTenant`) to support `tenantID:key=value(:k=v...)` identifiers.
> 
> Updates tenant resolution to **strip metadata for backward-compatible `TenantID`/`TenantIDs`**, and replaces subtenant APIs with `ExtractWithMetadata`/`ParseWithMetadata` that validate and parse metadata while enforcing single-orgID consistency. Tests and benchmarks are updated accordingly, and subtenant-related constants/helpers/interfaces are removed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 426308fc9c8e5f1d5feff6373750ccd21b120415. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tenant/metadata.go
+++ b/tenant/metadata.go
@@ -1,0 +1,164 @@
+package tenant
+
+import (
+	"fmt"
+	"iter"
+	"maps"
+	"slices"
+	"strings"
+)
+
+const (
+	// metadataSeparator separates the tenant ID from the metadata. The format is
+	// "tenantID:key=value" (e.g. "123456:product=k6"). The colon is not a valid
+	// character in tenant IDs, making it safe to use as a separator.
+	metadataSeparator = ':'
+	// metadataKVSeparator separates individual key value pairs of metadata. The
+	// format is "key=value" (e.g., "source=test"). The equal sign is not a valid
+	// character in tenant IDs, making it safe to use as a separator.
+	metadataKVSeparator = '='
+)
+
+var validMetadataChars [256]bool
+
+func init() {
+	validMetadataChars[metadataSeparator] = true
+	validMetadataChars[metadataKVSeparator] = true
+
+	for c := 'a'; c <= 'z'; c++ {
+		validMetadataChars[c] = true
+	}
+	for c := 'A'; c <= 'Z'; c++ {
+		validMetadataChars[c] = true
+	}
+	for c := '0'; c <= '9'; c++ {
+		validMetadataChars[c] = true
+	}
+	for _, c := range "-_" {
+		validMetadataChars[c] = true
+	}
+}
+
+type errMetadataUnsupportedCharacter struct {
+	pos      int
+	metadata string
+}
+
+func (e *errMetadataUnsupportedCharacter) Error() string {
+	return fmt.Sprintf(
+		"metadata '%s' contains unsupported character '%c'",
+		e.metadata,
+		e.metadata[e.pos],
+	)
+}
+
+type Metadata struct {
+	data map[string]string
+}
+
+func NewMetadata() Metadata {
+	return Metadata{}
+}
+
+// ParseMetadata from string without validating input. ValidMetadata must be used
+// to ensure input contains only allowed characters. The format for metadata is a
+// colon-delimited list of key value pairs.
+// Examples:
+// - key=value
+// - key=value:foo=bar
+func ParseMetadata(input string) (Metadata, error) {
+	if input == "" {
+		return Metadata{}, nil
+	}
+	var (
+		currentPair  string
+		hasMorePairs bool
+		remaining    = input
+		metadata     = NewMetadata()
+	)
+	for {
+		currentPair, remaining, hasMorePairs = stringsCut(remaining, metadataSeparator)
+		key, val, ok := stringsCut(currentPair, metadataKVSeparator)
+		if !ok {
+			return Metadata{}, fmt.Errorf("invalid key value pair %s", currentPair)
+		}
+		metadata.Set(key, val)
+		if !hasMorePairs {
+			break
+		}
+	}
+	return metadata, nil
+}
+
+// ValidMetadata returns an error if the metadata is invalid, nil otherwise.
+// Metadata must contain only lowercase letters, digits, '-', and '='.
+func ValidMetadata(s string) error {
+	for i := 0; i < len(s); i++ {
+		if !validMetadataChars[s[i]] {
+			return &errMetadataUnsupportedCharacter{metadata: s, pos: i}
+		}
+	}
+	if len(s) > MaxMetadataLength {
+		return fmt.Errorf("metadata too long: %d", len(s))
+	}
+	return nil
+}
+
+// Set a key value pair.
+func (m *Metadata) Set(key string, val string) {
+	if m.data == nil {
+		m.data = make(map[string]string)
+	}
+	m.data[key] = val
+}
+
+// Remove the value with key from m.
+func (m Metadata) Remove(key string) {
+	if m.data != nil {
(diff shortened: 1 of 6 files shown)


tenant/resolver.go
```diff
@@ -153,14 +153,9 @@ type Resolver interface {
...
-	// Returns empty subtenant if no subtenant is present.
-	// The orgID format is "tenantID:subtenantID".
-	SubtenantID(context.Context) (string, string, error)
-
-	// SubtenantIDs returns all subtenant IDs from the context. It should return
-	// a normalized list of ordered and distinct subtenant IDs.
-	SubtenantIDs(context.Context) ([]string, error)
+	// TenantWithMetadata returns the tenant ID and metadata from the context.
+	// The orgID format is "tenantID:metadata".
+	TenantWithMetadata(context.Context) (string, *Metadata, error)
```
tcard: **suggestion:** Returning `Metadata` (no pointer) will potentially save an allocation.

tenant/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+		e.metadata,
+		e.metadata[e.pos],
+	)
+}
+
+type Metadata struct {
+	data map[string]string
+}
+
+func NewMetadata() *Metadata {
```
tcard: Maps are already pointers, so making this return a `*Metadata` just adds another indirection and may prevent optimizations.

tenant/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+	return val, ok
+}
+
+// Encode the tenant metadata as a string.
+func (m *Metadata) Encode() string {
+	keys := make([]string, 0, len(m.data))
+	for k := range m.data {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
tcard: More compact as:

```suggestion
	keys := slices.Sorted(maps.Keys(m.data))
```

tenant/resolver.go
```diff
@@ -76,69 +76,51 @@ func parseTenantIDs(orgID string) ([]string, error) {
...
 }
 
-// SubtenantID returns the subtenant ID from the context, or an empty string if
-// no subtenant is present. The orgID format is "tenantID:subtenantID" (e.g., "123456:k6").
+// TenantWithMetadata returns the tenant ID and optional metadata from the context, or an empty string if
+// no metadata is present. The orgID format is "tenantID:metadata" (e.g., "123456:test=yes").
 //
 //nolint:revive
-func SubtenantID(ctx context.Context) (tenantID string, subtenantID string, _ error) {
+func TenantWithMetadata(ctx context.Context) (tenantID string, metadata *Metadata, _ error) {
```
tcard: **nitpick:** For consistency with `ExtractOrgID`, this could be called `ExtractTenantWithMetadata`. Or, since we're already in the `tenant` package, just `ExtractWithMetadata`.

tenant/resolver.go
```diff
@@ -76,69 +76,51 @@ func parseTenantIDs(orgID string) ([]string, error) {
...
+// ParseTenantWithMetadata returns the tenant ID and optional metadata from
+// orgID(s). Metadata is nil if no metadata is present. The orgID format is
+// "tenantID:metadata". The metadata string must contain alphanumeric characters
+// and "-" or "=". Metadata is a key-value pair separated by "=".
+//
+// Examples:
+// - 123456 (tenantID=12345, metadata=<nil>)
+// - 123456:a=b (tenantID=12345, metadata={a: b})
+// - 123456:a (tenantID=12345, metadata={a: b})
+func ParseTenantWithMetadata(orgIDs string) (string, *Metadata, error) {
```
tcard: **nitpick:** Could be reduced to `ParseWithMetadata`. Also, it would be nice to name the return `string` to make it more obvious.

```suggestion
func ParseWithMetadata(orgIDs string) (tenantID string, m *Metadata, err error) {
```

tenant/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Set a key value pair.
+func (m *Metadata) Set(key string, val string) {
+	m.data[key] = val
+}
+
+// Has checks whether a specific metadata key is present.
+func (m *Metadata) Has(key string) bool {
+	if m == nil {
```
tcard: We could save this check if we used plain `Metadata` everywhere.

tenant/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+func (m *Metadata) Get(key string) (string, bool) {
+	if m == nil {
+		return "", false
+	}
+	val, ok := m.data[key]
+	return val, ok
+}
+
+// Encode the tenant metadata as a string.
+func (m *Metadata) Encode() string {
```
tcard: We will typically want to append this to a tenant ID, so we'd save an allocation+copy if this method (or an additional one) either took a `tenantID string` or a `strings.Builder`.

Also, probably more importantly, there's currently no way of constructing a `<tenant ID>:<metadata>` string without hardcoding the `:` on the caller, who shouldn't have to know how those strings are encoded.

tenant/metadata.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+	if len(s) > MaxMetadataLength {
+		return fmt.Errorf("metadata too long: %d", len(s))
+	}
+	return nil
+}
+
+// Set a key value pair.
+func (m *Metadata) Set(key string, val string) {
```
tcard, line 108: How about we keep this as the only pointer method, and it initialized `m.data` if it's nil? This way, zero values would be useful, and we could do something like:

```go
tenantID, metadata, err := tenant.ParseWithMetadata(orgIDHeader)
metadata.Set("source", "k6") // metadata will typically be a zero Metadata{} here, it would still work
return metadata.EncodeWith(tenantID)
```

This would be more compact, efficient, and type-safer than the current API:

```go
tenantID, metadata, err := tenant.ParseWithMetadata(orgIDHeader)
if metadata == nil { // I may forget about this, but also it shouldn't be my problem
    metadata = tenant.NewMetadata()
}
metadata.Set("source", "k6")
return tenantID + ":" + metadata.Encode() // I shouldn't know ":" goes there
```

tenant/metadata.go
```diff
@@ -0,0 +1,144 @@
...
+	if m.data == nil {
+		m.data = make(map[string]string)
+	}
+	m.data[key] = val
+}
+
+// Has checks whether a specific metadata key is present.
+func (m Metadata) Has(key string) bool {
+	if m.data == nil {
```
tcard, line 124: **nitpick:** Not needed, a comma-ok read from a nil map will return zero, false.

Same for `Get`.

tenant/metadata.go
```diff
@@ -0,0 +1,151 @@
...
+
+func init() {
+	validMetadataChars[metadataSeparator] = true
+	validMetadataChars[metadataKVSeparator] = true
+
+	for c := 'a'; c <= 'z'; c++ {
+		validMetadataChars[c] = true
+	}
+	for c := 'A'; c <= 'Z'; c++ {
+		validTenantIdChars[c] = true
```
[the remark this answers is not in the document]
juliusmh: Good catch. Fixed.

PR #903, opened 2026-02-12
feat(server): always disable /debug/pprof/cmdline, return 404

colega:
The `/debug/pprof/cmdline` endpoint exposes the full command line of the running process, which may contain sensitive information such as credentials passed as flags. This unconditionally disables the endpoint by returning 404, with no configuration option to re-enable it.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to debug routing plus a test; behavior change only impacts the `pprof` cmdline endpoint.
> 
> **Overview**
> The server now **unconditionally disables** the `/debug/pprof/cmdline` endpoint by explicitly registering it to return `404`, while leaving other `/debug/pprof` routes served from `http.DefaultServeMux`.
> 
> Adds a regression test (`TestPprofCmdlineDisabled`) asserting `/debug/pprof/cmdline` returns `404` and that other pprof endpoints like `/debug/pprof/` and `/debug/pprof/heap` still return `200`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0508ab205a4fc1f12206e29ee34ef981fb52ede2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1069,6 +1069,8 @@ func TestPprofCmdlineDisabled(t *testing.T) {
 	var cfg Config
 	flagext.DefaultValues(&cfg)
 	setAutoAssignedPorts("tcp", &cfg)
+	reg := prometheus.NewPedanticRegistry()
+	cfg.Registerer, cfg.Gatherer = reg, reg
 
 	srv, err := New(cfg)
 	require.NoError(t, err)


server/server_test.go
```diff
@@ -1063,6 +1064,33 @@ func setAutoAssignedPorts(network string, cfg *Config) {
 	cfg.GRPCListenPort = 0
 }
 
+func TestPprofCmdlineDisabled(t *testing.T) {
+	cfg := Config{}
+	cfg.RegisterFlags(flag.NewFlagSet("", flag.PanicOnError))
```
56quarters: This is done with `flagext.DefaultValues(&cfg)` elsewhere

server/server_test.go
```diff
@@ -1063,6 +1064,33 @@ func setAutoAssignedPorts(network string, cfg *Config) {
...
+	setAutoAssignedPorts("tcp", &cfg)
+	cfg.Registerer = prometheus.NewPedanticRegistry()
+	cfg.Gatherer = prometheus.NewRegistry()
```
56quarters: Why are these different objects?

server/server_test.go
colega: Actually this file doesn't use `flagext.DefaultValues`, but I don't think it's worth discussing this so I replaced it.

server/server_test.go
colega: I removed them completely, the test passes without them.

56quarters (approved):

Issue #908, opened 2026-02-17
httpConnectionTTLMiddleware: misleading comment on removeIdleExpiredConnections


aknuds1:
## Description

In `middleware/http_connection_ttl_middleware.go`, the comment on `removeIdleExpiredConnections` says:

```go
// removeIdleExpiredConnections removes all expired idle cached connections, i.e.,
// the ones exceeding their own TTL.
```

But the implementation uses `m.maxTTL` as the idle timeout for all connections, not the per-connection `state.ttl`:

```go
func (m *httpConnectionTTLMiddleware) removeIdleExpiredConnections() {
    count := 0
    m.connectionsMu.Lock()
    for conn, state := range m.connections {
        if state.isIdleExpired(m.maxTTL) {   // uses maxTTL, not state.ttl
            delete(m.connections, conn)
            count++
        }
    }
    // ...
}
```

The behavior (using `maxTTL` as the idle threshold) is likely intentional for GC purposes — it's a reasonable upper bound for "this connection is gone and we can stop tracking it". But the comment claims it removes connections "exceeding their own TTL", which is inaccurate.

### Suggested fix

Update the comment to reflect the actual behavior, e.g.:

```go
// removeIdleExpiredConnections removes cached connections that have been idle
// (no requests seen) for longer than maxTTL, as a garbage collection mechanism.
```

PR #910, opened 2026-02-17
fix(middleware): prevent idle cleanup from resetting connection TTL

### LINKED ISSUE #907: httpConnectionTTLMiddleware: removeExpiredConnection unconditionally re-creates cleaned-up entries
## Description

In `middleware/http_connection_ttl_middleware.go`, `removeExpiredConnection` calls `connectionState(conn)` as its first step. `connectionState` always creates a new map entry if one doesn't exist:

```go
func (m *httpConnectionTTLMiddleware) removeExpiredConnection(conn string) bool {
    state := m.connectionState(conn)  // creates entry if missing
    if !state.isExpired() {
        return false
    }
    // ...
}
```

```go
func (m *httpConnectionTTLMiddleware) connectionState(conn string) *connectionState {
    // ...
    state, cachedConn := m.connections[conn]
    if cachedConn {
        state.lastSeen = now
    } else {
        ttl := m.calculateTTL(conn)
        state = &connectionState{ttl: ttl, created: now, lastSeen: now}
        m.connections[conn] = state  // always creates
    }
    // ...
}
```

This means that if a connection was previously cleaned up (e.g. by the idle ticker in `removeIdleExpiredConnections`), the next request from the same `RemoteAddr` will:

1. **Re-create** the entry with a fresh TTL and `created` timestamp
2. Increment `totalOpenConnections` (phantom open)
3. See `isExpired() == false` (just created) and return false — no `Connection: close` is sent

The net effect is that an expired connection that was garbage-collected silently gets a brand new TTL instead of being immediately signalled to close.

### Suggested fix

Separate "get-or-create" from "get-if-exists". In `removeExpiredConnection`, use a read-only lookup. Only create new entries in `Wrap` (or a dedicated tracking method) when the connection is not being evicted.

aknuds1:
**What this PR does**:

The connection TTL middleware has two independent code paths sharing a `connections` map: `removeExpiredConnection` (called per request, enforces TTL) and `removeIdleExpiredConnections` (background ticker, garbage-collects entries for idle connections). When a connection's TTL has expired but no request has arrived yet to actually close the connection (via `removeExpiredConnection`), the idle cleanup can fire first and delete the middleware's tracked connection state. The next request then creates a fresh entry and a new TTL, via `removeExpiredConnection`. This can repeat on every idle gap, allowing a TCP connection to live indefinitely without being closed and then rotated.

This PR fixes the problem through two-phase idle cleanup:

- `removeIdleExpiredConnections` now **marks** idle-expired entries on the first pass (`idleExpired = true`) and **deletes** them on the second pass only if no request arrived in between. This preserves the entry so that `removeExpiredConnection` can still find it, check the original `created` time, and correctly send `Connection: close`.

A deterministic test using `testing/synctest` reproducing the scenario is added.

Fixes #907 and #908

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> **Medium Risk**
> Changes core connection-tracking/cleanup behavior and timing interactions between request handling and a background goroutine; mistakes could lead to leaked map entries or incorrect connection closing/metrics.
> 
> **Overview**
> Fixes a bug where the background idle-connection cleanup could delete a tracked connection before a subsequent request arrived, allowing the next request to recreate state and effectively reset the connection’s TTL.
> 
> `removeIdleExpiredConnections` now performs **two-phase cleanup** by first marking idle-expired entries (`idleExpired`) and only deleting them on a later pass if they remain idle, ensuring `removeExpiredConnection` can still enforce the original TTL. Tests were updated to stop the middleware in cleanup, and a new deterministic `testing/synctest` case covers the “idle cleanup then request” scenario and asserts `Connection: close` is sent and metrics reflect the close reason.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 266e9130733459f3ab7fa1b1097939b28814a3b2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/middleware/http_connection_ttl_middleware_test.go
+++ b/middleware/http_connection_ttl_middleware_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"go.uber.org/atomic"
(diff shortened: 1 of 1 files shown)


middleware/http_connection_ttl_middleware.go
```diff
@@ -140,25 +148,24 @@ func (m *httpConnectionTTLMiddleware) removeExpiredConnection(conn string) bool
 	m.connectionsMu.Lock()
 	state, exists := m.connections[conn]
 	if !exists {
-		state = &connectionState{
+		m.connections[conn] = &connectionState{
```
pracucci: I think you can restore this function to the version in `main` to simplify the diff. There's no logic change here.

pracucci (commented):
I can't see the bug you're trying to fix here. The PR and issue description are not much clear to me, and I honestly can't see what we're fixing.

middleware/http_connection_ttl_middleware.go
aknuds1: I will check.

aknuds1: You're right @pracucci - this change was cosmetic, so I reverted it.

pracucci (approved):

aknuds1:
@pracucci I also found there is the edge case of two idleness ticks occurring after the TTL is exceeded, but before the evicting request occurs. Like your Claude session said though, it appears impractical to fix.

Issue #913, opened 2026-02-17
httpConnectionTTLMiddleware: Stop() not accessible via returned Interface


aknuds1:
## Description

`NewHTTPConnectionTTLMiddleware` in `middleware/http_connection_ttl_middleware.go` returns `middleware.Interface`, but the `Stop()` method that shuts down the background ticker goroutine is only defined on the unexported concrete type `*httpConnectionTTLMiddleware`.

```go
func NewHTTPConnectionTTLMiddleware(...) (Interface, error) {
    // ...
    // starts a background goroutine when maxTTL > 0
    go func() { ... }()
    return rpcLimiter, nil  // returns Interface, Stop() is inaccessible
}
```

When `maxTTL > 0`, callers have no way to stop the background goroutine without a type assertion to the unexported type (which they can't do from outside the package). The goroutine and its ticker will run for the lifetime of the process.

### Suggested fix

Either:
- Return a more specific interface that includes `Stop()`, or
- Export the concrete type so callers can type-assert, or
- Accept a `context.Context` and stop the goroutine when the context is cancelled

PR #914, opened 2026-02-17
fix(middleware): accept context.Context for goroutine lifecycle in httpConnectionTTLMiddleware

### LINKED ISSUE #912: httpConnectionTTLMiddleware: concurrent Stop() calls can panic
## Description

`Stop()` in `middleware/http_connection_ttl_middleware.go` uses a `select` on the `done` channel to check if it's already closed before calling `close()`:

```go
func (m *httpConnectionTTLMiddleware) Stop() {
    if m.done != nil {
        select {
        case <-m.done:
            // Already closed
        default:
            close(m.done)
        }
    }
}
```

This is not thread-safe. Two goroutines calling `Stop()` concurrently can both enter the `default` branch (neither sees the channel as closed yet), and one will panic on `close` of an already-closed channel.

### Impact

Low — `Stop()` is typically only called during shutdown. But if shutdown code has any concurrency (e.g., signal handler + graceful shutdown path), this can crash the process.

### Suggested fix

Use `sync.Once` to guarantee `close` is called exactly once:

```go
type httpConnectionTTLMiddleware struct {
    // ...
    stopOnce sync.Once
}

func (m *httpConnectionTTLMiddleware) Stop() {
    if m.done != nil {
        m.stopOnce.Do(func() {
            close(m.done)
        })
    }
}
```

aknuds1:
**What this PR does**:

`NewHTTPConnectionTTLMiddleware` returns `middleware.Interface` (which only has `Wrap()`), but the `Stop()` method for shutting down the background goroutine was only on the unexported concrete type. Callers outside the package could not stop the goroutine without an impossible type assertion.

This replaces the `done` channel and `Stop()` method with a caller-provided `context.Context`. The background goroutine stops when the context is cancelled. This is idiomatic Go and removes the need for a separate cleanup method.

Changes:
- Add `ctx context.Context` as the first parameter of `NewHTTPConnectionTTLMiddleware`
- Remove `done` and `ticker` fields from the struct (ticker is now a local variable in the goroutine)
- The goroutine selects on `ctx.Done()` instead of `m.done`
- Remove the `Stop()` method entirely

Fixes #912 and #913

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> **Low Risk**
> Small, localized API change to middleware construction and goroutine shutdown; behavior is otherwise unchanged aside from requiring callers to provide/cancel a context.
> 
> **Overview**
> Updates `NewHTTPConnectionTTLMiddleware` to take a caller-provided `context.Context` and uses `ctx.Done()` to stop the idle-connection cleanup goroutine.
> 
> Removes the internal `ticker`/`done` fields and the concrete-type `Stop()` method (which wasn’t reachable via the exported `middleware.Interface`), and updates tests to pass `t.Context()` when constructing the middleware.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4983d4f5e042999ee23ebfad576d64b170f13e11. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/middleware/http_connection_ttl_middleware.go
+++ b/middleware/http_connection_ttl_middleware.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"context"
 	"fmt"
 	"hash/fnv"
 	"net/http"
@@ -47,9 +48,6 @@ type httpConnectionTTLMiddleware struct {
 
 	totalOpenConnections   prometheus.Counter
 	totalClosedConnections *prometheus.CounterVec
-
-	ticker *time.Ticker
-	done   chan struct{}
 }
 
 // NewHTTPConnectionTTLMiddleware returns an HTTP middleware that limits the maximum lifetime, TTL,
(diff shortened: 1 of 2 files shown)


aknuds1:
Superseded by #915.

PR #915, opened 2026-02-17
chore: add HTTPConnectionTTLMiddleware interface

### LINKED ISSUE #913: httpConnectionTTLMiddleware: Stop() not accessible via returned Interface

pracucci:
**What this PR does**:

In this PR I'm adding the interface `HTTPConnectionTTLMiddleware` to expose `Stop()` too.

Fixes #913.

**Checklist**
- [x] Tests updated

### DIFF
--- a/middleware/http_connection_ttl_middleware.go
+++ b/middleware/http_connection_ttl_middleware.go
@@ -39,10 +39,11 @@ func (c *connectionState) isIdleExpired(timeout time.Duration) bool {
 }
 
 // HTTPConnectionTTLMiddleware is an HTTP middleware that limits the maximum lifetime
-// of TCP connections. It embeds Interface and adds a Stop method for cleanup.
+// of TCP connections.
 type HTTPConnectionTTLMiddleware interface {
 	Interface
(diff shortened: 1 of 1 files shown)


aknuds1:
I added that it fixes #913.

PR #916, opened 2026-02-18
chore: add config option to log memberlist beacons with high delay

pracucci:
**What this PR does**:

I've enabled the memberlist propagation delay tracker in some testing clusters at Grafana Labs. Typically the latency is very low (good!) but we have few random spikes to minutes. I can't understand it, but I don't have enough information to troubleshoot it.

In this PR:
- Add a config option to log memberlist beacons with high delay
- Add `published_by` to beacons, to know what node generated it

N/A

**Checklist**
- [ ] Tests updated

### DIFF
--- a/kv/memberlist/propagation_tracker_test.go
+++ b/kv/memberlist/propagation_tracker_test.go
@@ -1,7 +1,6 @@
 package memberlist
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"strings"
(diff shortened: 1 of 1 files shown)


kv/memberlist/propagation_tracker.proto
```diff
@@ -17,6 +17,8 @@ message PropagationDelayTrackerDesc {
 
 // Beacon represents a single propagation beacon published to the cluster.
 message BeaconDesc {
+    // Name of the node that published this beacon.
+    string published_by = 3;
```
narqo, line 21: (nit) can this go to the end, to keep the order of the definitions aligned with the order of protobuf fields

pracucci: It could but I preferred to keep the "published" fields close to each other.

PR #927, opened 2026-03-04
perf: pre-resolve Prometheus label observers in grpcStatsHandler to eliminate per-message WithLabelValues overhead

julienduchesne:
## Problem

`grpcStatsHandler.HandleRPC` is invoked by the gRPC runtime on **every single message** of every streaming RPC. For high-throughput RPCs like `QueryStream` this means millions of calls per second. The original code called `WithLabelValues(gRPC, fullMethodName)` on every `InPayload` and `OutPayload` event:

```go
g.receivedPayloadSize.WithLabelValues(gRPC, fullMethodName).Observe(float64(s.WireLength))
g.sentPayloadSize.WithLabelValues(gRPC, fullMethodName).Observe(float64(s.WireLength))
```

Even when the metric already exists in the vec, every `WithLabelValues` call pays label string hashing, an RLock on the internal metric map, and a hash map lookup. This appeared as **10.1% of total CPU** in cortex-prod-10 production profiles.

## Fix

Pre-resolve all labeled metric handles exactly **once per stream** in `TagRPC` (fires at stream open time, not per message) and cache the resolved `prometheus.Observer`/`prometheus.Gauge` handles in the context. `HandleRPC` reads the cached `*rpcObservers` and calls `.Observe()`/`.Inc()`/`.Dec()` directly.

The no-observer fallback path has been removed — `TagRPC` always runs before `HandleRPC` in the gRPC lifecycle, so the fallback was dead code.

## Benchmark Results (Apple M3 Pro, arm64)

Measured with `-count=6` and compared using `benchstat`:

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                             │    before     │              after              │
                             │    sec/op     │   sec/op     vs base            │
HandleRPC_InPayload-11          66.39n ± 26%   12.25n ± 5%  -81.56% (p=0.002 n=6)
HandleRPC_OutPayload-11         64.67n ± 17%   12.32n ± 4%  -80.95% (p=0.002 n=6)
HandleRPC_StreamRoundTrip-11     6.676µ ±  6%   1.266µ ± 4%  -81.04% (p=0.002 n=6)
geomean                          306.0n         57.58n       -81.18%

                             │    before     │              after              │
                             │     B/op      │    B/op     vs base             │
HandleRPC_InPayload-11          0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)
HandleRPC_OutPayload-11         0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)
HandleRPC_StreamRoundTrip-11    128.0 ± 0%     128.0 ± 0%       ~ (p=1.000 n=6)

                             │    before     │              after              │
                             │  allocs/op    │ allocs/op   vs base             │
HandleRPC_InPayload-11          0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)
HandleRPC_OutPayload-11         0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=6)
HandleRPC_StreamRoundTrip-11    2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=6)
```

**~81% reduction in CPU time** per message. Zero allocations per message remain zero. All existing tests pass.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 63e6df1389b04fca066aa9de6e1b0b883af91f23. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/middleware/grpc_stats.go
+++ b/middleware/grpc_stats.go
@@ -48,9 +48,8 @@ type grpcStatsHandler struct {
 type contextKey int
 
 const (
-	contextKeyMethodName contextKey = 1
-	contextKeyRouteName  contextKey = 2
-	contextKeyConnID     contextKey = 3
+	contextKeyRouteName    contextKey = 2
+	contextKeyConnID       contextKey = 3
 	contextKeyRPCObservers contextKey = 4
 )
 
@@ -64,8 +63,6 @@ type rpcObservers struct {
 }
 
 func (g *grpcStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
-	ctx = context.WithValue(ctx, contextKeyMethodName, info.FullMethodName)
-
 	// Pre-resolve all labeled metric handles once per stream so that HandleRPC
 	// (called on every message) can call Observe/Inc/Dec without any label
 	// hashing, map lookups, or mutex acquisitions inside the HistogramVec/GaugeVec.
@@ -80,8 +77,6 @@ func (g *grpcStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) c
 func (g *grpcStatsHandler) HandleRPC(ctx context.Context, rpcStats stats.RPCStats) {
 	obs, ok := ctx.Value(contextKeyRPCObservers).(*rpcObservers)
 	if !ok {
-		// Fallback: no pre-resolved observers (should not happen in normal usage).
-		g.handleRPCFallback(ctx, rpcStats)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


middleware/grpc_stats.go
```diff
@@ -48,45 +48,63 @@ type grpcStatsHandler struct {
...
+// eliminating per-message label hashing and mutex contention.
+type rpcObservers struct {
+	receivedPayloadSize prometheus.Observer
+	sentPayloadSize     prometheus.Observer
+	inflightRequests    prometheus.Gauge
+}
+
 func (g *grpcStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
-	return context.WithValue(ctx, contextKeyMethodName, info.FullMethodName)
+	// Pre-resolve all labeled metric handles once per stream so that HandleRPC
```
bboreham, line 66: Most gRPC calls are one per stream ("unary"), so this could perhaps be improved further.

PR #935, opened 2026-03-18
feat(flagext): add environment variable support for CLI flags

### LINKED ISSUE #790: Support for environment variables and configuration files?
Would there be interest in adding support for environment variables to the supported configuration options for dskit?

If you want flags to override environment variables right now, it's tricky when the flags have default values.

For example, `server.http-listen-port` has a default value of 80, and when you call flags.Parse(), it overwrites the registered config struct with that value if the flag wasn't set. This means you need to do some kind of "config merge" between any environment variables you loaded, flags the user explicitly set, and default values that *also* end up set on the registered cfg. You need to distinguish between explicity settings and defaults using `flag.Visit()`, which is cumbersome

If these cases were handled by dskit there would be less code in most calling applications, and one place to fix bugs. We'd probably have to make this an opt-in feature to maintain backwards compatibility with current behavior.

joeblubaugh:
**What this PR does**:

Allow CLI flags to be overridden by automatically generated environment variables, with configurable prefix and correct precedence: CLI flags > env vars > defaults.

Fixes #790

Includes AI-generated code

---

> [!NOTE]
> **Medium Risk**
> Changes CLI flag parsing behavior by introducing environment-variable overrides, which can alter runtime configuration precedence. Risk is mitigated by explicit CLI precedence and extensive unit tests, but could still affect deployments relying on previous defaults.
> 
> **Overview**
> **Adds environment-variable overrides for CLI flags.** New helpers derive env var names from flag names (uppercased with `.`/`-` → `_`, optional prefix) and set any *unset* flags from the environment, returning joined errors when values can’t be parsed.
> 
> **Extends parsing helpers to opt into env support.** Introduces `ParseFlagsAndArgumentsWithEnv` and `ParseFlagsWithoutArgumentsWithEnv` to apply env values after `Parse()`, preserving the precedence *CLI > env > defaults* and keeping the existing “no positional args” enforcement.
> 
> Adds comprehensive tests covering name mapping, precedence, empty values, custom flag types, multiple errors, and real env integration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 772be7c5e47962a3def4475242935d3bf47c511e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/flagext/env.go
+++ b/flagext/env.go
@@ -12,20 +12,21 @@ var flagNameReplacer = strings.NewReplacer(".", "_", "-", "_")
 
 // FlagToEnvVar converts a flag name to its corresponding environment variable name.
 // The flag name is uppercased and all dots, dashes, and underscores in the flag name
-// are replaced with (or kept as) underscores. The prefix is used as-is and should be
-// provided in uppercase. If prefix is non-empty it is prepended with an underscore
-// separator.
+// are replaced with (or kept as) underscores. The prefix is similarly upper-cased with
+// symbol substitution. If prefix is non-empty, an _ is placed between the flag name and prefix
 //
 // Note that distinct flag names can map to the same environment variable. For example,
 // "a.b" and "a-b" both map to "A_B". Callers should avoid registering flags whose
 // names differ only in dots, dashes, and underscores.
 //
-// For example, FlagToEnvVar("MYAPP", "server.http-listen-port") returns
-// "MYAPP_SERVER_HTTP_LISTEN_PORT".
+// For example, FlagToEnvVar("myapp-name", "server.http-listen-port") returns
+// "MYAPP_NAME_SERVER_HTTP_LISTEN_PORT".
 func FlagToEnvVar(prefix, flagName string) string {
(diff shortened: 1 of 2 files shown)


flagext/env.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"os"
+	"strings"
+)
+
+var flagNameReplacer = strings.NewReplacer(".", "_", "-", "_")
+
+// FlagToEnvVar converts a flag name to its corresponding environment variable name.
+// The flag name is uppercased and all dots, dashes, and underscores in the flag name
+// are replaced with (or kept as) underscores. The prefix is used as-is and should be
```
56quarters: Is there value in letting people use lower case here instead of converting the prefix to uppercase?

joeblubaugh: Good point. It's probably common to re-use the app name in multiple places: the metrics namespace, trace attributes, and probably env var prefixes. Most of those uses will be lower or mixed-case

PR #939, opened 2026-03-23
feat(rate-limiter): add method to remove rate limiters that have not been accessed after a specific cutoff time

juliusmh:
**What this PR does**:

This PR adds `RemoveStaleEntries` to `RateLimiter`, which cleans up per-tenant trackers that have not been accessed after a specified cutoff date.
Callers passing ephemeral tenant IDs can use it to prevent unbounded growth of the rate-limiter's state.

To be used with https://github.com/grafana/mimir/pull/14289.

n/a

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> **Medium Risk**
> Touches shared in-memory state and locking in the multi-tenant rate limiter; incorrect cutoff semantics or races could cause unexpected limiter recreation/behavior under load, though the change is small and covered by tests.
> 
> **Overview**
> Adds `RateLimiter.RemoveStaleEntries(cutoff)` to evict per-tenant limiter entries whose `recheckAt` precedes a caller-provided cutoff, preventing unbounded growth when tenant IDs are ephemeral.
> 
> Updates `recheckTenantLimiter` to safely handle tenants removed from the map (recreate the limiter if missing) and adds a unit test covering eviction and subsequent limiter reuse/recreation behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a9cc4d0b0a0785d87050f93a63613a57dc5afee9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/limiter/rate_limiter.go
+++ b/limiter/rate_limiter.go
@@ -66,6 +66,21 @@ func (l *RateLimiter) Burst(now time.Time, tenantID string) int {
 	return l.getTenantLimiter(now, tenantID).Burst()
 }
 
+// RemoveStaleEntries removes entries that have not been accessed since
+// the given cutoff time and returns the number of removed entries.
+func (l *RateLimiter) RemoveStaleEntries(cutoff time.Time) int {
+	l.tenantsLock.Lock()
+	defer l.tenantsLock.Unlock()
+	removed := 0
+	for tenantID, entry := range l.tenants {
+		if entry.recheckAt.Before(cutoff) {
+			delete(l.tenants, tenantID)
+			removed++
+		}
+	}
+	return removed
+}
+
 func (l *RateLimiter) getTenantLimiter(now time.Time, tenantID string) *rate.Limiter {
 	recheck := false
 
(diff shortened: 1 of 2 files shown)


limiter/rate_limiter.go
```diff
@@ -66,6 +66,21 @@ func (l *RateLimiter) Burst(now time.Time, tenantID string) int {
 	return l.getTenantLimiter(now, tenantID).Burst()
 }
 
+// RemoveStaleEntries removes entries that have not been accessed since
+// the given cutoff time and returns the number of removed entries.
+func (l *RateLimiter) RemoveStaleEntries(cutoff time.Time) int {
```
tcard, line 71: **suggestion:** This puts the burden of calling this periodically on the library user, who probably doesn't care much when this happens, as long as it happens from time to time.

How about we add a "remove stale" option to `NewRateLimiter` and if set, `getTenantLimiter` sometimes calls `RemoveStaleEntries` (which we may keep public, in case the caller _does_ care)? Maybe in a separate goroutine.

I'd suggest a functional option for this, which we use in other places in dskit too.

limiter/rate_limiter.go
```diff
@@ -66,6 +66,21 @@ func (l *RateLimiter) Burst(now time.Time, tenantID string) int {
...
+	l.tenantsLock.Lock()
```
tcard, line 72: **thought:** I wonder if it's worth having `tenantsLock` protect only the `tenants` map itself, and then have each `tenantLimiter` have its own lock. This would allow us to just fetch the list of tenants under `tenantsLock`, then compute which ones should be deleted under individual `tenantLimiter.mtx`, then finally take `tenantsLock` and recheck+remove, reducing the critical sections length.

PR #942, opened 2026-03-26
ring: clamp far-future heartbeat timestamps to prevent permanently stuck instances

domgoodwin:
**What this PR does**:
A crash can corrupt a ring instance's heartbeat timestamp to a
far-future value (e.g. year 2034). When this happens, the instance
becomes permanently stuck in the gossip ring because:

- The merge logic uses "highest timestamp wins", so no legitimate heartbeat (using time.Now()) can ever override the corrupted entry
- The localCAS path marks missing instances as LEFT with the current time, but other nodes reject this via gossip since their copy has the higher corrupted timestamp
- IsHeartbeatHealthy treats future timestamps as healthy (the time difference is negative, always within the timeout), so the instance can never be considered unhealthy or evicted

Fix by clamping timestamps that exceed now + 1 hour (maxAllowedClockDrift)
to the current time in both the merge function and IsHeartbeatHealthy.
The merge fix is self-healing: each node independently clamps corrupted
entries and gossips the correction, allowing normal operations to resume
on the next heartbeat cycle.

Fixes https://github.com/grafana/mimir/issues/13810

That issue is the same issue we encountered on our cluster: we had an 
instance with a heartbeat timestamp in 2034 in UP status, the only way
we could fix it was by having a fake member join, emit a DOWN status for
the broken instance with a timestamp into the future (to beat out the 2034
year).

**Checklist**
- [ ] Tests updated

---

> [!NOTE]
> **Medium Risk**
> Changes ring gossip merge and heartbeat health semantics by clamping future timestamps; mistakes could cause instances to be marked unhealthy or allow stale state to win, but the change is bounded by a 1h drift window and covered by targeted tests.
> 
> **Overview**
> Prevents corrupted far-future heartbeat timestamps from permanently winning ring gossip merges by introducing `maxAllowedClockDrift` (1h) and clamping any timestamp beyond `now + drift` back to `now`.
> 
> `mergeWithTime` now *self-heals* local corrupted entries (and gossips the correction via the returned change set) and clamps incoming entries before doing the "highest timestamp wins" comparison; `IsHeartbeatHealthy` now treats far-future heartbeats as unhealthy instead of always healthy.
> 
> Adds unit tests covering clamping behavior, merge outcomes (including `localCAS` marking missing instances `LEFT`), and acceptance of small legitimate clock skew.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec87a4cf52f6bc179ffe9a418ab9045220dd0797. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ring/merge_test.go
+++ b/ring/merge_test.go
@@ -480,6 +480,155 @@ func TestMergeMissingIntoLeft(t *testing.T) {
 	}
 }
 
+func TestMergeFarFutureTimestamp_IncomingClamped(t *testing.T) {
+	now := time.Now()
+	nowUnix := now.Unix()
+	farFuture := time.Date(2034, 1, 1, 0, 0, 0, 0, time.UTC).Unix()
+
+	// Local ring has instance with current timestamp.
+	local := &Desc{
+		Ingesters: map[string]InstanceDesc{
+			"Ing 1": {Addr: "addr1", Timestamp: nowUnix, State: ACTIVE, Tokens: []uint32{30, 40, 50}},
+		},
+	}
+
+	// Incoming ring has same instance with a corrupted far-future timestamp.
+	incoming := &Desc{
+		Ingesters: map[string]InstanceDesc{
+			"Ing 1": {Addr: "addr1", Timestamp: farFuture, State: ACTIVE, Tokens: []uint32{30, 40, 50}},
+		},
+	}
+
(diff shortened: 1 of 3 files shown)


pracucci:

```
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

Thanks!

PR #987, opened 2026-05-29
feat(log): sanitize control characters in user-controlled log values

duricanikolic:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated

### DIFF
--- a/log/sanitize.go
+++ b/log/sanitize.go
@@ -1,0 +1,284 @@
+package log
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"unicode"
+
+	"github.com/go-kit/log"
+)
+
+// Sanitizer transforms a string by removing or escaping characters that
+// could be used to inject log lines, fake terminal output, confuse log
+// parsers, or hide source via bidi overrides. Two implementations are
+// provided: DropControlChars and EscapeControlChars. Callers select one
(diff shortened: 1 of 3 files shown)


log/sanitize.go
```diff
@@ -0,0 +1,284 @@
...
+//	var mode string
+//	flag.StringVar(&mode, "log.sanitization", log.SanitizerNameDrop,
+//	    "Sanitizer for user-controlled log values: drop or escape.")
+//	// ... after flag.Parse() ...
+//	s, err := log.SanitizerFor(mode)
+//	if err != nil {
+//	    return err
+//	}
+//	logger = log.NewSanitizingLogger(logger, s)
+func SanitizerFor(name string) (Sanitizer, error) {
```
56quarters, line 42: "none" should be an option since I'm pretty sure this is going to come with a performance penalty.

56quarters (commented):
This code feels AI generated and I'd appreciate if it was cleaned up or rewritten before final review.

Overall I think instead of having an entirely custom logger implementation we start with `UserControlledValue` and selectively apply it to places we know need it. This avoids adding new code in our logging calls which are pretty hot.

We can add a custom logger that looks for control characters in the future if the special type isn't enough or is too error prone.

PR #995, opened 2026-06-03
fix(ring): always render read-only state on the ring status page

fedetorres93:
**What this PR does**:

The ring status page left the "Read-Only" column blank for instances that aren't read-only. Because the "Read-Only Updated" column still rendered the timestamp of the last read-only change, an instance that had switched back to read-write showed a lone timestamp next to a blank cell, which reads as a "stale" state.

This change always renders the boolean in the "Read-Only" column to remove this ambiguity.

Fixes #<issue number>

**Checklist**
- [x] Tests updated

---

> [!NOTE]
> **Low Risk**
> Display-only change to the ring status HTML template plus regression tests; no ring membership or read-only logic changes.
> 
> **Overview**
> Fixes ambiguous ring status UI where **Read-Only** was blank for non-read-only instances while **Read-Only Updated** could still show a timestamp (e.g. after switching back to read-write), which looked like a stale read-only state.
> 
> The **Read-Only** column in `ring_status.gohtml` now always renders the boolean (`true`/`false`) for every instance. **Read-Only Updated** formatting is unchanged: relative “ago” when read-only, formatted time when not read-only but a change timestamp exists, empty when there was never a read-only transition.
> 
> Adds an HTTP handler test covering read-only, formerly read-only, and never-read-only instances.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c455021c7ee5c0a1a7ae60f38d90ce678d67165. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ring/ring_http_test.go
+++ b/ring/ring_http_test.go
@@ -78,6 +78,67 @@ func TestRingPageHandler_handle(t *testing.T) {
 		}, `\s*`))), recorder.Body.String())
 	})
 
+	t.Run("always renders the read-only state and keeps the read-only updated timestamp", func(t *testing.T) {
+		readOnlyChange := now.Add(-time.Hour)
+		roRing := fakeRingAccess{
+			desc: &Desc{
+				Ingesters: map[string]InstanceDesc{
+					"read-only": {
+						Zone:                     "zone-a",
+						State:                    ACTIVE,
+						Addr:                     "addr-ro",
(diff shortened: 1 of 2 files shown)


seizethedave (approved):
LGTM!

PR #996, opened 2026-06-04
fix: cache: Allow additional discovery configuration for pooling, period

### LINKED ISSUE #991: Configurable memcached client dns update interval
Hi,

would you accept a PR making the dns provider update interval of the memcached client (https://github.com/grafana/dskit/blob/ee6783019c135d6d110d1f1cd2ec54ed8c465fd8/cache/memcached_client.go#L32) configurable?

The default value is causing some minor issues for us as systemd-resolved closes connections after 10s of inactivity (no query), so every update would receive a connection reset and open a new connection. As we have configured our resolv.conf with `attempts:1` this causes failures.

56quarters:
**What this PR does**:

Allow how often new cache servers are discovered to be configured and allow pooling of connections to the DNS servers used for discovery to be disabled.

This fixes an issue where systemd-resolved aggressively closes idle TCP connections after 10s. The combination of pooling and only doing discovery every 30s meant that this would always fail after startup unless retries were also configured.

Fixes #991

Required for https://github.com/grafana/mimir/issues/15565

**Checklist**
- [X] Tests updated

---

> [!NOTE]
> **Medium Risk**
> Changes default memcached DNS resolver pooling (now off) and breaks the `dns.NewProvider` call signature for downstream importers; mis-tuned lookup period could affect cache node discovery timing.
> 
> **Overview**
> Adds **Memcached** discovery knobs so operators can tune how often cache server addresses are re-resolved and whether the miekgdns resolver keeps idle TCP connections to upstream DNS.
> 
> **`addresses_lookup_period`** (default still 30s) drives the background refresh loop instead of a fixed constant, with validation that it must be positive. **`addresses_lookup_pool_size`** (default **0**) is passed into `dns.NewProvider` and the miekgdns pooling client so idle DNS connections can be disabled—addressing failures when resolvers (e.g. systemd-resolved) close idle TCP before the next 30s lookup.
> 
> `dns.NewProvider` is refactored to take resolver type and max idle connections up front; miekgdns pool sizes are **`uint`**, and tests cover pooling disabled vs enabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e46c4e385f4623db4759038d55ec2c5afb23a28b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cache/memcached_client.go
+++ b/cache/memcached_client.go
@@ -75,6 +75,14 @@ type MemcachedClientConfig struct {
 	// resolved with the DNS provider.
 	Addresses flagext.StringSliceCSV `yaml:"addresses"`
 
+	// AddressesLookupPeriod specifies how often addresses are resolved with the
+	// DNS provider.
+	AddressesLookupPeriod time.Duration `yaml:"addresses_lookup_period" category:"advanced"`
+
+	// AddressesLookupPoolSize specifies how many idle connections to the DNS provider
+	// are kept open. Use 0 to disable keeping any idle connections open.
+	AddressesLookupPoolSize uint `yaml:"addresses_lookup_pool_size" category:"advanced"`
+
 	// Timeout specifies the socket read/write timeout.
 	Timeout time.Duration `yaml:"timeout"`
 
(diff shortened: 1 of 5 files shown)


cache/memcached_client.go
```diff
@@ -143,6 +154,10 @@ func (c *MemcachedClientConfig) Validate() error {
 		return ErrMemcachedMaxAsyncConcurrencyNotPositive
 	}
 
+	if c.AddressesLookupPeriod <= 0 {
+		return ErrMemcachedAddressLookupPeriodNotPositive
```
aknuds1, line 158: Might it make sense to fall back to the default here, instead of returning an error, for backwards compatibility?

```suggestion
		c.AddressesLookupPeriod = dnsProviderUpdateInterval
```

aknuds1 (commented):
Please see question.

cache/memcached_client.go
56quarters: We could but this is already the behavior because we set the default as part of the CLI flags. I'd rather not have a default / fallback in two places.

56quarters: An example of this is [here in the corresponding Mimir PR](https://github.com/grafana/mimir/pull/15573/changes#diff-4235a2573c6bcba1dceb19097b113efd233483fa05ad12f7fffe935962d6f208)

aknuds1: Aha.

aknuds1 (commented):
I think you need to fix the call to `dns.NewProvider` in ring/example/local/local.go also.

56quarters:
> I think you need to fix the call to `dns.NewProvider` in ring/example/local/local.go also.

It's pulling in an older commit of dskit and so it doesn't have the new code. I'll update in a follow-up PR after this is in `main`.

alexweav (approved):
LGTM, straightforward

npazosmendez (approved):

PR #1010, opened 2026-06-25
docs(cache): correct WrapWithLRUCache expiry behavior in godoc

sueun-dev:
**What this PR does**:

The godoc for `WrapWithLRUCache` says:

> The LRU cache will also remove items from the underlying cache if they are expired.

That isn't what the code does. When an entry is expired in the LRU, `GetMultiWithError` only evicts it from the in-memory LRU and re-fetches it from the underlying cache:

```go
if item.ExpiresAt.After(now) {
    found[k] = item.Data
    continue
}
l.lru.Remove(k)
miss = append(miss, k)
```

The only place the underlying cache is deleted from is the explicit `Delete` method. The expiry path never calls `l.c.Delete`, so the underlying cache is never touched on expiry. The sentence has been inaccurate since the package was added.

This reworks the sentence to describe the actual behavior and adds a test that asserts the underlying cache is never asked to delete an expired entry (using `InstrumentedMockCache.CountDeleteCalls`).

None.

**Checklist**
- [x] Tests updated

### DIFF
--- a/cache/lru.go
+++ b/cache/lru.go
@@ -36,7 +36,7 @@ type Item struct {
 // WrapWithLRUCache wraps a given `Cache` c with a LRU cache. The LRU cache will always store items in both caches.
 // However it will only fetch items from the underlying cache if the LRU cache doesn't have the item.
 // Items fetched from the underlying cache will be stored in the LRU cache with a default TTL.
-// The LRU cache will also remove items from the underlying cache if they are expired.
+// Items that have expired in the LRU cache are evicted from it on read and re-fetched from the underlying cache.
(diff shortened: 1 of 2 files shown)


bboreham (commented):
Thanks for your contribution.

I think we can accept the clarification in the docstring, but not the test.
Exactly when evictions happen is an implementation detail.

If you can say a bit more about why you need this distinction made clear, maybe I would understand.

And the CLA.

bboreham (commented):
Thanks for your contribution.

I think we can accept the clarification in the docstring, but not the test.
Exactly when evictions happen is an implementation detail.

If you can say a bit more about why you need this distinction made clear, maybe I would understand.

And the CLA.

PR #1013, opened 2026-07-03
feat(ring): add PartitionRingWatchers and PartitionInstanceRings

pracucci:
**What this PR does**:

Adds two general-purpose helpers to the `ring` package:

- `PartitionRingWatchers` groups one or more `PartitionRingWatcher` and manages their lifecycle as a single service, exposing them by index. This is useful when a process watches several partition rings at once (e.g. one ring per shard, tenant group, or logical partition set) and wants to start, stop, and monitor them together.
- `PartitionInstanceRings` builds one `PartitionInstanceRing` per watcher in such a group, each pairing that watcher's partition ring with a shared instance ring.

These were originally written in Mimir and are generic enough to live in dskit, so other dskit consumers can reuse them. The dskit versions are decoupled from any application-specific ring naming: callers build the individual watchers (choosing their own names and KV keys) and pass them in.

(none)

**Checklist**
- [x] Tests updated

### DIFF
--- a/ring/partition_instance_rings.go
+++ b/ring/partition_instance_rings.go
@@ -8,7 +8,7 @@ import (
 // group, each pairing that watcher's partition ring with a shared instance ring. Rings are accessed
 // by the same index used by the underlying PartitionRingWatchers group.
 type PartitionInstanceRings struct {
-	// rings is indexed by position in the group. It always has at least one element.
+	// rings is indexed by position in the group.
 	rings []*PartitionInstanceRing
 }
 
--- a/ring/partition_instance_rings.go
+++ b/ring/partition_instance_rings.go
@@ -1,0 +1,33 @@
+package ring
+
+import (
+	"time"
+)
+
+// PartitionInstanceRings holds one PartitionInstanceRing per watcher in a PartitionRingWatchers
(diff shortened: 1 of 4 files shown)


ring/partition_instance_rings.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"time"
+)
+
+// PartitionInstanceRings holds one PartitionInstanceRing per watcher in a PartitionRingWatchers
+// group, each pairing that watcher's partition ring with a shared instance ring. Rings are accessed
+// by the same index used by the underlying PartitionRingWatchers group.
+type PartitionInstanceRings struct {
+	// rings is indexed by position in the group. It always has at least one element.
```
fionaliao: > It always has at least one element.

Is this an important property? If so, should we check for `len(watchers) > 0` in `NewPartitionInstanceRings`?

pracucci: Good question. I don't think it's an important property. We should just remove this part of the comment.

pracucci: Commented updated in https://github.com/grafana/dskit/pull/1013/commits/fe1e0f446ea89c4561bcc408c929831b4197c5c1

fionaliao (approved):

PR #1021, opened 2026-07-16
fix(loser): calling Push could result in seeing the same element twice

bboreham:
**What this PR does**:

Based on part of https://github.com/grafana/loki/pull/23295 from @jnewbigin

`loser.Tree.Push` flagged a full re-initialize on the next `Next()`, without advancing the previous winner.
A push after iteration had started therefore caused the current winner's already-returned value to be emitted a second time. 

**Checklist**
- [x] Tests updated


jnewbigin (approved):
LGTM

PR #1038, opened 2026-07-23
perf(ring): replace reflection-based proto.Clone in Desc.Clone with maps.Clone

julienduchesne:
## Summary

`Desc.Clone()` runs on the memberlist hot path (once per watcher notification, via `KV.get`), and its reflection-based `proto.Clone` gets expensive on large rings: in a CPU profile of a Mimir store-gateway in a cluster with thousands of memberlist nodes, this clone alone accounted for ~4.7% of total service CPU. Replace it with a plain `maps.Clone` of the `Ingesters` map, which is 4-8x faster and allocation-free per instance.

- This is not a semantics change: gogo's `proto.Clone` never deep-copied the instance internals. Non-nullable map values (`map[string]InstanceDesc`) fall into the shallow `default` branch of gogo's table-driven merge ([table_merge.go#L591-L624](https://github.com/gogo/protobuf/blob/v1.3.2/proto/table_merge.go#L591-L624)), so `Tokens` and `Versions` were already shared between the clone and the original.
- Document the actual contract on `Clone()` (independent map, read-only `Tokens`/`Versions`) instead of the previous "deep copy" comment, which was inaccurate.
- Pin the behavior with a test that requires every `InstanceDesc` field to be non-zero, so adding a field forces a review of its sharing semantics.

### Benchmarks

512 tokens per instance, Apple M3 Pro (`benchstat`, n=6):

```
                           │  proto.Clone  │              maps.Clone             │
                           │    sec/op     │    sec/op     vs base               │
Desc_Clone/instances=100        17.219µ ± 11%   4.230µ ±  7%  -75.44% (p=0.002 n=6)
Desc_Clone/instances=1000       223.55µ ±  8%   42.57µ ± 30%  -80.96% (p=0.002 n=6)
Desc_Clone/instances=10000      1814.8µ ±  3%   214.4µ ± 19%  -88.19% (p=0.002 n=6)
geomean                          191.2µ         33.80µ        -82.32%

                           │     B/op      │     B/op      vs base               │
Desc_Clone/instances=100         52.25Ki ± 0%   18.09Ki ± 0%  -65.37% (p=0.002 n=6)
Desc_Clone/instances=1000        744.4Ki ± 0%   288.1Ki ± 0%  -61.29% (p=0.002 n=6)
Desc_Clone/instances=10000       6.112Mi ± 0%   2.251Mi ± 0%  -63.18% (p=0.002 n=6)
geomean                          624.4Ki        229.0Ki       -63.32%

                           │   allocs/op   │  allocs/op   vs base                │
Desc_Clone/instances=100         213.000 ± 0%    5.000 ± 0%  -97.65% (p=0.002 n=6)
Desc_Clone/instances=1000       2024.000 ± 0%    7.000 ± 0%  -99.65% (p=0.002 n=6)
Desc_Clone/instances=10000      20083.00 ± 0%   35.00 ± 0%   -99.83% (p=0.002 n=6)
geomean                           2.053k         10.70        -99.48%
```

### DIFF
--- a/ring/model.go
+++ b/ring/model.go
@@ -2,6 +2,7 @@ package ring
 
 import (
 	"fmt"
+	"maps"
 	"math"
 	"sort"
 	"sync"
@@ -469,9 +470,21 @@ func (d *Desc) RemoveTombstones(limit time.Time) (total, removed int) {
 	return
 }
 
-// Clone returns a deep copy of the ring state.
+// Clone returns a copy of the ring state. The returned Desc and its Ingesters map are
+// safe to modify (entries can be added, removed or replaced), but reference-type fields
+// inside InstanceDesc (Tokens, Versions) share their underlying storage with the
+// original, so they must be treated as read-only.
+//
+// It's hand-written instead of using the reflection-based proto.Clone because it runs on
+// the memberlist hot path (once per watcher notification), where reflection is a
+// significant CPU cost on rings with thousands of instances. Sharing Tokens and Versions
+// storage is not a behavior change: gogo's proto.Clone shallow-copies non-nullable map
+// values, so it shared those fields too.
 func (d *Desc) Clone() memberlist.Mergeable {
-	return proto.Clone(d).(*Desc)
+	if d == nil {
+		return (*Desc)(nil)
+	}
+	return &Desc{Ingesters: maps.Clone(d.Ingesters)}
 }
 
 func (d *Desc) getTokensInfo() map[uint32]instanceInfo {
--- a/ring/model_test.go
+++ b/ring/model_test.go
@@ -1,6 +1,8 @@
 package ring
 
 import (
+	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


ring/model.go
```diff
@@ -469,9 +470,21 @@ func (d *Desc) RemoveTombstones(limit time.Time) (total, removed int) {
...
 
-// Clone returns a deep copy of the ring state.
+// Clone returns a copy of the ring state. The returned Desc and its Ingesters map are
+// safe to modify (entries can be added, removed or replaced), but reference-type fields
+// inside InstanceDesc (Tokens, Versions) share their underlying storage with the
+// original, so they must be treated as read-only.
+//
+// It's hand-written instead of using the reflection-based proto.Clone because it runs on
+// the memberlist hot path (once per watcher notification), where reflection is a
+// significant CPU cost on rings with thousands of instances. Sharing Tokens and Versions
```
56quarters, line 480: Seems weird to describe how this worked in the past in a comment.

56quarters (approved):
LGTM with a note that the comment about days of yore seems odd.

ring/model.go
julienduchesne: Yeah, I was on the fence about keeping this or not. On the side of keeping it is that someone might wonder: why not make this simpler and not use proto.Clone (the obvious option for a proto struct)?

bboreham: I think "Coded explicitly instead of the much slower reflection-based proto.Clone" would be fine.
Comments should describe why the code works the way it does now.
Description of how this changes compared to the past goes in git.

PR #1039, opened 2026-07-23
feat: gate: add a gate.Gate variant with a timeout

56quarters:
**What this PR does**:

Add a new `gate.Gate` implementation that can timeout after waiting for an execution slot. This is a port of the same logic from Mimir.

N/A

**Checklist**
- [X] Tests updated

### DIFF
--- a/gate/gate.go
+++ b/gate/gate.go
@@ -12,6 +12,8 @@ import (
 
 var ErrMaxConcurrent = errors.New("max concurrent requests inflight")
 
+var ErrGateTimeout = errors.New("timeout waiting for concurrency gate")
+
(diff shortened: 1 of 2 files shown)


56quarters:
[Mimir implementation](https://github.com/grafana/mimir/blob/9e1501cb8960ec6d499f288424101034a142c6bf/pkg/storegateway/bucket_stores.go#L526-L561).

56quarters:
The motivation for this is to allow us to try adding a timeout to the gate use in the Memcached client: it'd be nice to use more aggressive connection and concurrency limits in store-gateways in Mimir but doing that today means we block requests indefinitely waiting on the cache.

julienduchesne (approved):

PR #1040, opened 2026-07-23
perf(ring): skip token comparison in RingCompare for slices sharing storage

julienduchesne:
## Summary

`RingCompare` runs on every ring watcher notification and compares all tokens of all instances element-wise (512 × N on typical rings), even for pure heartbeat updates where no tokens changed. On the same profile that motivated #1038 (Mimir store-gateway, thousands of memberlist nodes), it accounted for ~6% of total service CPU. Skip the element-wise comparison when both slices point at the same backing array.

- Consecutive ring states handed to a watcher are clones of the memberlist KV store's value. The store's merge only replaces map entries of changed instances and `Desc.Clone()` shares `Tokens` backing arrays, so for unchanged instances both states reference the same storage — pointer equality (`unsafe.SliceData`) proves token equality without looking at the elements.
- Correctness doesn't depend on the sharing: slices with different storage but equal content still compare equal through the existing element-wise loop. A length mismatch is caught before the pointer check, so prefix-aliased slices are handled.
- The benchmark covers both provenances: `token storage=shared` (heartbeat-style update, the steady-state case for memberlist-backed rings) and `token storage=unshared` (independently decoded states, e.g. consul/etcd backends), which exercises the fallback.

### Benchmarks

512 tokens per instance, Apple M3 Pro (`benchstat`, n=6, interleaved runs):

```
                                                  │  before   │            after                    │
                                                  │  sec/op   │    sec/op     vs base               │
RingCompare/instances=100/token storage=shared      28.194µ ± 29%   3.553µ ± 30%  -87.40% (p=0.002 n=6)
RingCompare/instances=100/token storage=unshared     26.23µ ± 47%   26.98µ ± 24%        ~ (p=0.589 n=6)
RingCompare/instances=1000/token storage=shared     216.42µ ± 10%   31.90µ ± 18%  -85.26% (p=0.002 n=6)
RingCompare/instances=1000/token storage=unshared    200.1µ ± 44%   211.7µ ±  3%        ~ (p=0.065 n=6)
RingCompare/instances=10000/token storage=shared    3552.9µ ±  8%   264.3µ ±  7%  -92.56% (p=0.002 n=6)
RingCompare/instances=10000/token storage=unshared   3.125m ± 12%   3.483m ± 16%        ~ (p=0.180 n=6)
```

Both paths are allocation-free (0 B/op before and after).

### DIFF
--- a/ring/model.go
+++ b/ring/model.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"sync"
 	"time"
+	"unsafe"
 
 	"github.com/gogo/protobuf/proto"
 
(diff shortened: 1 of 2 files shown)


56quarters (approved):
Nice.

PR #1043, opened 2026-07-24
fix(runtimeconfig): preserve provider order in config hash

aknuds1:
**What this PR does**:

Follow-up to #1033 that makes the MapLoader runtime-config fingerprint respect provider order.

Runtime-config providers are merged from left to right, so reversing providers can change the effective configuration. The original MapLoader fingerprint stored source hashes in a map and sorted provider names, allowing different effective configurations to expose the same `runtime_config_hash`.

This change:

- keeps provider hashes in `LoadPath` order for reload detection and fingerprinting;
- hashes an unambiguous sequence of length-prefixed provider names and fixed-size content digests, preserving duplicate provider names;
- adds an end-to-end regression test that verifies both merge precedence and the exported hash.

MapLoader continues to receive the merged map directly without YAML re-encoding. There are no public API changes.

**Benchmarks**:

Command for each sample:

```
go test ./runtimeconfig -run '^$' -bench '^BenchmarkManagerLoadConfig/MapLoader$' -benchmem -benchtime=5x
```

Environment: Apple M4 Pro, `darwin/arm64`, six interleaved samples per revision. The same benchmark-only MapLoader harness was applied to base `a26df7e5` and head `c06088d1`; the harness is not part of this PR.

| Metric | Toni base | Head | Change |
| --- | ---: | ---: | ---: |
| Time | 96.85 ms/op | 96.46 ms/op | statistically unchanged (`p=0.818`) |
| Bytes allocated | 80.73 MiB/op | 80.70 MiB/op | statistically unchanged (`p=0.180`) |
| Allocations | 1.341M/op | 1.341M/op | about 18 fewer allocations/op (`p=0.002`) |

Validation:

- `go test ./runtimeconfig`
- `go test -race ./runtimeconfig`
- `go test ./...`
- `go vet ./runtimeconfig`
- `make lint`
- `git diff --check origin/runtimeconfig-optimize-unmarshal...HEAD`

Follow-up to #1033.

**Checklist**
- [x] Tests updated

### DIFF
--- a/runtimeconfig/manager.go
+++ b/runtimeconfig/manager.go
@@ -5,12 +5,13 @@ import (
 	"compress/gzip"
 	"context"
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
 	"net/http"
-	"sort"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -37,6 +38,11 @@ type Loader func(r io.Reader) (interface{}, error)
 
 // MapLoader loads the configuration from a map.
 type MapLoader func(m map[string]interface{}) (interface{}, error)
+
+type providerHash struct {
+	name   string
+	digest [sha256.Size]byte
+}
 
 // Config holds the config for an Manager instance.
 // It holds config related to loading per-tenant config.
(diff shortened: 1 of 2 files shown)


runtimeconfig/manager.go
```diff
@@ -301,19 +301,14 @@ func (om *Manager) loadConfig(ctx context.Context) error {
...
+func combinedFilesHash(hashes []providerHash) [sha256.Size]byte {
 	h := sha256.New()
-	for _, name := range names {
-		_, _ = io.WriteString(h, name)
-		_, _ = io.WriteString(h, "=")
-		_, _ = io.WriteString(h, hashes[name])
-		_, _ = io.WriteString(h, ";")
+	var nameLength [8]byte
+	for _, providerHash := range hashes {
+		binary.BigEndian.PutUint64(nameLength[:], uint64(len(providerHash.name)))
```
tcard, line 308: Why is this writing the provider name length into the hash?

tcard (approved):

runtimeconfig/manager.go
aknuds1: The idea is to prevent different tuples from producing the same hash. I added a corresponding test case.

PR #1047, opened 2026-08-03
fix(memberlist): stop key worker goroutines when obsolete keys are removed

aldernero:
### What this PR does

Fixes a goroutine leak in the memberlist KV client.

`NotifyMsg` spawns one worker goroutine per distinct KV key (`getKeyWorkerChannel` / `processValueUpdate`) to process incoming value updates. The worker was only ever stopped at full KV shutdown — nothing removed the `workersChannels` entry when a key was deleted and later purged from the store by `cleanupObsoleteEntries`. In clusters where keys are created and deleted over time (e.g. per-tenant keys), every retired key leaks one goroutine until the process restarts.

We spotted this in production: per-pod goroutine counts grow linearly (~+15%/week, r>0.8 correlation with pod age) and reset at every rollout. A goroutine profile shows `processValueUpdate` holding >50% of all goroutines in a process where the memberlist KV client dominates, with the count matching live keys right after a restart and exceeding it as keys churn. The leak affects every memberlist-enabled process; it's just less visible against larger goroutine baselines.

Changes:

- `cleanupObsoleteEntries` now collects the keys it removes from the store and stops their workers (`stopKeyWorkers`): the map entry is deleted and the update channel closed, letting the worker drain buffered updates and exit.
- The channel send in `NotifyMsg` moved under `workersMu` (into `enqueueKeyUpdate`) so closing a worker channel cannot race with a sender. `NotifyMsg` already acquired this mutex once per message via `getKeyWorkerChannel`, so this adds no new lock acquisition — it only extends the critical section by a non-blocking send.
- If a removed key is seen again later, a new worker is spawned on demand, same as for a brand-new key.

### Which issue(s) this PR fixes

n/a

### Checklist
- [x] Tests updated (`TestCleanupObsoleteEntriesStopsKeyWorkers`, passes with `-race`)
- [ ] `CHANGELOG.md` updated (no changelog in this repo)

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -1553,11 +1553,51 @@ func (m *KV) processValueUpdate(workerCh <-chan valueUpdate, key string) {
 				m.broadcastNewValue(key, mod, version, update.codec, false, deleted, updated)
 			}
 
+			if version == 0 && !m.keyExists(key) {
+				// The update didn't (re)create the key in the store — e.g. a tombstone
+				// received for a key that was already removed by cleanupObsoleteEntries,
+				// which is a no-op merge. The cleanup will therefore never stop this
+				// worker, so it deregisters itself instead of staying around forever.
+				if m.tryDeregisterKeyWorker(key, workerCh) {
+					return
+				}
+			}
+
 		case <-m.shutdown:
 			// stop running on shutdown
 			return
 		}
 	}
+}
+
+func (m *KV) keyExists(key string) bool {
+	m.storeMu.Lock()
+	defer m.storeMu.Unlock()
+
+	_, ok := m.store[key]
+	return ok
+}
+
+// tryDeregisterKeyWorker removes the worker's entry from workersChannels, and returns
+// true if the worker must stop. It returns false if the worker has to keep running,
+// either because more updates were enqueued to it in the meantime, or because it's not
+// the registered worker for the key anymore (stopKeyWorkers removed it already, in
+// which case the closed channel stops the worker instead).
+func (m *KV) tryDeregisterKeyWorker(key string, workerCh <-chan valueUpdate) bool {
+	m.workersMu.Lock()
+	defer m.workersMu.Unlock()
+
+	if ch, ok := m.workersChannels[key]; !ok || ch != workerCh {
+		return false
+	}
+
+	// Sends happen while holding workersMu, so the channel length cannot change here.
+	if len(workerCh) > 0 {
+		return false
+	}
+
+	delete(m.workersChannels, key)
+	return true
 }
 
 // GetBroadcasts is method from Memberlist Delegate interface
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -1479,16 +1479,20 @@ func (m *KV) NotifyMsg(msg []byte) {
 		return
 	}
 
-	ch := m.getKeyWorkerChannel(kvPair.Key)
-	select {
-	case ch <- valueUpdate{value: kvPair.Value, codec: codec, messageSize: len(msg), deleted: kvPair.Deleted, updateTime: updateTime(kvPair.UpdateTimeMillis)}:
-	default:
+	update := valueUpdate{value: kvPair.Value, codec: codec, messageSize: len(msg), deleted: kvPair.Deleted, updateTime: updateTime(kvPair.UpdateTimeMillis)}
+	if !m.enqueueKeyUpdate(kvPair.Key, update) {
 		m.numberOfDroppedMessages.Inc()
 		level.Warn(m.logger).Log("msg", "notify queue full, dropping message", "key", kvPair.Key)
 	}
 }
 
-func (m *KV) getKeyWorkerChannel(key string) chan<- valueUpdate {
+// enqueueKeyUpdate hands the update over to the worker goroutine for the given key,
+// spawning the worker if it doesn't exist yet. It returns false if the worker's queue
+// is full and the update was dropped.
+//
+// The channel send happens while holding workersMu. This guarantees that no update
+// can be sent to a channel after stopKeyWorkers has closed it.
+func (m *KV) enqueueKeyUpdate(key string, update valueUpdate) bool {
 	m.workersMu.Lock()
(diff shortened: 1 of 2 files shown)


kv/memberlist/memberlist_client.go
```diff
@@ -1500,13 +1504,24 @@ func (m *KV) getKeyWorkerChannel(key string) chan<- valueUpdate {
 
 		m.workersChannels[key] = ch
 	}
-	return ch
```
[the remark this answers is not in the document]
aldernero: Good catch. Fixed in bf0e9bc.

When a processed update results in a no-op merge (`version == 0`) and the key is absent from the store (the late-tombstone-after-purge case, and also failed merges for unknown keys), the worker now deregisters itself and stops (`tryDeregisterKeyWorker`) instead of waiting for a cleanup that will never come.

The deregistration is race-free under the same invariant the PR already relies on: it runs while holding `workersMu`, only if this worker is still the registered one for the key, and only if its update channel is empty. The channel length can't change there, since all sends also hold `workersMu`. If more updates were enqueued meanwhile, the worker keeps running; if it was already replaced, the closed channel stops it instead.

Added a test case: a `Deleted: true` pair gossiped for a key not in the store must not leave a worker behind.

kv/memberlist/memberlist_client.go
```diff
@@ -1538,13 +1553,53 @@ func (m *KV) processValueUpdate(workerCh <-chan valueUpdate, key string) {
 				m.broadcastNewValue(key, mod, version, update.codec, false, deleted, updated)
 			}
 
+			if version == 0 && !m.keyExists(key) {
```
56quarters, line 1556: Should this be an `else if` branch of the conditional above so that we only shut down this worker if merging returned a non-error result?

kv/memberlist/memberlist_client_test.go
```diff
@@ -2016,6 +2016,72 @@ func keyValuePair(t *testing.T, key string, codec codec.Codec, value interface{}
...
+		return len(kv.workersChannels)
+	}
+
+	// Receiving an update for a key spawns a worker goroutine for that key.
+	kv.NotifyMsg(marshalKeyValuePair(t, key, c, &data{
+		Members: map[string]member{
+			"a": {Timestamp: time.Now().Unix(), State: JOINING},
+		}}))
+
+	require.Eventually(t, func() bool { return numKeyWorkers() == 1 }, 5*time.Second, 10*time.Millisecond)
```
56quarters: This test seems like it depends on timing a lot. Maybe we can use `synctest` here to help?

kv/memberlist/memberlist_client_test.go
```diff
@@ -2016,6 +2016,72 @@ func keyValuePair(t *testing.T, key string, codec codec.Codec, value interface{}
...
+	flagext.DefaultValues(&cfg)
+	cfg.TCPTransport.BindAddrs = getLocalhostAddrs()
+	cfg.Codecs = append(cfg.Codecs, c)
+	cfg.ObsoleteEntriesTimeout = 100 * time.Millisecond
+
+	kv := NewKV(cfg, log.NewNopLogger(), &staticDNSProviderMock{}, prometheus.NewPedanticRegistry())
+	require.NoError(t, services.StartAndAwaitRunning(context.Background(), kv))
+	defer services.StopAndAwaitTerminated(context.Background(), kv) //nolint:errcheck
+
+	numKeyWorkers := func() int {
```
56quarters: Instead of having test code take the lock, can we add this as a private method to the KV?

kv/memberlist/memberlist_client_test.go
```diff
@@ -2016,6 +2016,72 @@ func keyValuePair(t *testing.T, key string, codec codec.Codec, value interface{}
...
+
+	// If the key is seen again, a new worker is spawned.
+	kv.NotifyMsg(marshalKeyValuePair(t, key, c, &data{
+		Members: map[string]member{
+			"b": {Timestamp: time.Now().Unix(), State: JOINING},
+		}}))
+
+	require.Eventually(t, func() bool { return numKeyWorkers() == 1 }, 5*time.Second, 10*time.Millisecond)
+
+	// A tombstone for a key that is not in the store is a no-op merge which doesn't
```
56quarters: Can you rephrase this comment? It seems like we're testing that we keep a worker around for the existing test key so the number of workers should be `1` and then that `NotifyMsg()` starts another worker but it immediately shuts down.

56quarters (commented):
Production code LGTM but there are a few changes to the tests that I think would be helpful.

kv/memberlist/memberlist_client.go
aldernero: I kept this as a standalone check intentionally: a failed merge for a key that isn't in the store (`err != nil`, `version == 0`, key absent) orphans the worker the same way the no-op tombstone does, since the cleanup will never run for a key that was never stored. Deregistering is safe in both cases because it only happens when the worker's queue is empty, and the next update for the key just spawns a fresh worker. I expanded the comment in f455af6 to spell out both cases. Happy to restrict it to the non-error path if you prefer.

kv/memberlist/memberlist_client_test.go
aldernero: Done in f455af6. One wrinkle: the started KV can't run inside a bubble because the transport goroutines block in Accept syscalls, which keeps the fake clock from ever advancing. So the test now drives `enqueueKeyUpdate` (what `NotifyMsg` uses) on an unstarted KV, with the two broadcast queues stubbed since they're normally created on service start. In exchange everything is deterministic: tombstone aging uses the fake clock and `synctest.Wait()` replaces the polling.

kv/memberlist/memberlist_client_test.go
aldernero: Done in f455af6, added `numKeyWorkers()` as a private method on `KV` in the test file.

kv/memberlist/memberlist_client_test.go
aldernero: Done in f455af6, reworded along those lines.

56quarters (approved):
Nice. I haven't been able to get this new test to flake in many 1000s of runs.

PR #1050, opened 2026-08-05
fix(memberlist): don't track zone-aware routing skips while joining

pitasi:
## what
don't increase `memberlist_client_zone_aware_routing_select_nodes_skipped_total` while the local node is still joining the cluster.

## why
a just-started node hasn't discovered every zone's bridge yet, so it skips zone-aware routing for a few seconds. that's expected, but the metric can't tell it apart from a real bridge outage.

during a rollout there's always some node in that state, so `sum(rate(..._skipped_total[...])) > 0` never returns to zero and `MimirMemberlistZoneAwareRoutingAutoFailover` fires for the whole rollout. saw it flap for over an hour on a mimir cell today, with all bridges healthy the entire time.

## notes
the failover behaviour is unchanged, only the metric is gated. the gate opens after the join attempt whether it succeeded or not, so a node that fails to join doesn't hide its failovers forever.

### DIFF
--- a/kv/memberlist/memberlist_client.go
+++ b/kv/memberlist/memberlist_client.go
@@ -323,8 +323,9 @@ type KV struct {
 	localBroadcasts  *memberlist.TransmitLimitedQueue // queue for messages generated locally
 	gossipBroadcasts *memberlist.TransmitLimitedQueue // queue for messages that we forward from other nodes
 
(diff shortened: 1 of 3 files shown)


pracucci (approved):
Makes sense to me.

PR #1058, opened 2026-08-14
perf(cache/lru): narrow global lock scope

francoposa:
**What this PR does**:

**Narrows LRU lock scope on `GetMulti`.** This comes after we observed high tail latency on the index-header bucket reader with thousands of goroutines waiting on the local LRU lock.

Previously, if any of the multiple keys missed the local LRU cache, the LRU held a global lock while it backfilled all the missed keys from the remote (memcached). This blocked _every_ other LRU operation while one or more calls to memcached were made, even on completely unrelated keys because the lock is not sharded by keyspace in any way.

Now we only hold the global LRU lock when checking the local LRU, or when updating it _after_  fetching missing keys.
In the worst case, multiple callers that are on the same key will now miss and go to memcached rather than waiting in line for the first one to fill it in. But this is a trivial downside compared to waiting in line behind locks for unrelated keys.

**Scope**
Only affects Mimir store-gateways - LRU wrapper is not used in other dskit consumers.
Within Mimir store-gateways, it only affects the existing in-memory LRU wrapper for chunks object attributes and the new use of the LRU for the index-header bucket reader.

**Benchmark results**

Before - jumping from 8 to 24 cores gives less than 2% improvement:
```
BenchmarkLRUCache_GetMulti_ConcurrentMissesWithSlowBackend-8     	   10000	   1075138 ns/op
BenchmarkLRUCache_GetMulti_ConcurrentMissesWithSlowBackend-24    	   10000	   1056089 ns/op
```

After - 8-core number is ~8x faster than its baseline, 24-core number is ~24x faster than its baseline
```
BenchmarkLRUCache_GetMulti_ConcurrentMissesWithSlowBackend-8     	   10000	    133101 ns/op
BenchmarkLRUCache_GetMulti_ConcurrentMissesWithSlowBackend-24    	   10000	     44102 ns/op
```

Fixes #<issue number>

**Checklist**
- [ ] Tests updated


cache/lru.go
```diff
@@ -186,6 +192,7 @@ func (l *LRUCache) GetMultiWithError(ctx context.Context, keys []string, opts ..
 			})
 			found[k] = v
 		}
+		l.mtx.Unlock()
```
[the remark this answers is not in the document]
francoposa: This particular cache is only ever used with TSDB blocks, whose contents never change. A KV cannot be outdated.

cache/lru.go
```diff
@@ -174,10 +173,17 @@ func (l *LRUCache) GetMultiWithError(ctx context.Context, keys []string, opts ..
 		miss = append(miss, k)
 
 	}
+	l.mtx.Unlock()
 	l.hits.Add(float64(len(found)))
 
+	// Fetch misses from the remote backing cache without holding the mutex lock.
+	// The mutex lock is not sharded by key and therefore blocks every other concurrent operation.
+	// Tradeoff: may result in extra trips to the remote cache and extra local LRU set operations
+	// in the case that multiple callers experience local LRU miss on the same keys concurrently.
```
chencs, line 182: Do we have any way of knowing how often this happens/identifying if it's happening a lot in practice?

francoposa: No measurable difference in memcached traffic when this runs is our best indicator. 
The case in the store gateway would have to be multiple concurrent goroutines querying the same identical chunk of the same block and missing at the same time, it’s not really worth worrying about in reality.
I just wanted to make clear what the effect is from a conceptual cache tradeoff angle.

chencs: Yeah, my concern is that query patterns can influence this, right? So it would be good, if memcached starts acting up in the wild, if we could trace that back to this tradeoff. Though I guess "false" cache misses would have to be an order of magnitude worse than baseline for this to cause meaningful problems. I'm willing to gamble on that not being the case if figuring out how to track it is too onerous.

francoposa: What is in main right now is massively worse than some theoretical extra local cache misses. What is in main is a complete global lock on every LRU operation every time it needs to fill in a cache miss.
This was the root cause of us still getting multi-minute tail latencies after fixing HTTP/2 - 5k+ parked goroutines all waiting on the same global lock.

We've been running for over a week with the fix and there are no issues and tail latency is much better.

chencs (approved):

francoposa:
Adding in some extra context & reasoning from Slack discussion - w.r.t the potential for "extra" local LRU misses adding read load to memcached:

1. The store-gateway query gate is the de facto limiter for Mimir. 200, or 500 whichever we use now is the absolute max. You would only get to up 500 potential cache misses on the same key if every single query was for the same tenant and hit the same block for the same 16000-byte range in the same instant.
2. The approach of not blocking every goroutine on an in-process global lock is how every other usage of memcached in Mimir already works
3. We've already run with no LRU at all which is strictly worse on memcached load and everything was fine

PR #1061, opened 2026-08-18
feat(runtimeconfig): fork mapstructure to use as MapLoader

tcard:
**What this PR does**:

We have a PR open to optimize mapstructure in order to use it to load merged runtimeconfig into the target value:

* https://github.com/go-viper/mapstructure/pull/195

The upstream repo seems dead, so let's bring it here under `runtimeconfig/mapstructure`.

Without the optimizations, mapstructure is even slower than marshaling the config into YAML, and then unmarshal that into the target, so there would be no point to runtimeconfig.MapLoader.

Follow up of #1033 

**Checklist**
- [x] Tests updated


runtimeconfig/mapstructure/internal/errors/join_go1_19.go
colega, line 1: I think we can delete this.

runtimeconfig/mapstructure/reflect_go1_19.go
colega, line 1: We can also delete this

tcard: 0658262

runtimeconfig/mapstructure/internal/errors/join_go1_19.go
tcard: 0658262

colega (approved):

juliusmh (approved):

PR #1078, opened 2026-09-07
fix: use GetFirstAddressOf for fallback advertise address in memberlist

KR-Ravindra:
## Problem

`memberlist` members that bind to `0.0.0.0` without an explicit advertise address fail to start on hosts whose only IPv4 addresses are outside the RFC 1918 ranges:

```
failed to create memberlist: Failed to get final advertise address: no private IP address found, and explicit IP not provided
```

This happens for example with Kubernetes pods on a cluster CIDR that is not in `10/8`, `172.16/12` or `192.168/16`. Such members currently need an explicit per-member `-memberlist.advertise-addr`, which is awkward to inject into every pod, while the same deployment on a private pod CIDR works out of the box.

## Root cause

`kv/memberlist/tcp_transport.go:416-422` (`FinalAdvertiseAddr`, `0.0.0.0` branch) resolves the advertise address with `sockaddr.GetPrivateIP()`, which only considers RFC 1918 addresses, and returns an error when it finds none. The IPv6 branch right below (`::`) already resolves the address with `netutil.GetFirstAddressOf(nil, t.logger, true)`, so IPv6-only members without a private address work while IPv4 ones do not.

## Fix

In the `0.0.0.0` branch, when `GetPrivateIP()` returns no address, fall back to `netutil.GetFirstAddressOf(nil, t.logger, false)`, i.e. the first non-loopback, non-link-local IPv4 address of the host, like the IPv6 branch does. Behaviour is unchanged when a private IP exists (it is still preferred), when the lookup itself fails (the error is returned, now wrapped with `%w`), and when the transport is bound to a specific address or an advertise address is given.

`sockaddr.GetPrivateIP` is referenced through a package-level `var getPrivateIP` so that tests can simulate a host without a private address.

## How tested

New `TestFinalAdvertiseAddr_BoundToZeroZeroZeroZero` in `kv/memberlist/tcp_transport_test.go`:

- `GetPrivateIP` returns a private IP: that IP is advertised (existing behaviour).
- `GetPrivateIP` returns no address: a non-nil, non-loopback, non-unspecified IP equal to `netutil.GetFirstAddressOf(nil, logger, false)` is advertised, no error.
- `GetPrivateIP` returns an error: the error is returned (`errors.Is` succeeds).

Before the fix, the fallback case fails with `no private IP address found, and explicit IP not provided` (and the error case fails `errors.Is` because the error was formatted with `%v`). After the fix all cases pass, and `go test ./kv/memberlist/...`, `go vet ./kv/memberlist/...` and `gofmt -l kv/memberlist` are clean.

## Links

There is no dskit issue for this. The scenario is memberlist members running on a pod or host network whose addresses are not in the RFC 1918 private ranges, for example Kubernetes clusters whose pod CIDR is a public or otherwise non-private range.

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/kv/memberlist/tcp_transport.go
+++ b/kv/memberlist/tcp_transport.go
@@ -36,6 +36,9 @@ const (
 const zeroZeroZeroZero = "0.0.0.0"
 const colonColon = "::"
 
+// getPrivateIP is a variable so that tests can simulate hosts without a private IP address.
+var getPrivateIP = sockaddr.GetPrivateIP
+
 // TCPTransportConfig is a configuration structure for creating new TCPTransport.
 type TCPTransportConfig struct {
 	// BindAddrs is a list of IP addresses to bind to.
@@ -413,12 +416,18 @@ func (t *TCPTransport) FinalAdvertiseAddr(ip string, port int) (net.IP, int, err
 			// Otherwise, if we're not bound to a specific IP, let's
 			// use a suitable private IP address.
 			var err error
-			ip, err = sockaddr.GetPrivateIP()
+			ip, err = getPrivateIP()
 			if err != nil {
-				return nil, 0, fmt.Errorf("failed to get interface addresses: %v", err)
+				return nil, 0, fmt.Errorf("failed to get interface addresses: %w", err)
 			}
 			if ip == "" {
-				return nil, 0, fmt.Errorf("no private IP address found, and explicit IP not provided")
+				// No RFC 1918 address on this host (e.g. pods on a non-private
+				// cluster CIDR): fall back to the first non-loopback address,
+				// as the IPv6 case already does.
+				ip, err = netutil.GetFirstAddressOf(nil, t.logger, false)
+				if err != nil {
+					return nil, 0, fmt.Errorf("no private IP address found, and explicit IP not provided: %w", err)
+				}
 			}
 
 			advertiseAddr = net.ParseIP(ip)
--- a/kv/memberlist/tcp_transport_test.go
+++ b/kv/memberlist/tcp_transport_test.go
@@ -1,6 +1,7 @@
 package memberlist
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -18,6 +19,7 @@ import (
 	"github.com/grafana/dskit/concurrency"
 	"github.com/grafana/dskit/crypto/tls"
 	"github.com/grafana/dskit/flagext"
+	"github.com/grafana/dskit/netutil"
 )
 
 func TestTCPTransport_WriteTo_ShouldNotLogAsWarningExpectedFailures(t *testing.T) {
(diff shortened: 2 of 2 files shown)


KR-Ravindra:
Round 1 self-review (AI agent operated by me; I review every revision). This replaces #1075. The production change is now the right shape (fall back to `netutil.GetFirstAddressOf(nil, t.logger, false)`, mirroring the `::` branch), but it is not ready:

1. **`getFinalAdvertiseAddrIP` is dead code.** `FinalAdvertiseAddr` does not call it; it re-implements the branch inline. The six new tests therefore only exercise `net.ParseIP` on string literals and never touch the code path this PR changes. Either make `FinalAdvertiseAddr` use the helper and pass the two lookups (`sockaddr.GetPrivateIP`, `netutil.GetFirstAddressOf`) in as injectable funcs so a test can drive the no-private-IP branch for real, or drop the helper and the test file.
2. **gofmt fails on both files**: `gofmt -l kv/memberlist` lists `tcp_transport.go` (trailing tabs on the blank lines in the helper) and `final_advertise_addr_test.go` (struct field alignment). Lint will fail.
3. **Fallback error swallows the cause.** Use `fmt.Errorf("no private IP address found, and explicit IP not provided: %w", err)` so the `GetFirstAddressOf` failure is visible.
4. **PR body**: remove the `repo:`/`title:` header lines; remove the paragraph about #1075 (it was closed unmerged, so nothing upstream "incorrectly used" the bound address); link a real report of the failure (grafana/loki#8634 is the long-running one); make "How tested" describe what the tests actually cover.

CLA signature is on me.

KR-Ravindra:
recheck

KR-Ravindra:
recheck

KR-Ravindra:
Round 2 self-review (AI agent operated by me; I review every revision). This revision must be redone. **Changes requested:**

1. **The commit deletes ~80 lines of unrelated code**: everything in `registerMetrics` after `outgoing_streams_total` (the outgoing-stream-errors, packets sent/received/dropped, unknown-connections, sent/received bytes counters) is gone. Those fields are then nil at runtime. Restore the function exactly as on `main`; the PR must touch only the `zeroZeroZeroZero` branch of `FinalAdvertiseAddr`.
2. **The test file from the first revision was dropped**, so nothing exercises the fallback. Add a test in `tcp_transport_test.go` that makes `sockaddr.GetPrivateIP` return empty (bind to `0.0.0.0` on a host/netns with only non-RFC1918 addresses, or factor the private-IP lookup behind a small function variable that the test overrides) and asserts the advertised address is the first non-loopback address, plus the error path when none exists.
3. Keep the `%w` wrapping and the `GetFirstAddressOf(nil, t.logger, false)` fallback, which are correct. `gofmt`, `go vet ./kv/memberlist/...`, `go test ./kv/memberlist/...` must pass before pushing.
4. One signed-off commit on current `main`; the PR body still needs the `Problem / Root cause / Fix / How tested` structure without the paragraph about the closed #1075.

KR-Ravindra:
Round 3 self-review (AI agent operated by me; I review every revision). I rebuilt the branch from current main as a single commit: the fallback now lives in the `0.0.0.0` branch itself, mirroring the IPv6 case; errors wrap with `%w`; `registerMetrics` and everything else are untouched. Tests cover the fallback, the unchanged private-IP path, and error propagation via a small package-level seam for `sockaddr.GetPrivateIP`. Package tests, vet and gofmt pass. Marking ready.

PR #1079, opened 2026-09-07
feat(runtimeconfig): let a config source tolerate being unreadable

replay:
**What this PR does**:

Currently a `Manager` refuses to start when it cannot read one of its sources, and it never reads that source again afterwards. A process that reads its config over HTTP therefore cannot start while the config server is down. It stays that way until someone restarts it.

An entry in `-runtime-config.file` can now end with JDBC-style semicolon options that say what to do when the source cannot be read:

- `;optional-on-startup` lets the process start without the source. A failure after startup still fails the reload, as it does today.
- `;optional-keep-last-value-on-failure` also lets the process start without the source. A failure after startup keeps the value the source supplied last, and the other sources keep updating.

Known options are peeled from the right, so a URL path parameter such as `;jsessionid=ABC` stays part of the path.
A source that the Manager has never read supplies nothing, so the merge uses the other sources on its own.
An entry without an option behaves as it does today.

The new `runtime_config_source_last_reload_successful` metric reports each source on its own. It matters because `optional-use-last-value` lets the reload succeed, which leaves `runtime_config_last_reload_successful` at `1`.

Related to this issue: https://github.com/grafana/databases-sre/issues/458

**Checklist**
- [x] Tests updated


runtimeconfig/manager.go
```diff
@@ -142,22 +162,37 @@ func New(cfg Config, configName string, registerer prometheus.Registerer, logger
...
 
 	mgr.Service = services.NewBasicService(mgr.starting, mgr.loop, mgr.stopping)
 	return &mgr, nil
 }
 
+// sourceLabel returns the "source" metric label value. Providers whose name can hold
+// credentials supply a redacted form.
+func sourceLabel(p provider) string {
+	if n, ok := p.(interface{ NameForMetrics() string }); ok {
+		return n.NameForMetrics()
```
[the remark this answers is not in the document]
replay: fixed in 45c3fe020cc3f54b367f4bb5767be23c523a6ce6

runtimeconfig/manager.go
```diff
@@ -216,48 +251,86 @@ func (om *Manager) loop(ctx context.Context) error {
...
+			err = errors.Wrapf(err, "read %q", p.Name())
+		}
+
+		if err != nil {
+			om.sourceLoadSuccess.WithLabelValues(sourceLabel(p)).Set(0)
+
+			if !policy.tolerates(initial) {
 				om.configLoadSuccess.Set(0)
-				return errors.Wrapf(err, "preprocess %q", p.Name())
+				return err
```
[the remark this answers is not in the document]
replay: fixed in a4b61b56654e2dad6ece5c0a63e41cdae0d636ad

runtimeconfig/manager.go
```diff
@@ -132,8 +146,14 @@ func New(cfg Config, configName string, registerer prometheus.Registerer, logger
 
 	var httpClient *http.Client
 	var httpDuration *prometheus.HistogramVec
-	for _, p := range cfg.LoadPath {
-		if isURL(p) {
+	for _, entry := range cfg.LoadPath {
+		src, err := parseSource(entry)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
replay: fixed in 9960fd703f6a8f1cd9451f84e5b161bb18152d8b

runtimeconfig/config_source.go
```diff
@@ -0,0 +1,113 @@
...
+	"slices"
+	"strings"
+)
+
+// sourceParameter is a per-source parameter parsed from a LoadPath entry.
+type sourceParameter string
+
+const (
+	parameterOptionalOnStartup              sourceParameter = "optional-on-startup"
+	parameterOptionalKeepLastValueOnFailure sourceParameter = "optional-keep-last-value-on-failure"
```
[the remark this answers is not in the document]
replay: updated the PR description

colega (dismissed):

runtimeconfig/manager.go
```diff
@@ -142,11 +173,15 @@ func New(cfg Config, configName string, registerer prometheus.Registerer, logger
...
 			}
-			mgr.providers = append(mgr.providers, newHTTPProvider(p, httpClient, httpDuration))
+			cs.provider = newHTTPProvider(cs.path, httpClient, httpDuration)
 		} else {
-			mgr.providers = append(mgr.providers, newFileProvider(p))
+			cs.provider = newFileProvider(cs.path)
 		}
+
+		// Create the series up front, at 0: nothing has been read yet.
+		mgr.sourceLoadSuccess.WithLabelValues(cs.provider.Name()).Set(0)
```
tinitiuset: Should we avoid exposing raw URLs as metric labels here? `provider.Name()` can contain user-info, query parameters, and fragments, potentially leaking credentials or tokens through `/metrics`.

tinitiuset (dismissed):

runtimeconfig/manager.go
replay: In an earlier version of this PR the URL was sanitized, but this introduces the problem that multiple providers could end up having the same label value if the only differentiating factor is something that gets sanitized away.
At that point, I kind of thought "nobody should be putting secrets into these URLs anyway", so I prioritized avoiding collisions and stopped sanitizing... I'm honestly not sure what's better though

replay: what if we make the name a sanitized version of the URL + a hash of the unsanitized version? I think that would solve both cases.

replay: Went with sanitizing rather than the hash: whoever reads `/metrics` sees the sanitized URL too, so they know every part of the hash input except the secret, which makes a short token brute-forceable and leaves the "don't put secrets in the URL" caveat in place anyway — and a hash suffix makes the label unreadable for whoever got paged. The label is now the URL without userinfo (dropped entirely, since `url.Redacted()` keeps the username), query, and fragment, with the LoadPath index appended only to the entries whose sanitized names actually collide. `runtime_config_http_request_duration_seconds` uses the same value now, so the two metrics can be read against each other. A secret in a path segment still reaches `/metrics`, and a URL that `url.Parse` rejects is now an error from New instead of falling back to the raw URL.

replay: done in 6a80f9b682c2bae7881f00c34fbdbd94222995e1

replay: This repo dismissed your approval when I pushed, PTAL.
I've set the PR to auto-merge.

tinitiuset (approved):

PR #1083, opened 2026-09-09
feat(server): add EnableOpenMetricsTextCreatedSamples option to dskit server.

sophiewaldman:
**What this PR does**:
Adds a config flag for EnableOpenMetricsTextCreatedSamples.
Without this option, the experimental `start_timestamp_zero_ingestion` flag doesn't do anything, because the target will not emit `_created` timestamps in the first place.

We need this experimental feature enabled for Loki, to improve metric accuracy.

Tested by running Loki locally with `EnableOpenMetricsTextCreatedSamples` manually set to true.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -254,7 +254,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Throughput.Unit, "server.throughput.unit", "samples_processed", "Unit of the server throughput metric, for example 'processed_bytes' or 'samples_processed'. Observed values are gathered from the 'Server-Timing' header with the 'val' key. If set, it is appended to the request_server_throughput metric name.")
 	cfg.ClusterValidation.RegisterFlagsWithPrefix("server.cluster-validation.", f)
 	f.BoolVar(&cfg.CreateNewTraces, "server.create-new-traces", false, "Creates new traces for each call rather than continuing the existing trace. A span link is used to allow navigation to the parent trace. Only works when using Open-Telemetry tracing.")
-	f.BoolVar(&cfg.EnableOpenMetricsTextCreatedSamples, "server.enable-open-metrics-text-created-samples", false, "Specifies if this handler should add, extra, synthetic created timestamps for counters, histograms and summaries, which for the current version of OpenMetrics are defined as extra series with the same name and \"_created\" suffix. Only applies if -server.register_instrumentation is set to true.")
+	f.BoolVar(&cfg.EnableOpenMetricsTextCreatedSamples, "server.enable-open-metrics-text-created-samples", false, "Specifies if this handler should emit start timestamps for counters, histograms and summaries over OpenMetrics 1.0, which are defined as extra series with the same name and \"_created\" suffix. Only applies if -server.register_instrumentation is set to true.")
 }
 
 func (cfg *Config) Validate() error {
--- a/server/server.go
+++ b/server/server.go
@@ -179,8 +179,9 @@ type Config struct {
 	// existing trace when the function returns true. A span link will be used
 	// to connect to any existing trace. It only works if using Open-Telemetry
 	// tracing.
-	PublicEndpointFn func(*http.Request) bool `yaml:"-"`
-	CreateNewTraces  bool                     `yaml:"create_new_traces"`
+	PublicEndpointFn                    func(*http.Request) bool `yaml:"-"`
+	CreateNewTraces                     bool                     `yaml:"create_new_traces"`
+	EnableOpenMetricsTextCreatedSamples bool                     `yaml:"enable_open_metrics_text_created_samples"`
 }
 
 type Throughput struct {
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -253,9 +254,13 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.Throughput.Unit, "server.throughput.unit", "samples_processed", "Unit of the server throughput metric, for example 'processed_bytes' or 'samples_processed'. Observed values are gathered from the 'Server-Timing' header with the 'val' key. If set, it is appended to the request_server_throughput metric name.")
 	cfg.ClusterValidation.RegisterFlagsWithPrefix("server.cluster-validation.", f)
 	f.BoolVar(&cfg.CreateNewTraces, "server.create-new-traces", false, "Creates new traces for each call rather than continuing the existing trace. A span link is used to allow navigation to the parent trace. Only works when using Open-Telemetry tracing.")
+	f.BoolVar(&cfg.EnableOpenMetricsTextCreatedSamples, "server.enable-open-metrics-text-created-samples", false, "Specifies if this handler should add, extra, synthetic created timestamps for counters, histograms and summaries, which for the current version of OpenMetrics are defined as extra series with the same name and \"_created\" suffix. Only applies if -server.register_instrumentation is set to true.")
```
krajorama: Couple of comments related to this:
- The name "created timestamp" has been replaced in the past year with "start timestamp" to align with OpenTelemetry. But of course we couldn't change OpenMetrics 1.0.
- The created timestamp isn't synthetic, it's recorded when the counter is actually created (or reset) in the application.
- Scraping OpenMetrics 1.0 isn't efficient when this feature is enabled, so you pay extra in CPU. You're much better off scraping with `PrometheusProto` (see `scrape_protocols` in https://prometheus.io/docs/prometheus/latest/configuration/configuration/) which supports start timestamps without any additional setting.
- We also changed [OpenMetrics 2.0](https://prometheus.io/docs/guides/open_metrics_2_0_migration/#start-timestamps) to not have this problem, but support is only experimental at the moment.

```suggestion
	f.BoolVar(&cfg.EnableOpenMetricsTextCreatedSamples, "server.enable-open-metrics-text-created-samples", false, "Specifies if this handler should emit start timestamps for counters, histograms and summaries over OpenMetrics 1.0, which are defined as extra series with the same name and \"_created\" suffix. Only applies if -server.register_instrumentation is set to true.")
```

krajorama: Isn't the flag written with dash '-' ?

```suggestion
	f.BoolVar(&cfg.EnableOpenMetricsTextCreatedSamples, "server.enable-open-metrics-text-created-samples", false, "Specifies if this handler should add, extra, synthetic created timestamps for counters, histograms and summaries, which for the current version of OpenMetrics are defined as extra series with the same name and \"_created\" suffix. Only applies if -server.register-instrumentation is set to true.")
```

server/server.go
```diff
@@ -253,9 +254,13 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 }
 
 func (cfg *Config) Validate() error {
+	if cfg.EnableOpenMetricsTextCreatedSamples && !cfg.RegisterInstrumentation {
+		return fmt.Errorf("server.enable-open-metrics-text-created-samples can only be used if server.register_instrumentation is set to true")
```
krajorama:
```suggestion
		return fmt.Errorf("server.enable-open-metrics-text-created-samples can only be used if server.register-instrumentation is set to true")
```

krajorama (commented):
LGTM, minor comments on the docs side and it would be nice to have a test to go with this.

server/server.go
sophiewaldman: Yes, thank you! Fixed.

server/server.go
sophiewaldman: Fixed.

sophiewaldman:
@krajorama what sort of test would you like to see here? I didn't see any similar tests that exist already.

server/server.go
```diff
@@ -580,15 +585,21 @@ func RegisterInstrumentation(router *mux.Router) {
 	RegisterInstrumentationWithGatherer(router, prometheus.DefaultGatherer)
 }
 
-// RegisterInstrumentationWithGatherer on the given router.
-func RegisterInstrumentationWithGatherer(router *mux.Router, gatherer prometheus.Gatherer) {
+// RegisterInstrumentationWithGatherer on the given router and set EnableOpenMetricsTextCreatedSamples.
```
aknuds1: The doc comment should still start with the name of the documented function.
```suggestion
// RegisterInstrumentationWithGathererAndCreatedSamples registers metrics and pprof handlers.
// If enableOpenMetricsTextCreatedSamples is true, OpenMetrics 1.0 responses include _created samples.
```

server/server.go
```diff
@@ -580,15 +585,21 @@ func RegisterInstrumentation(router *mux.Router) {
...
 	router.Handle("/metrics", promhttp.HandlerFor(gatherer, promhttp.HandlerOpts{
-		EnableOpenMetrics: true,
+		EnableOpenMetrics:                   true,
+		EnableOpenMetricsTextCreatedSamples: enableOpenMetricsTextCreatedSamples,
 	}))
 	router.Handle("/debug/pprof/cmdline", http.NotFoundHandler())
 	router.PathPrefix("/debug/pprof").Handler(http.DefaultServeMux)
 }
 
+// RegisterInstrumentationWithGatherer on the given router.
```
aknuds1:
```suggestion
// RegisterInstrumentationWithGatherer registers metrics and pprof handlers.
```

aknuds1 (commented):
The `RegisterInstrumentationWithGathererAndCreatedSamples` doc comment should be fixed (see suggestion). I also think there should be regression tests, to cover the new behaviour. Would you like me to provide the test code?

aknuds1 (dismissed):
LGTM thanks, assuming my documentation comment suggestions are applied.

aknuds1 (approved):

ywwg (approved):

